PR #31, opened 2023-10-31
Hide non-cm jobs and departments in character setup

DrSmugleaf:
image


QuietlyWhisper:
Check build action fail

PR #491, opened 2023-11-14
Id sprite fix

QuietlyWhisper:
## About the PR

Layers ID cards over backpack so you can still see them...


QuietlyWhisper:
id layering should be pushed upstream later weeeeh

Issue #30, opened 2023-10-31
assign access to roles


QuietlyWhisper:
Only the roles not yet created are missing access at this point, which is probably going to be done at the same time anyways

PR #612, opened 2023-11-21
CM14 access levels

QuietlyWhisper:
Do not use these access levels for PRs.
Use the access groups that will be included which should be assigned rather than individual access when the pr is merged

Required for

- #616
- #30


Tunguso4ka:
door access groups.

Resources/Prototypes/_CM14/Access/Groups/medical_access_groups.yml
```diff
@@ -0,0 +1,21 @@
+﻿- type: accessGroup
+  id: CMCECMO
```
Tunguso4ka, line 2: you forget to remove "CE"
CM **CE** CMO

QuietlyWhisper: thanks

DrSmugleaf (approved):

PR #717, opened 2023-11-30
Port round end sounds

Tunguso4ka:
## About the PR
Ports 7 round end sounds from CM13

- [x] It works, trust me.


QuietlyWhisper (approved):
I'm sure cm13 doesn't actually have the licenses for these but surely it's fine

Issue #714, opened 2023-11-29
Ducks found Issues


MaloTV:
# Description
Hello this is a list of issues i found while mapping i will update this list every time i find a new issue or one gets fixed

<details><summary>

### Desert floor variants have to many rocks :/ [#709]

</summary>

_How they are_

![image](https://github.com/CM-14/CM-14/assets/70515562/2201036b-ca78-4d20-a44e-4dd00318f51f)

_How they should be_

![image](https://github.com/CM-14/CM-14/assets/70515562/6e2ce6f1-1b04-44d0-aac5-96907814e394)

</details>

QuietlyWhisper:
reopen this if anything new gets added

MaloTV:
Reopen?

diraven:
This seems to be taken care of?

Vermidia:
Going to assume this is done.

PR #724, opened 2023-12-01
fix sandbag height

QuietlyWhisper:
resolves an issue on 
- #714
@MaloTV 
image


thymol-blue:
The DamageOverlay for the sandbags needs to be edited as well or else it will move when damaged

QuietlyWhisper:
thanks

QuietlyWhisper:
> I think it is so shifted up because the damage overlay "leaks" down so it needs to be shifted up like that. Not sure how I was going to fix it so I left it how it was 😅

Yup. I fixed this by just... clipping the extra details, unfortunately

PR #742, opened 2023-12-10
Radio channels colors, MarineCommon as Common

Tunguso4ka:
## About the PR
adds color for Intel radio channel,
changes color for Echo radio channel,
makes MarineCommon - common.


Content.Shared/Chat/SharedChatSystem.cs
```diff
@@ -23,7 +23,7 @@ public abstract class SharedChatSystem : EntitySystem
     public const char DefaultChannelKey = 'h';
 
     [ValidatePrototypeId<RadioChannelPrototype>]
-    public const string CommonChannel = "Common";
+    public const string CommonChannel = "MarineCommon";
```
QuietlyWhisper, line 26: smug will probably want this to be a server variable but it's fine for now

QuietlyWhisper (approved):

PR #761, opened 2023-12-20
Ported human screams

Tunguso4ka:
## About the PR
Ports human male and female screams.

## Media

https://github.com/CM-14/CM-14/assets/71643624/18f2e36c-12ab-462c-9fef-4eb83200ad1e

(scary)


JoeHammad1844:
*warcry

QuietlyWhisper (changes requested):
I think the best way to do this will to be to make a new `CMMobHuman` that inherits from `MobHuman`, this way we can just change the spawn prototype in one line, less potential conflicts.

Then just make a _CM14 screams.yml for the humans

in `Resources\Prototypes\Species\human.yml`
we would update `prototype: MobHuman` to be `prototype: CmMobHuman`

QuietlyWhisper (approved):
if it works this is correct

PR #763, opened 2023-12-24
Pheromones cleanup

nikthechampiongr:
## About the PR

Moved a bunch of constants in the XenoPheromonesSystem in their respective Pheromones Components. Also made all the pheromone components SessionSpecific. This resolves 2 outstanding TODOs on the system.

## Why / Balance

This allows pheromone effects to be modified without affecting every other xeno(should that ever be necessary), it also follows the ECS philosophy more if that matters.

Components that have info about bonuses the xenos have no longer get leaked to clients that should not know about them like marines.
## Technical details

I just moved every relevant constant to their components. For SessionSpecific it's all handled by a single function with a generic argument over the component which just checks whether the entity receiving them is a xeno.
## Media

## Breaking changes

**Changelog**

no cl no fun

### DIFF
--- a/Content.Shared/_CM14/Xenos/Pheromones/XenoFrenzyPheromonesComponent.cs
+++ b/Content.Shared/_CM14/Xenos/Pheromones/XenoFrenzyPheromonesComponent.cs
@@ -1,6 +1,7 @@
 ﻿using Content.Shared.Damage.Prototypes;
 using Content.Shared.FixedPoint;
 using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype.List;
 using Robust.Shared.Utility;
 using static Robust.Shared.Utility.SpriteSpecifier;
 
@@ -22,7 +23,7 @@ public sealed partial class XenoFrenzyPheromonesComponent : Component
     [DataField]
     public FixedPoint2 MovementSpeedModifier = 0.1;
 
-    [DataField, ValidatePrototypeId<DamageGroupPrototype>]
+    [DataField(customTypeSerializer: typeof(PrototypeIdListSerializer<DamageGroupPrototype>))]
     public List<string> DamageTypes = new () { "Blunt", "Slash", "Piercing" };
 
 
--- a/Content.Shared/_CM14/Xenos/Pheromones/XenoFrenzyPheromonesComponent.cs
+++ b/Content.Shared/_CM14/Xenos/Pheromones/XenoFrenzyPheromonesComponent.cs
@@ -16,20 +16,15 @@ public sealed partial class XenoFrenzyPheromonesComponent : Component
     [DataField]
     public FixedPoint2 Multiplier;
 
-    [DataField, ValidatePrototypeId<DamageTypePrototype>]
-    public string DamageTypeOne = "Blunt";
-
-    [DataField, ValidatePrototypeId<DamageTypePrototype>]
-    public string DamageTypeTwo = "Slash";
-
-    [DataField, ValidatePrototypeId<DamageTypePrototype>]
-    public string DamageTypeThree = "Piercing";
-
     [DataField]
     public float AttackDamageModifier = 1.1f;
 
     [DataField]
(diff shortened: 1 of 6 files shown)


Content.Shared/_CM14/Xenos/Pheromones/XenoFrenzyPheromonesComponent.cs
```diff
@@ -14,4 +16,16 @@ public sealed partial class XenoFrenzyPheromonesComponent : Component
...
     public FixedPoint2 Multiplier;
+
+    [DataField]
+    public float AttackDamageModifier = 1.1f;
+
+    [DataField]
+    public FixedPoint2 MovementSpeedModifier = 0.1;
+
+    [DataField(customTypeSerializer: typeof(PrototypeIdListSerializer<DamageGroupPrototype>))]
+    public List<string> DamageTypes = new () { "Blunt", "Slash", "Piercing" };
```
DrSmugleaf, line 27: This should be DamageTypePrototype, not DamageGroupPrototype (Brute is a group, Blunt/Slash/Piercing are types) 
ProtoId<DamageTypePrototype> can be used instead so you don't need to use a custom type serializer.
You might need to update some usages after this is done.

```suggestion
    [DataField]
    public List<ProtoId<DamageTypePrototype>> DamageTypes = new () { "Blunt", "Slash", "Piercing" };
```

Content.Shared/_CM14/Xenos/Pheromones/XenoWardingPheromonesComponent.cs
```diff
@@ -14,4 +15,12 @@ public sealed partial class XenoWardingPheromonesComponent : Component
 
     [DataField]
     public FixedPoint2 Multiplier;
+
+    [DataField]
+    public string DamageTypeOne = "Bloodloss";
+
+    [DataField, ValidatePrototypeId<DamageTypePrototype>]
+    public string DamageTypeTwo = "Asphyxiation";
```
DrSmugleaf, line 23: These two should be in a list as with XenoFrenzyPheromonesComponent.

DrSmugleaf (changes requested):
Some small changes and it should be good to merge, thanks for the PR!
Might need to check if being a marine and controlling a xeno makes you get existing pheromones components or not, if not then you would need to dirty all pheromones components when someone becomes a xeno, but that can be done later.

nikthechampiongr:
> Might need to check if being a marine and controlling a xeno makes you get existing pheromones components or not

I can test that but what do you mean by this? When does a marine control a xeno?

nikthechampiongr:
Oh if it's just about new xenos not getting the pheromones I can test that one second.

nikthechampiongr:
Ok actually thanks a lot for the heads up. I fucked up the function for SessionSpecific, had to use args.Player not uid. Forgot uid is for the actual entity with the component.

nikthechampiongr:
Alright so I did as much testing as I could locally, however I am not 100% certain it works perfectly. Specifically, having 2 clients connected and making one a xeno will lead to the human client knowing about the pheromone component. For some reason it won't know about the other pheromone components once the xeno starts emitting them. Putting the xeno pheromone component to some random mob will lead to the xeno client knowing about while the human does not which is how it's supposed to work. I did all the testing with View Variables.

nikthechampiongr:
> Alright so I did as much testing as I could locally, however I am not 100% certain it works perfectly. Specifically, having 2 clients connected and making one a xeno will lead to the human client knowing about the pheromone component. For some reason it won't know about the other pheromone components once the xeno starts emitting them. Putting the xeno pheromone component to some random mob will lead to the xeno client knowing about while the human does not which is how it's supposed to work. I did all the testing with View Variables.

This does also mean that a human becoming a xeno works perfectly fine.

nikthechampiongr:
Ok I will need to actually hunt down the XenoPheremonesComponent bug that makes it known to the client even when it shouldn't.

nikthechampiongr:
Nvm there is no point. It will do what it needs to do even if clients get the initial state of it.

DrSmugleaf (approved):

PR #809, opened 2024-01-05
Steel

ike709:
Adds steel for https://github.com/CM-14/CM-14/issues/746

Also sets the recipes to use the new prototype.

image

Small first PR to make sure I'm not doing things wrong.


thymol-blue:
Thanks for updating the barricade graphs!

DrSmugleaf (approved):
Thanks for the PR!

Issue #794, opened 2024-01-03
Missing general vendor fills


DrSmugleaf:
- [ ] #793 
- [ ] #792 
- [ ] #791 
- [ ] #790 
- [ ] #789 
- [ ] #788

PR #810, opened 2024-01-06
Add all MRE components and contents, water bottle, matchbooks and packet, fortune cookie, add MRE to vendors

DrSmugleaf:
## About the PR
See #794 

## Media
https://github.com/CM-14/CM-14/assets/10968691/19d1789a-95d5-4c50-8e57-7a0ae7248f68


JoeHammad1844:
can the matches and cigs fit in the helmet?

DrSmugleaf:
> can the matches and cigs fit in helmets?

Not yet, none of the helmets have a storage component.

QuietlyWhisper:
> can the matches and cigs fit in helmets?

Adding onto this, do you know if its an itemslot or a full storage? Single item slots can be replaced instead of using storage, so you can just take and place into it, much like how the captains sabre and knife/pistol jackboots work now

JoeHammad1844:
helmets are supposed to have 3 slots. by the way are slots supported in the current inventory system?

QuietlyWhisper:
> helmets are supposed to have 3 slots. by the way are slots supported in the current inventory system?

That would just be a regular inventory then

JoeHammad1844:
> > helmets are supposed to have 3 slots. by the way are slots supported in the current inventory system?
> 
> That would just be a regular inventory then

actually its more like 2 slots, the third is reserved for an accessory like broken NVG or a camo net

PR #964, opened 2024-02-01
Moved cans yml

Tunguso4ka:
## About the PR
We had two Consumable dirs. This PR fixes it by moving cans.yml from Consumable/Drink to Consumables/Drink.


QuietlyWhisper (approved):
Thanks

PR #999, opened 2024-04-02
Fixed pill size

### LINKED ISSUE #1001: 16 pills don't fit in a pill canister currently
## Description

Dunno if there's supposed to be that many, but the canister can just be made bigger if that's the case..

It would come with the issue of the inventory being stupidly long but I'm not sure if there's a better solution other than making the pills 1x1, which isn't really bad with our inventory changes but looks inconsistant with the rest of the game.

Tunguso4ka:
## About the PR
CM Pills was to big for CM Pill Canister. So i fixed it.

resolves #1001

### DIFF
--- a/Resources/Prototypes/Catalog/Fills/Boxes/general.yml
+++ b/Resources/Prototypes/Catalog/Fills/Boxes/general.yml
@@ -11,7 +11,7 @@
   - type: Storage
     maxItemSize: Small
     grid:
-    - 0,0,4,1
+    - 0,0,2,2
   - type: Sprite
     state: box
   - type: Tag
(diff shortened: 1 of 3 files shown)


Tunguso4ka:
okay, it seems we need to accept our build errors and live with them, because i dont want to fix every storage container from upstream. I will revert fix of Upstream box.

QuietlyWhisper (changes requested):
is the actual issue that there's too many pills in one canister? if it's supposed to be 16, then just make the pill canister larger, and keep the inventory 1x2 items like CM13
image

QuietlyWhisper:
I think we could also alternatively just make custom item sizes?

QuietlyWhisper:
We're also working on fix for upstream sizes

PR #1002, opened 2024-04-04
Automatically expand the space of any storages that are too small on spawn, adjust item sizes

DrSmugleaf:
## About the PR
It now warns instead of erroring.
This doesn't distinguish between upstream and cm inventories but it should some day, so that we know when a cm inventory is setup wrong.

## Media
image
image


QuietlyWhisper (approved):
Epico

PR #1003, opened 2024-04-04
Fill medical belts

Tunguso4ka:
## About the PR

Fills medical belt and ERT medical belt with autoinjectors, traumapacks and cm ointments.
Leaved here todo for yet not implemented drugs.


QuietlyWhisper (dismissed):
Probably fine

PR #1186, opened 2024-04-12
Weapon accuracy very bad when not wielded

Froffy025:
## About the PR

One handed firing of most weapons is a lot more punishing now. Weapons used improperly are no longer able to be effectively aimed, and recoil is much more difficult to control, with higher per-shot kickback and lower recovery. Pistols can be effective with one hand, but only when a shooter can take their time recover from recoil. One-handed recoil is dangerous to everything around a Marine.

Wielded numbers remain mostly the same. 

## Technical details

- requires https://github.com/space-wizards/space-station-14/pull/26915
## Media

## Breaking changes

**Changelog**

:cl:
- One-handed recoil adjusted to be less accurate and more dangerous. Marines should remember to properly wield their firearms before engaging in combat.


QuietlyWhisper:
Unsure if this is accuracy or recoil, but when I said recoil it meant the physical camera movement when you shoot one handed in cm13 which usually throws the camera to one side.

Regardless though this seems important as well, I'll ask smug to look at it.

Froffy025:
oooh. that'd probably involve making a camera recoil component somehow and using it in Content.Client/Camera/CameraRecoilSystem.cs but i don't know how to do that

QuietlyWhisper:
I'm assuming these are the videos
https://discord.com/channels/1168210010233376858/1168210011823026270/1228359411974541373

Content.Shared/Wieldable/WieldableSystem.cs
```diff
@@ -90,6 +90,8 @@ private void OnGunRefreshModifiers(Entity<GunWieldBonusComponent> bonus, ref Gun
         {
             args.MinAngle += bonus.Comp.MinAngle;
             args.MaxAngle += bonus.Comp.MaxAngle;
+            args.AngleDecay += bonus.Comp.AngleDecay;
+            args.AngleIncrease += bonus.Comp.AngleIncrease;
```
DrSmugleaf, line 94: This is good to keep but this should go on an upstream PR since it modifies an upstream file.
This is probably a better solution than just cranking up the camera recoil very high alone, for one because it doesn't make people throw up and two because a modified client cannot cheat modified angle values on the gun.
From the videos on Discord though the unwielded spread is too high, this is how shooting an M41A in cm13 looks unwielded vs wielded with a firearms skill of 1 (rifleman):

https://github.com/CM-14/CM-14/assets/10968691/5511b16d-185b-4c33-be0a-de7b3344425f

With a firearms skill of 0 it is a lot worse depending on where you are aiming and hard to see, but the accuracy penalty is less severe than in the videos since it doesn't quite shoot sideways:

https://github.com/CM-14/CM-14/assets/10968691/4f64c34c-2d0f-496c-8bf2-1ed9f25a0e8f

Content.Shared/Weapons/Ranged/Components/GunWieldBonusComponent.cs
```diff
@@ -17,4 +17,10 @@ public sealed partial class GunWieldBonusComponent : Component
     /// </summary>
     [ViewVariables(VVAccess.ReadWrite), DataField("maxAngle"), AutoNetworkedField]
     public Angle MaxAngle = Angle.FromDegrees(-43);
+
+    [ViewVariables(VVAccess.ReadWrite), DataField("angleDecay"), AutoNetworkedField]
+    public Angle AngleDecay = Angle.FromDegrees(-43);
+
+    [ViewVariables(VVAccess.ReadWrite), DataField("angleIncrease"), AutoNetworkedField]
+    public Angle AngleIncrease = Angle.FromDegrees(-43);
```
DrSmugleaf, line 25: This should also go upstream

DrSmugleaf (changes requested):

QuietlyWhisper:
Now we just wait on
- requires https://github.com/space-wizards/space-station-14/pull/26915

Froffy025:
why is he so broken

Froffy025:
trying again on #1239 because the checks here don't like me

PR #1187, opened 2024-04-12
Fix multiplier issues with marine orders

nikthechampiongr:
## About the PR

This PR fixes what I can only describe as an unfathomable skill issue on my part when I originally wrote this code. Which I can only attribute to sleep deprivation and stupidity. 

The way modifiers were applied for the hold older made is so you received MORE damage. 

This pr also standardizes the way orders apply their modifier. 

Lastly it just makes some fields in the marineorders component VVWrite which may be useful for admins.

## Why / Balance

It bug.

## Technical details

Now all modifiers are applied as original_value * (1 * modifier).

Also orders have default values properly.

## Media

## Breaking changes

N/A
**Changelog**

no cl no fun


DrSmugleaf (approved):
There's a bug where AssignMultiplier doesn't set the value but multiply the existing one, I don't think that would work properly with multiple people giving out orders, but that's for another PR.

nikthechampiongr:
> There's a bug where AssignMultiplier doesn't set the value but multiply the existing one, I don't think that would work properly with multiple people giving out orders, but that's for another PR.

I think that might have been a conscious decision on how to handle orders being stacked but I cannot recall. Either way it was a bad decision on my part. I'll fix that as well.

PR #1235, opened 2024-04-16
Make radio more readable on new chat

Skarletto:
## About the PR

Some radio channels were painfully unreadable with newchat, so I changed the colors of all radios, though most radios only got a little brighter. I think MP radio is the one I'm most unhappy with since it's kinda drab, but at least it's easier to tell which radio is which between MP and Alpha.
## Why / Balance

As we don't use the same kind of chat background as CM13 does, radio colors would benefit from an overhaul. This improves visibility and clarity. Also, we have like 4 brown~orange radios, so I tried making them more different for visibility. I don't know what Marine Common is, but I was assuming it were the Almayer radio, so I made it gray like on CM13.
## Technical details

nuffin major
## Media

gddhd
image (1)
image
image
image

## Breaking changes

nun

**Changelog**

:3

### DIFF
--- a/Resources/Prototypes/_CM14/radio_channels.yml
+++ b/Resources/Prototypes/_CM14/radio_channels.yml
@@ -13,49 +13,49 @@
   name: chat-radio-marine-common
   keycode: "g" #; and g in cm13
   frequency: 1480
-  color: "#C38312"
+  color: "#a0a0a0"
 
 - type: radioChannel
   id: MarineCommand
   name: chat-radio-marine-command
   keycode: "v" #v in cm13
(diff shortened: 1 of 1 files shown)


QuietlyWhisper:
Looks a lot better yeah!

Skarletto:
image
I'm having trouble with coloring MP, if anyone has suggestions lmk, but for now I will keep experimenting until I'm satisfied with the coloring

Skarletto:
I'm going to assume tests failing are not my fault

QuietlyWhisper (approved):
Is good thanks

PR #1244, opened 2024-04-16
Change all item and storage sizes to be consistent with CM13

### LINKED ISSUE #1147: Item and storage size fixes
CM13 has static item sizes that would be very easy for us to replicate, they use the same style of system we're doing for 14 so it works out, we just need to do a checkup on all the item sizes.

DrSmugleaf:
## About the PR
Resolves #1147 
Also adds a way for storage to fit items that normally wouldn't fit inside, so that webbings can fit magazines.
Adds generic tags for rifle/pistol/revolver/sniper magazines.

## Media

**Changelog**
:cl:
- tweak: Changed all item and storage sizes to be consistent with CM13.


QuietlyWhisper:
Storage fill fails my beloathed

PR #1300, opened 2024-04-22
update cmdev

QuietlyWhisper:
Currently broken due to multiple duplicated UID


QuietlyWhisper:
also includes minor entity replacements
image

PR #2201, opened 2024-06-02
Passive speed modifiers while walking on weeds

Errant-4:
## About the PR
Fixes https://github.com/CM-14/CM-14/issues/1980

Mobs walking on weeds will have their speed increased or reduced, depending on whether or not they have XenoComponent
Speed multipliers can be set separately for xeno/outsider, and per weeds prototype

## ISSUE

I cleaned up the code with the last wip commit but it broke the HiveLord's Resin Walker ability in a somewhat exploitable way: when they step off weeds with the ability active, they will lose the weeds speed multiplier but keep the ResinWalker multiplier until they turn off the ability

**Changelog**

:cl: Errant
- add: Walking on resin/weeds now speeds up xenos and slows down everyone else.


DrSmugleaf (approved):
Thankd for the PR!

PR #2224, opened 2024-06-03
VP 78 pistol and magazine

Errant-4:
## About the PR
Basic implementation of VP 78 pistol and its magazine. It works, but its (assumed) original stats cannot be replicated without system updates we can't have different ROF for different firemodes, or different damage for guns that fire the same cartridges

Additionally, the pistol ROF values shown on the CM-SS13 wiki make **no sense** so for now I just gave it the same ROF as the 88M4

Part of  #1551

## Media


Errant-4:
Okay I guess the wiki was prosaic about it being automatic, but if it's semi-auto, do we set it's rate of fire for burst mode? Meaning it will also be a fast firing semi-auto?

QuietlyWhisper (approved):
thanks

PR #2321, opened 2024-06-06
Folded stasis bag into medical vendor

MetalSage:
## About the PR

Folded stasis bag replace unfolded stasis bag into medical vendor


QuietlyWhisper (approved):
Thanks

Issue #2333, opened 2024-06-07
[BUG] Red Loading indicator on the wrong layers most weapons


amo020:
## Description
Minor issue with the layering on loading indicators of various weapon. This is a sprite work with a little bit of state management.
See affected list below:

##Scope
| Weapon | Red ammo indicator position | Mag-0 |
| --- | --- |
| M41AE2 | base | Ok |
| M39 | base | Ok |
| M79 | base | Missing |
| M92 | base | Missing |

## Step to reproduce
* Grab a loaded weapon (from the list above)
* Unload the weapon

## Expected behaviour
The red loading indicator update to show no mag.

## Observed behaviour
Red indicator stays on.

amo020:
I will try and solve this issue if you allow me.

amo020:
Working on it here: https://github.com/amo020/CM-14/tree/Fix-red-ammo-indicator-layering

amo020:
Correcting the M92 blinking behaviour require it's own issue

PR #2334, opened 2024-06-07
Fix red ammo indicator for guns

amo020:
## About the PR
Moved the red ammo indicator of the M41AE2, M4RA, M39, M79 to the mag layer. Like done for the M41A
As specified in #2333

## Technical details
A design decision have to be taken for the M92 as there is a strange layout in there.

## Media

**Recording :**

https://github.com/CM-14/CM-14/assets/167775671/19ea11ff-5f95-4371-9972-bc588e08d2ae


amo020:
Fixed all but the M92 as the blink is not that easy to implement for me.

DrSmugleaf:
I'll need to check how it works in 13/14 to see if we can avoid the sprite change easily, so that we stay closer to cmss13 and make porting any future sprites easier in the future.

PR #2336, opened 2024-06-08
Adds a popup for CMItemSlots cooldown

### LINKED ISSUE #2328: cm itemslots does not display cooldown message

Just-a-Unity-Dev:
Resolves #2328

image
image


DrSmugleaf:
I need to check how this works in CM13 since there is no popup there for magazine boxes (which is what prompted the github issue to be created), but there is still a cooldown.
The popup was originally intended for the knife rig, handled in line 315.
Either way, putting this in the handler for ItemSlotEjectAttemptEvent wouldn't work, since that would trigger any time CanEject is called, even if a null popup is passed, which includes when interaction verbs are created upon right clicking an entity.

Just-a-Unity-Dev:
alright

DrSmugleaf:
We made magazine boxes have no cooldown in the end since It would have affected every player. In the future the cooldown stuff probably needs to be moved to the cm holster component or its own since I don't think we are going to be using it in any other case, but for now it's fine staying in cm item slots component

Issue #2307, opened 2024-06-06
The M52CE2 needs weapon mounting for accuracy


gostt899:
## Description
The M52CE2 heavy pulse rifle no longer fires in automatic and is instead now single action when it should be automatic.

**Reproduction**
Grab the heavy pulse rifle. hold down M1 and it only fires one shot.

**Video**

https://github.com/CM-14/CM-14/assets/137981496/718a60ce-c747-4859-b0d9-305ee73d3b4c

Dutch-VanDerLinde:
this is fixed

PR #2339, opened 2024-06-09
Makes the HPR full-auto, fixes the attempted quick equip on wielding pulse rifles.

Siigiil:
## About the PR
The HPR now fires in full- instead of semi-auto.
The mk2 and the HPR no longer snap to your armour for a brief moment when you wield them.

## Why / Balance
#2307

## Media
thing

**Changelog**
:cl:
- fix: The HPR fires in full-auto again.
- fix: The mk2 and the HPR no longer try to snap to your armour when being wielded.

### DIFF
--- a/Resources/Prototypes/_CM14/Entities/Objects/Weapons/Guns/Rifles/uscm_rifles.yml
+++ b/Resources/Prototypes/_CM14/Entities/Objects/Weapons/Guns/Rifles/uscm_rifles.yml
@@ -7,7 +7,7 @@
   - type: Item
     size: Large
   - type: Clothing
-    quickEquip: true
+    quickEquip: false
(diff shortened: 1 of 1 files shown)


QuietlyWhisper:
This can't be full auto until we have weapon mounting to account for the accuracy, it won't be used in the playtest for that reason, thanks for the PR I'll go update the issue

Siigiil:
> This can't be full auto until we have weapon mounting to account for the accuracy, it won't be used in the playtest for that reason. Thanks for the PR I'll go update the issue

I'll close this for now, then. The quick equip thing was fixed separately, and full-auto can just be toggled on in whatever eventual PR adds mounting — it's just two lines of code.

PR #2385, opened 2024-06-10
Fix Ordnance Technician description

LankLTE:
image

I asked myself "What's the easiest PR I could do?" This is a one character change. Fixes the localization line for the Ordnance Technician being incorrect.


DrSmugleaf (approved):
ty king

PR #2396, opened 2024-06-10
Removes ammonia from rotting corpses and perishables

### LINKED ISSUE #2355: Bodies produce miasma
## Description

Title. Shouldn't happen, can lead to areas being gassed. (i.e Shuttle, which transports bodies)

K-Dynamic:
## About the PR
Corpses and perishable no longer produce ammonia.

## Why / Balance
Fixes #2355 - atmos takes a backseat in CM14, the purple stank is distracting.

## Technical details
Ammonia itself is still in-game, set mol generation to 0 in PerishableComponent.cs.

## Media
image

- [x] I have added screenshots/videos to this PR showcasing its changes ingame

## Breaking changes
Hopefully none since ammonia gas is still in-game, just not generated by rot.

**Changelog**

:cl:
- tweak: Bodies and perishables no longer produce ammonia

### DIFF
--- a/Content.Server/Atmos/Rotting/RottingSystem.cs
+++ b/Content.Server/Atmos/Rotting/RottingSystem.cs
@@ -1,9 +1,11 @@
 using Content.Server.Atmos.EntitySystems;
+using Content.Server.Body.Components;
 using Content.Server.Temperature.Components;
 using Content.Shared.Atmos;
 using Content.Shared.Atmos.Rotting;
 using Content.Shared.Damage;
 using Robust.Server.Containers;
+using Robust.Shared.Physics.Components;
 using Robust.Shared.Timing;
 
 namespace Content.Server.Atmos.Rotting;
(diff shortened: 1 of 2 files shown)


DrSmugleaf (changes requested):
This should be a "CreateMiasma" DataField on PerishableComponent or RottingComponent that disables miasma generation when set to false, and upstreamed later. The reason is we want to avoid modifying upstream code where possible to avoid merge conflicts.

QuietlyWhisper:
You can set it to 0 on the yml of the mobs, then you dont have to edit the .cs files if thats what you wanted to try

K-Dynamic:
Is it better to make an upstream PR that adds a data field and modifies both RottingSystem.cs and PerishableComponent.cs, or PR in this repo yml that sets mol generation to 0?

K-Dynamic:
> After discussing further with QuietlyWhisper, I have elected to upstream the data field approach so other forks can also prevent miasma generation if needed

After discussing merge conflicts with DrSmugleaf and some upstream feedback, I've reopened this PR

Whilst I could use yaml and set the CM14 organic base species mol gas generation to 0, any SS14 mobs spawned using upstream organic base will end up polluting the map with miasma

DrSmugleaf (approved):

DrSmugleaf:
Thanks for the PR!

Issue #2384, opened 2024-06-10
Deconstructing various structures yields base-game metal, not CM metal.


SphiraI:
## Description
Title

**Reproduction**
- Find "metal" dismantlable structure
- Deconstruct
- Cry at lack of correct metal type

QuietlyWhisper:
this issue includes structures like tables and chairs that may be dropping upstream steel due to parenting

PR #2400, opened 2024-06-10
Fixes material drops from destroyed constructible objects, makes the construction steps identical to those in CM13

### LINKED ISSUE #2389: De/Construction inconsistencies
- CM14 reinforced girders turn into SS14 steel sheets when destroyed.
- Barricades give 2 metal when deconstructed, and 3 when destroyed.

Siigiil:
## About the PR
Girders and other structures marines can build now drop the CM14 versions of materials instead of the vanilla SS14 ones.
All current recipes made identical to the ones in CM13. Excluding the APC, the turnstiles and the windoors:
- I can't really make APCs be built like CM13 ones in a way that won't require editing when/if we get swapping power cells and such, so for now they're just closer, not identical.
- The turnstiles don't appear to be constructible in CM13, so I left them as is.
- I changed the final construction for windoors from using a crowbar to using a screwdriver to be consistent with the usual pattern of "screwdriver to secure board, crowbar to pull it out."

Reinforced girder removed from the construction menu.

Airlock assemblies now properly change their sprites during construction.

Closes #2389

## Why / Balance
From what I understand, the plan is to attempt to get CM14 as close to CM13 as possible. Hence the changes to material numbers and construction steps.

As far as I can tell, marines can only build normal girders in CM13, and both normal walls and rwalls are built using them. So, since the reinforced girder doesn't actually *do* anything, I removed it from the crafting menu until we find a use for it.

## Media
Airlocks in different stages of construction:
airlocks

**Changelog**

:cl: Sigil
- remove: Reinforced girder removed from crafting menu, as it currently has no use.
- tweak: Where possible, changed the material requirements and construction steps for objects marines can build to match their CM13 versions.
- fix: Made girders and other constructible objects drop the CM14 versions of materials.
- fix: Made catwalks snap to the grid properly.
- fix: Made airlocks properly change their appearance during construction.

### DIFF
--- a/Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
+++ b/Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
@@ -40,11 +40,11 @@
           CMShardGlass:
             min: 1
             max: 1
-      - !type:SpawnEntitiesBehavior
-        spawn:
-          CableApcStack1:
-            min: 1
-            max: 1
+#      - !type:SpawnEntitiesBehavior
+#        spawn:
+#          CableApcStack1: #TODO CM14 APC cable
+#            min: 1
+#            max: 1
   - type: Construction
     graph: CMWindoor
     node: assembly
--- a/Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
+++ b/Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
@@ -37,9 +37,14 @@
         acts: ["Destruction"]
       - !type:SpawnEntitiesBehavior
         spawn:
-          CMSheetMetal1:
+          CMShardGlass:
             min: 1
-            max: 3
+            max: 1
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          CableApcStack1:
+            min: 1
+            max: 1
   - type: Construction
     graph: CMWindoor
     node: assembly
@@ -62,9 +67,14 @@
         acts: ["Destruction"]
       - !type:SpawnEntitiesBehavior
         spawn:
-          SheetPlasteel1: #todo cm14 look up
+          CMShardGlass:
             min: 1
-            max: 2
+            max: 1
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          CableApcStack1:
+            min: 1
+            max: 1
   - type: Construction
     graph: CMWindoor
     node: assemblySecure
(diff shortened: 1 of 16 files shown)


ike709:
Does this also fix https://github.com/CM-14/CM-14/issues/2384 ?

Siigiil:
> Does this also fix #2384 ?

It does. We have a very limited number of deconstructible objects for now, and I'm pretty sure I fixed all of them. Didn't notice that issue yesterday, thanks for pointing it out.

Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
```diff
@@ -37,9 +37,14 @@
...
-          CMSheetMetal1:
+          CMShardGlass:
             min: 1
-            max: 3
+            max: 1
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          CableApcStack1:
+            min: 1
+            max: 1
```
QuietlyWhisper, line 47:
```suggestion
/*      - !type:SpawnEntitiesBehavior
        spawn:
          CableApcStack1:
            min: 1
            max: 1 TODO CM14 APC cable */ 
```

Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
```diff
@@ -62,9 +67,14 @@
...
-          SheetPlasteel1: #todo cm14 look up
+          CMShardGlass:
+            min: 1
+            max: 1
+      - !type:SpawnEntitiesBehavior
+        spawn:
+          CableApcStack1:
             min: 1
-            max: 2
+            max: 1
```
QuietlyWhisper, line 77:
```suggestion
/*      - !type:SpawnEntitiesBehavior
        spawn:
          CableApcStack1:
            min: 1
            max: 1 TODO CM14 APC cable */ 
```

QuietlyWhisper (commented):
Looks fine, we'll probably find out if anything was wrong during the playtest

Commenting out the wire drops so we can get cm14 wire in instead

QuietlyWhisper:
- also removing #2384 from fixes list because that issue includes unintended drops like tables and chairs giving upstream steel due to parenting

QuietlyWhisper (changes requested):
you'll have to manually add the second comment block because I cannot do it from github

Resources/Prototypes/_CM14/Entities/Structures/Doors/Windoors/assembly.yml
Siigiil: Commented out. Also fixed a bug with windoor deconstruction that I missed. Remember that YAML uses `#` for comments and doesn't support block comments.

QuietlyWhisper (approved):
Looks good hopefully we didn't miss anything important

PR #2403, opened 2024-06-11
Fix misprediction where a bullet would be "lost" on wielding

dffdff2423:
## About the PR

Fixes https://github.com/CM-14/CM-14/issues/2343

## Why / Balance

Bugfix. I elaborated a bit more in Technical Details.

## Technical details

Given that the cycle is not being handled server-side I assume that the
intended behavior is for the gun to not cycle when you press z. I think this is
a better approach than binding two actions to the same key because wielding is the
more frequent action. If this is not the case let me know and I am happy to
figure out a better solution.

## Media

**Changelog**

:cl: Aquif
- fix: Guns will no longer pretend to cycle when you try to wield it by pressing z.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
@@ -36,7 +36,6 @@ public abstract partial class SharedGunSystem
         if (args.Handled)
             return;
 
-        ManualCycle(uid, component, TransformSystem.GetMapCoordinates(uid), args.User);
         args.Handled = true;
     }
 


XxSWAG-MASTER-OVERLORDxX:
This is also a issue upstream. could this be used to fix the issue upstream?

dffdff2423:
Maybe? I thought that it actually cycled upstream where here it does not. I will test it now.

dffdff2423:
Just tested it. The behavior is different upstream. On CM z toggles whether or not you are wielding it. Upstream, z acts differently depending on when it is used. It seems like it also has a form of this bug, but I don't think it is exactly the same one.

DrSmugleaf:
I need to check this, but as far as I know this is a mispredict (where the client thinks it should cycle, but the server does not), this should be fixed by checking what code is running in client that the server is not, and making both run the same code.
Otherwise this would remove the ability to cycle entirely, which is not intended.
It might be a little involved to find what the difference is, since you would need to breakpoint all handlers of UseInHandEvent related to guns, but this should also be upstreamed when the fix is found.

geraeumig:
Maybe this happens due to virtual item spawning always failing on the client's side: https://github.com/space-wizards/space-station-14/blob/c79384c5fd0156cea5db6db9576674fab28e534c/Content.Shared/Inventory/VirtualItem/SharedVirtualItemSystem.cs#L219

DrSmugleaf:
> Maybe this happens due to virtual item spawning always failing on the client's side: [space-wizards/space-station-14@`c79384c`/Content.Shared/Inventory/VirtualItem/SharedVirtualItemSystem.cs#L219](https://github.com/space-wizards/space-station-14/blob/c79384c5fd0156cea5db6db9576674fab28e534c/Content.Shared/Inventory/VirtualItem/SharedVirtualItemSystem.cs#L219)

I think so, there's a PR for that here https://github.com/space-wizards/space-station-14/pull/28840

dffdff2423:
Ah, right. I have been sick the past few days or I would have looked into it myself. I am going to close this because it is probably an upstream issue.

PR #2405, opened 2024-06-11
Fixes hypersleep pods

### LINKED ISSUE #2359: No Cryopods (Alamyer)
## Description

Can't use them to cryo, can't climb in. Nothing.

**Screenshots**

image

Siigiil:
## About the PR
Hypersleep chambers can now be used to go into cryo. No shared storage like in CM13 yet, just what we have from base SS14, but I'd say better this than having no option except to go SSD when you need to leave the round.

Closes #2359

## Technical details
Extraneous code removed from the systems that manage the pods — there's already a component for changing an object's sprite if it's full and another component for leaving an object on pressing a movement key.

The actual cryostorage is handled by the same component as in the base game. If we want to use CM13's shared cryo inventory, we'll probably have to either completely reimplement it or figure out a workaround.

The hypersleep chamber slot name was changed from `cm_hyper_sleep` to `cm-hypersleep`, as from what I can tell, the usual convention for this uses dashes.

## Media
pods
podui

**Changelog**
:cl: Sigil
- add: Hypersleep pods now function as cryo pods.

### DIFF
--- a/Content.Client/_CM14/Marines/HyperSleep/HyperSleepChamberSystem.cs
+++ b/Content.Client/_CM14/Marines/HyperSleep/HyperSleepChamberSystem.cs
@@ -5,19 +5,4 @@ namespace Content.Client._CM14.Marines.HyperSleep;
 
 public sealed class HyperSleepChamberSystem : SharedHyperSleepChamberSystem
 {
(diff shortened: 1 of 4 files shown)


DrSmugleaf (approved):
Thanks for the PR! Didn't know about ExitContainerOnMove

PR #2408, opened 2024-06-12
88 mod damage modifier

liltenhead:
## About the PR
Adds a 1.2 damage multiplier to the 88 Mod 4 sidearm.

## Why / Balance
It is this way in CM13, makes it more competitive with the other pistols. 

## Media


DrSmugleaf (approved):
Thanks for the PR!

PR #2426, opened 2024-06-12
Add scoping action

geraeumig:
## About the PR

- Added `Scope` component, which adds an action to use optics.
- Added `Scope` component to `CMM42ASniperRifle`

## Why / Balance

Feature exists in CM13.

Still TODO - not planned in this PR:
 - Ability to toggle between different zoom values
 - Support for bigger zoom values, right now recommended maximum is ~1.25

## Technical details

For toggling partially reused the shield (`BlockingComponent`) code.

Spawns invisible `CMScopingChunkLoader` for PVS subscription to load far chunks.
Reuses currently unused `CameraRecoilComponent.BaseOffset` for client eye offset.
Offset and zoom are set on the server and then autonetworked to the client, to try ensure that chunk loading and perspective setting happen simultaneously. **- Is there a better solution?**

## Media

https://github.com/CM-14/CM-14/assets/171753363/2021947c-5c21-4a8c-b4b5-f86fa0e09796

## Breaking changes

**Changelog**

:cl:
- add: Added ability to scope in with M42A rifle


Content.Server/Camera/CameraRecoilSystem.cs
```diff
@@ -1,16 +1,29 @@
...
+    {
+        base.Update(frameTime);
+
+        var query = AllEntityQuery<EyeComponent, CameraRecoilComponent>();
+        while (query.MoveNext(out var uid, out var eye, out var recoil))
+        {
+            // apparently this is required for aim to work properly
+            _eye.SetOffset(uid, recoil.BaseOffset, eye);
+        }
+    }
```
DrSmugleaf, line 28: I need to look into why this is necessary, but this code should go under a new system in the Content.Server/_CM14 folder, we don't want to modify upstream code wherever possible, specially systems. You could move it to ScopeSystem.

Content.Server/_CM14/Scoping/ScopeSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+            var cardinalVector = xform.LocalRotation.GetCardinalDir().ToVec();
+            var targetOffset = cardinalVector * ((smallestViewpointSize * scopeComponent.Zoom - 1) / 2);
+            cameraRecoilComponent.BaseOffset = targetOffset;
+            Dirty(user, cameraRecoilComponent);
+
+            _eye.SetZoom(user, Vector2.One * scopeComponent.Zoom, true);
+            if (TryComp<ActorComponent>(user, out var actorComp))
+            {
+                // add cardinal vector, until better pvs handling is introduced here
+                var loaderId = Spawn("CMScopingChunkLoader", _transformSystem.GetMapCoordinates(xform).Offset(targetOffset + cardinalVector * 2));
```
DrSmugleaf, line 32: `"CMScopingChunkLoader"` should be made into a field inside ScopeComponent of type EntProtoId. That lets it be changed if necessary in yaml, and also validated to exist as an entity.

Content.Shared/Camera/CameraRecoilComponent.cs
```diff
@@ -13,5 +14,6 @@ public sealed partial class CameraRecoilComponent : Component
     /// <summary>
     ///     Basically I needed a way to chain this effect for the attack lunge animation. Sorry!
     /// </summary>
+    [ViewVariables, AutoNetworkedField]
```
DrSmugleaf, line 17: This isn't used anywhere upstream I think, but if this is needed it should be made as a separate PR upstream. Same with the AutoGenerateComponentState attribute.

Content.Shared/_CM14/Scoping/ScopeComponent.cs
```diff
@@ -0,0 +1,46 @@
...
+    public EntityCoordinates LastScopedAt;
+
+    /// <summary>
+    /// Helper entity, which is spawned to load far chunks
+    /// </summary>
+    [ViewVariables, AutoNetworkedField]
+    public EntityUid? PvsLoader;
+
+    [DataField(customTypeSerializer: typeof(PrototypeIdSerializer<EntityPrototype>))]
+    public string ScopingToggleAction = "CMActionToggleScope";
```
DrSmugleaf, line 42: This can be made of type EntProtoId, then you don't need to specify the custom type serializer and other code can have more information about what this field is supposed to represent. One example would be being able to index it in IPrototypeManager without having to remember to specify EntityPrototype directly.
```suggestion
    [DataField]
    public EntProtoId ScopingToggleAction = "CMActionToggleScope";
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+    [Dependency] private readonly SharedActionsSystem _actionsSystem = default!;
+    [Dependency] private readonly ActionContainerSystem _actionContainer = default!;
+    [Dependency] private readonly SharedTransformSystem _transformSystem = default!;
+    [Dependency] private readonly SharedHandsSystem _handsSystem = default!;
+    [Dependency] private readonly SharedPopupSystem _popupSystem = default!;
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
+    [Dependency] private readonly INetManager _net = default!;
+
+    /// <inheritdoc/>
+    public override void Initialize()
```
DrSmugleaf, line 25: You don't need the <inheritdoc/> here, I think all IDEs should infer it. The doc doesn't have particularly useful information anyways.
```suggestion
    public override void Initialize()
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+    {
+        args.AddAction(ref component.ScopingToggleActionEntity, component.ScopingToggleAction);
+    }
+
+    private void OnToggleAction(EntityUid uid, ScopeComponent component, ToggleActionEvent args)
+    {
+        if (args.Handled)
+            return;
+
+        if (!(_handsSystem.TryGetActiveItem(args.Performer, out var heldItem) && heldItem == uid))
```
DrSmugleaf, line 85: This check can be simplified so that it's easier to understand
```suggestion
        if (!_handsSystem.TryGetActiveItem(args.Performer, out var heldItem) || heldItem != uid)
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+        {
+            var msgError = Loc.GetString("cm-action-popup-scoping-user-must-hold", ("scope", Name(uid)));
+            if (_gameTiming.IsFirstTimePredicted && _gameTiming.InPrediction)
+            {
+                _popupSystem.PopupEntity(msgError, args.Performer, args.Performer);
+            }
```
DrSmugleaf, line 91: This can be replaced with PopupClient. It should do the same check that you are doing here, but check that the popup shows properly in-game.
```suggestion
            _popupSystem.PoupClient(msgError, args.Performer);
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        if (component.IsScoping)
+            StopScoping(uid, component, args.Performer);
+        else
+            StartScoping(uid, component, args.Performer);
+
+        args.Handled = true;
+    }
+
+    private void OnShutdown(EntityUid uid, ScopeComponent component, ComponentShutdown args)
```
DrSmugleaf, line 104: For all of these event handlers you can replace `EntityUid uid, ScopeComponent component, ComponentShutdown args` with `Entity<ScopeComponent> ent, ref ComponentShutdown args` which is the preferred way to handle events.
It couples the component to the entity uid so it doesn't get accidentally passed around with the wrong entity uid and component combination, and it makes you not have to remember if an event is handled by ref or not. See all usages of SubscribeLocalEvent for ones that you can convert.

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+            Dirty(user, scopeUserComp);
+        }
+
+        var msgUser = Loc.GetString("cm-action-popup-scoping-user", ("scope", Name(item)));
+
+        _actionsSystem.SetToggled(component.ScopingToggleActionEntity, true);
+        if (_gameTiming.IsFirstTimePredicted && _gameTiming.InPrediction)
+        {
+            _popupSystem.PopupEntity(msgUser, user, user);
+        }
```
DrSmugleaf, line 130: Same as above.
```suggestion
        _popupSystem.PopupClient(msgUser, user, user);
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        var msgUser = Loc.GetString("cm-action-popup-scoping-stopping-user", ("scope", Name(item)));
+
+        _actionsSystem.SetToggled(component.ScopingToggleActionEntity, false);
+
+        if (_gameTiming.IsFirstTimePredicted && _gameTiming.InPrediction)
+        {
+            _popupSystem.PopupEntity(msgUser, user, user);
+        }
```
DrSmugleaf, line 159:
```suggestion
        _popupSystem.PopupClient(msgUser, user, user);
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -0,0 +1,193 @@
...
+
+        var heldItems = _handsSystem.EnumerateHeld(user, hands).ToArray();
+
+        foreach (var heldItem in heldItems)
+        {
+            if (HasComp<ScopeComponent>(heldItem))
+            {
+                return;
+            }
+        }
```
DrSmugleaf, line 185: This should be able to enumerate the hands directly so that you don't need to allocate a new array.
```suggestion
        foreach (var heldItem in _handsSystem.EnumerateHeld(user, hands))
        {
            if (HasComp<ScopeComponent>(heldItem))
            {
                return;
            }
        }
```

Content.Shared/_CM14/Scoping/SharedScopeSystem.User.cs
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    private void OnMove(EntityUid uid, ScopeUserComponent component, ref MoveEvent args)
+    {
+        if (!TryComp<ScopeComponent>(component.ScopingItem, out var scopeComponent))
+            return;
+
+        if (!_transformSystem.InRange(scopeComponent.LastScopedAt, args.NewPosition, 1.5f))
+            UserStopScoping(uid, component);
+    }
```
DrSmugleaf, line 25: `MoveEvent` is a very expensive event to subscribe to since it will trigger every single time an entity move and make physics slower to tick in general.
If it works, use what SharedWatchXenoSystem uses instead, `MoveInputEvent`.

Content.Shared/_CM14/Scoping/SharedScopeSystem.User.cs
```diff
@@ -0,0 +1,55 @@
...
+
+    private void OnParentChanged(EntityUid uid, ScopeUserComponent component, ref EntParentChangedMessage args)
+    {
+        UserStopScoping(uid, component);
+    }
+
+    private void OnInsertAttempt(EntityUid uid, ScopeUserComponent component, ContainerGettingInsertedAttemptEvent args)
+    {
+        UserStopScoping(uid, component);
+    }
```
DrSmugleaf, line 35: Are these two intended for cases when the rifle is dropped from the user's hands? If so `UnequippedHandEvent` should suffice.
```suggestion
    private void OnParentChanged(EntityUid uid, ScopeUserComponent component, ref EntParentChangedMessage args)
    {
        UserStopScoping(uid, component);
    }

    private void OnInsertAttempt(EntityUid uid, ScopeUserComponent component, ContainerGettingInsertedAttemptEvent args)
    {
        UserStopScoping(uid, component);
    }
```

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Client/_CM14/Scoping/ScopeSystem.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    [Dependency] private readonly ContentEyeSystem _eye = default!;
+
+    protected override void StartScopingCamera(EntityUid user, ScopeComponent scopeComponent)
+    {
+    }
+
+    protected override void StopScopingCamera(EntityUid user, ScopeComponent scopeComponent)
+    {
+        if (TryComp<CameraRecoilComponent>(user, out var cameraRecoilComponent))
```
DrSmugleaf, line 18: This is mostly a style thing but it does matter in other APIs, we usually put the type next to out, instead of as a generic parameter explicitly.
Feel free to change it in other cases if you want to.
```suggestion
        if (TryComp(user, out CameraRecoilComponent? cameraRecoilComponent))
```

DrSmugleaf (changes requested):

Content.Shared/_CM14/Scoping/SharedScopeSystem.User.cs
geraeumig: Those are intended to be triggered when the user is put in a container (felinid) or a box, would only `OnParentChanged` suffice?

geraeumig: Ah I see the code to which this comment was meant, will fix

Content.Server/_CM14/Scoping/ScopeSystem.cs
geraeumig: Thanks for the info, changed

Content.Server/Camera/CameraRecoilSystem.cs
geraeumig: Reverted the upstream changes

Content.Shared/Camera/CameraRecoilComponent.cs
geraeumig: Upstream changes reverted

Content.Shared/_CM14/Scoping/ScopeComponent.cs
geraeumig: Much better, thanks!

Content.Client/_CM14/Scoping/ScopeSystem.cs
geraeumig: Sure, changed here and in other places

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Thanks, updated

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Fixed

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Thanks for info, fixed

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Fixed

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Thanks for the info, changed here too

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
geraeumig: Fixed, thanks!

Content.Shared/_CM14/Scoping/SharedScopeSystem.User.cs
geraeumig: Actually won't anchor the user, as it might have an unintended consequence of players using their scopes to avoid being dragged off - will think of a different solution, for now will just unscope if being dragged.

Content.Server/_CM14/Scoping/ScopeSystem.cs
```diff
@@ -0,0 +1,57 @@
...
+        var scopeToggleEvent = new CMScopeToggleEvent(GetNetEntity(user), targetOffset);
+        RaiseNetworkEvent(scopeToggleEvent, user);
+
+        if (TryComp(user, out ActorComponent? actorComp))
+        {
+            // add cardinal vector, until better pvs handling is introduced here
+            var loaderId = Spawn(scopeComponent.PvsLoaderProto, _transformSystem.GetMapCoordinates(xform).Offset(targetOffset + cardinalVector * 2));
+            scopeComponent.PvsLoader = loaderId;
+            _viewSubscriber.AddViewSubscriber(loaderId, actorComp.PlayerSession);
+        }
```
DrSmugleaf, line 40: I'm looking into this more closely now, I think you should be able to do what "Watch Xenomorph" does to move the player's view:
```csharp
_eye.SetTarget(watcher, watch, watcher);
_viewSubscriber.AddViewSubscriber(watch, player);
```

That should let you avoid having to change eye offsets and having to send a networked event.

DrSmugleaf (changes requested):

DrSmugleaf:
One issue with how this works is that it allows you to see through walls since the FOV is not drawn from your player but from new viewpoint. I have no idea how we would fix this but I will try to figure it out.
https://github.com/CM-14/CM-14/assets/10968691/bbd2630f-1880-4438-82e2-2e98896b79c2

DrSmugleaf:
I have fixed the issue above now with [`4acf4be` (#2426)](https://github.com/CM-14/CM-14/pull/2426/commits/4acf4bedb7a630b76fb0088a2db17457807b55ec), but there is one remaining bug that if positioned correctly, the user can see past their PVS range and not see any entities loaded.

https://github.com/CM-14/CM-14/assets/10968691/f005ac4f-ddd7-4ebd-b41a-d166f619fe28

That should be fixed by https://github.com/space-wizards/RobustToolbox/pull/5228 eventually but for now I'm going to slightly lower the offset until that is fixed.
The zoom value and offset will need to be adjusted once that is fixed anyways, since the real values from CM13 are an 11 tile offset and 24x24 view in 4x, and a 6 tile offset and normal view in 2x. Right now we are doing a mix of both and only one setting.

geraeumig:
Maybe `scoping.EyeOffset = targetOffset;` and `_contentEye.SetZoom(user, Vector2.One * scope.Comp.Zoom, true);` for enabling the scope should be performed on the server's side only, in order to avoid a visual bug on client's side, where unloaded chunk is displayed. 
Presumably the new chunk loads with the same datapacket as the update to offset and zoom variables on the client's side.

Content.Shared/_CM14/Scoping/SharedScopeSystem.cs
```diff
@@ -74,115 +80,92 @@ private void OnToggleAction(Entity<ScopeComponent> ent, ref ToggleActionEvent ar
...
 
-        component.IsScoping = false;
-        Dirty(item, component);
+        var msgUser = Loc.GetString("cm-action-popup-scoping-stopping-user", ("scope", Name(scope.Owner)));
+        _popup.PopupClient(msgUser, user, user);
 
-        return true;
+        _actionsSystem.SetToggled(scope.Comp.ScopingToggleActionEntity, false);
+        _contentEye.ResetZoom(user);
+        _contentEye.SetZoom(user, Vector2.One * scope.Comp.Zoom, true);
```
geraeumig, line 161: `_contentEye.ResetZoom(user);` should be enough.
```suggestion
```

DrSmugleaf: Oops, fixed in [`c91a0f5` (#2426)](https://github.com/CM-14/CM-14/pull/2426/commits/c91a0f51f44bfce6a20a2494f2bedfa1c28b48dd)

PR #2427, opened 2024-06-13
Optimise pheromones

metalgearsloth:
On playtest this was 2x as expensive as physics.
RefreshMoveSpeedModifiers is still pretty expensive and you would need some thread-safe wrapper for it if you wanted that fixed too.

I tested with 25 queens in close proximity 15 running recovery pheromones for 30 seconds. I cbfd getting release working so profile is on debug but it should still be somewhat indicative.

Left before Right after:

image

Profile was before only running it every second which would also have a huge performance impact.

Still want to:
- Change the lookupflags to approx to make it significantly faster.
- ~~Add NextUpdate fields so they don't all run at the same time.~~

### DIFF
--- a/Content.Shared/_CM14/Xenos/Pheromones/SharedXenoPheromonesSystem.cs
+++ b/Content.Shared/_CM14/Xenos/Pheromones/SharedXenoPheromonesSystem.cs
@@ -9,8 +9,12 @@ using Content.Shared.Mobs.Systems;
 using Content.Shared.Movement.Systems;
 using Content.Shared.Popups;
 using Content.Shared.Weapons.Melee.Events;
+using Microsoft.Extensions.ObjectPool;
+using Robust.Shared.Collections;
 using Robust.Shared.GameStates;
+using Robust.Shared.Threading;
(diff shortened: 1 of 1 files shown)


Content.Shared/_CM14/Xenos/Pheromones/SharedXenoPheromonesSystem.cs
```diff
@@ -230,13 +243,30 @@ public override void Update(float frameTime)
                 Dirty(uid, pheromones);
             }
 
-            _receivers.Clear();
-            _entityLookup.GetEntitiesInRange(xform.Coordinates, pheromones.PheromonesRange, _receivers);
```
metalgearsloth, line 234: Okay so I wanted to add like a "NextUpdate" field but it looks like there's already one for plasma so we could just hook into that if that's okay smug?

DrSmugleaf: Sur

DrSmugleaf (approved):
banger pr ty

PR #2477, opened 2024-06-15
thirst reduces at the same rate as hunger

### LINKED ISSUE #2439: Humanoids should get thirsty slower or smaller amounts of water should fill more thirst (We have water canteens)

QuietlyWhisper:
This should make thirst occur 10x slower than normal, not sure how much it will actually affect but there's no thirst in cm anyways so I'm not very concerned.

- fix #2439 

**Changelog**

:cl: Whisper

- tweak: Greatly reduced the speed thirst occurs

### DIFF
--- a/Resources/Prototypes/_CM14/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/_CM14/Entities/Mobs/Species/base.yml
@@ -62,7 +62,7 @@
   - type: Marine
   - type: Hunger
   - type: Thirst
-    baseDecayRate: 0.01666666666f
+    baseDecayRate: 0.01666666666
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/_CM14/Entities/Mobs/Species/base.yml
```diff
@@ -62,6 +62,7 @@
   - type: Marine
   - type: Hunger
   - type: Thirst
+    baseDecayRate: 0.01666666666f
```
QuietlyWhisper, line 65:
```suggestion
    baseDecayRate: 0.01666666666
```

PR #2478, opened 2024-06-15
Run pheromones receivers every tick

metalgearsloth:
Fixes https://github.com/CM-14/CM-14/issues/2361

Oversight on my part. Given we only update the lookup on every plasma usage we'll cache the receivers and re-run them every tick.


Content.Shared/_CM14/Xenos/Pheromones/SharedXenoPheromonesSystem.cs
```diff
@@ -227,14 +227,18 @@ public override void Update(float frameTime)
...
 
         var query = EntityQueryEnumerator<XenoActivePheromonesComponent, XenoPheromonesComponent, TransformComponent>();
+        _pheromonesJob.Receivers.Clear();
         _pheromonesJob.Pheromones.Clear();
 
         while (query.MoveNext(out var uid, out var active, out var pheromones, out var xform))
         {
+            _pheromonesJob.Pheromones.Add((uid, active, pheromones, xform));
+
             // We'll only update pheromones receivers whenever plasma gets used.
```
metalgearsloth, line 237: You know I realise if it's remcompdeferred this should be lower so have to fix that.

DrSmugleaf: Wouldn't it be possible to add it prematurely and then do RemoveAt if it's supposed to be removed? In line 248.

PR #2530, opened 2024-06-17
Added a shared empty layer state for storages

amo020:
## About the PR
* Added a shared empty layer state for storages
* Set grenades packets prototypes to use empty layers

## Media
Recording

https://github.com/CM-14/CM-14/assets/167775671/28bf82ab-da9a-4c61-b2d7-e13fe921cef0

This is the minor fix for https://github.com/CM-14/CM-14/issues/2411

## Technical detail

This might have been done more elegantly by expanding the `BagOpenVisuals` and setting it to a `CMStorageVisuals` but this is my first C# change and wanted to make it simple.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
@@ -39,23 +39,23 @@ namespace Content.Shared.Storage.EntitySystems;
 
 public abstract class SharedStorageSystem : EntitySystem
 {
-    [Dependency] private   readonly IPrototypeManager _prototype = default!;
+    [Dependency] private readonly IPrototypeManager _prototype = default!;
     [Dependency] protected readonly IRobustRandom Random = default!;
     [Dependency] protected readonly ActionBlockerSystem ActionBlocker = default!;
-    [Dependency] private   readonly EntityLookupSystem _entityLookupSystem = default!;
-    [Dependency] private   readonly SharedAppearanceSystem _appearance = default!;
+    [Dependency] private readonly EntityLookupSystem _entityLookupSystem = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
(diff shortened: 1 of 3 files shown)


Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
```diff
@@ -39,23 +39,23 @@ namespace Content.Shared.Storage.EntitySystems;
...
+    [Dependency] private readonly InventorySystem _inventory = default!;
     [Dependency] protected readonly SharedItemSystem ItemSystem = default!;
-    [Dependency] private   readonly SharedPopupSystem _popupSystem = default!;
-    [Dependency] private   readonly SharedHandsSystem _sharedHandsSystem = default!;
-    [Dependency] private   readonly SharedStackSystem _stack = default!;
+    [Dependency] private readonly SharedPopupSystem _popupSystem = default!;
+    [Dependency] private readonly SharedHandsSystem _sharedHandsSystem = default!;
+    [Dependency] private readonly SharedStackSystem _stack = default!;
     [Dependency] protected readonly SharedTransformSystem TransformSystem = default!;
-    [Dependency] private   readonly SharedUserInterfaceSystem _ui = default!;
```
DrSmugleaf, line 58: Reformatting changes like these are fine for our code but should not be done on upstream code since it will cause merge conflicts when we update from upstream.

DrSmugleaf (changes requested):
I need to look into this more since I think this could be upstreamed (in which case we have both a PR here, and one upstream with the same code changes)

Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
amo020: A yes.
It triggered me because one was not indented like the others. i'll revert that.

ShadowCommander:
This can probably be done with the current sprite layers similar to #2747

PR #2543, opened 2024-06-19
Fix rifle and smg magazines

Tunguso4ka:
## About the PR

It was impossible to insert AP and Extended magazines.
(Tag component removes parent`s Tags)


QuietlyWhisper (changes requested):
Guns need to have whitelisted tag to accept the different mags, not other way around, it's done like this so you can't put an `AP` mag into a `regular` ammo box

QuietlyWhisper (approved):
Thank you! The same bug was with magazine boxes and apparently I didn't notice that so that was fixed too with this pr

PR #2553, opened 2024-06-21
Add xeno infection stages & popups

### LINKED ISSUE #2089: Infected marines should get popups and other effects as the infection progresses

Vermidia:
## About the PR

Adds xeno infection stages and symptoms to marines. 

## Why / Balance

Fixes #2089

## Technical details

In SharedXenoParasiteSystem in the update loop, it now assigns a stage to the VictimInfectedComponent, using some math to get the percentage of how close the victim is to burst * how many stages of infection there are (I put this at 5). If the current stage is greater than or equal a defined stage (with the more severe ones taking priority) it'll roll for symptoms, which can cause marines to shake and get knocked down, scream, sneeze, take a little damage, and usually pops up a message, etc.

Changes the xeno HUD to use the same stages defined in the infection stages.

I tried to replicate the chances of said actions occurring in CM13, but most of the values likely need adjusting.

## Media

https://github.com/CM-14/CM-14/assets/32827189/9c8a6e61-a44d-4175-9156-b0a6d5581b02

## Breaking changes

probably not?
**Changelog**

:cl:
- add: Marines will now experience symptoms of being infected.


Content.Shared/_RMC14/Xenonids/Parasite/VictimInfectedComponent.cs
```diff
@@ -71,4 +73,55 @@ public sealed partial class VictimInfectedComponent : Component
...
+    public ProtoId<EmotePrototype> SneezeId = "Sneeze";
+
+    [DataField, AutoNetworkedField]
+    public ProtoId<EmotePrototype> CoughId = "Cough";
+
+    [DataField, AutoNetworkedField]
+    public ProtoId<EmotePrototype> ScreamId = "Scream";
+
+    [DataField, AutoNetworkedField]
+    public DamageSpecifier InfectionDamage = new() { DamageDict = new() { { "Blunt", 1 } } };
```
DrSmugleaf, line 126: These fields should be moved to a separate component or not be networked, otherwise when Dirty gets called or this component gets networked all of these fields will be sent to the client every time, making the component state very large.

Vermidia:
Unnetworked everything new but current stage since that may be used for visuals later

QuietlyWhisper (changes requested):
The lower stages popups seem extremely rare in testing, and it would be nice if it was very apparent someone was about to burst, like constantly shaking during the last minute or so until the burst happens. 
Ideally you want a message to always appear at certain times in case by chance you never get a feedback about infection.

Vermidia:
https://github.com/RMC-14/RMC-14/assets/32827189/fbbc58f6-1aa2-4111-bb09-77bd3da9d134

Increased the chance of middling symptoms (and the muscle ache especially), as well as increase the shakes chance slightly.

Also implemented the suggestion of always getting a message right before bursting (here, on the stage right before the final one)

Content.Shared/_RMC14/Xenonids/Parasite/VictimInfectedEmoteEvent.cs
```diff
@@ -0,0 +1,6 @@
+using Content.Shared.Chat.Prototypes;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared._RMC14.Xenonids.Parasite;
+
+public record struct VictimInfectedEmoteEvent(ProtoId<EmotePrototype> Emote);
```
DrSmugleaf, line 6: I missed this on the first review sorry, this needs [ByRefEvent] added and then you can raise it by ref so it doesn't get copied (for performance).

Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -361,17 +370,102 @@ public override void Update(float frameTime)
...
+                if (stage >= infected.FinalSymptomsStart)
+                {
+                    if (_random.Prob(infected.MajorPainChance * frameTime))
+                    {
+                        var message = Loc.GetString("cm-xeno-infection-majorpain-" + _random.Pick(new List<string> { "chest", "breathing", "heart" }));
+                        _popup.PopupEntity(message, uid, uid, PopupType.SmallCaution);
+                        if (_random.Prob(0.5f))
+                        {
+                            var ev = new VictimInfectedEmoteEvent(infected.ScreamId);
+                            RaiseLocalEvent(uid, ev);
```
DrSmugleaf, line 401: This is where you need to change it to be raised by ref with the change above.
```suggestion
                            RaiseLocalEvent(uid, ref ev);
```

Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -361,17 +370,102 @@ public override void Update(float frameTime)
...
+                    {
+                        _popup.PopupEntity(Loc.GetString("cm-xeno-infection-muscle-ache"), uid, PopupType.SmallCaution);
+                        if (_random.Prob(0.2f))
+                            _damage.TryChangeDamage(uid, infected.InfectionDamage, true, false);
+                    }
+                    else if (_random.Prob(infected.SneezeCoughChance * frameTime))
+                    {
+                        var emote = _random.Pick(new List<ProtoId<EmotePrototype>> { infected.SneezeId, infected.CoughId });
+                        var ev = new VictimInfectedEmoteEvent(emote);
+                        RaiseLocalEvent(uid, ev);
```
DrSmugleaf, line 426: Same here.
```suggestion
                        RaiseLocalEvent(uid, ref ev);
```

DrSmugleaf (changes requested):

DrSmugleaf (approved):
Thanks for the PR!

PR #2565, opened 2024-06-23
Temporary add headsets to med personnel

Tunguso4ka:
## About the PR

Title, temporary adds headsets to med cryo spawn points.


QuietlyWhisper (approved):
Thanks!

PR #2571, opened 2024-06-23
Adds integrated attachments, fixes a few bugs with attachments.

Siigiil:
## About the PR
Added support for integrated and roundstart attachments. None are currently in the game. All you need to add a roundstart attachment is to add it in a `ContainerFillComponent`.

Integrated attachments are made by adding one as a roundstart attachment and setting the slot's `Locked` field to true.

Fixed toggleable attachments that are only supposed to work when attached not locking out properly.

Fixed actions not being removed when swapping attachments.

Fixed underbarrel shotgun offset.

## Technical details
Damage modifiers have been moved to `AttachableData.cs` from `AttachableUI.cs`, as they have nothing to do with any UIs.

The slot dictionary in `AttachableHolderComponent` now uses a new class: `AttachableSlot`, instead of a whitelist. The new class contains the whitelist from before and the `Locked` field, which prevents the slot's contents from being modified. This is used for integrated attachments.

`CMAttachableX` in all attachment tags was replaced with `CMAttachmentX`, as the tags should be consistent with the items' IDs.

## Media
Attachment offsets fixed for the new main rifle:
attachments

**Changelog**

:cl:
- fix: Fixed attachment actions not being properly removed.
- fix: Fixed the attachment offsets for the M54.


Content.Shared/_RMC14/Attachable/AttachableHolderSystem.cs
```diff
@@ -263,7 +265,7 @@ private void OnAttachDoAfter(EntityUid uid, AttachableHolderComponent component,
         if (!TryComp(args.Target, out AttachableHolderComponent? holder) ||
             !HasComp<AttachableComponent>(args.Used))
             return;
-
+
```
DrSmugleaf, line 268:
```suggestion

```

Content.Shared/_RMC14/Attachable/AttachableHolderSystem.cs
```diff
@@ -360,7 +372,7 @@ private void OnDetachDoAfter(EntityUid uid, AttachableHolderComponent component,
 
         if (string.IsNullOrEmpty(slotId))
             CanAttach(holder, attachableUid, ref slotId);
-
+
```
DrSmugleaf, line 375:
```suggestion

```

Content.Shared/_RMC14/Attachable/AttachableHolderSystem.cs
```diff
@@ -279,38 +281,48 @@ private void OnAttachDoAfter(EntityUid uid, AttachableHolderComponent component,
...
+        RaiseLocalEvent(holder, ref holderEv);
+        
+        Dirty(holder);
+    }
 
     //Detaching
     public void StartDetach(Entity<AttachableHolderComponent> holder, string slotId, EntityUid userUid)
     {
-        if (TryGetAttachable(holder, slotId, out var attachable))
+        if (TryGetAttachable(holder, slotId, out var attachable) && !holder.Comp.Slots[slotId].Locked)
```
DrSmugleaf, line 325: Is Slots guaranteed to have a slot with slotId here? Feels like it should be a try get instead specially since this is a public method.

DrSmugleaf (changes requested):
For this I think we should keep the -a just to keep it the same as 13's, even if it is inconsistent with our style, just to make future future importing easier. The rest of the changes look good though.

Content.Shared/_RMC14/Attachable/AttachableHolderSystem.cs
Siigiil: It is *not*. Good catch, I completely missed this somehow. Fixed now.

Content.Shared/_RMC14/Attachable/AttachableHolderSystem.cs
Siigiil: I have no idea how these got there, but they should be gone now.

Content.Shared/_RMC14/Attachable/AttachableData.cs
```diff
@@ -0,0 +1,53 @@
...
+
+    [DataField]
+    public float FireRate = 1.0f;
+
+    [DataField]
+    public float ProjectileSpeedFlat = 0;
+
+    [DataField]
+    public float ProjectileSpeedMultiplier = 1.0f;
+}
```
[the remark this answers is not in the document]
Siigiil: I'll do it in the final attachment PR. Though this will likely be on Sunday, as I will *probably* be too busy to do much coding tomorrow, and *definitely* will be on Friday and Saturday.

DrSmugleaf (approved):

PR #2595, opened 2024-06-24
fix(SharedIvDripSystem): Fix blood bags from spamming when out of range.

CaasGit:
## About the PR

Blood bags will spam the screen with text if you click on a target that is out of range. Move the check up a bit as suggested in the past to ensure you're within range of the target before popping up the text.

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Blood bags do not print text when you're clearly out of range.


DrSmugleaf (approved):
Thanks for the PR!

PR #2611, opened 2024-06-24
Move faster in lighter armors and slower in heavier armors

krose124:
## About the PR
Change speed values for Light, Medium, and Heavy marine armors.

## Why / Balance
The purpose is to increase the speed difference to be more noticeable. I think, armor selection should have trade-offs such that heavy armors have better armor values but move more slowly. While this was already implemented, in game the difference in speed felt negligible. As a result, I propose increasing the difference in speed multipliers in armor.

## Technical details

Yaml edits on the speed modifiers, I also explicitly added the speed modifier on the light armor to be 1. I assume if absent it would take the speed multiplier from the parent.

## Media

I cannot run SS13 to check what the speed differences are in CM13, in addition the local server does not seem to work for me. Although, the latter is probably just a skill issue on my part. I think a video comparison should be shown of the speed of the character whilst wearing armor that changes speed.

I think this change should require a showcase, I am unable to produce one at this time.

## Breaking changes

This only changes one yaml file, I do not think anything should break.

**Changelog**

:cl:
- tweak: Parent marine armor (the medium one) speed mult .9 -> .85
- tweak: EOD armor (the heavy one) speed mult .8 -> .75
- add: Explicit speed value on M3-L (the light one) speed multiplier as 1

While, I like the idea behind this I am not certain about it. I would like for this to be used as a way for us to debate this topic of armor movement speed. In addition, I realize this is super low priority, but given my busy schedule "I'm doing my part!" as they say.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
@@ -103,9 +103,7 @@
   - type: CMArmor
     armor: 25
     explosionArmor: 20
-  - type: ClothingSpeedModifier
-    walkModifier: 1
-    sprintModifier: 1
+  
 
 - type: entity
   parent: CMArmorM3Medium
@@ -193,6 +191,9 @@
   - type: CMArmor
     armor: 15
     explosionArmor: 20
+  - type: ClothingSpeedModifier
+    walkModifier: 1
+    sprintModifier: 1
   - type: Storage
     grid:
     - 0,0,3,1 # 2 slots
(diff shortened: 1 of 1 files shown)


krose124:
those last two commits are me suffering with git and github

DrSmugleaf (changes requested):
Testing this in cmss13 it seems like the intended multipliers are 0.851 for light armor, 0.784 for medium armor and 0.727 for heavy armor, but the actual values are 0.814 0.68 0.68, making medium and heavy the same.

For now set the multipliers of light to 0.814, medium to 0.727 and heavy to 0.68, to use the actual values you get through gameplay without making the medium armor worse.

Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
```diff
@@ -103,6 +103,7 @@
   - type: CMArmor
     armor: 25
     explosionArmor: 20
+
```
QuietlyWhisper, line 106:
```suggestion
```
extra newline

krose124: This should be resolved

QuietlyWhisper (approved):
Should be good thank you

PR #2615, opened 2024-06-25
fix(SharedIVDripSystem): Overall clean up and updates to Blood Bag.

CaasGit:
## About the PR
This now check to see if the target for the blood bag has a MarineComponent as used elsewhere. There's also some Rider suggestions cleaned up. You should not now attempt to poke random objects or have issues tossing blood bags into disposals.

## Media

- [ x ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:
- fix: blood bags no longer attempt to poke random objects

### DIFF
--- a/Content.Shared/_RMC14/Medical/IV/SharedIVDripSystem.cs
+++ b/Content.Shared/_RMC14/Medical/IV/SharedIVDripSystem.cs
@@ -79,7 +79,7 @@ public abstract class SharedIVDripSystem : EntitySystem
         UpdateIVAppearance(iv);
     }
 
-    private static void OnIVDripCanDrag(Entity<IVDripComponent> iv, ref CanDragEvent args)
+    private void OnIVDripCanDrag(Entity<IVDripComponent> iv, ref CanDragEvent args)
     {
         args.Handled = true;
     }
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Medical/IV/SharedIVDripSystem.cs
```diff
@@ -79,30 +79,28 @@ private void OnIVDripAfterHandleState(Entity<IVDripComponent> iv, ref AfterAutoH
         UpdateIVAppearance(iv);
     }
 
-    private void OnIVDripCanDrag(Entity<IVDripComponent> iv, ref CanDragEvent args)
+    private static void OnIVDripCanDrag(Entity<IVDripComponent> iv, ref CanDragEvent args)
```
DrSmugleaf, line 82: Rider might have suggested this since there's no instance data being used, but by convention we keep all methods that are handling events non-static
```suggestion
    private void OnIVDripCanDrag(Entity<IVDripComponent> iv, ref CanDragEvent args)
```

DrSmugleaf (changes requested):

Content.Shared/_RMC14/Medical/IV/SharedIVDripSystem.cs
DrSmugleaf: No need to include the ignore comment, Rider shouldn't be showing you that anyways if it's properly importing the project's .editorconfig

DrSmugleaf (approved):

PR #2618, opened 2024-06-25
Fixes grenade launcher projectiles not passing over objects and creatures.

Siigiil:
## About the PR
The two GLs we have implemented now properly launch their grenades *over* the objects between the shooter and the target.

## Why / Balance
Because bug. And I need this to finish implementing attachments.

## Media
arc

**Changelog**

:cl: Sigil
- fix: The grenade launcher now properly fires over creatures and objects.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
@@ -1,4 +1,7 @@
-﻿using Robust.Shared.GameStates;
+﻿using Content.Shared.Physics;
+using Robust.Shared.GameStates;
+using Robust.Shared.Physics.Dynamics;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
 
 namespace Content.Shared._RMC14.Weapons.Ranged;
 
@@ -7,8 +10,8 @@ namespace Content.Shared._RMC14.Weapons.Ranged;
 public sealed partial class AmmoFixedDistanceComponent : Component
 {
     /// <summary>
-    /// If set to true, this makes shots from the weapon pass over creatures and objects.
+    /// This will be applied to the projectiles fired by the weapon.
     /// </summary>
-    [DataField, AutoNetworkedField]
-    public bool HighArc = false;
+    [DataField(customTypeSerializer: typeof(FlagSerializer<CollisionLayer>)), AutoNetworkedField]
+    public int CollisionLayer = (int) CollisionGroup.ThrownItem;
 }
--- a/Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
@@ -1,7 +1,4 @@
-﻿using Content.Shared.Physics;
-using Robust.Shared.GameStates;
-using Robust.Shared.Physics.Dynamics;
-using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+﻿using Robust.Shared.GameStates;
 
 namespace Content.Shared._RMC14.Weapons.Ranged;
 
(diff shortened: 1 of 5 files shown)


Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
```diff
@@ -2,6 +2,13 @@
...
 [Access(typeof(CMGunSystem))]
-public sealed partial class AmmoFixedDistanceComponent : Component;
+public sealed partial class AmmoFixedDistanceComponent : Component
+{
+    /// <summary>
+    /// If set to true, this makes shots from the weapon pass over creatures and objects.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool HighArc = false;
+}
```
metalgearsloth, line 14: How come this can't use a collision layer?

QuietlyWhisper (dismissed):
.

Content.Shared/_RMC14/Weapons/Ranged/CMGunSystem.cs
```diff
@@ -97,9 +97,16 @@ private void OnAmmoFixedDistanceShot(Entity<AmmoFixedDistanceComponent> ent, ref
...
             comp.FlyEndTime = time + TimeSpan.FromSeconds(direction.Length() / gun.ProjectileSpeedModified);
+            comp.HighArc = ent.Comp.HighArc;
         }
     }
 
+    private void OnStartCollide(Entity<ProjectileFixedDistanceComponent> ent, ref StartCollideEvent args)
+    {
+        if (!ent.Comp.HighArc)
+            StopProjectile(ent);
+    }
```
DrSmugleaf, line 108: This makes the projectile stop when it starts to collide, not pass over objects and structures.
This isn't a property of the grenade launcher itself either, it's a property of the grenade itself when thrown, so the grenades just need their collision changed in YAML.

DrSmugleaf (changes requested):

Content.Shared/_RMC14/Weapons/Ranged/AmmoFixedDistanceComponent.cs
Siigiil: Good question. Probably because it's thirty degrees in here and my brain is overheating. Fixing now.

Content.Shared/_RMC14/Weapons/Ranged/CMGunSystem.cs
DrSmugleaf: Also the grenades already don't collide with mobs, but this doesn't change that either.

Siigiil: Just tested again on master. They do collide with mobs.
Pretty sure setting the collision in the grenade YAML will make *all* thrown grenades pass over objects. Which isn't what we want, since normal throws are supposed to collide with obstacles. High-toss has a delay, a limited range and isn't an action we've implemented yet.

Changed it to set the grenade's collision layer when it's fired instead.

DrSmugleaf: Oh ok I see now, I was confused with the PR name and code. The fixture collision layer change is not doing what you want though, as it is already set to 0.
You were probably looking for the collision mask, but even then, you shoud be grabbing the fixture that you want by name instead of the first one.
Is there even any case where grenade launcher grenades should stop on collision though? If not, just removing the StartCollideEvent subscription like you did and reverting all other code will work like you want.

DrSmugleaf: Looking into this more I see why you did it that way originally, but I don't think the AmmoFixedComponent is doing anything on the grenade prototype itself. So you can just remove it from there, keep the StartCollideEvent subscriber removed, and revert all other code to get what you want.

Siigiil: Changed according to the latest comment. Didn't even notice grenades had AmmoFixedComponent. Why that was the case is beyond me.

I'm confused why my previous solution was bad. The point was to have a datafield you can change on the launcher to either make the grenades fired from it pass over objects or impact them, in case we ever get GLs that fire straight ahead.

The default value was set so that when it was applied to the projectile, it would impact people and tall objects. From my testing, it worked pretty much perfectly.

QuietlyWhisper (approved):
Thanks

PR #2619, opened 2024-06-25
Perspective stairs, handrails, platforms, and more

Cu1r:
## About the PR
Added quite a few mapping features related to elevation changes

## Why / Balance
Necessary for all the maps we have in the works right now

## Media
platforms blahh

## Breaking changes
The small platform corner piece under entity CMPlatformCorner has been moved to CMPlatformCornerSmall (to fit with railing naming conventions) and has been replaced with an L-shaped platform piece. The small corner piece also had a few rotational and collision issues that needed to be addressed.
No maps in the repo currently use CMPlatformCorner, however Solaris Ridge is up for review right now and does. This will not make Solaris Ridge unloadable (since CMPlatformCorner is still in use), however those pieces will need to be remapped with CMPlatformCornerSmall. Even without this change the rotation of those elements would have needed to be remapped since they were ported weirdly.


QuietlyWhisper (changes requested):
Change IDs from `CM..` to `RMC..`

Also I'm not sure if stairs are supposed to be destructible? Was that a mistake or intended?

QuietlyWhisper (approved):
Thanks

diraven:
> Change IDs from `CM..` to `RMC..`
> 
> Also I'm not sure if stairs are supposed to be destructible? Was that a mistake or intended?

Apparently, [that](https://github.com/RMC-14/RMC-14/pull/2619/commits/731f800831531300015f372b56e36a507dacf262#diff-5ea39aba9bd8051fd68652abd7b8d32de03587dbd722af04460cff13931817b1L15) also made some railings indestructible: https://github.com/RMC-14/RMC-14/issues/3351

PR #2630, opened 2024-06-25
Disable collision check for folding barricades

### LINKED ISSUE #2366: You can't close cades if there's a barricade or a person on the same tile

dffdff2423:
## About the PR

Fixes #2366

## Why / Balance

I assume this is the intended behavior based on the above issue being on the
roadmap list. If it is not let me know and I am happy to make a different fix.

When I tested this there is not really anything too strange that happens
because of it. The worst that can happen is you are "snapped" to above or below
the barricade when it closes and you need to be on a very particular spot for
this to happen.

## Technical details

One-Line YAML change

## Media

## Breaking changes

**Changelog**

:cl: Aquif
- fix: Folding barricades can now be closed while there is people on the same tile as them.


DrSmugleaf (approved):
Thanks for the PR!

Issue #901, opened 2024-01-15
Imidazoline


DrSmugleaf:

DrSmugleaf:
https://github.com/RMC-14/RMC-14/pull/2642

PR #2642, opened 2024-06-26
Port Meralyne, Imidazoline and Clonexadone from CM13.

Zarq0:
## About the PR

Ports #893, #901 and #905 from CM13.

## Media

**Changelog**

:cl:
- add: Ported Meralyne, Imidazoline and Clonexadone from CM13.

### DIFF
--- a/Resources/Prototypes/_RMC14/Reagents/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Reagents/medicine.yml
@@ -498,24 +498,26 @@
       effects:
       - !type:EqualHealthChange
         damage:
-        - Brute: -1.5
+        - Brute: -0.75
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
           min: 15
         damage:
           groups:
-            Burn: 1.5
+            Burn: 0.375
       - !type:HealthChange
         conditions:
         - !type:ReagentThreshold
           min: 25
         damage:
           groups:
-            Burn: 9
-            Toxin: 3
+            Burn: 3.75
+            Toxin: 1.5
 
 # TODO RMC14 other effects
+# TODO RMC14 progressive eye healing
+# TODO RMC14 brain damage over 50u
 - type: reagent
   parent: [ CMReagent, CMDylovene ]
   id: CMImidazoline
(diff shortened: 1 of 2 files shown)


Resources/Prototypes/_RMC14/Reagents/medicine.yml
```diff
@@ -484,3 +484,82 @@
...
+      metabolismRate: 0.2
+      effects:
+      - !type:EqualHealthChange
+        conditions:
+        - !type:Temperature
+          max: 170
+        damage:
+        - Toxin: -6
+        - Brute: -6
+        - Burn: -6
```
DrSmugleaf, line 565: This one sucks to calculate but, cryometabolizing level 6 (potency 3) means it will boost reagents by half of the level, so 3.
Neogenetic 3 is (3+3) = 6 levels and 3 potency, so twice Meralyne, which is -3 Brute, however this also passes the extra threshold check of >2, adding half its potency as extra brute healing half as often, so -3.75.
Anticorrosive 3, 6 levels 3 potency, so 3x Kelotane, which is -1.5, and same calculations, so -2.25.
Antitoxic 3, 6 levels 3 potency, 3x Dylovene -3, no extra healing step this time.
```suggestion
        damage:
        - Toxin: -3.75
        - Brute: -2.25
        - Burn: -3 // TODO RMC14 heal cloneloss
```

DrSmugleaf (changes requested):
Easiest way to test that these numbers are correct 100% is testing it in a local development server in ss13, I haven't had the time to yet but I think these should be correct

Zarq0:
I added the comments to the top of the reagents due to the engine not liking them next to damage and healing values.

Siigiil:
The metabolism rates are wrong. SS13 life ticks happen once every two seconds. SS14 metabolism cycles happen once every second. So a metabolism rate of 0,2 here is twice as fast as in CM13.

Zarq0:
I just copied the metabolism rates from the SS13 ones, had no idea that it was the case. I'll work on it.

Zarq0:
Just checked and that is indeed the case, thanks for letting me know.

DrSmugleaf:
> I added the comments to the top of the reagents due to the engine not liking them next to damage and healing values.

Oh oops that's because I gave you the wrong kind of comment for YAML, with // instead of #

Zarq0:
> Oh oops that's because I gave you the wrong kind of comment for YAML, with // instead of #

Alright, I'll keep it in mind for the future, I have made the changes requested on the review.

QuietlyWhisper (approved):
Thanks

PR #2658, opened 2024-06-28
Medicine icon labels and associated logic

### LINKED ISSUE #2162: Autoinjectors should show a text prefix for their chemical on their icon while held

blueDev2:
## About the PR

Add icons to medical containers for both within an inventory and in slots (including hands)
Fixes #2162

## Media

image

## Breaking changes

**Changelog**

:cl: blueDev2
- add: Medical containers now have prefix icons


Content.Client/UserInterface/Systems/Storage/Controls/ItemGridPiece.cs
```diff
@@ -136,12 +143,12 @@ protected override void Draw(DrawingHandleScreen handle)
                 var offset = size * 2 * new Vector2(x - boundingGrid.Left, y - boundingGrid.Bottom);
                 var topLeft = PixelPosition + offset.Floored();
 
-                if (GetTexture(adjustedShape, new Vector2i(x, y), Direction.NorthEast) is {} neTexture)
```
DrSmugleaf, line 139: These formatting fixes should not be made, at least not here downstream

DrSmugleaf (changes requested):
Thanks for the PR!
The changes to SlotControl and ItemGridPiece should be upstreamed in case upstream wants them, but we could merge these here in the meantime. I will check this more after we release today

Content.Client/UserInterface/Systems/Storage/Controls/ItemGridPiece.cs
blueDev2: I don't recall ever doing anything with formatting, but I did do some copying and pasting since I was working on this pre-RMC renaming

blueDev2:
> Thanks for the PR! The changes to SlotControl and ItemGridPiece should be upstreamed in case upstream wants them, but we could merge these here in the meantime. I will check this more after we release today

I was thinking about it, but I couldn't really imagine a situation where default ss-14 would want these labels
Like the main point of the labels is that RMC can get pretty hectic and a fast way to see what each med is can be crucial

But the regular ss-14 is more chill

DrSmugleaf:
I'll check this again tomorrow, you're probably right upstream has no use for it but I'll ask just in case

Issue #2586, opened 2024-06-23
We 100% need a cooldown on sound emotes


Skarletto:
brief is just YawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawnsYawns because marines think it's HILARIOUS to spam sounds

my brain

Dutch-VanDerLinde:
It's done

PR #2660, opened 2024-06-28
New MP Whistle now mutes anyone in an area who can't give orders

Vermidia:
## About the PR

This new red whistle will now mute anyone (except the user) in an area (radius 5, same as it's whistle range) who cannot give orders and is a marine.

## Why / Balance

Mutes emote spammers and chatting, allowing briefings to be a little more peaceful. Kinda tackles the problem in #2586 but not quite in the way requested.

The full suggestion also would've made it not effect MP but they currently don't have any unique components so it currently does effect them.

## Technical details

Add new system ApplyStatusToAreaonActivationSystem which currently only activates when something is used inhand. Applies a status in an area. Has blacklist and whitelist functionality.

## Media

https://github.com/RMC-14/RMC-14/assets/32827189/6b93c528-fe8b-4c19-bb00-5d05980d35f3

## Breaking changes

Nah

**Changelog**

:cl:
- add: MP Whistle will now mute marines who hear them, except for anyone who can give orders.

### DIFF
--- a/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
+++ b/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
@@ -1,0 +1,18 @@
+
+namespace Content.Server._RMC14.Abilities;
+
+[RegisterComponent]
+public sealed partial class ApplyStatusToAreaOnActivateComponent : Component
+{
+    [DataField]
+    public string Key = default!;
+
+    [DataField]
+    public string Component = String.Empty;
+
+    [DataField]
+    public float Time = 1.0f; //In Seconds
+
+    [DataField]
+    public float Range = 1;
+}
--- a/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
+++ b/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
@@ -21,7 +21,7 @@ public sealed partial class ApplyStatusToAreaOnActivateComponent : Component
     public EntityWhitelist? Blacklist = null;
 
     [DataField]
-    public EntityWhitelist? WhiteList = null;
+    public EntityWhitelist? Whitelist = null;
 
     [DataField]
     public bool ApplyToSelf = false;
--- a/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
+++ b/Content.Server/_RMC14/Abilities/ApplyStatusToAreaOnActivateComponent.cs
@@ -1,3 +1,4 @@
+using Content.Shared.Whitelist;
 
 namespace Content.Server._RMC14.Abilities;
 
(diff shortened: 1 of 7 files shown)


skypatrol21:
I use the whistle a lot to get attention and give orders both as FTL and SL while in combat. I feel like this would just make the item useless in combat. I think spammers is an admin issue that should be addressed with a boink, especially since the server is suppose to be mrp. Muting marines in combat might prevent important conversations from happening while not in yet, I can already see a JTAC being muted by the whistle and delaying an airstrike or mortar barrage by precious seconds.

Vermidia:
The main problem is there can be a few spammers and they can be hard to manage. I could make the range or duration much shorter, it's mostly supposed to be a quick attention grabber.

Vermidia:
Ok, made a specific MP whistle that's red and does the muting effect. For now, all MP roles will start with one on their necks, but it should probably move into the vendor when that's done.

QuietlyWhisper:
Smug wants to think about it, but this can also go well with policing skill for the mp which is not done yet!

QuietlyWhisper:
Oh actually you're probably right, the skill is there it's just not used for anything

Vermidia:
Ok, added skill support to the component (not in the prettiest way tbh, UseInHandEvent isn't great) that'll only have the status effect go off if the user has 2> police skill.

QuietlyWhisper:
Can you provide more examples of use case

skypatrol21:
As an MP whistle, this is actually a really cool idea mb for not understanding it at first.

Vermidia:
Other than silencing people during briefing, it'd ideally probably be used to stop arguments or rioters for a brief amount of time, at least enough for the MP to try to get a word in.
You wouldn't really use it when trying to arrest someone because you probably need them to speak their case.

QuietlyWhisper:
Can you ensure this does not affect other MP or the user?

QuietlyWhisper:
I'm trying to test this but the effect doesn't appear to work

Vermidia:
Yeah I broke it for a bit. Atm it works off a whitelist and blacklist. I can try to have a "skill blacklist" too.

Skarletto:
I don't think emote spammers should be an MP issue
Interesting item, but I don't think giving MPs an item that can essentially mute RP is a good idea. I of course don't believe emote spammers are RP, but that shouldn't be fixed by an ingame item. I don't see any reason to mute arguments if people are having them at all.
Mutinies are rare, and usually those are dealt with using violence, not words.

DrSmugleaf (changes requested):
Same opinion as Skarlet, I don't feel like this actually addresses an issue or that it gives the MPs something they need, as cool of an idea as it seems to me too, it just makes the MP very oppressive to anyone that happens to be within range of them.
If it's trying to fix emote spam, that should be fixed with a delay on emotes, if it's trying to fix verbally disruptive captives, that should be fixed with isolation in the brig or in the case of treasonous talk, the corresponding punishment.
The cooldown is also too low, 3 second cooldown for a 5 second mute is a perma mute. It also makes it hard for anyone nearby to communicate with the MP or in general, and someone just spamming out chat while getting arrested which is the use-case for this item isn't even a big deal, while preventing someone from explaining themselves and being very frustrating to play against.
I don't think the concept is unworkable, but I don't really see how those things could be fixed either.

Vermidia:
Yeah that's fair. I'll close this for now.

Issue #2646, opened 2024-06-27
GunDamageModifier is not applied until an extended barrel is attached at least once


snebl:
## Description

While testing weapons in the dev area i noticed that the extended barrel attachment seemed to have a damage increase that remains on a weapon even after you remove it.

Example: The battle rifle does 40 damage to a drone normally but will do 56 with an extended barrel HOWEVER a battle rifle with its extended barrel then removed also does the same 56 damage.

EDIT: I figured out that the assault rifle and battle rife are **supposed** to have a base damage increase. So the real issue is that this increase does not get applied at all unless you attach an extended barrel.

**Reproduction**

1.) Put an extended barrel on any applicable weapon
2.) Take it off
3.) Your weapon now deals more damage (the correct amount) despite now being identical to a fresh one 

**Screenshots**

![image](https://github.com/RMC-14/RMC-14/assets/96957003/583fa418-a329-47ec-99ed-b03c8563cee8)

**Additional context**

Siigiil:
This appears to be a bug with the battle rifle itself not applying its +40% damage modifier. The extended barrel doesn't have a damage modifier, but attaching it causes the gun to properly apply its own, which then stays.

snebl:
> The extended barrel doesn't have a damage modifier

Thats interesting. I assumed it was intended and almost mentioned how weird it was that the kind of substantial damage buff isn't mentioned in the attachment's description.

I should mention the assault rifle also got a damage buff when doing this AND that this bug seems to only happen with the extended barrel. The recoil suppressor had a damage reduction and something like a smart scope did not affect the damage.

snebl:
Updated the issue a bit for clarity.

PR #2669, opened 2024-06-30
Fix GunDamageModifier

snebl:
## About the PR

The assault rifle, battle rifle, and other weapons that use the `GunDamageModifier` component should now deal the correct damage.

## Why / Balance

Fixes [#2646](https://github.com/RMC-14/RMC-14/issues/2646).

## Technical details

Basically calls `RefreshGunDamageMultiplier` when a gun spawns in and not just when something attaches to it.
No clue what i'm doing for the most part but it seems to work without error. I'm sorry if i have committed some kind of sin here, i have never touched ss14 code before.

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Fixed damage modifiers not being applied on marine firearms.


Siigiil:
This will fire every time a gun shoots. So every time every marine gun shoots, it'll recalculate this modifier.
Would be better to just subscribe to `MapInitEvent` and do it there. That way it will only fire once, when the gun spawns in. And then whenever the relevant attachments are modified.

snebl:
@Siigiil Like this?

Siigiil:
> @Siigiil Like this?

Yes.

QuietlyWhisper (approved):
Thanks

snebl:
image

PR #2672, opened 2024-06-30
Add blood packs to requisitions

a-person5660:
## About the PR
this PR adds blood packs to the medical crate in requisitions

## Why / Balance
currently there is no way to get more blood packs, this causes problems in longer drawn out rounds where medical runs out of blood and can no longer heal people with blood loss, this means that the marines are nearly guaranteed to lose if the round goes too long.

## Media

**Changelog**
🆑 
- tweak: added blood packs to the medical crate in requisitions


Resources/Prototypes/_RMC14/Catalog/Fills/Crates/medical.yml
```diff
@@ -73,6 +73,7 @@
       - id: CMPillCanisterKelotane
       # - id: CMPillCanisterTramadol
       # - id: CMPillCanisterPeridaxon
+      - id: Bloodpack
```
QuietlyWhisper, line 76: This needs to be an RMC bloodpack but I also don't think a single pack in this purchase is going to make a difference, maybe look into having a blood pack crate made instead? I'd need to see if something like that exists in 13...

QuietlyWhisper (changes requested):

Resources/Prototypes/_RMC14/Catalog/Fills/Crates/medical.yml
a-person5660: should i make it have multiple blood packs or make a new crate specifically for them?

a-person5660:
failed check isnt my fault

thymol-blue:
I believe blood packs should be refillable at the medical vendor, not ordered at requisitions. I might be mistaken though.

QuietlyWhisper:
> I believe blood packs should be refillable at the medical vendor, not ordered at requisitions. I might be mistaken though.

That's probably something we missed in testing and makes a lot of sense, in which case the 13 implementation would be prefered

QuietlyWhisper:
I think we decided to close this, the marines will be required to give blood donations instead of buying more blood.

JackSage:
> I think we decided to close this, the marines will be required to give blood donations instead of buying more blood.

You can't inject blood into blood packs 
image

PR #2674, opened 2024-06-30
enable cat ears and tail markings for humans

Afrokada:
## About the PR
This PR enables the usually unavailable cat ears and tail for humans :)

## Why / Balance
We have short cat people now, why not have normal height cat people?

**Changelog**

:cl: cooldolphin
- add: Enabled cat ears for humans.

### DIFF
--- a/Resources/Prototypes/Species/human.yml
+++ b/Resources/Prototypes/Species/human.yml
@@ -45,10 +45,10 @@
       points: 1
       required: false
     Tail: # the cat tail joke
-      points: 0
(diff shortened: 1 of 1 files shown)


Afrokada:
okay, I don't understand the build fail here :(

Kadeo64:
> okay, I don't understand the build fail here :(

past two PRs aside from this one are test failing too

QuietlyWhisper:
We'd rather people just play felinid. Sorry.

PR #2690, opened 2024-07-01
Remove all red crosses

MACMAN2003:
## About the PR

Greenified all the red crosses that I could find.
## Why / Balance

Because the ICRC is a bunch of haters.
## Media

## Breaking changes

**Changelog**

I don't think players should be concerned about compliance with the geneva conventions


Kadeo64:
why is geneva filled with fun killers? first the geneva conventions, then the red cross.

QuietlyWhisper (changes requested):
Looks good, just two first aid sprites not needed

JoeHammad1844:
does anything actually happen if we keep the red crosses? CM13 has had them for YEARS and NOTHING seems to have happened

JoeHammad1844:
also the white cross with the red background doesnt count as a red cross, since the cross is white

MACMAN2003:
better to be safe and green than red and smited by the killjoys at the ICRC

QuietlyWhisper (approved):
Thanks!

PR #2694, opened 2024-07-01
Optimize server-side comps for invincible walls

mirrorcult:
## About the PR

changed the parent for invincible walls away from `CMBaseStructureCorrodible` which includes a lot of things that are unnecessary

since things parented from these walls make up the majority of any gameplay map (lv-426 is like, mostly rock walls) its best for them to be lean. not sure how much it actually affects memory usage because that map is not in the repo.

among things removed:
- corrodible (it just set corrodible to false anyway)
- clickable (shouldnt be necessary since you cant interact with these walls?)
- staticprice/pullable (self-explanatory)
- radiationblocker (to my knowledge cm does not have radiation mechanics?)
- tag (only tagged as structure which is kind of erroneous to begin with and unused here)

can probably get away with removing placementreplacement in gameplay somehow since its only relevant for mapping

## Media

b4

image

afta

image

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Walls/walls.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Walls/walls.yml
@@ -301,14 +301,6 @@
   - type: Tag
     tags:
     - Wall
-  - type: Destructible
-    thresholds:
-    - trigger:
-        !type:DamageTrigger
-        damage: 3000
-      behaviors:
-      - !type:DoActsBehavior
-        acts: [ "Destruction" ]
   - type: IconSmooth
     key: walls
     base: jungle_veg
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Base/base_structure.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Base/base_structure.yml
@@ -42,7 +42,6 @@
 # BaseWall
 - type: entity
   abstract: true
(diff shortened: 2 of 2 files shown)


Resources/Prototypes/_RMC14/Entities/Structures/Walls/walls.yml
```diff
@@ -301,14 +301,6 @@
...
     tags:
     - Wall
-  - type: Destructible
-    thresholds:
-    - trigger:
-        !type:DamageTrigger
-        damage: 3000
-      behaviors:
-      - !type:DoActsBehavior
-        acts: [ "Destruction" ]
```
mirrorcult, line 311: this is for 'jungle vegetation', this was removed since they arent even damageable to begin with being invincible. are they intended to be damageable? if so should be parented to something else instead of this change

mirrorcult:
these are an optimal target for shared components if we ever get that in engine

Resources/Prototypes/_RMC14/Entities/Structures/Walls/walls.yml
DrSmugleaf: Checked and yeah this is fine, they have a damage_cap of 3000 in cmss13 but they are also a hull wall so not actually destroyable.

DrSmugleaf (approved):
ty boss

PR #2724, opened 2024-07-01
Various emote fixes

Vermidia:
## About the PR

Ghosts can't use felinid emotes (Think only dead felinids could? Either way they have a proper whitelist/blacklist now)
Slimes now use the RMC versions of screams (They have their own collections)
Added missing gasp sound collections for RMC humans
Changed the sound collection names to RMC

## Why / Balance

Bugs, missing stuff, eventual consistency

## Technical details

N/A

## Media

## Breaking changes

Nope
**Changelog**
nah

### DIFF
--- a/Resources/Prototypes/_RMC14/Voice/speech_emotes.yml
+++ b/Resources/Prototypes/_RMC14/Voice/speech_emotes.yml
@@ -2,6 +2,12 @@
   id: Hiss
   name: rmc-emote-name-hiss
   category: Vocal
+  whitelist:
+    components:
+    - Vocal
+  blacklist:
+    components:
+    - BorgChassis
   available: false
   icon: Interface/Actions/scream.png
   chatMessages: ["rmc-emote-hiss"]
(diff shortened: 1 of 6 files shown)


Resources/Prototypes/_RMC14/Voice/speech_emotes.yml
```diff
@@ -2,6 +2,12 @@
   id: Hiss
   name: rmc-emote-name-hiss
   category: Vocal
+  whitelist:
+    components:
+    - Vocal
+  blacklist:
+    components:
+    - BorgChassis
```
QuietlyWhisper, line 10: There's probably a better way to do this, I'm not sure why you're blacklisting the borg chassis component?
You should probably look into how it's prevented on upstream emotes?

QuietlyWhisper (changes requested):
Tests are failing because the IDs are not valid, it seems you forgot to change these three
image

Resources/Prototypes/_RMC14/Voice/speech_emotes.yml
Vermidia: I copied it from the other upstream emotes, they all whitelist vocal and blacklist borgchasis (though that's probably not needed here in this case).

The whitelist prevents the emotes from being used and showing up for ghosts, since ghosts don't have the vocal component.

QuietlyWhisper (approved):
Sounds good thanks!

PR #2734, opened 2024-07-02
adds screech shockwave

vero5123:
## About the PR
title

## Why / Balance
visually pleasing

## Technical details
If you're wondering why I didn't just use the previous shockwave shader and just mess with the parameters to create the effect; it would be because It just didn't look good and it wasn't what I was looking for. The previous shockwave overlay can be generic enough to be used for various explosions but not for this since it's more of a pulsating wave with a different set of parameters & math to create the effect. Hence why I implemented a whole new separate overlay and component since the old fields didn't match to the new shader parameters. 

## Media

https://github.com/RMC-14/RMC-14/assets/73014819/8373173e-1737-4a1f-bbde-e916dc96d3c7

- [x] I have added screenshots/videos to this PR showcasing its changes ingame

## Breaking changes
N/A

**Changelog**

:cl:
- add: Adds shockwave overlay to queen's screech

### DIFF
--- a/Content.Client/Entry/EntryPoint.cs
+++ b/Content.Client/Entry/EntryPoint.cs
@@ -17,6 +17,7 @@ using Content.Client.Replay;
 using Content.Client.Screenshot;
 using Content.Client.Singularity;
 using Content.Client._RMC14.Explosion;
+using Content.Client._RMC14.Xenonids;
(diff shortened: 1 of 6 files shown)


QuietlyWhisper (approved):
seems fine thanks

PR #2738, opened 2024-07-02
Change pistol pouch inheritance to remove unnecessary components

ShadowCommander:
## About the PR

Found this while messing with pouch visualizers. Tested in game. No changes seen.

## Why / Balance

Pistol pouch did not use any of the components provided by CMPouchClosed, so it was changed to inherit from the parent CMPouch.

## Media

## Breaking changes

**Changelog**


QuietlyWhisper (approved):
thanks

PR #2749, opened 2024-07-02
Fix Bodybags always having a tag

Vermidia:
## About the PR

Bodybags no longer have the labels always visible. Also fixed them upso both bags should support the paperLabel enum properly

## Why / Balance

Pretty sure bug?

## Technical details

All yml
Removed generic visualizer from stasis bag (the one from bodybag works fine)
Changed the name and mapping of the bodybag state (originally bodybag_label") to work with paper labels
Made sure stasis bags also use the same sprites for labels from bodybag.rsi
(Also copied the label sprite for invoice since now those exist and can currently be put on bags, just to be safe)

## Media

bags

## Breaking changes

Hopefully not?
**Changelog**
no cl smol bug


DrSmugleaf (changes requested):
Is the rename of bodybag_label to invoice intended? Doesn't seem to be used

Vermidia:
It was renamed to paper and invoice, just so the sprite wouldn't be missing if someone tried to attach a cargo invoice, since the bags currently can take any paper.

QuietlyWhisper (approved):
thanks

PR #2754, opened 2024-07-02
Rework storage interaction

ShadowCommander:
## About the PR

This fixes interacting with items that are inside a storage entity. Stacks now stack when clicked on if they have space, and bottles can be poured into without having to hold them both in your hands.

## Technical details

The removal of `args.Handle()` here:

https://github.com/RMC-14/RMC-14/blob/25fbb58102d4f7d6b90a637f94d8c1faa9e9c43d/Content.Client/UserInterface/Systems/Storage/StorageUIController.cs#L234-L242

Caused the insertion [code in StorageContainer.KeyBindDown](https://github.com/RMC-14/RMC-14/blob/873199f94484fb7304ed37dd4b793906639b1100/Content.Client/UserInterface/Systems/Storage/Controls/StorageContainer.cs#L521-L526) to always run before any of the interaction code and handle the UI event in KeyBindDown, when the interaction code runs in KeyBindUp.

The event that ran interact using code started in [ItemGridPiece.KeyBindUp](https://github.com/RMC-14/RMC-14/blob/25fbb58102d4f7d6b90a637f94d8c1faa9e9c43d/Content.Client/UserInterface/Systems/Storage/Controls/ItemGridPiece.cs#L246) and went to StorageUIController.OnPieceUnpressed, which raised a [StorageInteractWithItemEvent](https://github.com/RMC-14/RMC-14/blob/25fbb58102d4f7d6b90a637f94d8c1faa9e9c43d/Content.Client/UserInterface/Systems/Storage/StorageUIController.cs#L324-L326)

The event goes to [SharedStorageSystem.OnInteractWithItem](https://github.com/RMC-14/RMC-14/blob/a28f20d423829877b8d769c1bdf73079dce3243e/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs#L621) to call SharedInteractionSystem.InteractUsing.

---

To keep the insertion on non-interaction while also allowing item interaction on entities in storage, I changed [SharedInteractionSystem.InteractUsing](https://github.com/RMC-14/RMC-14/blob/a28f20d423829877b8d769c1bdf73079dce3243e/Content.Shared/Interaction/SharedInteractionSystem.cs#L915-L921) and [InteractDoAfter](https://github.com/RMC-14/RMC-14/blob/a28f20d423829877b8d769c1bdf73079dce3243e/Content.Shared/Interaction/SharedInteractionSystem.cs#L947) to return a bool that returns true when any of the interaction events are handled. If not then it returns false.

And then checked the return value in SharedStorageSystem.OnInteractWithItem to determine whether to insert.

https://github.com/RMC-14/RMC-14/blob/222c6a4492c36ad4c37a9497150624d21a3cc457/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs#L621-L628

## Media

https://github.com/RMC-14/RMC-14/assets/10494922/025fe04c-ca53-4b53-a7f8-aaa2a9392d76

## Breaking changes

Changed InteractUsing and InteractDoAfter to return a bool.

Refactored PlayerInsertHeldEntity to take in Entity<Component> arguments instead of EntityUid and Components.

**Changelog**

:cl:
- fix: You can use items in your hand on other items in storage. If there is no interaction, it will insert the item instead.


DrSmugleaf (changes requested):
Looks like most of this code except for StorageUIController could/should be upstreamed

DrSmugleaf (approved):
Thanks for the PR!

PR #2768, opened 2024-07-02
Explosions now stun and throw mobs

vero5123:
## About the PR
Title

## Why / Balance
Similar to CM13, explosions should stun mobs in close vicinity, xenos shouldn't be able to dance around nades, same goes for marines. This makes nades a double edged sword where they are now relatively more impactful if used correctly but if you screw up then the ff damage can also be significant. 

## Technical details
We can maybe make this an optional parameter instead, so that some mobs like queen and crusher can potentially be immune to these effects.  

## Media

https://github.com/RMC-14/RMC-14/assets/73014819/80a2da72-ac73-47be-9ed6-d9478063e315

## Breaking changes
N/A

**Changelog**
:cl:
- add: Explosions will stun & toss mobs in close vicinity

### DIFF
--- a/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
+++ b/Content.Server/Explosion/EntitySystems/ExplosionSystem.Processing.cs
@@ -468,7 +468,7 @@ public sealed partial class ExplosionSystem
         {
             var pos = _transformSystem.GetWorldPosition(xform);
             var adjustedThrowForce = throwForce;
-            if (HasComp<MobStateComponent>(uid)) 
+            if (HasComp<MobStateComponent>(uid))
             {
                 adjustedThrowForce /= 5;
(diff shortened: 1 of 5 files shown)


vero5123:
maybe the overall quantity of nades can be reduced as well, but I mean that probably will happen regardless once more weapons & kit choices are added.

JoeHammad1844:
nothing wrong with lots of nades

QuietlyWhisper (changes requested):
This should probably be in a configurable component for mobs instead of in the base code, but the effect it self looks great

vero5123:
as discussed in the technical details section I think having a field value for whether or not the mob is affected by the explosion might be a good idea, however, I don't think the actual adjusted force should be a configurable parameter.

vero5123:
I'm going to refactor the nade's effects on mobs entirely.

vero5123:
i'll get back to this later, unless someone beats me to it.

PR #2788, opened 2024-07-03
Attachments, wield delay, wield slowdown

Siigiil:
## About the PR
Added all **standard** attachments excluding the bipod, the underbarrel extinguisher, the underbarrel flamers and attachments specific to weapons that we may later be renaming. But all the standard stuff like the rail flashlight, the agrip, the vgrip, the UGL and so on are in.

Implemented breech-loaded weapons. Currently it's just UGL. Unique action to open the breech. After loading, close the breech and fire. Then open and close it again to cock the weapon and ready it for firing again.

Made the field strip menu a keybind. There's no default, so if you want to have a hotkey for it, bind it yourself.

Fixed the default keybindings for using attachments.

Made scopes properly work with attachment hotkeys.

Certain scopes now allow movement when scoped.

Implemented wield delay - after picking up a gun, there's a short delay before you can wield it. Certain attachments may modify this.

Implemented wield slowdown - when wielding some weapons, you will be slowed. Certain attachments may modify the slowdown.

## Why / Balance
Because porting.

## Technical details
I'll be adding the bipod in a later PR, as it needs a bunch of stuff to work properly, including an upstream PR to make it possible to add and remove firing modes.

Changed how unique action works: it no longer needs `UniqueActionComponent` to send out the event. The component now only adds the verb for unique action to the entity.

Renamed all attachment-related IDs and key functions to use "RMC".

Converted everything in AttachableData to record structs.

Changed how scope attachments work to make it possible for them to properly interact with other attachment modifiers. They're now activated through AttachableToggleableSimpleActivateComponent. This required adding a field to scopes that can allow them to be used in hand, but since we're adding binoculars eventually, this was needed anyway.

Changed how roundstart attachments work - they're now handled entirely by AttachmentHolderSystem. This was changed because with ContainerFillComponent, we still needed to go over the newly added attachments to make sure all the containers were set up properly, and it required adding two more components manually, which seems excessively fiddly.

Changed how weapon and many other modifiers work. They are now handled by AttachmentModifiersSystem, which is split between multiple files. Axed most of the modifier components, replaced each group with a single component that has a list of modifier sets. Each modifier set contains its modifiers and a set of conditions under which they are applied.

Changed ranged weapon modifiers. Added modifiers for accuracy and the other things we don't have implemented yet. This is to have them ready on the attachments and make it easier on us when we implement all of that. Replaced all the firing angle modifications with a single variable: scatter. It increases both the minimum and maximum angle of your shots.
Conversion from CM: multiply CM scatter by 2. This is because CM gets the actual angle of every shot as a random number between `-angle` and 'angle`, while SS14 uses `-0.5*angle` and `0.5*angle`

**Wield slowdown:** The slowdown in CM13 is handled with move delay, which we don't have. This matters because move delay can't be converted into a single multiplier - the speed change depends on the target's current speed. Since the only good way to alter player speed right now is to subscribe to  `RefreshMovementSpeedModifiersEvent`, our only option is to use multipliers.

The way I'm handling this is applying different slowdowns for walking and running. The multipliers are precalculated using the *default* human movement speeds in RMC: 2.5f for walking and 4.5f for running.
The formula is as follows: `1 / (1 / SS14_MOVE_SPEED + SS13_MOVE_DELAY / 10) / SS14_MOVE_SPEED`

`1 / SS14_MOVE_SPEED` gets us the default move delay in seconds. The SS13 move delay is divided by 10 because in SS13 it's expressed in deciseconds. This is then turned back into movement speed and divided by the original movement speed to get the multiplier.

Wield slowdown is then reduced by the marine's worn armour. The reduction is stored in `WieldSlowdownCompensationComponent`. In CM, this is the `movement_compensation` variable on clothes.

**Attachment visuals:** All attached attachment sprites have been moved so that they don't all need unique offsets. We shouldn't need to use the AttachmentVisualsComponent offset field for anything other than attachments that are too big to properly place them in their 32x32 sprite or those that use their normal inventory state when attached.

## Media
S8 scope zoom:
S8

Miniscope zoom:
S4

New agrip sprite:
agrip

New vgrip sprite:
vgrip

New folding stock sprites:
folding

New solid stock sprite:
solid

Attachments (ANGLED GRIP, VERTICAL GRIP AND M54 STOCKS OUTDATED, SEE ABOVE):
attachments

## Breaking changes
Changed a bunch of attachment-related IDs to use RMC instead of CM. Pretty sure I got all mentions of them, so this shouldn't break anything.

**Changelog**
:cl: Sigil, Cepha
- add: Added all **standard** attachments, excluding the bipod, the underbarrel extinguisher, the underbarrel flamers and most attachments specific to weapons that are unimplemented.
- add: Added the missing default keybinds for attachment activation. Your keybinds will be reset.
- add: You can now open the attachment strip menu with a hotkey as well as the verb. No default keybinding.
- tweak: Certain scopes now allow movement. These being the smartscope and the miniscope.
- fix: Scopes now properly work with attachment hotkeys.
- add: Implemented wield delay. You will need to wait a short while after picking up a weapon before you can wield it. Shooting is unaffected. Attachments can affect this.
- add: Added wield slowdown. Wielding certain weapons will slow you down. Attachments can affect this.
- tweak: Added new sprites for the agrip and vgrip. Sprited by Cepha.
- tweak: Added new sprites for the M54C solid and folding stocks. Sprited by Sigil.


Siigiil:
> image isnt this part supposed to be the stock?

It's part of the gun's sprite. No sprites for it unfolded exist. So for now I used the standard CM sprites. We'll probably update them once we get something that both looks good and works for the gun. There are a few options in the works. Not sure if it's going to be in this PR, though.

JoeHammad1844:
yeah we could just shop that part off the gun later

JoeHammad1844:
also please tell me the bayonet actually works

Siigiil:
> also please tell me the bayonet actually works

I can certainly *tell* you that.

We're waiting for this: https://github.com/space-wizards/space-station-14/pull/29484

JoeHammad1844:
finally!

UnicornOnLSD:
humongous work, absolutely amazing, thanks for the work 👍

MACMAN2003:
how long is the wield delay

Siigiil:
> how long is the wield delay

Depends on the gun. 0.6 seconds for the M54 and shotgun. Less for the SMG, much less for the SPR and pistols. Much more for the sniper spec rifle.

Siigiil:
Added the new sprites. This is ready for review.

DrSmugleaf:
> * tweak: Certain scopes now allow movement. These being the smartscope and the miniscope.

Is this outdated? I checked 13 and your yaml and it (correctly) does not seem to allow movement.

Siigiil:
> > * tweak: Certain scopes now allow movement. These being the smartscope and the miniscope.
> 
> Is this outdated? I checked 13 and your yaml and it (correctly) does not seem to allow movement.

What do you mean? Both allow movement.

Miniscope in CM13:
miniscope
Smartscope in CM13:
smartscope

The scope component on the miniscope in my implementation:
miniscope14

In both cases, the smartscope inherits allowing movement. I tested in CM13 and they both allow movement.

Siigiil:
Conflicts should be resolved now.

DrSmugleaf:
> In the latter case, the smartscope inherits allowing movement. I tested in CM13 and they both allow movement.

Oh, yeah that is on latest master. Still, it doesn't allow movement sideways or backwards, only forwards.

https://github.com/RMC-14/RMC-14/assets/10968691/d9e34a5a-529b-44a9-838d-c8338da1b8f0

Siigiil:
> > In the latter case, the smartscope inherits allowing movement. I tested in CM13 and they both allow movement.
> 
> Oh, yeah that is on latest master. Still, it doesn't allow movement sideways or backwards, only forwards. Is that the case here too?

ShadowCommander reminded me that direction locking exists in CM. If you lock yourself *in the direction you're aiming*, you can move. I completely forgot about this mechanic.

The latest commit is a bunch of bugfixes for scopes.

DrSmugleaf (approved):
Thanks for the PR!

PR #2801, opened 2024-07-04
Make VirtualItem not block attacking

ShadowCommander:
## About the PR

Allow attacking while a pull VirtualItem is in hand.

## Why / Balance

Allows xenos to more easily shove marines when dragging them.

## Technical details

The check for item in hand cancelled the attack if the item didn't have a MeleeWeaponComponent. This change bypasses the cancel if the held item has a VirtualItemComponent.

## Media

https://github.com/RMC-14/RMC-14/assets/10494922/8eeb0cda-9b36-4012-84c7-a2006106808e

## Breaking changes

**Changelog**

:cl:
- tweak: Players can now use melee attacks and shoves while dragging an entity in their active hand.

### DIFF
--- a/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
+++ b/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
@@ -13,6 +13,7 @@ using Content.Shared.Hands;
 using Content.Shared.Hands.Components;
 using Content.Shared.Interaction;
 using Content.Shared.Inventory;
+using Content.Shared.Inventory.VirtualItem;
 using Content.Shared.Item.ItemToggle.Components;
 using Content.Shared.Physics;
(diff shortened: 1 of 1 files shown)


MACMAN2003:
i feel like this will make tackle spam even more of an uncounterable strat for the marines

ShadowCommander:
Xenoids can already tackle spam with their other hand. This is really only a skill check for players that are new to playing Xenoid. If we want to make tackle spam less oppressive, there are other things to tweak that'd work better than preventing this.

DrSmugleaf:
> i feel like this will make tackle spam even more of an uncounterable strat for the marines

I'm going to make it so immature queens dont get screech (like 13) which should help with some of that

DrSmugleaf (approved):
Thanks for the PR!
This should be upstreamed too when you can

PR #2811, opened 2024-07-04
Locks viewport size to 15x15

Siigiil:
## About the PR
As the title says, this locks the viewport to being 15x15, instead of the previous variation from 15x15 to 21x15. Also hides the viewport width slider from the settings menu.

## Why / Balance
With a viewport wider than it is tall, players need to constantly rotate their camera to make the most of their view distance and any ranged options they might have.

*Not* doing this puts you at a disadvantage against anyone who *is* doing it, especially if they have a ranged weapon.
*Doing* it can be both disorienting and plain obnoxious with the camera frequently spinning.

Also, scopes are more effective along the wider part of the viewport and give you an even longer range than is normal for them.

CM13 never added a widescreen viewport for a reason.

## Media
Viewport and settings menu:
viewport

**Changelog**

:cl:
- remove: You can no longer change your viewport size. It's locked to 15x15

### DIFF
--- a/Content.Client/Options/UI/Tabs/GraphicsTab.xaml.cs
+++ b/Content.Client/Options/UI/Tabs/GraphicsTab.xaml.cs
@@ -75,7 +75,7 @@ public sealed partial class GraphicsTab : Control
         ViewportScaleSlider.Visible = !ViewportStretchCheckBox.Pressed;
         IntegerScalingCheckBox.Visible = ViewportStretchCheckBox.Pressed;
         ViewportVerticalFitCheckBox.Visible = ViewportStretchCheckBox.Pressed;
-        ViewportWidthSlider.Visible = !ViewportStretchCheckBox.Pressed || !ViewportVerticalFitCheckBox.Pressed;
+        ViewportWidthSlider.Visible = false;// !ViewportStretchCheckBox.Pressed || !ViewportVerticalFitCheckBox.Pressed;
     }
 
     private void UpdateViewportWidthRange()
--- a/Content.Shared/CCVar/CCVars.cs
+++ b/Content.Shared/CCVar/CCVars.cs
@@ -1653,10 +1653,10 @@ namespace Content.Shared.CCVar
             CVarDef.Create("viewport.minimum_width", 15, CVar.REPLICATED | CVar.SERVER);
 
         public static readonly CVarDef<int> ViewportMaximumWidth =
-            CVarDef.Create("viewport.maximum_width", 21, CVar.REPLICATED | CVar.SERVER);
+            CVarDef.Create("viewport.maximum_width", 15, CVar.REPLICATED | CVar.SERVER);
 
         public static readonly CVarDef<int> ViewportWidth =
-            CVarDef.Create("viewport.width", 21, CVar.CLIENTONLY | CVar.ARCHIVE);
+            CVarDef.Create("viewport.width", 15, CVar.CLIENTONLY | CVar.ARCHIVE);
 
         public static readonly CVarDef<bool> ViewportVerticalFit =
             CVarDef.Create("viewport.vertical_fit", true, CVar.CLIENTONLY | CVar.ARCHIVE);


Tunguso4ka:
We probably should make a vote about this in the discord.

DrSmugleaf (changes requested):
Problem with this is we use the chat much less than cmss13, we have a lot more popups for example which aren't relayed to chat. I'm not saying they should, we intentionally made most of the game that way to keep chat to actual chat messages. But it makes the game feel cramped when playing with such a small width.
We also don't have a large list of buttons with tabs on the top right, and even with Ike's idea to put the PDA UI there (which doesn't really apply well to rmc14), there's still not enough information there to take up ~50% of the screen.
This also doesn't work in any form with a modified client, which can get a much bigger advantage this way than they would normally.
I think it would be a better solution to make the camera rotation keys more accessible, instead of having the default be numpad keys, so that if someone really wants to rotate the camera to see an extra ~3 tiles in one direction (mostly applicable to snipers, with scopes its about ~5), then they can.
Other than that, we could maybe think about revisiting this some day if we get more important information that we can put there, like ammo, health info, some special actions (specially for the Queen), but even then I doubt we could make this work well enough.

Siigiil:
I will close this for now and hope that it will be revisited in the future, then. While we could change the default keybindings, they *are* rebindable, so people can just do it themselves.

diraven:
Just curious - what's wrong bout having 2 black vertical bars in the viewport without making chat section wider?

Siigiil:
> Just curious - what's wrong bout having 2 black vertical bars in the viewport without making chat section wider?

From my perspective — nothing. I just prefer to not have wasted space on the screen, so I stretched the chat out. Others could keep it narrow if they wanted to. The size of the chat window isn't set anywhere and can be adjusted freely.

diraven:
image

> Problem with this is we use the chat much less than cmss13, we have a lot more popups for example which aren't relayed to chat. I'm not saying they should, we intentionally made most of the game that way to keep chat to actual chat messages. But it makes the game feel cramped when playing with such a small width. We also don't have a large list of buttons with tabs on the top right, and even with Ike's idea to put the PDA UI there (which doesn't really apply well to rmc14), there's still not enough information there to take up ~50% of the screen. This also doesn't work in any form with a modified client, which can get a much bigger advantage this way than they would normally. I think it would be a better solution to make the camera rotation keys more accessible, instead of having the default be numpad keys, so that if someone really wants to rotate the camera to see an extra ~3 tiles in one direction (mostly applicable to snipers, with scopes its about ~5), then they can. Other than that, we could maybe think about revisiting this some day if we get more important information that we can put there, like ammo, health info, some special actions (specially for the Queen), but even then I doubt we could make this work well enough.

I mean, most of this argument seems to refer to chat taking a lot of real estate. Which may not necessarily be the case.

Furthermore, there are ultrawide monitors, and those might have those black regions anyways. There is just no way to adjust the viewport in a way that would fit all sizes not favoring any particular hardware setup other then make it standard 15x15 for everybody.

Having wider viewport then it is higher does indeed have pretty significant balance implications. 15 vs 21 view distance is over 30% higher. That's a lot...

DrSmugleaf:
> > Problem with this is we use the chat much less than cmss13, we have a lot more popups for example which aren't relayed to chat. I'm not saying they should, we intentionally made most of the game that way to keep chat to actual chat messages. But it makes the game feel cramped when playing with such a small width. We also don't have a large list of buttons with tabs on the top right, and even with Ike's idea to put the PDA UI there (which doesn't really apply well to rmc14), there's still not enough information there to take up ~50% of the screen. This also doesn't work in any form with a modified client, which can get a much bigger advantage this way than they would normally. I think it would be a better solution to make the camera rotation keys more accessible, instead of having the default be numpad keys, so that if someone really wants to rotate the camera to see an extra ~3 tiles in one direction (mostly applicable to snipers, with scopes its about ~5), then they can. Other than that, we could maybe think about revisiting this some day if we get more important information that we can put there, like ammo, health info, some special actions (specially for the Queen), but even then I doubt we could make this work well enough.
> 
> I mean, most of this argument seems to refer to chat taking a lot of real estate. Which may not necessarily be the case.
> 
> Furthermore, there are ultrawide monitors, and those might have those black regions anyways. There is just no way to adjust the viewport in a way that would fit all sizes not favoring any particular hardware setup other then make it standard 15x15 for everybody.
> 
> Having wider viewport then it is higher does indeed have pretty significant balance implications. 15 vs 21 view distance is over 30% higher. That's a lot...

You are right, I did mostly approach it from the "wider chat" point of view though since the alternative is black bars (nothing of acutal gameplay importance in that space, so wasted space). They can be filled with some UI elements, like inventory, actions, status effects, etc, but that doesn't necessarily apply well to every HUD layout (separated for example has the actions along the top instead of vertically on the left).
For ultrawide monitors I recognize we aren't accomodating them already, but the most common aspect ratio is 16:9 which is the demographic my comment was aimed at. Granted, anything other than a 4:3 CRT won't even approach the intended square ratio.
This is also just about accomodating people that are used to the wider viewport from upstream and other ss14 servers, if a good solution is ever devised in upstream that results in the viewport being square then we could consider doing the same here.
As for the 30% difference that's why I wanted to make the default keybinds easier to access, so that players can make a choice if they want to use it to their advantage. It doesn't help that a lot of the fighting angles on LV are vertical, as opposed to Solaris which has a lot more horizontal ones, but one could always choose to leave it rotated 90º in that case where most engagements will be in that direction.

diraven:
In that case it really boils down to rotating inconvenience to see 15% farther vs inconvenience of having vertical black bars. A question of priorities and what game tries to be. Whether competitive (and thus balancing) is more important then utilizing all the screen real estate on most common ratios.

It's really up to hosts to decide. Or put up a vote for the community if they want. Not much more to it.

UnicornOnLSD:
Right, aware it's closed, but putting this here regardless.

If the reason WHY we would make the main game window on only half the screen is because some people take advantage of camera rotation, why not just lock said rotation ? I believe people would, hypothetically enjoy this way more than having half of their screen taken by the chat box.

PR #2818, opened 2024-07-04
Update Savannah: Requi venders

Tunguso4ka:
## About the PR

- Added requisitions vender
- One new disposal unit
- Removed maint door to the space.

**Changelog**

:cl:
- add: Added requisition venders to the Savannah.


DrSmugleaf (changes requested):
You don't need to update the fake almayer file don't worry, otherwise it will keep failing tests when the becomes station station name gets overriden.

PR #2819, opened 2024-07-04
Xenoid variant of Rouny

Kadeo64:
## About the PR
Adds an actual Xenoid variant of Rouny. It's, of course, a runner reskin.
## Why / Balance
we're called "ROUNY's marine corps" yet our only Rouny is a shitty simplemob that doesn't have any unique abilities. This Rouny is just a reskin of Runner (but to make it more fun when it's spawned in admeme rounds it stuns in two shoves and said stun lasts for 20 seconds)
## Technical details
Parents from runner. It has a suffix saying "RMC14", to separate it from the normal salvage simplemob Rouny, which I have kept in for Wizden compatibility reasons. If an admin spawns a normal rouny in, bully them. 
## Media

## Breaking changes

**Changelog**
:cl:
- add: ROUNY IS FINALLY HERE!!


Kadeo64:
idk what test fails are being caused by i'm not smart enough for that

Zarq0:
> idk what test fails are being caused by i'm not smart enough for that
Test fails are because everything after description should be under components iirc

Zarq0:
you also should add mob thresholds for the states

Kadeo64:
> you also should add mob thresholds for the states, just take a look at other xenos

I thought that would be handled by runner parent, the only change to the states I've made is the sprite path 

> Test fails are because everything after description should be under components iirc

oh yeah I'm stupid I'll fix that when I get back to my computer

Zarq0:
i think it does get handled that way not sure i havent made an entity before

Kadeo64:
okay test fail is being caused by rsi I think but I'm not going to be able to deal with that right now since I'm coding on steam deck.

Kadeo64:
god i hate github let me reopen this PR with a different branch

PR #2825, opened 2024-07-05
Resin Explosion Modifier and HEDP/HEFA Adjustments

Kaga-404:
## About the PR

Ported equivalents to explosive damage modifiers for resin walls and doors. Similar to the melee 8.3x and xeno 4x damage modifiers, doors and walls normally take 15x explosive damage. This is reduced to 12.5x and 7.5x for resin walls and doors respectively.

Because HEDP and HEFA will be massively more effective at clearing out resin like in CM13 now, their explosive damage has been tuned down. They're still about twice as strong as CM damage-wise, but this has been reduced down from approximately 4x as strong. Previously could do up to 360 damage on a direct hit, now caps out at 200.

Not sure why the damage distribution was at 9 per intensity in the RMC explosion type, so I changed it to 10 per intensity for ease of conversion. If there's a specific reason for this please let me know, the numbers can be adjusted pretty easily with a bit of math, but 10 per intensity matches up 1:10 with CM explosion severity.

## Why / Balance

Noticed that the feel of HEDP was very off, particularly because it was both very lethal and would clear weeds (even through walls) in a wide circle, yet couldn't break down walls, whereas in CM13, HEDP is a fairly weak combat option that only deals up to around 100 raw damage, has a very small damage radius, and is primarily used for AOE damage, stunning smaller xenos with launchers or cooked nades, and clearing resin.

Couldn't model HEDP explosions exactly because SS14's explosion system doesn't seem to support that kind of falloff as far as I could tell. In CM, HEDP falls off from 100 to 75 to 25 and then 0.

Here, I just used a step of 6 instead, so 20 to 14 to 8 to 2, so it still has a similar 2.5 radius clearing zone.

Finally, kept the base explosion damage for both grenades higher than CM's (HEDP deals 200 instead of 100, HEFA deals 80 instead of 40) because they can't stun yet. Once explosion stuns are implemented I'd recommend dropping HEDP to (10, 3, 200) and HEFA to (4, 2, 110). The total intensity is kept because that seems to affect the number of entities an explosion can hit.

Total Intensity was dropped for HEDP here because 400 was messing up falloff somehow. I don't understand the inner workings of SS14 explosions enough yet to figure out why, but 200-300 seemed alright. 200 can still clear walls up to 4 tiles away if unobstructed, but does very little or no damage to mobs, there was a lot of weird inconsistency with that in testing compared to cell-based CM explosions.

Walls absorb the explosions well, and this seems to interact with Total Intensity too. At 400, it would consistently blow through 3 layers, at 200 it can be 'muffled' if it's inside and only hit the close walls, while still being able to hit walls out to 4-tiles if exposed. The overall reduction in intensity and Total Intensity *should* decrease the maximum radius in resin areas, especially through walls, and reduce things like weeds being destroyed through walls without the the walls soaking the intensity first.

## Media

## Breaking changes

**Changelog**

:cl: Kaga
- add: Added proper resin wall explosion damage modifiers. They should be much easier to blow up, now.
- tweak: Reduced HEDP and HEFA base explosion power to account for the resin-breaking buff.


DrSmugleaf (changes requested):
Thanks for the PR!

> Not sure why the damage distribution was at 9 per intensity in the RMC explosion type, so I changed it to 10 per intensity for ease of conversion. If there's a specific reason for this please let me know, the numbers can be adjusted pretty easily with a bit of math, but 10 per intensity matches up 1:10 with CM explosion severity.

The numbers should have been accurate for damage to xenos themselves, though if this is still 2x as strong as cmss13 then it needs to be adjusted to not be as well. I can get the exact numbers from cmss13 later, but whichever those are should be followed for xenos/humans/resin etc.

DrSmugleaf (approved):
Thanks for the PR!

diraven:
It seems that all the resin structures are now one-shot by grenades. Is that intentional and how it works in CM13?

DrSmugleaf:
> It seems that all the resin structures are now one-shot by grenades. Is that intentional and how it works in CM13?

Yes, HEDP one-shots normal resin walls from a tile away in 13

diraven:
> > It seems that all the resin structures are now one-shot by grenades. Is that intentional and how it works in CM13?
> 
> Yes, HEDP one-shots normal resin walls from a tile away in 13

It's much more then 1 tile away. And it's not just normal resin. It pretty much deconstructs everything in a huge radius.

Interstingly enough, I've tested it locally and it seems to behave properly. While in-game it destroys pretty much everything.

Here is a video of one grenade pretty destroying everything including thick resin walls in a huge radius around it.


https://github.com/user-attachments/assets/4160727d-47ea-4f93-80a2-0e93a8f63d89

diraven:
> > It seems that all the resin structures are now one-shot by grenades. Is that intentional and how it works in CM13?
> 
> Yes, HEDP one-shots normal resin walls from a tile away in 13

It's much more then 1 tile away. And it's not just normal resin. It pretty much deconstructs everything in a huge radius.

Interstingly enough, I've tested it locally and it seems to behave properly. While in-game it destroys pretty much everything.

Here is a video of one grenade pretty much destroying everything including thick resin walls in a huge radius around it. You can see thick wall entity being destroyed in the bottom right corner.

https://github.com/user-attachments/assets/4160727d-47ea-4f93-80a2-0e93a8f63d89

diraven:
Another one. With a lot of thick resin walls and doors now.

https://github.com/user-attachments/assets/c72a5970-88bd-4a2c-81b8-9e56774d1bc5

PR #2837, opened 2024-07-05
Gives barricades a use delay

SphiraI:
## About the PR
Makes any toggle-able barricades have a small delay upon use

## Why / Balance
To prevent the semi-often case of people spamming them for the ratchet sound, and so people can't try and cheese their instant blocking methods.

## Technical details
Adds `UseDelay` to the barricades proto.

**Changelog**
:cl: Sphiral
- tweak: Barricades now have a 0.7 use delay


DrSmugleaf:
0.7 might be a little too high but I need to think on it, I'll test how it feels locally later

QuietlyWhisper (approved):
thanks

Issue #2823, opened 2024-07-04
Sandbag barricades shouldn't be unanchorable


Skarletto:
the point is it's a quick cade to place down if you need

you can deconstruct it with an entranching tool or a shovel

moving a stack of bags like you could with a normal barricade is silly

UnicornOnLSD:
yeah but what if I rotate it wrong and I unanchor it from the ground with it's imaginary bolts to set it right >:( (valid issue)

Skarletto:
you shovel it and place it again (skill issue)

PR #2843, opened 2024-07-05
Sandbag barricades no longer have invisible bolts and can now be deconstructed

snebl:
## About the PR

Makes it so sandbag barricades no longer have an anchorable component. They now have to be deconstructed with the entrenching tool (simply click and wait for the do after).

Once deconstructed the amount of sandbags you get back depends on how damaged the barricade was. The values should match up with cmss13. [This is what i referenced.](https://discord.com/channels/1168210010233376858/1168210011823026270/1258819675278807170)

https://github.com/snebl/RMC-14/blob/5226556bf0915e4391a9168fe62a585b7a8ce017/Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs#L94

## Why / Balance

Resolves [#2823](https://github.com/RMC-14/RMC-14/issues/2823)

## Technical details

Adds `BarricadeSandbagComponent.cs` and `SandbagDismantleDoAfterEvent.cs`.
Most changes are in `EntrenchingToolSystem.cs`. It can now handle the use of the entrenching tool on a sandbag barricade.

## Media

image

## Breaking changes

**Changelog**

:cl:
- add: Sandbag barricades can now be deconstructed with the entrenching tool.

### DIFF
--- a/Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
+++ b/Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
@@ -83,29 +83,29 @@ public sealed class EntrenchingToolSystem : EntitySystem
 
         args.Handled = true;
 
-        if (_net.IsServer)
+        if (_net.IsClient)
+            return;
+
+        if (!TryComp(args.Target, out BarricadeSandbagComponent? barricade))
+            return;
+        var full = Spawn(barricade.Material, EntityManager.GetCoordinates(args.Coordinates));
+
+        var bagsSalvaged = barricade.MaxMaterial;
+        if (bagsSalvaged <= 0 && TryComp(full, out FullSandbagComponent? fullSandbag))
+            bagsSalvaged = fullSandbag.StackRequired;
+        if (TryComp(args.Target, out DamageableComponent? damageable))
+            bagsSalvaged -= Math.Max((int) damageable.TotalDamage / barricade.MaterialLossDamageInterval - 1, 0);
+
+        Del(args.Target);
+
+        if (bagsSalvaged <= 0)
         {
-            if (!TryComp(args.Target, out BarricadeSandbagComponent? barricade))
-                return;
-            var full = Spawn(barricade.Material, EntityManager.GetCoordinates(args.Coordinates));
-
-            var bagsSalvaged = barricade.MaxMaterial;
-            if (bagsSalvaged <= 0 && TryComp(full, out FullSandbagComponent? fullSandbag))
-                bagsSalvaged = fullSandbag.StackRequired;
-            if (TryComp(args.Target, out DamageableComponent? damageable))
-                bagsSalvaged -= Math.Max((int) damageable.TotalDamage / barricade.MaterialLossDamageInterval - 1, 0);
-
-            Del(args.Target);
-
-            if (bagsSalvaged <= 0)
-            {
-                Del(full);
-                return;
-            }
-
-            if (TryComp(full, out StackComponent? fullStack))
-                _stack.SetCount(full, bagsSalvaged, fullStack);
+            Del(full);
+            return;
         }
+
+        if (TryComp(full, out StackComponent? fullStack))
+            _stack.SetCount(full, bagsSalvaged, fullStack);
     }
 
     private void OnDoAfter(Entity<EntrenchingToolComponent> tool, ref EntrenchingToolDoAfterEvent args)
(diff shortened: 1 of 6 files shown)


Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
```diff
@@ -48,10 +50,62 @@ private void OnAfterInteract(Entity<EntrenchingToolComponent> tool, ref AfterInt
...
+    private void OnSandbagDismantleDoAfter(Entity<EntrenchingToolComponent> tool, ref SandbagDismantleDoAfterEvent args)
+    {
+        if (args.Cancelled || args.Handled)
+            return;
+
+        args.Handled = true;
+
+        if (_net.IsServer)
+        {
+            var full = Spawn("CMSandbagFull", EntityManager.GetCoordinates(args.Coordinates));
```
DrSmugleaf, line 88: CMSandbagFull here should be a datafield on the barricade of type EntProtoId, if nothing else so that it automatically gets validated to exist, but also in case it ever needs to be changed from YAML

Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
```diff
@@ -48,10 +50,62 @@ private void OnAfterInteract(Entity<EntrenchingToolComponent> tool, ref AfterInt
...
+
+            var bagsSalvaged = 0;
+            if (TryComp(full, out FullSandbagComponent? fullSandbag))
+                bagsSalvaged = fullSandbag.StackRequired;
+            if (TryComp(args.Target, out DamageableComponent? damageable))
+                bagsSalvaged -= Math.Max((int) damageable.TotalDamage / 75 - 1, 0);
```
DrSmugleaf, line 94: This 75 should also be a datafield on the barricade comp

Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
```diff
@@ -48,10 +50,62 @@ private void OnAfterInteract(Entity<EntrenchingToolComponent> tool, ref AfterInt
...
     }
 
+    private void DismantleSandbagBaricade(Entity<EntrenchingToolComponent> tool, ref AfterInteractEvent args)
+    {
+        if (TryComp(tool, out ItemToggleComponent? toggle) && !toggle.Activated)
+            return;
+
+        _popup.PopupClient(Loc.GetString("cm-entrenching-dismantle"), args.User, args.User);
+
+        var ev = new SandbagDismantleDoAfterEvent(GetNetCoordinates(args.ClickLocation));
```
DrSmugleaf, line 71: This should be try get net coordinates with a return and maybe a log warning, so that if a client passes a malicious input it doesn't error

Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
```diff
@@ -48,10 +50,62 @@ private void OnAfterInteract(Entity<EntrenchingToolComponent> tool, ref AfterInt
...
+    }
+
+    private void OnSandbagDismantleDoAfter(Entity<EntrenchingToolComponent> tool, ref SandbagDismantleDoAfterEvent args)
+    {
+        if (args.Cancelled || args.Handled)
+            return;
+
+        args.Handled = true;
+
+        if (_net.IsServer)
```
DrSmugleaf, line 86: You can reverse this condition to if is client return to reduce the indentation on the rest of the method

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
snebl: There is no `TryGetNetCoordinates` method in `EntitySystem` or anywhere else i could find, only a `TryGetNetEntity` which wouldn't work here as far as i can tell. There are two other instances of just `GetNetCoordinates` being used within the same file before i had made any changes.

I'm going to mark this as resolved but if you want i can make changes to all three instances i just need to know how.

snebl:
Finished requested changes and i also added a 3rd data field for manually setting the max stack size received in case there are ever multiple kinds of sandbag-like barricades.

Content.Shared/_RMC14/Entrenching/EntrenchingToolSystem.cs
DrSmugleaf: Nevermind, I checked now and apparently I hallucinated that method

DrSmugleaf (approved):
Thanks for the PR!

snebl:
image

PR #2847, opened 2024-07-06
Adds command tablet

tsp01:
## About the PR
This PR creates the CO tablet, which is functionally just a mobile ground-side operations console. It is literally the exact same UI as the Console, primarily because they both do the exact same thing. Since it uses the same UI as the Console, I also renamed the UI to remove the word Console on the announcement pop-up. 

## Why / Balance
This was added mostly because it existed in CM, but also because commanders need a way to get everyone's attention at a moments notice, even if they aren't in CiC. This is especially true given that there is currently no way to see whats going on ground side without being ground side

## Technical details
No C#, just yaml and mapping

## Media
Screenshot 2024-07-05 at 8 44 31 PM

**Changelog**

:cl:
- add: Added Command Tablet
- tweak: changed UI name for groundside operations console to "Groundside Operations"


JoeHammad1844:
isn't CIC on the second floor, as such its completely inaccessible?

tsp01:
> isn't CIC on the second floor, as such its completely inaccessible?

We don't have a second floor. CiC(as well as the brig) is west of the hanger.

Tunguso4ka:
> We don't have a second floor. CiC(as well as the brig) is west of the hanger.

almayer.yml that presented in the repo is not actually a Almayer, thats Savannah.

Resources/Prototypes/_RMC14/Entities/Objects/Devices/command_tablet.yml
```diff
@@ -0,0 +1,18 @@
+- type: entity
+  parent: BaseItem
+  id: CMCotablet
```
Tunguso4ka, line 3:
```suggestion
  id: RMCTabletCO
```

please, use `RMC` instead of `CM`.

Resources/Maps/_RMC14/almayer.yml
Tunguso4ka, line 1: Thats not actually Almayer. almayer.yml in the repo is a copy of savannah.yml. We have it here so Integration checks can pass.

Tunguso4ka (changes requested):

Resources/Maps/_RMC14/almayer.yml
tsp01: interesting, should I remove the mapping change?

Tunguso4ka: Yes, that would be better. We dont use this map file on the server.

Resources/Prototypes/_RMC14/Entities/Objects/Devices/command_tablet.yml
```diff
@@ -0,0 +1,18 @@
...
+  id: RMCTabletCO
+  name: command tablet
+  description: A portable command interface used by top brass, capable of issuing commands over long ranges to their linked computer. Built to withstand a nuclear bomb.
+  components:
+  - type: Sprite
+    sprite: _RMC14/Objects/Devices/command_tablet.rsi
+    state: cotablet
+  - type: Item
+    size: Normal
+  - type: ActivatableUI
```
QuietlyWhisper, line 12:
```suggestion
  - type: ActivatableUI
  - type: ActivatableUIRequiresAccess
  - type: AccessReader
    access: [ [ "CMAccessCommand" ] ]
```
This should only be usable by command access

QuietlyWhisper (changes requested):

Resources/Maps/_RMC14/almayer.yml
tsp01: Done.

Resources/Prototypes/_RMC14/Entities/Objects/Devices/command_tablet.yml
```diff
@@ -0,0 +1,21 @@
...
```
QuietlyWhisper, line 12:
```suggestion
  - type: ActivatableUI
    key: enum.MarineCommunicationsComputerUI.Key
```

Resources/Prototypes/_RMC14/Entities/Objects/Devices/command_tablet.yml
```diff
@@ -0,0 +1,22 @@
...
+    key: enum.MarineCommunicationsComputerUI.Key
+  - type: ActivatableUIRequiresAccess
+  - type: AccessReader
+    access: [ [ "CMAccessCommand" ] ]
+    key: enum.MarineCommunicationsComputerUI.Key
```
DrSmugleaf, line 17: The linter fail is because this doesn't need to be here, there is no key property in AccessReader.
It is needed in ActivatableUI and UserInterface though.
```suggestion
```

DrSmugleaf (approved):
Thanks for the PR!

tsp01:
> Thanks for the PR!

Thank you as well, and have an excellent day.

PR #2703, opened 2024-07-01
Disposable Rocket Launcher

KittenColony:
## About the PR
Introducing the **Rapid Anti Structure / Kinetic Explosive Launcher**, or **RAS/KEL**. _(patent pending)_ 

An Infantry-portable disposable anti-structure launcher. Featuring a limited yield rocket in a cheap tube, allowing you to **THUMP!** and dump. Similar in design to classics such as the M79 LAW and RPG-27

## Why / Balance
It's something new for marines to play around with, as well as opening up new opportunities and strategies, for both Marines to use, and Xenoids to counter. It acts as both a breaching tool against hives and an AT option against heavy xenoids.

The primary drawback is the single use nature, and the difficulty carrying them, as it takes either the suit storage or back slot to carry. If someone was particularly insane they could bring four with them, but it would be their only weapon, barring the potential for a pistol.

## Media

[ PROPAGANDA PHOTOGRAPHY WILL BE PRESENTED SOON]

- [ ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl:

* Add: New disposable rocket launchers for Grunts to play with.

PR #2856, opened 2024-07-06
Increase the cooldown of the Queen's screech to 2.5 minutes

Mattshark89:
## About the PR

See title. It used to be fifty seconds, now it's 150, effectively tripling its cooldown.

## Why / Balance

The queen's screech at the moment is an incredibly strong ability, able to immobilize a squad and guarantee at least one marine capture for the Queen herself (or several if grouped up with runners and a couple of bullet tanks). Coming from a frequent Queen player myself, the captures you can get from this feel very free. While the ability is nice for the Xenoid side (and the gameplay pacing as a whole) in breaking stalemates, this screech happens a little too often. There isn't really any marine counterplay either besides just standing back so far away from the Xenoid hive behind cades (focusing on suppressive fire) that nobody on either side is able to gain any ground.

I like how strong the Queen feels; she should be a major threat. But the screech (in my opinion) should be used more as a trap or a surprise attack. In screech's current state it's no surprise anymore given how often we see it on the frontline. This cooldown should help it feel a little more "special" from both sides.

## Technical details

The useDelay field has been increased from 50 to 150.

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: The Queen's screech delay has been increased from 50 seconds to 150 seconds.


Mattshark89:
As always, this is just how I feel about the ability. I'm open to discussion.

aspiringLich:
I feel like the main problem with the screech is the effort differential on the Xeno and Marine's part. Marines can push a choke for 5 minutes, slowly fortifying, and expending a lot of ammo and resources, and the queen can stroll in, screech, grab 2 marines, and undo all that work in 30 seconds.

DrSmugleaf (changes requested):
The Screech ability is intended to be a very strong ability that causes breakthroughs, we won't be changing the cooldown of it from 50 seconds as that is what it is in cmss13. 150 seconds is a ridiculous amount.

A good tweak to make it to it though would be to make it not affect marines through walls. It doing so is not intended.

https://github.com/RMC-14/RMC-14/assets/10968691/a4f9fc23-1902-4bc8-8338-2e3d2d1214f7

Here's a video of it in cmss13 not working through walls as intendede
https://github.com/RMC-14/RMC-14/assets/10968691/e69b4258-fb94-4130-8bfc-d9ca0d79934d

That would require doing a in range unobstructed check that is blocked by walls in the code with the entities found.

If you want something that doesn't require code, another one would be to implement the SADAR kit for weapon spec, that directly counters aggressive play from the Queen. You could port the sprites and/or write the YAML for the items. Here's the wiki page for it: https://cm-ss13.com/wiki/Weapons_Specialist#Demolitionist_Loadout

Mattshark89:
> 150 seconds is a ridiculous amount.
Gotcha. I wasn't comparing it to CM13's balance.
> That would require doing a in range unobstructed check that is blocked by walls in the code with the entities found.
I assume this only applies for solid walls? I think I know a way to do this (or at least what SS14 code to look at).
> If you want something that doesn't require code, another one would be to implement the SADAR kit for weapon spec, that directly counters aggressive play from the Queen. You could port the sprites and/or write the YAML for the items. Here's the wiki page for it: https://cm-ss13.com/wiki/Weapons_Specialist#Demolitionist_Loadout
I saw the WIP PR for the disposable rocket launcher (https://github.com/RMC-14/RMC-14/pull/2703). I reckon the demo kit's RPG could benefit from that merge.

So the 150 cooldown won't be happening. Do you want this PR closed or renamed something something "screech balance"?

DrSmugleaf:
> I assume this only applies for solid walls and not cades. I think I know a way to do this (or at least what SS14 code to look at).

Yes, I think so, should be doable depending on the collision layers that are passed to InRangeUnobstructed unless there's a better way to do it.

> I saw the WIP PR for the disposable rocket launcher (#2703). I reckon the demo kit's RPG could benefit from that merge.

I talked with KittenColony about it, we can reuse some of the code if relevant to replicate the SADAR but we don't want to merge something like a disposable rocket launcher until we have everything from 13 ported, at which point we can consider new additions like that.

> So the 150 cooldown won't be happening.

Unfortunately (for the marines) no, sorry.

---

Also, I forgot to say, another code change is to make it so the xenos don't get any evolution points 5 minutes after round-start like in cmss13. The code for it is mostly there, but it needs to give the Queen a popup when that threshold of time has passed, a confirmation dialog when exiting the ovipositor, and ideally some kind of way for xenos to tell that they aren't getting points because their Queen isn't in ovi. This should probably be in the way of the evolution window always being visible, listing current/max points, and in red say "The Queen isn't on their ovipositor!"

This would change the dynamic of Queen screech from a purely "use it whenever off cooldown" to a high risk high reward trade-off for the Queen, as any second spent in the front-lines trying to be aggressive is a second not spent giving her xenos evolution points.

Mattshark89:
> Yes, I think so, should be doable depending on the collision layers that are passed to InRangeUnobstructed unless there's a better way to do it.

I could compare it to SS14's ninja dash `InRangeUnOccluded` function too. I think the screech obstruction right now is looking like the most likely thing I could contribute at the moment.

> I talked with KittenColony about it, we can reuse some of the code if relevant to replicate the SADAR but we don't want to merge something like a disposable rocket launcher until we have everything from 13 ported, at which point we can consider new additions like that.

Makes sense. I might look at weapon porting later.

> any second spent in the front-lines trying to be aggressive is a second not spent giving her xenos evolution points.

Okay, this does change up things a bit. I didn't realize Xenos simply couldn't evolve if their Queen didn't put in her ovi time. That would reduce the frequency of screeches for sure (at least once people begin to realize its implications).

Mattshark89:
> A good tweak to make it to it though would be to make it not affect marines through walls. It doing so is not intended.

Figured it out! (Finally, a change of mine that isn't just a stat shift)

https://github.com/RMC-14/RMC-14/assets/32344718/6a72da9c-78a1-42c3-bbdf-1ee347595493

Is this worth a new PR or should I just rename this one?

Mattshark89:
I'll make this into a new PR. For simplicity's sake.

PR #2861, opened 2024-07-06
Add remote lockable doors for Dropships (Draft because it keeps crashing and I would like some help with that)

blueDev2:
## About the PR

Add new component, LockableDropshipDoorComponent, which has a field for a localized name.
All doors with the same localized name are bolted and unbolted together from the navigation computer.

Added YMLs for the new doors of current Dropships
## Media

I was unable to test this component as the YMLs for the dropships are private. Also the game kept crashing when I tried to even search up the navigation computer console

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Client/_RMC14/Dropship/DropshipNavigationBui.cs
+++ b/Content.Client/_RMC14/Dropship/DropshipNavigationBui.cs
@@ -1,7 +1,11 @@
-﻿using Content.Client.Message;
+using Content.Client.Message;
 using Content.Shared._RMC14.Dropship;
+using Content.Shared.Doors.Components;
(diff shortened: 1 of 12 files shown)


blueDev2:
So it builds properly, but when trying to run the server file, it keeps on crashing with this error message:
`[FATL] unhandled: System.NotSupportedException: Type Robust.Shared.GameObjects.EntityUid is not marked as Serializable`
I have absolutely no clue what this error is trying to tell me.

vero5123:
https://docs.spacestation14.com/en/robust-toolbox/netcode/net-entities.html

blueDev2:
DrSmugleaf has already created a PR to fix the issue

PR #2877, opened 2024-07-07
Fix snouts not being hidden when a mask is equipped

Dutch-VanDerLinde:
## About the PR
hideLayersOnEquip is needed in HumanoidAppearance so the layers are actually hidden

## Media
Before:
image

After:
image
image

**Changelog**
no

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Species/reptilian.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Species/reptilian.yml
@@ -7,6 +7,10 @@
   components:
   - type: HumanoidAppearance
     species: Reptilian
+    hideLayersOnEquip:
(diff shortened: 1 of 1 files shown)


DrSmugleaf (commented):
Thanks for the PR!
I need to check if this has any bugs, I know upstream did this at some point but then stopped since it was buggy.

Dutch-VanDerLinde:
> Thanks for the PR! I need to check if this has any bugs, I know upstream did this at some point but then stopped since it was buggy.

upstream has the same implementation as this PR

QuietlyWhisper:
Sterile mask will probably need a displacement or species specific resprite but other than that this seems fine
image
image

QuietlyWhisper (approved):

PR #2886, opened 2024-07-08
Adds 5 hours required time to play parasite

Vermidia:
## About the PR

To play a parasite you now need at least 5 hours of Xenonid time

## Why / Balance

Parasites are often called suicidal by xenonids by how often they rush to the front and die. Some of these are probably salty marines, but a lot are also probably new players that don't know exactly what to do yet. Putting a time requirement limits them to players that already know the xenonid game loop and should hopefully have them already know what to do as a parasite.

Even though this limits parasites, lesser drones will still be able to be taken freely, as they are a less crucible compared to the limited eggs the Queen currently produces, and already are a sort of 'fodder' extra role anyone should be able to take.

## Technical details

N/A

## Media

paratime

## Breaking changes

**Changelog**

:cl:
- tweak: Playing a parasite now requires 5 hours of xenonid time

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/parasite.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/parasite.yml
@@ -7,6 +7,10 @@
   components:
   - type: GhostRole
     name: cm-job-name-xeno-parasite
+    requirements:
+    - !type:DepartmentTimeRequirement
+      department: CMXeno
+      time: 18000 #5 hours
   - type: Sprite
     sprite: _RMC14/Mobs/Xenonids/Parasite/parasite.rsi
   - type: MobState


DrSmugleaf (changes requested):
I definitely see the problem with parasites suiciding, but I'm not sure if this is the best way to fix it.
I think 13 deals with it by having a cooldown between taking over parasites or just xenos in general for each player, so for example if you take a parasite and die, you would need to wait x amount of minutes before being able to take over another again.
I'd rather we implement that (which needs xeno queue), specially since 5 hours is rather high, I'll look into the details of it today or tomorrow

Vermidia:
That's fair. I did think 5 hours was a bit high, but if you're already working on time between xenos/xeno queue this won't be needed so I'll close it.

PR #2915, opened 2024-07-08
Fix health analyzer display

Vermidia:
## About the PR

The health analyzer's bar correctly now is fully filled at 100%
Also, negative %'s now scale with death threshold if its different from a mob's crit threshold

## Why / Balance

Bug, more intuitive that -100% = Dead

## Technical details

HealthBar.MaxValue is always 100 now (we always give it a value out of 100 so no need to change with threshold)
There's now a check if damage is under 0 and we have a dead threshold and it's not equal to the previous gotten threshold, to change the threshold to dead - crit so it scales correctly with the death state of the mob.

## Media

Almost crit
Dead

## Breaking changes

No?
**Changelog**
nah


DrSmugleaf (approved):
Thanks for the PR!

PR #2926, opened 2024-07-09
Window frame dump

Cu1r:
## About the PR
Added all the window frame textures, a base, and the variant I wanted to map.

## Why / Balance
Frames are the next step toward partially destructible and repairable windows

## Media
windowframe


DrSmugleaf (approved):
Thanks for the PR!

PR #2941, opened 2024-07-10
Demolitionist Specialist

Kaga-404:
## About the PR

Demo Spec ported, main gameplay features complete.
Several backend updates to gun code to make it flexible enough to get this to work. Probably extremely unrefined, but it works.
Documentation for the GunSystem and BallisticAmmoProvider stuff that I had to dig through to get this to work.

Also a few minor fixes because they showed up while I was coding. CM-ported projectile textures being upside down, grenade spec not being able to put launcher on backslot, and the grenade belts not being available though they were marked in all the vendors.

Spec Kits are now limited to 2 copies of each. This can be lowered to 1 with a YML change once the other 2 spec kits are in.

Missing Features:
- Spent Shell breakdown: Spent shells need a component to break down into 2 metal sheets.
- Assisted Reload: Someone with a shell in hand within 45 degrees behind the spec while they have the M5 wielded should be able to click on the spec to load in 3s, twice as fast.
- Smoke: Firing should generate 2 smoke clouds, one on top of the spec, 1 behind. This is mostly cosmetic and can be tweaked if necessary for the SS14 engine, since pixel movement makes tile smoke work a lot weirder for this.
- Backblast and Hearing Protection: **Requires Ear Damage/Deafness.** M3-T Helmet should give hearing protection. Not having it on will make you deaf if you fire the M5, and damage + stun for 12 seconds + deafen if you're within 1m, 45 degrees behind when it fires. M3-T should block all effects of backblast, allowing for assisted reloads to be safe.
- WP: **Needs tile fire first.**
- Custom Rockets: **Requires full chemistry and tilefire port.** All the icons except the warheads are implemented already with this.
- Binoculars and C4: Missing parts of kit.

Additional Suggestions:
- Range adjustments for RMC14: Notes are below, increasing short ranges by 10-20%
## Why / Balance

Roadmap, porting.

## Technical details

Don't think I can summarize this at a high level, but what I've messed with are:

1. There's a new AmmoSource variable on ProjectileComponents. There were already Shooter and Weapon vars to keep it from hitting itself, AmmoSource is the same thing except for big casings. This was needed because the M5 drops its casings and that was causing collision issues.
2. AmmoSource is set during ShootOrThrow in GunSystem, in all cases. It's a nullable EntityUid so it's fine even if something's weird about a special projectile or ammo. like stuff firing itself, or deleted cartridges.
3. Its only purpose is to be another entity for PreventCollideEvent to ignore for Projectiles in SharedProjectileSystem
4. Renamed AmmoFixedDistanceComponent to ShootAtFixedPointComponent. The old name kept tripping up because AmmoComponents usually have Ammo in their name, like CartridgeAmmo, ChemicalAmmo, AmmoComponent, and so on. FixedDistance also felt like it was supposed to be a component that would make a weapon fire a specific distance based on a var, not just precisely where you clicked.
5. Added DemoSpecWhitelist, simple empty component to be spec skill.
6. Added **MaxFixedRange** to ShootAtFixedPointComponent and ProjectileComponent. This var defaults to 0, which means no max range. The max range only applies when firing using the ShootAtFixedPointsComponent, it's not a general purpose range limiter. It functions by taking the minimum nonzero between the projectile source with SAFP and the projectile, which becomes the range cap in tiles. Then it calculates normally to the clicked point, and then truncates that vector's length directly with the range cap. **The actual in-world range seems to come out to +1 of the configured number for some  unknown reason, might be something with coordinates. What this means is that the direct conversions were given -1 range, which is about the same for direct impact, but is slightly worse for max range shots because tile-based movement meant hitting a tile guaranteed full damage on it, but here it can explode short on collide.**
7. Added **Arc** to SAFP and ProjectileFixedDistanceComponent. I didn't change that name because it was more accurate to what it did, it's a projectile component, and it gives it a fixed distance. Also added documentation to better indicate that it's pretty much a timer on a projectile (which doesn't need a ProjectileComponent, just anything fired, so it works on ballistic non-ammo like HEDP too), which is applied by SAFP and directly uses the final calculated distance to measure how long it'll take to travel there with the projectilespeed. Once it gets there, the component deletes itself, which signals to stop the projectile in place even if it hasn't hit anything.
8. **Arc** is in the form of two variables, one on SAFP (guns) called ShootArc and the other on ProjectileFixed called ProjectileArcing. They're simple booleans, and ShootArc directly sets ProjectileFixed. Then, the actual PreventCollideEvent checks are on the projectile instead, until the component is deleted. It functions by checking if a collision is with an impassable object through a basic bit flag, and then marking the event to cancel the collision by args.Cancelled = true if it's not impassable and ProjectileArcing is true.
9. BallisticAmmoProvider changes:
- Loading non-BAP items into a BAP can now have a DoAfter, Set InsertDelay > 0
- Unloading a BAP can now have a DoAfter. This just used the existing Cycle function, as a direct port of using another hand seemed excessive for very little benefit. Set CycleDelay > 0, make sure that Cycleable is toggled. I also technically updated the 'UseInHandEvemt' function of BAPs which is supposed to toggle a cycle too like the verb, and made both it and the verb pass the relevant vars into a new BallisticCycleDelayCheckfunction instead of dircetly going to ManualCycle, but that only really matters for things like emptying shells from a shotgun box.
- FillDelay was converted from a directly defined TimeSpan.FromSeconds(0.5) to a double, which is then converted into TimeSpans are needed. This was also used for basically every other timer-related variable in this project, so they could be customized in prototypes. There's probably better ways to do this but I'm not aware of any.
- Cycling and Inserting now have associated DoAfterEvents, and logic for that.
- **Fixed a bug with the existing DoAfterEvent of loading a BAP from another BAP.** Previously, if you moved during the DoAfterEvent channel, it would cancel but there was no logic to check args.Cancelled, so it would just call the full OnDoAfter functions and complete the action. This wasn't really that impactful since the timer was always 0.5s before, and it was only relevant to moving ammo between mags or reloading shotguns, but with the M5 having a 6-second cooldown that's been fixed.
10. Made a substitute 'TriggerOnStop' by adding a 0.2s TriggerOnTimer on rockets, which means even if they land short without a collision it should still explode fast enough to see seamless. Ideally this will be a component in the future.
11.  AP rockets don't rely on the explosion code directly, they have hardcoded damage that's roughly equivalent instead. Current explosions might never be able to substitute, so having an explosive damagetype or something of that sort that can be reduced by ExplosiveArmor would nice.
12. The explosion numbers are balanced around 5 brute/5 burn, and xenos having the proper modifier from the HEDP/HEFA PR, so they'll be about 45% effective in terms of damage for now.
13. **SS14 viewrange looks like 9 tile radius, while SS13 is 7.5**; this difference means ranges might need to be tweaked. Right now, the range numbers are exactly like source. Source HE is max_range 7, which is self + 6 tiles, while MaxFixedRange = 6 which is 6 guaranteed squares + 1 depending on where you are standing on your tile, but it will line up at 7m. My suggested changes are increasing all range by 10-20% to compensate for the 20% increase in view radius and the difference of tile movement. This is _including_ the hidden +1, so with 10% MaxFixedRange = 6 should become MaxFixedRange = 7.4, MaxFixedRange = 7 should become 8.6, and 5 becomes 6.2
14. There's other few missing bits I'll list in this PR for the future. Most of them don't interact with the direct demo mechanics and are just flavor, or depend on features we don't have yet.
## Media

## Breaking changes

I have no idea what to do for this part. I guess AmmoFixedDistanceComponent becoming ShootAtFixedPointComponent, the function renames in CMGunSystem and SharedGunSystem.Ballistics, the changing of FillDelay to a double var, and renaming a few prototypes?
**Changelog**

:cl: Kaga
- add: Added the M5 Anti-Tank Launcher, the Demolitionist Weapon Specialist, and their associated gear.
- add: Added Demolitionist ammo to the requisition console.
- add: Added M276 pattern grenade belts to most vendors they're supposed to be in. They're empty, but can hold a decent amount of grenades.
- add: Made limited max range possible for some weapons, which can differ between ammo types. Currently, this only applies to the M5-ATL.
- fix: Fixed some bullets appearing backwards.
- tweak: Weapons specialists are now limited to 2 kits of each type, meaning that there can ever only be 2 snipers, 2 grenadiers or 2 demolitionists. 13's limit of 1 per will be introduced once we have more specialist kits.


JoeHammad1844:
GOOD HITS SADAR
THE SL WAS A VERY DANGEROUS XENO

please tell me the SADAR can gib marines

JoeHammad1844:
by the way weild delay has been merged, you can remove that from missing features

DrSmugleaf:
I need to look through the code still, but before this is published we'll need the mechanic from 13 that limits weapons spec kits vended to 1 per specialist (2 per for us for now since we will only have 3 and the sniper is unfinished at the moment), assuming that 13 still has that mechanic

Content.Shared/Projectiles/ProjectileComponent.cs
```diff
@@ -26,6 +26,12 @@ public sealed partial class ProjectileComponent : Component
     [DataField, AutoNetworkedField]
     public EntityUid? Weapon;
 
+    /// <summary>
+    ///     Ammo used to shoot this projectile.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid? AmmoSource;
```
DrSmugleaf, line 33: This field shouldn't be necessary from looking at the rest of the code, if you need more information you should be able to put it on the projectile with new components as well
That gets rid of most of the ammo code changes

Content.Shared/Projectiles/ProjectileComponent.cs
```diff
@@ -73,4 +79,12 @@ public sealed partial class ProjectileComponent : Component
...
     [DataField]
     public bool DamagedEntity;
+
+    /// <summary>
+    /// Sets the maximum range for a projectile fired with ShootAtFixedPointComponent.
+    /// This can be set on both the Projectile and ShootAtFixedPoint Components.
+    /// The default value is null for no cap. The minimum value between the two is used.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public float? MaxFixedRange;
```
DrSmugleaf, line 89: I added a new component that should be able to replace this with the Boiler PR https://github.com/RMC-14/RMC-14/pull/3258/files, ProjectileMaxRangeComponent

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -28,18 +28,60 @@ protected virtual void InitializeBallistic()
         SubscribeLocalEvent<BallisticAmmoProviderComponent, InteractUsingEvent>(OnBallisticInteractUsing);
         SubscribeLocalEvent<BallisticAmmoProviderComponent, AfterInteractEvent>(OnBallisticAfterInteract);
         SubscribeLocalEvent<BallisticAmmoProviderComponent, AmmoFillDoAfterEvent>(OnBallisticAmmoFillDoAfter);
+        SubscribeLocalEvent<BallisticAmmoProviderComponent, DelayedAmmoInsertDoAfterEvent>(OnBallisticDelayedAmmoInsertDoAfter);
+        SubscribeLocalEvent<BallisticAmmoProviderComponent, DelayedCycleDoAfterEvent>(OnBallisticDelayedCycleDoAfter);
```
DrSmugleaf, line 32: These two should be moved to be subscriptions on CMGunSystem

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -51,15 +93,29 @@ private void OnBallisticInteractUsing(EntityUid uid, BallisticAmmoProviderCompon
...
+            {
+                BreakOnMove = true,
+                BreakOnDamage = false,
+                NeedHand = true
+            });
+        }
+        else // If there's no custom InsertDelay on this component, immediately load the ammo in. Shotguns and mag-filling.
+        {
+            ManualLoad(uid, component, args.Used, args.User);
+        }
```
DrSmugleaf, line 111: This should probably be upstreamed, need to ask Sloth

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -51,15 +93,29 @@ private void OnBallisticInteractUsing(EntityUid uid, BallisticAmmoProviderCompon
...
         args.Handled = true;
-        UpdateBallisticAppearance(uid, component);
-        Dirty(uid, component);
     }
 
+    /// <summary>
+    /// Interacting with a BallisticAmmoProvider with another one, to transfer ammo.
+    /// Uses FillDelay, defaulting to 0.5s
+    /// </summary>
```
DrSmugleaf, line 118: If it's upstream code not commenting it or commenting it with an upstream PR is better, for conflicts

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
```diff
@@ -113,7 +179,7 @@ private void OnBallisticAmmoFillDoAfter(EntityUid uid, BallisticAmmoProviderComp
                 args.User);
             return;
         }
-
+        // Simulates using a single ammo entity on the other BAP, loading it in.
```
DrSmugleaf, line 182: Same with this

Content.Shared/_RMC14/Vendors/SharedCMAutomatedVendorSystem.cs
```diff
@@ -129,6 +129,63 @@ protected virtual void OnVendBui(Entity<CMAutomatedVendorComponent> vendor, ref
             Dirty(actor, user);
         }
 
+        if (section.SharedSpecLimit is { } globalLimit)
+        {
+            if (HasComp<RMCVendorSpecialistComponent>(vendor))
+            {
```
DrSmugleaf, line 135: These two checks can be placed inside the same if statement to reduce indentation

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/m5spec.yml
```diff
@@ -0,0 +1,133 @@
...
+  parent: RMCProjectileRocket84mm
+  noSpawn: true
+  components:
+  - type: Projectile
+    impactEffect: BulletImpactEffect
+    damage:
+      types:
+        CMArmorPiercing: 50
+        Brute: 160
+        Burn: 150
```
DrSmugleaf, line 111: holy shit

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/m5spec.yml
```diff
@@ -0,0 +1,133 @@
...
+        # 10 brute damage + 150 brute and 150 burn to simulate direct explosive damage.
+        # Temporary mechanic to make it consistent with intended main target damage.
+        # TODO RMC14 Explosive damagetype that gets reduced by ExplosionArmor in CMArmorSystem and converts to Brute and Burn.
+    MaxFixedRange: 5 # 6 tiles
```
DrSmugleaf, line 115: YML fields are camelCase
```suggestion
    maxFixedRange: 5 # 6 tiles
```

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/m5spec.yml
```diff
@@ -0,0 +1,133 @@
...
+  id: RMCProjectileRocket84mmAntiArmor
+  name: 84mm anti-armor rocket
+  parent: RMCProjectileRocket84mm
+  noSpawn: true
+  components:
+  - type: Projectile
+    impactEffect: BulletImpactEffect
+    damage:
+      types:
+        CMArmorPiercing: 50
```
DrSmugleaf, line 109: Is this a leftover from old code? The damage type doesn't seem to exist (not that it should either)
```suggestion
```

DrSmugleaf (changes requested):
SADAR MAJOR

I'll look into getting rid of most of the upstream changes here in the future, for now I want this for tomorrow, the comments are mostly notes for myself unless you want to go through them yourself

Content.Shared/_RMC14/Vendors/RMCVendorSpecialistComponent.cs
```diff
@@ -0,0 +1,16 @@
...
+/// <summary>
+/// Component used by Specialist Vendors to restrict duplicate kits. Stores successful vends based on Entry int.
+/// Since only spec kits need this for porting, all logic specifically uses this component.
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(SharedCMAutomatedVendorSystem))]
+public sealed partial class RMCVendorSpecialistComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public Dictionary<int, int> GlobalSharedVends = new Dictionary<int, int>();
```
DrSmugleaf, line 15: This doesn't account for weapons specs cryoing and freeing up their gear

DrSmugleaf (changes requested):

PR #2946, opened 2024-07-10
Increase default projectile speed from 41 to 51

DrSmugleaf:
## About the PR
Should be good now with https://github.com/RMC-14/RMC-14/pull/2889
This does mean any further projectile speed increases )extended barrel for example) don't work until we increase physics.maxlinvelocity, but the default was slightly lower than 13's.

## Media

**Changelog**
:cl:
- tweak: Default projectile speed has been increased by 24%


UnicornOnLSD:
WOOOOOO THANKS SMUG

PR #2957, opened 2024-07-10
fix m1984 regular mags accepting 9mm AP

QuietlyWhisper:
## Why / Balance

1984

**Changelog**

:cl: Whisper

- fix: Fixed M1984 accepting 9mm AP in regular mags.


DrSmugleaf (approved):
1984

PR #2973, opened 2024-07-11
Actual Wield Delay

Kaga-404:
## About the PR

Wield delay now also applies when you wield, and prevents shooting.

There should be multiple types of wield delays from, from worse accuracy, worse spread, or just no shooting at all.

This only adds no shooting for now because this is directly relevant to the Demolitionist Spec PR, which needs is supposed to have to wait after wielding to fire. Future updates should be done when accuracy and spread modifiers are properly in.

## Why / Balance

Attachments update only added delay on equipping, which made it so you can't wield until the timer is up.

Wield delay is actually supposed to be used twice, once when equipping and once when wielding, which should cause various effects depending on the type.

## Media

## Breaking changes

**Changelog**

:cl: Kaga
- fix: Wield delay now applies to shooting after wielding.


DrSmugleaf (approved):
Thanks for the PR!

PR #2989, opened 2024-07-11
Fixes attachment size modifiers

Siigiil:
## About the PR
Attachments can no longer permanently increase or reduce the size of the thing they're attached to. They can no longer make the size invalid.

## Why / Balance
Because bug.

## Technical details
Changed how size modifiers work to allow for things other than attachments to apply those modifiers. Also made the code less shit.

## Media

**Changelog**
:cl: Sigil
- fix: Attachments that modify the size of their holder now properly apply and remove their modifiers.


DrSmugleaf (approved):
Thanks for the pr!

PR #2990, opened 2024-07-11
Port the XM88

### LINKED ISSUE #875: XM88

### LINKED ISSUE #1136: XM88 Heavy Rifle crate (XM88 x2) Cost: $4000

krose124:
## About the PR
Adds the XM88 in general, based off the cm13 wiki.

## Why / Balance

Porting, more specifically:
Resolves #1136 
Resolves #875
## Technical details

This should add ammo, gun, req purchase, xm88 unique mechanic, and more.
## Media

image
image

https://github.com/user-attachments/assets/49bf81c2-105d-422a-affd-c628b5f55dca

https://github.com/user-attachments/assets/2012c9e1-20b6-43db-8a28-0f6a6bb7d83d

 This does not add the XM88 stock (afaik its only available to the SL kit for the XM88 and this does not implement that kit).
 The SOCOM box is supposed to have 60 handfuls in it, however it causes great lag when the xm88 crate is spawned. As a result of that finding, I lowered it to 30 handfuls (still a considerable amount of ammo) and the spawning no longer lags.

## Breaking changes

**Changelog**
:cl: krose124, DrSmugleaf
- add: Added XM88 heavy rifle and rig. The XM88 has more AP, damage, and faster reload when hitting every shot, but will lose all AP if any shot is missed. They can be bought in the Squad Leader vendor and in Requisitions.


krose124:
While I am currently thinking about this, I think the crate texture for both the XM88 and the SOCOM bullets are missing currently.

krose124:
I assume the overhauled sounds ( #3106 ) will get merged before this because I'm busy(cause thesis and I'm moving soon too) so I plan on stealing the sounds from the scoped rifle for the xm88

krose124:
I might have accidentally broke the HPR in the merge

krose124:
Rifle.yml in ammo magazines 
"Item with the same key has already been added" 
I guess this also presents the opportunity to check the yml again

krose124:
from Sigil on discord:
"Also, you'll need to make the XM88 use RMCSelectiveFireComponent for all the values it lets you set, instead of setting them in the GunComponent."

krose124:
458 box is bigger than the shotgun boxes, so I cant use that rsi, Im gonna make an rsi for the 458 stuffs

krose124:
I am not going to do the XM88 stock here, nor will I implement Deerb's idea of using the shotgun stock. My reasoning for the XM88 stock is, the difference made is marginal given the XM88 stock does not change the accuracy while the shotgun stock would increase accuracy significantly. If the community thinks the stock should be implemented, it should be trivial. Although, this is more of a maintainer question as it pertains to the accuracy of CM13, but there do exist ideas that would change the XM88 possibly for the better.

krose124:
As of the most recent commit, the XM88 is fully textured (including ammo box, attachment offsets, magazine, etc.), purchasable by req (XM88 crate for 4k), shoots (80 damage 0 AP) but the unique mechanic is not fully implemented.

krose124:
This is like 90% done, but I'm too busy to finish this in August at least, Smug said they would finish it. If for some reason it's not, then I'll do it once I have time again.

krose124:
Ok, should be a merge and update in that commit. Time to go from there

krose124:
I don't know what those errors in the yaml and build are for, it seems to work fine on my machine.

krose124:
Build was canceled?

DrSmugleaf (changes requested):
This is missing the belt rig, targeting relay and padded stock, and the shooting sound / racking sound
I fixed the bullet count to be parity and the firerate increase not applying

Carolyn3114:
changelog could just be
"add: ap stack mechanic"
and "add: xm88 can now be bought by cargo, which contains 2 XM88s and 2 SOCOM bullet boxes" instead of 4 of the same thing

DrSmugleaf (approved):

PR #2995, opened 2024-07-11
Larva Embryo dies if host becomes unreviviable + more incubation modifiers

Vermidia:
## About the PR

When a host dies and "Rots" the embryo inside the host now dies.

For monkeys/any non humanoid host, if they die at all the embryo dies with them (In CM13 it looks like they just stop processing but never start back up? Either way it's probably fineee)

Also added the missing incubation modifiers. Nesting is 1.5 (so half as long), and stasis bags for stage 4 are 0.44 instead of 0.33 (less effective).

## Why / Balance

CM13 feature parity, gives Xenonids a reason not to immediately kill someone after being infected, while also not effecting balance too much hopefully.

## Technical details

Just checking rot + marine component and deleting

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: Larva dies if a host becomes unreviviable
- add: Nesting speeds up incubation, and stasis bags are less good against late stage infections


Dutch-VanDerLinde:
if this is added, i believe being either hugged/nested should keep slightly out of crit so xenonids are still able to get new larva without a host succumbing to bloodloss etc

Vermidia:
I could try to add the nesting incubation speed increase in here maybe...I'll try

DrSmugleaf:
Is this how it works in 13 then? When I was testing it it was confusing to me, sometimes it would burst past death, sometimes not, is the criteria unrevivable?

Vermidia:
Yep! I checked the code and it checks if the grace revival period has passed or not, basically.

ShadowCommander:
The code from CM13 for multipliers is here:

[hive_status.dm#L50](https://github.com/cmss13-devs/cmss13/blob/cffeff95cb6794c7ec203f0d195d0d84cd89b33d/code/modules/mob/living/carbon/xenomorph/hive_status.dm#L50)
```js
var/larva_gestation_multiplier = 1
```

[Embryo.dm#L87-L103](
https://github.com/cmss13-devs/cmss13/blob/cffeff95cb6794c7ec203f0d195d0d84cd89b33d/code/modules/mob/living/carbon/xenomorph/Embryo.dm#L87-L103)

```js
if(affected_mob.in_stasis || affected_mob.bodytemperature < 170)
	if(stage < 5)
		counter += 0.33 * hive.larva_gestation_multiplier * delta_time
	if(stage == 4) // Stasis affects late-stage less
		counter += 0.11 * hive.larva_gestation_multiplier * delta_time
else if(HAS_TRAIT(affected_mob, TRAIT_NESTED)) //Hosts who are nested in resin nests provide an ideal setting, larva grows faster
	counter += 1.5 * hive.larva_gestation_multiplier * delta_time //Currently twice as much, can be changed
else
	if(stage < 5)
		counter += 1 * hive.larva_gestation_multiplier * delta_time

if(stage < 5 && counter >= per_stage_hugged_time)
	counter = 0
	stage++
	if(iscarbon(affected_mob))
		var/mob/living/carbon/affected_carbon = affected_mob
		affected_carbon.med_hud_set_status()
```

Each stage is 90 seconds, with 5 infection stages for 450 seconds as the editable configured number.

[Embryo.dm#L18-L22](https://github.com/cmss13-devs/cmss13/blob/cffeff95cb6794c7ec203f0d195d0d84cd89b33d/code/modules/mob/living/carbon/xenomorph/Embryo.dm#L18-L22)
```js
var/per_stage_hugged_time = 90 //Set in Initialize due to config
```

```js
per_stage_hugged_time = CONFIG_GET(number/embryo_burst_timer) / 5
```

[game_options.dm#L126-L129](https://github.com/cmss13-devs/cmss13/blob/cffeff95cb6794c7ec203f0d195d0d84cd89b33d/code/controllers/configuration/entries/game_options.dm#L126-L129)
```js
/datum/config_entry/number/embryo_burst_timer
	min_val = 1
	config_entry_value = 450
	integer = TRUE
```

So, stasis is 0.33 (272.7 seconds/stage, 204.5 for stage 4), 1,295.45 for infection in stasis.

ShadowCommander:
The relevant code for death is here:

[Embryo.dm#L61-L67](https://github.com/cmss13-devs/cmss13/blob/cffeff95cb6794c7ec203f0d195d0d84cd89b33d/code/modules/mob/living/carbon/xenomorph/Embryo.dm#L61-L67)
```js
			var/mob/living/carbon/human/affected_human = affected_mob
			if(world.time > affected_human.timeofdeath + affected_human.revive_grace_period) //Can't be defibbed.
				var/mob/living/carbon/xenomorph/larva/larva_embryo = locate() in affected_mob
				if(larva_embryo)
					larva_embryo.chest_burst(affected_mob)
				qdel(src)
				return FALSE
```

Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -394,8 +398,15 @@ public override void Update(float frameTime)
 
             if (infected.BurstAt > time)
             {
-                // TODO RMC14 make this less effective against late-stage infections, also make this support faster incubation
-                if (infected.IncubationMultiplier < 1)
+                // Embryo dies if unrevivable when dead
+                // Kill the embryo if we've rotted or are non-humanoid
+                if (_mobState.IsDead(uid) && (!HasComp<MarineComponent>(uid) || _rotting.IsRotten(uid)))
```
ShadowCommander, line 403: `!HasComp<MarineComponent>(uid)` This should probably check simplemob somehow instead of marine for the case of non-marine humanoids in the future like predators.

Content.Shared/_RMC14/Medical/Stasis/CMStasisBagSystem.cs
```diff
@@ -55,7 +55,14 @@ private void OnInStasisRemove(Entity<CMInStasisComponent> ent, ref ComponentRemo
...
         if (ent.Comp.Running)
-            args.Multiply(ent.Comp.IncubationMultiplier);
+        {
+            // less effective in late stages
+            var multiplier = ent.Comp.IncubationMultiplier;
+            if (TryComp<VictimInfectedComponent>(ent, out var infection) && infection.CurrentStage >= infection.FinalSymptomsStart)
+                multiplier += (multiplier / 3);
+
+            args.Multiply(multiplier);
+        }
```
ShadowCommander, line 65: It's probably better to add the speedup to the embryo code instead of here.

Or maybe pass the stage into the GetInfectedIncubationMultiplierEvent instead of getting the component here.

Content.Shared/_RMC14/Xenonids/Construction/Nest/XenoNestSystem.cs
```diff
@@ -187,6 +192,7 @@ private void OnNestSurfaceDoAfter(Entity<XenoNestSurfaceComponent> ent, ref Xeno
         _transform.SetLocalRotation(victim, direction.ToAngle());
 
         _standing.Stand(victim, force: true);
+        _parasite.RefreshIncubationMultipliers(victim);
```
ShadowCommander, line 195: Can you move this to a <XenoNestedComponent, ComponentAdded> subscription instead?

ShadowCommander (changes requested):

Kadeo64:
I think that if we're going to have this, devouring or nesting should stabilize someone (stop bleeding, possibly inject tricord?) I know CM13 has devour stabilizing.

Vermidia:
I checked CM13, there doesn't seem to be any stabilization on devour. You take acid damage while in a xeno's stomach, and if are somehow in there long enough, actually dissolve completely.

Nesting will already speed up incubation quite a bit and it takes awhile for people to both bleed out and become unrevivable (5 minutes for the latter) so I think it should be balanced enough rn.

DrSmugleaf (approved):
Thanks for the PR!

Issue #3000, opened 2024-07-11
Actually, something is wrong with Queen's screech


Mattshark89:
## Description

Something's been bugged since the screech wall penetration PR https://github.com/RMC-14/RMC-14/pull/2866

In some cases it seems to be blocked by... nothing?

**Reproduction**

See video below

**Screenshots**

https://github.com/user-attachments/assets/a8b1b2b1-23f7-4ea1-8569-20eb25e4f375

**Additional context**

I can still hug the marines, so `InRangeUnobstructed` should be resulting to `true`.

Mattshark89:
...Y'know, I think I found the cause, and given its severity I'll call @DrSmugleaf directly into this one.

I think... marines are considered Opaque. 💀 

https://github.com/user-attachments/assets/43e3b20b-1a72-427c-ad2f-4575ed201d80

DrSmugleaf:
> I think... marines are considered Opaque. 💀

Incredible
Apparently so, so the fix is checking that any blockers are both Opaque and Impassable I think, I don't know if the flag in InRangeUnobstructed checks that something is both or either, but at worst you can pass in your own Ignore predicate.

Mattshark89:
I've just finished experimenting with `InRangeUnOccluded` as an alternative to `InRangeUnobstructed` (used in ninja dash) right now. It appears to work exactly as expected, even fixing the weird windows-being-blockers issue.

However, that is borrowed from the Examine System (Shared) rather than the Shared Interaction System, so I'm unsure if using it can be considered good practice.

https://github.com/user-attachments/assets/c01579be-865c-4896-8139-c538f2f7ec33

I'll save this fix off to the side so I can experiment with Unobstructed more.

DrSmugleaf:
> Or if you want the UnOccluded fix, let me know and I'll get a PR up.

Sounds fine, I'd need to check the code but if you can't find any edge cases it's probably ok

Mattshark89:
fixed by https://github.com/RMC-14/RMC-14/pull/3002

PR #3002, opened 2024-07-12
Fix Queen's screech not properly knocking down marines

Mattshark89:
## About the PR

Some people noticed the queen's screech wasn't behaving as expected, often leaving several people unstunned.

The queen's screech now uses occlusion (what you can see) rather than obstruction (what doesn't have something physical in the way)

fixes https://github.com/RMC-14/RMC-14/issues/3000

## Why / Balance

Apparently, marines are considered "opaque". Standing in the same tile as another marine made you both immune to the screech since you were occluding one another.

Still has parity with CM13's screech, at least intuitively. You can screech through transparent objects there.

## Technical details

Swaps out `SharedInteractionSystem`'s `InRangeUnobstructed` to `ExamineSystemShared`'s `InRangeUnOccluded` (the check used for SS14's ninja dash, for example).

## Media

For showcasing purposes *only*, my local build has the screech cooldown reduced.

Before the bugfix:

https://github.com/user-attachments/assets/b2fe77bc-fd48-4dbb-846d-02ea381f9467

After the bugfix:

https://github.com/user-attachments/assets/3b4353df-d60a-4f54-b9ef-65728c10c40f

## Breaking changes

**Changelog**

:cl:
- fix: The Queen's screech is no longer blocked by marines. Apparently, marines are opaque.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Screech/SharedXenoScreechSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Screech/SharedXenoScreechSystem.cs
@@ -1,9 +1,8 @@
 ﻿using Content.Shared._RMC14.Marines;
 using Content.Shared._RMC14.Xenonids.Plasma;
 using Content.Shared.Coordinates;
-using Content.Shared.Interaction;
+using Content.Shared.Examine;
 using Content.Shared.Mobs.Systems;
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
Thanks for the PR! And the bug investigation

PR #3008, opened 2024-07-12
More Military Police belts and small unbloating

Tunguso4ka:
## About the PR

Added:
- Military Police belt
- MP SPP belt and filled variant
- MP Marshal belt and filled variant

Fixed:
- MP Belt had a wrong id.

Also, unbloated belts.yml.


DrSmugleaf (approved):
Thanks for the PR!

PR #3018, opened 2024-07-12
Fix defib working on xenos

### LINKED ISSUE #3004: Marines can defib xenos
What why does this work lmao, might even be able to revive them but I didnt want to risk it

0xRainy:
## About the PR

Fixes #3004 
Added a check for XenoComponent to defib

Not sure if I should use the popup text I wrote or the unrevivable text

## Why / Balance

While funny, probably shouldn't happen

## Media

Screenshot_20240712_092319

## Breaking changes

**Changelog**

### DIFF
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -9,7 +9,6 @@ using Content.Server.PowerCell;
 using Content.Server.Traits.Assorted;
 using Content.Shared._RMC14.Damage;
 using Content.Shared._RMC14.Medical.Defibrillator;
-using Content.Shared._RMC14.Xenonids;
 using Content.Shared.Damage;
 using Content.Shared.DoAfter;
 using Content.Shared.Interaction;
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -113,14 +113,11 @@ public sealed class DefibrillatorSystem : EntitySystem
         if (!component.CanDefibCrit && _mobState.IsCritical(target, mobState))
             return false;
 
-        if (HasComp<CMDefibrillatorBlockedComponent>(target))
+        if (TryComp(target, out CMDefibrillatorBlockedComponent? block))
         {
-            if (TryComp(target, out CMDefibrillatorBlockedComponent? block))
-            {
-                if (user != null)
-                    _popup.PopupEntity(Loc.GetString(block.Popup), uid, user.Value);
-                return false;
-            }
+            if (user != null)
+                _popup.PopupEntity(Loc.GetString(block.Popup), uid, user.Value);
+            return false;
         }
 
         return true;
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -50,6 +50,7 @@ public sealed class DefibrillatorSystem : EntitySystem
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
     [Dependency] private readonly SharedMindSystem _mind = default!;
     [Dependency] private readonly CMDefibrillatorSystem _cmDefibrillator = default!;
+    [Dependency] private readonly IEntityManager _entityManager = default!;
 
     /// <inheritdoc/>
     public override void Initialize()
@@ -113,10 +114,11 @@ public sealed class DefibrillatorSystem : EntitySystem
         if (!component.CanDefibCrit && _mobState.IsCritical(target, mobState))
             return false;
 
-        if (HasComp<XenoComponent>(target))
+        if (HasComp<CMDefibrillatorBlockedComponent>(target))
         {
+            // var text = _entityManager.GetComponent<CMDefibrillatorBlockedComponent>(target).Popup;
             if (user != null)
-                _popup.PopupEntity(Loc.GetString("defibrillator-xeno"), uid, user.Value);
+                _popup.PopupEntity(_entityManager.GetComponent<CMDefibrillatorBlockedComponent>(target).Popup, uid, user.Value);
             return false;
         }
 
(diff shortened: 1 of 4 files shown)


Dutch-VanDerLinde:
better way: Make DefibBlock component with custom popup field, if they have this component then it blocks defibbing and sends the popup in the component's field

prevents hardcoding defibs to xenos and allows for a custom popup

0xRainy:
> better way: Make DefibBlock component with custom popup field, if they have this component then it blocks defibbing and sends the popup in the component's field
> 
> prevents hardcoding defibs to xenos and allows for a custom popup

I turned it into a component but trying to figure out how to do the popup

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -136,13 +146,13 @@ public bool TryStartZap(EntityUid uid, EntityUid target, EntityUid user, Defibri
...
-                BreakOnMove = !component.AllowDoAfterMovement,
-                DuplicateCondition = DuplicateConditions.SameEvent,
-            });
+        {
+            BlockDuplicate = true,
+            BreakOnHandChange = true,
+            NeedHand = true,
+            BreakOnMove = !component.AllowDoAfterMovement,
+            DuplicateCondition = DuplicateConditions.SameEvent,
+        });
```
DrSmugleaf, line 155: To keep merge conflicts low we don't do code reformatting on upstream files
```suggestion
            {
                BlockDuplicate = true,
                BreakOnHandChange = true,
                NeedHand = true,
                BreakOnMove = !component.AllowDoAfterMovement,
                DuplicateCondition = DuplicateConditions.SameEvent,
            });
```

Content.Shared/_RMC14/Medical/Defibrillator/CMDefibrillatorBlockedComponent.cs
```diff
@@ -0,0 +1,11 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Medical.Defibrillator;
+
+[RegisterComponent, NetworkedComponent]
+[Access(typeof(CMDefibrillatorSystem))]
+public sealed partial class CMDefibrillatorBlockedComponent : Component
+{
+    [DataField]
+    public string Popup = string.Empty;
```
DrSmugleaf, line 10: This should be of type LocId to use the localization message you made

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
```diff
@@ -78,6 +78,8 @@
     - Piercing
     - Heat
     - Shock
+  - type: CMDefibrillatorBlocked
+    popup: "Why would you want to bring this thing back!"
```
DrSmugleaf, line 82: Likewise this should be changed to use the loc id
```suggestion
    popup: defibrillator-xeno
```

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -112,6 +114,14 @@ public bool CanZap(EntityUid uid, EntityUid target, EntityUid? user = null, Defi
         if (!component.CanDefibCrit && _mobState.IsCritical(target, mobState))
             return false;
 
+        if (HasComp<CMDefibrillatorBlockedComponent>(target))
+        {
+            // var text = _entityManager.GetComponent<CMDefibrillatorBlockedComponent>(target).Popup;
+            if (user != null)
+                _popup.PopupEntity(_entityManager.GetComponent<CMDefibrillatorBlockedComponent>(target).Popup, uid, user.Value);
+            return false;
+        }
```
DrSmugleaf, line 123: This can be changed to a TryComp so that you don't need to get the component twice.
```suggestion
        if (TryComp(target, out CMDefibrillatorBlockedComponent? block))
        {
            if (user != null)
                _popup.PopupEntity(block.Popup, uid, user.Value);
            return false;
        }
```

DrSmugleaf (changes requested):
Thanks for the PR!

0xRainy:
Thanks for the feedback, fixed.

Not sure how, or if I should, try and set some default for the component to use if a loc isn't set for a prototype that has the component.

Resources/Locale/en-US/medical/components/defibrillator.ftl
```diff
@@ -1,4 +1,5 @@
-﻿defibrillator-not-on = The defibrillator isn't turned on.
+defibrillator-not-on = The defibrillator isn't turned on.
 defibrillator-no-mind = No intelligence pattern can be detected in patient's brain. Further attempts futile.
 defibrillator-rotten = Body decomposition detected: resuscitation failed.
 defibrillator-unrevivable = This patient is unable to be revived due to a unique body composition.
+defibrillator-xeno = Why would you want to bring this thing back!
```
Dutch-VanDerLinde, line 5: Why would you want to bring this thing back! -->  Why would you want to bring this thing back!?
 
 so its a question

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -112,6 +113,16 @@ public bool CanZap(EntityUid uid, EntityUid target, EntityUid? user = null, Defi
...
 
+        if (HasComp<CMDefibrillatorBlockedComponent>(target))
+        {
+            if (TryComp(target, out CMDefibrillatorBlockedComponent? block))
+            {
+                if (user != null)
+                    _popup.PopupEntity(Loc.GetString(block.Popup), uid, user.Value);
+                return false;
+            }
+        }
```
DrSmugleaf, line 124: You're checking the same thing twice here, TryComp also does a HasComp implicitly.
```suggestion
        if (TryComp(target, out CMDefibrillatorBlockedComponent? block))
        {
            if (user != null)
                _popup.PopupEntity(Loc.GetString(block.Popup), uid, user.Value);
            return false;
        }
```

DrSmugleaf (approved):
Thanks for the PR!

PR #3021, opened 2024-07-12
fix barricades not anchoring again after being unanchored

### LINKED ISSUE #2975: Unable to anchor barricades
## Description

Suspecting it's #2894

It's calling Anchor, but then calling Unanchor in 

https://github.com/RMC-14/RMC-14/blob/7f775c8eaec9427ff8e3077ee1c0598d2e70e08b/Content.Shared/_RMC14/Construction/RMCConstructionSystem.cs#L99-L103

**Reproduction**

- Construct barricade
- Wrench to unanchor
- Wrench again to try to anchor
- Barricade is not anchored

**Screenshots**

**Additional context**

MendaxxDev:
## About the PR
Closes #2975. I don't really know if this was here for a reason and removing it will break everything, but by doing so it's possible to actually create a barricade, unanchor it, move it around and then anchor it again for good (that is, it can't be pulled even after being anchored like before)

## Why / Balance
To make anchoring of barricades work as intended

## Media

## Breaking changes

**Changelog**
:cl: MendaxxDev, DrSmugleaf
- fix: Fixed not being able to anchor barricades outside of dropships.

### DIFF
--- a/Content.Shared/_RMC14/Construction/RMCConstructionSystem.cs
+++ b/Content.Shared/_RMC14/Construction/RMCConstructionSystem.cs
@@ -33,7 +33,6 @@ public sealed class RMCConstructionSystem : EntitySystem
 
         SubscribeLocalEvent<RMCDropshipBlockedComponent, MapInitEvent>(OnMapInit);
         SubscribeLocalEvent<RMCDropshipBlockedComponent, AnchorAttemptEvent>(OnAnchorAttempt);
-        SubscribeLocalEvent<RMCDropshipBlockedComponent, UserAnchoredEvent>(OnUserAnchored);
     }
 
     private void OnConstructionAttempt(ref RMCConstructionAttemptEvent ev)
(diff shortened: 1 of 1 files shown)


Dutch-VanDerLinde:
believe that was for preventing the anchoring of barricades on dropships

Dutch-VanDerLinde:
so make it unanchor when its on a dropship instead of ALWAYS unanchoring

MendaxxDev:
Is there a reason on why it should not be possible to make cades on dropships?

Dutch-VanDerLinde:
because maintainers dont want there to be buildable barricades on dropships, you use the dropship's door bolts for defense

and for CM13 parity i believe

Dutch-VanDerLinde:
you also cant build on dropships currently, so building it elsewhere then anchoring it on the ship would be a way to circumvent it

MendaxxDev:
Mhh, I can kinda see why this was the fix then. I have no idea on how to discriminate whether anchoring is occuring inside a dropship or not. I guess I'll keep this open and I'll investigate it, maybe in the meantime someone comes up with an idea.

DrSmugleaf:
> Mhh, I can kinda see why this was the fix then. I have no idea on how to discriminate whether anchoring is occuring inside a dropship or not. I guess I'll keep this open and I'll investigate it, maybe in the meantime someone comes up with an idea.

Need to get the transform component of the barricade and then check whether or not its grid uid has a dropship component, I forgot to do that when implementing it

DrSmugleaf (approved):
See [`2b16f4d` (#3021)](https://github.com/RMC-14/RMC-14/pull/3021/commits/2b16f4d18adaff608d23af47f327fbf732c51091) for the dropship check

PR #3042, opened 2024-07-13
Update Savannah: Remove poddoors, add more bureaucracy, lathes

Tunguso4ka:
## About the PR

- Removed PodDoors
- Added more paper, pens, clipboards, folders and hand labelers
- Added lathes
- Disabled some APC, because they are bugged
- Added Jones!
- Added new hijack destination


DrSmugleaf (approved):
Thanks for the PR!

PR #3061, opened 2024-07-14
First set of admin tweaks

TsjipTsjip:
## About the PR

- Adds High Command access level, ID holobadge and airlock.
- Adds the admin access configurator, a new prototype parenting off of upstream's universal access configurator with all of the CM access levels.
- Updates the aghost loadout with these new items, adding a health analyzer as that was in the Admin PDA, which is now gone. (It was upstream's health analyzer anyways so...)

## Why / Balance

Admin QoL.

## Media

Access configurator:
image

Airlock and ID card in spawn panel:
image

## Breaking changes

**Changelog**

Zippo/Zilch.

### DIFF
--- a/Resources/Prototypes/Roles/Jobs/Fun/misc_startinggear.yml
+++ b/Resources/Prototypes/Roles/Jobs/Fun/misc_startinggear.yml
@@ -175,12 +175,12 @@
   id: MobAghostGear
   equipment:
     back: ClothingBackpackSatchelHolding
-    id: AdminPDA
+    id: CMIDCardHighCommand # RMC-14: Changed from AdminPDA
   storage:
     back:
-    - GasAnalyzer
-    - trayScanner
-    - AccessConfiguratorUniversal
+    # RMC-14: Replaced loadout
+    - CMHealthAnalyzer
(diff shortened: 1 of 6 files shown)


Resources/Prototypes/Roles/Jobs/Fun/misc_startinggear.yml
```diff
@@ -175,12 +175,12 @@
   id: MobAghostGear
   equipment:
     back: ClothingBackpackSatchelHolding
-    id: AdminPDA
+    id: CMIDCardHighCommand # RMC-14: Changed from AdminPDA
```
diraven, line 178: These changes will likely create conflicts on upsteam merge if upstream ends up changing them. I'd imagine it may be a good idea to inherit this prototype and use that instead of overriding base game one.

TsjipTsjip: I'm gonna leave this up to Smug or Whisper. The comments document the exact change made, if this causes a merge conflict it should be a no-brainer to fix.

QuietlyWhisper (approved):

PR #3067, opened 2024-07-14
Almayer is now fallback map, added Almayer to default map pool

Tunguso4ka:
## About the PR

- Added `fallback: true` to Almayer proto
- Removed `fallback: true` from Saltern proto
- Added Almayer to Default Map Pool
- Changed minPlayers from 0 to 90 on Almayer

### DIFF
--- a/Resources/Prototypes/Maps/Pools/default.yml
+++ b/Resources/Prototypes/Maps/Pools/default.yml
@@ -2,6 +2,7 @@
   id: DefaultMapPool
   maps:
   - Savannah
+  - Almayer
 
 - type: gameMapPool
   id: DefaultMapPoolAlmayer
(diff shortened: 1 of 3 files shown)


QuietlyWhisper:
I tried making savannah fallback, setting savannah min pop to 50 to force almayer to be picked but it still falls back to savannah so something must be wrong with almayer

QuietlyWhisper:
image
???

QuietlyWhisper:
it might just be because almayer is secretly savannah and I'm getting gaslit so I'll try more tests

QuietlyWhisper:
That should be enough info for smug to look into it at least

DrSmugleaf (approved):
saltern major victory

PR #3078, opened 2024-07-15
Revolver Fixes

Kaga-404:
## About the PR

Makes revolvers in RMC function properly. They no longer cycle on wield, will spin when unique-action is pressed, and don't spawn spent rounds. Revolvers that cannot be wielded will still cycle on use in hand.

Extra Notes:
- Russian Roulette doesn't work because SS14 always shows the current chamber next to hands. Completely unimportant to port since it has nothing to do with the main game and is mostly an old bit of SS13 code.
- Don't think revolver cylinders can be manually cycled in CM13 except for just shooting, but we may want to just have that as a QOL in RMC14 since the code already exists. **That is not included in this PR since this is a basic port and fix to make revolvers work as intended.**
- If you do decide to add cycling back to RMC14 revolvers after this, can maybe directly copy what was done with BallisticAmmoProviders like the grenade launchers, adding a verb that allows cycling manually.

## Why / Balance

So revolvers work as intended.

## Technical details

Gun code is haunted. Revolvers are worse.

Revolvers literally pretend that a round exists at multiple points and spawns spent casings at multiple points, including while shooting and while emptying a revolver. This even ignores whatever was actually loaded or shot, using the FillPrototype instead, if AmmoSlots and Chambers aren't both null.

My head hurts.

For anyone refactoring this stuff in the future: RevolverAmmoComponent.Chambers[] actually indicates if a chamber is loaded with a spent round, unspent round, or no round. True-False does is not filled or empty Chamber, it's unspent or spent.

## Media

## Breaking changes

**Changelog**

:cl: Kaga
- fix: Fixed wielding and caseless ammo for revolvers.
- add: Added cylinder spinning as the default RMC14 revolver unique action.


Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Revolver.cs
```diff
@@ -393,10 +395,12 @@ private void OnRevolverTakeAmmo(EntityUid uid, RevolverAmmoProviderComponent com
                 args.Ammo.Add((spawned, EnsureComp<AmmoComponent>(spawned)));
 
                 if (cartridge.DeleteOnSpawn)
+                    component.AmmoSlots[index] = null;
                     component.Chambers[index] = null;
```
DrSmugleaf, line 399: Only the first statement is inside the if since it is missing curly braces. This and line 403 should also be upstreamed.

Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Revolver.cs
```diff
@@ -26,17 +28,17 @@ protected virtual void InitializeRevolver()
...
-        if (!_useDelay.TryResetDelay(uid))
+        if (!_useDelay.TryResetDelay(gun)) // cannot wield after cycling. Can cycle after wielding tho? Can shoot regardless.
             return;
 
-        Cycle(component);
-        UpdateAmmoCount(uid, prediction: false);
-        Dirty(uid, component);
+        Cycle(gun.Comp);
+        UpdateAmmoCount(gun, prediction: false);
+        Dirty(gun, gun.Comp);
```
DrSmugleaf, line 41: This should be a field on the component that makes UseInHand happen or not, and the UniqueActionEvent a subscription in one of our systems for merge conflicts.

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf (approved):
Thanks for the PR!

PR #3089, opened 2024-07-16
Add markdown notices to Almayer files stating they are private.

PursuitInAshes:
## About the PR
Adds two notices (Essentially the same), to the Almayer map prototype and Savannah map file.

```
# This file is a duplicate of the Savannah. The actual Almayer map file is currently considered private content by RMC14, along with the map file for LV624.
# This file is to serve as a solution for a test fail that would result from the lack of its existence otherwise.
# If you wish for this file, or LV624, to be public you may recreate them yourself and PR it to the repo.
```

## Why / Balance
Recent and previous issues regarding clarity on their current status. This will allow for future direction to these files for clarity.

## Media
N/A.

## Breaking changes
N/A. Haven't tested ingame, but markdown should not cause a issue.


QuietlyWhisper (approved):
thanks

PR #3114, opened 2024-07-17
Fix disabler shape

Vermidia:
## About the PR

Change disabler shape to be the default gotten from item size

## Why / Balance

Had a wonky 3-tile shape due to the parent

## Media

gun

## Breaking changes

N/A
**Changelog**

no cl smol


DrSmugleaf (approved):
Thanks for the PR!

PR #3115, opened 2024-07-17
Webbing Storage fixes

### LINKED ISSUE #3111: Pill packets should be able to fit inside of webbing
CM13 parity

Vermidia:
## About the PR

You can put surgical cases, mres, and pill packets/canisters into both webbings now

Also changed surgical tray size to large to match cm-13

## Why / Balance

Fixes #3111, CM-13 parity

## Technical details

Gave the 5-slot webbing their own whitelists for ignoring size, and made the brown one parent from the black one since they are currently the same (and not sure if they're supposed to be different at all)

## Media

https://github.com/user-attachments/assets/944cb875-9bf1-47d5-8042-622b42010b0a

## Breaking changes

hopefully not
**Changelog**

:cl:
- fix: Webbings of both kinds can now properly store MREs, pill packets, pill canisters, and surgical kits.

### DIFF
--- a/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
+++ b/Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
@@ -916,7 +916,7 @@ public abstract class SharedStorageSystem : EntitySystem
         }
 
         if (TryComp<StorageComponent>(insertEnt, out var insertStorage)
-            && GetMaxItemSize((insertEnt, insertStorage)) > maxSize
+            && GetMaxItemSize((insertEnt, insertStorage)) >= maxSize
             && !CMStorage.IgnoreItemSize((uid, storageComp), insertEnt))
         {
             reason = "comp-storage-too-big";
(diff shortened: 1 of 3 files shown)


Dutch-VanDerLinde:
why just the 3 slot webbing

Vermidia:
Too big pretty much
Same reason why you can't put an MRE in there but can put the components.

Vermidia:
Actually I'm going to look at this...
Webbing (3-slot) needs the whitelist for magazines, but both should only have small items...but can't

Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
```diff
@@ -916,7 +916,7 @@ public void UpdateAppearance(Entity<StorageComponent?, AppearanceComponent?> ent
         }
 
         if (TryComp<StorageComponent>(insertEnt, out var insertStorage)
-            && GetMaxItemSize((insertEnt, insertStorage)) >= maxSize
+            && GetMaxItemSize((insertEnt, insertStorage)) > maxSize
```
DrSmugleaf, line 919: This check is for nested storages so you don't get infinite storage, 13 has a similar check.
For this we just have IgnoreContentsSize to override it when we know that it's safe, that's the preferred solution

DrSmugleaf (changes requested):

Content.Shared/Storage/EntitySystems/SharedStorageSystem.cs
Vermidia: I see, I'll change other stuff to make sure to change the other stuff to use that then

DrSmugleaf (approved):
Thanks for the PR!

PR #3117, opened 2024-07-18
window layering fix

musicmanvr:
literally 1 component missing it's so over

## About the PR

I added 1 component to directional windows to make their clipping work like barricades.
## Why / Balance

It looked like ass
## Media

- [ X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

Please ignore how the after is first i'm stupid

image

also works with any direction like it should too
## Breaking changes

**Changelog**

:cl:
- fix: Directional windows no longer mess up layering when you press against them.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Windows/directional_windows.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Windows/directional_windows.yml
@@ -11,12 +11,14 @@
       fix1:
         shape:
           !type:PhysShapeAabb
-          bounds: "-0.49,-0.49,0.49,-0.40"
+          bounds: "-0.49,-0.49,0.49,-0.38"
         density: 1500
         mask:
         - FullTileMask
         layer:
         - GlassLayer
+  - type: RotationDrawDepth
+    southDrawDepth: OverMobs
   - type: Sprite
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_RMC14/Entities/Structures/Windows/directional_windows.yml
```diff
@@ -11,12 +11,14 @@
       fix1:
         shape:
           !type:PhysShapeAabb
-          bounds: "-0.49,-0.49,0.49,-0.40"
+          bounds: "-0.49,-0.49,0.49,-0.38"
```
DrSmugleaf, line 14: Is this to make it draw properly when hugging against the windoor?

DrSmugleaf (changes requested):
I forgor I made this component for cades ty for the PR

Resources/Prototypes/_RMC14/Entities/Structures/Windows/directional_windows.yml
musicmanvr: Im not sure why i did this, I think it was to make the hitbox in line with the barricade

DrSmugleaf (approved):

PR #3122, opened 2024-07-18
Feat/extended-gamemode

0xRainy:
## About the PR

Added Extended gamemode

I'm opening this as a draft because I have no idea if the way I've done this is the right way

## Why / Balance

I'm tired of entering commands to stop the round ending when testing 😮‍💨
## Technical details

I basically copied the DistressSignal mode and ripped out some parts of it to make the round last until both xenos and marines are all dead.

I'm not sure if all of the code I left in is needed or if some of the code I removed is actually necessary for certain situations in a round.

I've tested it as well as I can with 1 player on the server

## Media

## Breaking changes

**Changelog**

:cl:
- add: Extended gamemode

### DIFF
--- a/Content.Server/_RMC14/Rules/CMExtendedRuleComponent.cs
+++ b/Content.Server/_RMC14/Rules/CMExtendedRuleComponent.cs
@@ -1,0 +1,79 @@
+using Content.Shared._RMC14.Weapons.Ranged.IFF;
+using Content.Shared.Roles;
+using Robust.Shared.Audio;
+using Robust.Shared.Configuration;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Server._RMC14.Rules;
+
+[RegisterComponent, AutoGenerateComponentPause]
+[Access(typeof(CMExtendedRuleSystem))]
+public sealed partial class CMExtendedRuleComponent : Component
+{
+    [DataField]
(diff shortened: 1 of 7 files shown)


Kadeo64:
you know what would be really fun? if extended allowed xenos and marines to talk to eachother. I don't know if that's directly possible though

0xRainy:
> you know what would be really fun? if extended allowed xenos and marines to talk to eachother. I don't know if that's directly possible though

Could probably do this with a new component like `UniversalTranslator` and a small change to https://github.com/RMC-14/RMC-14/blob/dfb5a1654290fc72aa1a916cf2b554908dfa5cc1/Content.Server/_RMC14/Chat/Chat/CMChatSystem.cs#L36-L37
https://github.com/RMC-14/RMC-14/blob/dfb5a1654290fc72aa1a916cf2b554908dfa5cc1/Content.Server/_RMC14/Chat/Chat/CMChatSystem.cs#L55-L56

then add the component when they spawn, might work.. there may be a better way of doing it though

DrSmugleaf (changes requested):
Thanks for the PR!
Need to go through and de-duplicate this code because just copy pasting the bits of distress signal without the end-game round part isn't ideal
Could have a boolean on the game rule for whether or not to check for round end conditions instead

0xRainy:
> Thanks for the PR! Need to go through and de-duplicate this code because just copy pasting the bits of distress signal without the end-game round part isn't ideal Could have a boolean on the game rule for whether or not to check for round end conditions instead

Some real life stuff came up but I will work on it this week, thanks for the feedback

0xRainy:
Gonna close this, I havent been able to figure it out.

PR #3123, opened 2024-07-18
Makes Wehny into a temporary boss 

Kadeo64:
## About the PR
Wehnies will now despawn after 2 minutes. 
## Why / Balance
Smugleaf said that Wehny is too OP, even for a shitpost boss xeno. This makes Wehny into a temporary entity, so admins don't need to worry about a single xenonid destroying the entire server. Only most of the server. (also tackle has been made better, but only slightly). 2 minutes was chosen since that's roughly the time it takes for the lobby to load after the round ends, letting admins have a wehny terrorize the server for the full two minutes. 
## Technical details
simple YML work
## Media

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/wehny.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/wehny.yml
@@ -75,10 +75,12 @@
       groups:
         Brute: 50
   - type: Tackle
-    threshold: 5
-    stun: 8
+    threshold: 1
+    stun: 15
   - type: MovementSpeedModifier
     baseWalkSpeed: 5.55
     baseSprintSpeed: 20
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/wehny.yml
```diff
@@ -75,10 +75,12 @@
       groups:
         Brute: 50
   - type: Tackle
-    threshold: 5
-    stun: 8
+    threshold: 4
+    stun: 15
```
DrSmugleaf, line 79: chief this is a stun longer than the queen's for only one hit more on a much faster attack rate
```suggestion
    threshold: 4
    stun: 15
```

DrSmugleaf (changes requested):

PR #3124, opened 2024-07-18
bring back admin PDA, put ID card inside admin PDA

QuietlyWhisper:
For round timer and base health analyzer for deeper health debugging


QuietlyWhisper:
- https://github.com/RMC-14/RMC-14/pull/3124

PR #3137, opened 2024-07-18
make sandbags cheaper

deltanedas:
## About the PR
$2000 -> $1500, similar for the kit that has entrenching tools (which i think are extremely overpriced but whatever)

also fixed a typo lit -> kit

## Why / Balance
- sandbags are far less versatile than metal so they shouldnt cost the same
- they need filling instead of just being able to use it immediately
- they cant be used on the almayer since no sand

**Changelog**
:cl:
- tweak: Sandbags are now cheaper to buy at Requesitions.

### DIFF
--- a/Resources/Prototypes/_RMC14/Catalog/Cargo/requisitions_catalog.yml
+++ b/Resources/Prototypes/_RMC14/Catalog/Cargo/requisitions_catalog.yml
@@ -98,9 +98,9 @@
         crate: CMCrateClothingOfficerOutfit
     - name: Engineering
       entries:
-      - cost: 2000
+      - cost: 1500
         crate: CMCrateEmptySandbags
-      - cost: 3000
+      - cost: 2500
         crate: CMCrateSandbagsConstructionKit
       - cost: 2000
         crate: CMCrateMetalSheets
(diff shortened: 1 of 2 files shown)


DrSmugleaf (changes requested):
These need their missing functionality added: building a partial sandbag barricade when less sandbags are used than the ones required for a barricade, and piling more on on a partial sandbag barricade, not a buff in cost. This wouldn't address the cost for it in the comtech's rack, either.

DrSmugleaf:
@deltanedas I just realized this probably never reached you since you aren't on the Discord, but relating to this and the req PR, we are aiming to port 13 1:1 specially when a missing feature would address a problem we have.
The exceptions to this are when that feature would be very difficult to implement or made to work properly with prediction (RNG tackles, tile movement) or we can improve on it without giving up anything in exchange (mostly rule enforcement stuff made into mechanics that 13 hasn't gotten to for one reason or another).
Be careful looking at the wiki though since that often has outdated information, I usually check most things in-game in a local 13 server and ask the 13 veterans.

deltanedas:
wasnt aware of it :trollfacd:

PR #3139, opened 2024-07-18
added ammo box to requesition munitions vendor

### LINKED ISSUE #3092: req munitions vendor should use boxes
## Description
every round gamers just pack it all into boxes anyway so just save the trouble + offering multiple hundred individual mags is weird. swap "gameplay" of spam clicking hundreds of times to pack everything for deployment for instead, John Marine asks for X mags so you take a box and remove some then leave it on a rack very good

also entity spam, if boxes works like magazines do with bullets. spawn mags only when theyre taken and not immediately which is good for performance (from a brief look i dont think so but still good)

deltanedas:
## About the PR
no mags were removed just added free boxes because cm13 has that

## Why / Balance
fixes #3092

## Media

**Changelog**
:cl:
- tweak: Requesitions can now get boxes of ammo in their munitions vendor.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Boxes/magazine_boxes.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Boxes/magazine_boxes.yml
@@ -40,8 +40,6 @@
   abstract: true
   parent: CMBoxMagazineBase
   id: CMBoxMagazineAPBase
-  name: generic magazine box (? x ?)"
-  description: A box of magazines.
   components:
   - type: GenericVisualizer
     visuals:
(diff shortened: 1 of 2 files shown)


DrSmugleaf (changes requested):
This isn't an either or, req is supposed ot have both individual mags and boxes.

https://github.com/user-attachments/assets/39f6222d-ea20-4474-9ab8-2737f750ba8a

See requisitions.dm for all the starting amounts.

deltanedas:
oh so should i reduce the numbers of mags to compensate or just have the same number of boxes to mags as cm13

deltanedas:
right now i just copied numbers from the video

DrSmugleaf:
I'll check once I get back home but it might be random, I know some start with 0 at least

DrSmugleaf:
> oh so should i reduce the numbers of mags to compensate or just have the same number of boxes to mags as cm13

I'm not sure if we ever increased them to compensate but both mag and box counts need be the same as 13

deltanedas:
oh thats cursed as hell, maybe my original implementation was more mag-accurate then

DrSmugleaf:
I'll try to code it later so the req vendors can auto-box mags inside them

DrSmugleaf:
This should be done with the auto-boxing functionality instead

PR #3155, opened 2024-07-20
Barbed Wire Application Fixes

### LINKED ISSUE #3138: barricade wiring issues
## Description
- if multiple people try wiring a cade at once the extra attempts go through and delete wire for no reason
- if you click multiple cades with wire it cancels the doafter. this is very annoying and should either allow you to wire multiple at once or just not cancel it. if you really want to stop wiring a cade you can always move

blueDev2:
## About the PR

Add DoAfterAttemptEvent for the barbed wire application DoAfter to check if the barricade targeted is already wired. If it is, cancel the DoAfter.

Add a target to the DoAfterArgs so that the DoAfter System recognizes events that target different barricades as different DoAfterEvents, preventing cancellation due to "duplicate" events.

Fixes #3138

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Multiple people wiring the same barricade no longer wastes barbed wire.
- tweak: Attempting to wire another cade while already wiring does not cancel the original wiring.

### DIFF
--- a/Content.Shared/_RMC14/Barricade/SharedBarbedSystem.cs
+++ b/Content.Shared/_RMC14/Barricade/SharedBarbedSystem.cs
@@ -46,13 +46,14 @@ public abstract class SharedBarbedSystem : EntitySystem
         if (!component.IsBarbed && HasComp<BarbedWireComponent>(args.Used))
         {
             var ev = new BarbedDoAfterEvent();
-            var doAfterEventArgs = new DoAfterArgs(EntityManager, args.User, component.WireTime, ev, uid, uid, used: args.Used)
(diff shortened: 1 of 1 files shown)


QuietlyWhisper (changes requested):
They are intended to only be able to be wired one at a time.

However, in 13 you're able to queue doafters, which we don't have code for iirc. You're not able to barb two things at the same time but you can queue the next action it seems.

blueDev2:
> They are intended to only be able to be wired one at a time.
> 
> However, in 13 you're able to queue doafters, which we don't have code for iirc. You're not able to barb two things at the same time but you can queue the next action it seems.

Right then I'll just make it so you can only apply 1 at a time and remove the cancellation from attempting to wire 2 things.

DrSmugleaf (approved):
Thanks for the PR!

PR #3160, opened 2024-07-20
Separate attachment activate and deactivate delays.

aleios:
## About the PR

Added the ability to set both an activation and deactivation delay independent of one another to attachments.
To showcase, have changed the smart scope to use a delay on activation but have a zero delay on deactivation.

There's some whitespace changes too which seem to be more correct with the rest of the project formatting?

## Why / Balance

This might be useful for future attachments and getting it in early means less file edits. Does mean some fields are duplicates... but there might be a solution that I'm not seeing to that through serialization?

As for the smart scope change, have found myself not hitting the button to deactivate instead opting to unwield or perform some other interrupt action. Ends up being way faster than the action especially with higher latency.

Straight up unwield is still faster for movement anyways but can remove from the PR if that part is unwanted.

## Media

[rmc-attachment-delay.webm](https://github.com/user-attachments/assets/02f1f8f8-2dd7-4cad-8b40-46b1f5005470)

## Breaking changes

* `DoAfter` field is renamed to `ActivateDelay` as well as adding a new field named `DeactivateDelay`.

**Changelog**

* add: Attachment activation and deactivation delays.
* tweak: S8 Smart-Scope zero deactivation delay.

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Components/AttachableToggleableComponent.cs
+++ b/Content.Shared/_RMC14/Attachable/Components/AttachableToggleableComponent.cs
@@ -46,7 +46,10 @@ public sealed partial class AttachableToggleableComponent : Component
     public bool WieldedOnly = false;
 
     [DataField, AutoNetworkedField]
-    public float DoAfter;
+    public float ActivateDelay;
+
+    [DataField, AutoNetworkedField]
+    public float DeactivateDelay;
 
     [DataField, AutoNetworkedField]
     public bool DoAfterNeedHand = true;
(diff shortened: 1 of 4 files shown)


Siigiil:
This is handled in the upcoming attachments PR. Which I will make as soon as I get the last few sprites I need.

aleios:
Ah ok that's good news then. I'll close the PR then if that's the case.

Yeah, it saw that it has a doAfter in 13. I think just nobody bothered to do the active check and it's not exactly the most high priority thing ever right?

Siigiil:
> Yeah, it saw that it has a doAfter in 13. I think just nobody bothered to do the active check and it's not exactly the most high priority thing ever right?

Probably. I'll ask smug later if we can remove the deactivation doafter for scopes. Doesn't make any sense for that to exist, and, as you said in the PR, you can just bypass it by unwielding anyway.

PR #3174, opened 2024-07-20
Windows break now and more

### LINKED ISSUE #2504: Windows are not breakable by reasonable means
Video, left to right: a reinforced window (it drops 1 rod and takes 4 hits)), two normal windows (takes two hits each).

https://github.com/RMC-14/RMC-14/assets/122584947/a980f189-a284-4690-b27e-d9c1fd341cc4

Deconstructing a windowsil gives you 2 metal

Cu1r:
## About the PR
All windows break now and revert to their window frame variants. You can replace the window glass or deconstruct the window entirely if you wish. Window frames make great fortifications because they have 600 health and you can climb over them, but that also means nobody should be allowed to build them from scratch.
Window smoothing still has problems. That is purely visual and is not the scope of this PR. This PR covers all of the important gameplay elements associated with window interaction.
Sanitized a lot of things including indestructible windows being vulnerable to explosions and corrosion, problems with entity parenting, textures, etc.

## Why / Balance
Closes #2504 
Breaking a window and climbing through/removing the frame is important for troop movements.

## Technical details
Not sure what the deconstruction/construction steps should be in depth other than what I've tried to glean from cm13 code (the issue only shows windows frames being deconstructed), so if anyone wants to supply more information or change it later in another PR please do so.

## Media
https://www.youtube.com/watch?v=nz-Y0iYn0mk

- [X] I have added screenshots/videos to this PR showcasing its changes ingame

**Changelog**
:cl: Cu1r
- fix: Fixed windows not having any destructible functionality.
- add: Windows now leave vaultable frames when smashed and can be repaired or deconstructed entirely.


DrSmugleaf (approved):
Thanks for the PR!

Skarletto:
only issue with this is that the window should be decon'd with a wrench not a welder, otherwise splendid

Skarletto:
Should be easy compared to making this all over again :trollface: teehee

PR #3175, opened 2024-07-20
Make bloodpacks injectable & drawable

Vermidia:
## About the PR

Bloodpacks can now be draw or injected with a syringe to add or remove reagents.

The iv system was modified so that it'll only transfer the allowed reagents out of the bloodpack (as it does in cm-13)
## Why / Balance

Fixes discord issue, CM-13 parity

## Technical details

Had to add TransferableReagents to bloodpack comp. IV system now seperates non-transferable reagents into a tempory solution before it tries to transfer, then adds them back to the original bloodbag.

## Media

https://github.com/user-attachments/assets/e2ed949d-fe57-4f2b-a467-2f88feea8798

## Breaking changes

No?
**Changelog**

:cl:
- add: You can now inject and draw from blood bags, but they will only inject blood in them


DrSmugleaf (approved):
Thanks for the PR!

Issue #2118, opened 2024-05-13
Pheromones, evolutions and choose resin structure should use a radial menu


DrSmugleaf:

Dutch-VanDerLinde:
resin construction uses the same window in CM13, only pheromones are radial menu

PR #3191, opened 2024-07-21
Make xenonid phermones use radial menu

Dutch-VanDerLinde:
## About the PR
see: #2118

CM13 parity

## Media
image

**Changelog**
:cl:
- tweak: The xeno pheromones menu is now a radial menu rather than a window


lzk228:
would be good to also add little descriptions to every type of pheromone

DrSmugleaf (changes requested):
This needs the help icon from 13 and to be in the same order
image

This is what the help icon prints to chat:
```
Pheromones provide a buff to all Xenos in range at the cost of some stored plasma every second, as follows:
Frenzy (Red) - Increased run speed, damage and chance to knock off headhunter masks.
Warding (Green) - While in critical state, increased maximum negative health and slower off weed bleedout.
Recovery (Blue) - Increased plasma and health regeneration.
```

Dutch-VanDerLinde:
image
image

was able to get the help button done but i'm not sure on how to change the order of the buttons

DrSmugleaf:
> was able to get the help button done but i'm not sure on how to change the order of the buttons

That should depend on what order the buttons are added to parent in the code, you might need to manually set the order of xeno pheromones instead of iterating the enum values

Dutch-VanDerLinde:
@DrSmugleaf 
image
Got the radial buttons to appear in the correct order

DrSmugleaf (approved):
Thanks for the PR!

PR #3194, opened 2024-07-21
Update ChatSanitizationManager.cs

SolidusSnek:
## About the PR

Testing on line 80 if putting a period in will add punctuation correctly to emote/text replacement - will replicate for all emotes if so.

## Technical details

Instead of "Someone salutes", should be "Someone salutes."
## Media

## Breaking changes

**Changelog**


SolidusSnek:
> Um... why?

To keep it consistent with the emote menu which is correctly adding punctuation at the end of each emote.

Issue #3150, opened 2024-07-19
Gun size mania


UnicornOnLSD:
## Description

It's been reported three times on the discord now, because of how gun size is calculated now, it seems you can fit guns in ridiculous spaces where you shouldn't
**Reproduction**

have gun
try to put it in helmet
works ?
try to put a long shotgun in webbing
works ?
**Screenshots**

![image](https://github.com/user-attachments/assets/b03136d2-6fb1-4aed-ad4d-3eabea83dbf7)
![image](https://github.com/user-attachments/assets/85aec7e1-9dc7-4abc-9de9-d654bf231bce)
![image](https://github.com/user-attachments/assets/4c6b2f0f-e7ed-4dfc-9a4d-cb3299a71e3a)

**Additional context**

it's 100% linked to how gun size is now calculated based on attachments it has, with #2788

deltanedas:
specifically caused by adding a stock

Siigiil:
#3197 seems to have fixed this. At least I couldn't reproduce it in the round I played earlier today, and there haven't been any new reports. So I'm reasonably certain it's gone.

PR #3197, opened 2024-07-22
Gun melee, CM scatter, bipod, magharn

### LINKED ISSUE #1049: M41AE2 HPR Magazines crate (HPR ammo box x2) Cost: $2000

### LINKED ISSUE #3121: Xenos can strip attachments off weapons
## Description

Xenos can take off attachments on weapons, say like rail flashlights...
(they can turn said light off)

**Reproduction**

**Screenshots**

**Additional context**

### LINKED ISSUE #3162: The arm brace makes the m63 unable to fire in burst fire mode
- Spawn M63
- Put in arm brace
- Put in mag
- Enable arm brace
- Switch to semi auto
- Switch to burst
- Won't fire

Siigiil:
## About the PR
This adds the magharn and the bipod. Also adds firemode-dependent gun stats and changes scatter calculation to match how it's done in CM. All guns have been changed to use this system.

For full details, see the changelog.

Closes #3121, closes #1049 and closes #3162.

*Should* also deal with #3150, but I can only reproduce the bug on the live server, so I will close this one manually after the PR is merged and I have confirmation that it's fixed.

## Why / Balance
Because porting.

## Technical details
For upstream file changes, see the next section.

RMCSelectiveFireComponent and RMCSelectiveFireSystem now handle the majority of gun stats.
Use RMCSelectiveFireComponent to set values defined in it instead of GunComponent.
Don't use GunWieldBonus. It's scuffed, and we need wielding to change the *base* values of the gun, so this is now contained in RMCSelectiveFireComponent.

Burst scatter calculation doesn't work quite like CM for a few reasons: in CM, burst scatter instantly jumps to the maximum for the burst amount, and the direction your character is spraying in tracks the target you clicked on to shoot.
The latter can't really be implemented here, so if we also used the former, burst fire would both have more scatter and, usually, shoot slower than every other firemode. And would, therefore, be even more dogshit than it already is.
So instead I made the scatter ramp up. The hard cap for burst scatter in CM is reached at a 6-shot burst, so this is the number of shots it takes to reach maximum scatter. Bursts will have the same scatter peak as in CM, but will ramp up to it instead of instantly spraying in random directions.

AngleDecay is now always set to 0, as in CM, scatter is instantly reset to 0 after you're done shooting a burst or full-auto spray.

AngleIncrease is now automatically set for any firemode that contains a ShotsToMaxScatter value in its modifier set. Otherwise, this is set to 0, since in CM scatter only ramps up in full-auto and has a concrete number of shots before it reaches its peak.

I'll see what I can do to condense the modifier refresh methods in RMCSelectiveFireSystem in the next PR.

The attachment examine window currently won't properly display modifiers if their conditions are identical in everything but the contents of their whitelist or blacklist. I'll fix this in the next PR. We don't have any attachments that need it anyway, so it's not an issue yet.

## Upstream file changes
This changes two sets of upstream files:

The first set is the client MeleeWeaponSystem and a new component: AltFireMeleeComponent. This mirrors the changes from here: https://github.com/space-wizards/space-station-14/pull/29484 because I'm tired of waiting for it to get merged. This adds melee to guns.

The second set includes GunComponent.cs and SharedWeaponSystem.Interactions.cs. The first now gives access to RMCSelectiveFireSystem, while the latter sends out RMCFireModeChangedEvent when a weapon's firemode is changed.
The former allows RMCSelectiveFireSystem to make changes to a weapon's GunComponent, which is necessary, as this system needs to be able to change the weapon's *base* stats. This could have been accomplished by adding a bunch of setter methods to SharedWeaponSystem, but I decided against it as this lets us have more control over the component without me having to make an upstream PR to add a method for every little change we might need.
The latter is needed to allow RMCSelectiveFireSystem to detect when a gun's firemode is changed and alter its values accordingly.

## Media
Different weapon stats for different firemodes:

semiauto
burst
fullauto

Attachment modifiers:
attachment

Melee with guns:

https://github.com/user-attachments/assets/8df74695-e3af-48a5-8cc0-f4962ed2e685

**Changelog**

:cl: Sigil
- add: Ranged weapons can now be used to attack in melee. Right-click to do so. The base damage for most guns is 5, but attachments may modify it.
- add: A new attachment has been added: the magnetic harness. If your weapon is knocked out of your hands or you drop it, this attachment will make the gun snap to your armour. It fits on the M54C, the M4SPR, the M77, the M83, the M42A2 and the M63.
- add: A new attachment has been added: the bipod. You can examine it ingame to see the modifiers. It fits on the M54C, the M4SPR and the M96S.
- add: A new attachment has been added: the M44 magnum sharpshooter stock. It fits on the M44 revolver. While inactive, the stock prevents the weapon from being fired. When active, it strongly reduces wielded scatter.
- add: Attachments can now be toggled through the verb menu. Toggling attachments on items you don't have equipped or held takes longer.
- add: Cycling firemodes now has a keybinding. The default is control+spacebar. This can be changed in the options menu.
- add: The BFA now adds burst fire to weapons that don't have it, like it's supposed to.
- add: Examining attachments how lets you check their modifiers. Click the question mark icon in the corner of the examine window to do so.
- add: Guns can now have different stats depending on the selected firemode. Most of the time, semi-auto will have lower scatter than the other modes, while burst fire will fire slower overall, but have lower scatter than full-auto.
- add: Implemented CM-style scatter calculation. It only increases when firing on burst or full-auto. It's set to zero the moment you stop shooting. Full-auto and burst fire generally have higher scatter and are more adversely affected by the gun not being wielded.
- tweak: The UGL can now only be fired if the gun it's attached to is wielded.
- tweak: Added use delays for attachments. You can no longer spam toggle rail and grip lights.
- tweak: All implemented weapon stats adjusted to match their CM counterparts.
- fix: The smartpistol can now hold the barrel charger, the extended barrel, the recoil compensator, the suppressor, the rail light, the reflex sight, the red-dot sight, the burst fire assembly and the laser sight. Like it was supposed to from the start.
- fix: The UGL and the masterkey no longer uncock if you shoot the gun they're attached to.
- fix: The arm brace no longer causes an error when its user is handcuffed.
- fix: The bayonet no longer deals 20 extra damage when used as a knife.
- fix: Guns with size-changing attachments should no longer spontaneously shrink down and fit in your pockets.
- fix: Xenos are no longer capable of removing attachments from items.
- fix: RDS removed from the SG equipment rack, as its only modifiers are for an unimplemented mechanic and it's not supposed to be publicly available yet.
- fix: The rail light now fits on the smartgun.
- fix: The MK2 UGL can no longer shoot further than 7 tiles away. The UGL no longer overshoots your cursor.


Siigiil:
The code is ready to be reviewed. Made this as a draft to make sure it's not merged before I get the HPR sprites. Once I have them, I'll add them in, fix the attachment offsets on it, and make a video that showcases the thing, and *then* this will be ready for merging.

Siigiil:
I have no idea why it's saying the conflicts are still there. Pretty sure I fixed them.

QuietlyWhisper:
We should just drop the HPR addition for now so cepha can make the sprites on their own time, we can add the HPR later

Siigiil:
> We should just drop the HPR addition for now so cepha can make the sprites on their own time, we can add the HPR later

Removed it from public access.

DrSmugleaf (approved):
Thanks for the PR!

PR #3210, opened 2024-07-22
flares despawn shortly after expiring

QuietlyWhisper:
## Why / Balance

Garbage cleanup, unlike empty mags these serve no purpose but to take up space on the screen, among performance reasons I suppose.

## Technical details

Adding onto UpdateVisualizer, whenever the item progresses to the spent sprite, the TimedDespawnComponent is given so the item is deleted after the default 5 seconds have passed, similar to current blood puddles.

I think this is the least obtrusive way to add this functionality without touching more than this one upstream file.

I also do not know much about coding.

/shrug

**Changelog**

:cl: Whisper
- add: Spent flares will now despawn.

### DIFF
--- a/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
+++ b/Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
@@ -13,6 +13,7 @@ using Robust.Server.GameObjects;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Player;
+using Robust.Shared.Spawners;
 using Robust.Shared.Utility;
 
 namespace Content.Server.Light.EntitySystems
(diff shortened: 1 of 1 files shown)


Content.Server/Light/EntitySystems/ExpendableLightSystem.cs
```diff
@@ -137,6 +138,7 @@ private void UpdateVisualizer(Entity<ExpendableLightComponent> ent, AppearanceCo
                     _appearance.SetData(ent, ExpendableLightVisuals.Behavior, string.Empty, appearance);
                     var ignite = new IgnitionEvent(false);
                     RaiseLocalEvent(ent, ref ignite);
+                    EnsureComp<TimedDespawnComponent>(ent);
```
DrSmugleaf, line 141: This needs to be under an event handler for a new component and IgnitionEvent checking for Ignite false

DrSmugleaf (changes requested):

ShadowCommander:
Spent flares are one of the items I used for game sense for checking of any marines have been through this area and how long it's been since then. I'm not sure exactly how useful that is, but it's a small thing that I think is at least a tiny bit useful.

I'm not super against despawning them, but just wanted to provide a reason why they shouldn't be if it wasn't known before.

PR #3213, opened 2024-07-22
Fix medical rig description

Vermidia:
## About the PR

Changes the medical rig description to be updated to the latest CM-13 one, which gets rid of the ammo pistol part.

## Why / Balance

CM-13 one doesn't store pistol ammo, so they changed the desc. This just mirrors it.

## Technical details

N/A

## Media

## Breaking changes

nope
**Changelog**

nah


DrSmugleaf (approved):
Thanks for the PR!

PR #3215, opened 2024-07-22
Larva now transfers control to the burst victim

Dutch-VanDerLinde:
## Why / Balance
CM13 parity

also more forgiving for marines who were captured, as they aren't round removed

## Media

https://github.com/user-attachments/assets/11955401-1d00-42c3-9881-e8642402529b

## Breaking changes

**Changelog**
:cl:
- tweak: Bursted larvas will now be under control of the burst victim.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
@@ -25,6 +25,8 @@ using Content.Shared.Rounding;
 using Content.Shared.Standing;
 using Content.Shared.StatusEffect;
 using Content.Shared.Stunnable;
+using Content.Shared.Mind;
+using Content.Shared.Mind.Components;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Containers;
 using Robust.Shared.Network;
@@ -57,6 +59,8 @@ public abstract class SharedXenoParasiteSystem : EntitySystem
     [Dependency] private readonly DamageableSystem _damage = default!;
     [Dependency] private readonly StatusEffectsSystem _status = default!;
     [Dependency] private readonly SharedRottingSystem _rotting = default!;
+    [Dependency] private readonly SharedMindSystem _mind = default!;
+    [Dependency] private readonly ActorSystem _actor = default!;
 
     public override void Initialize()
     {
@@ -247,7 +251,7 @@ public abstract class SharedXenoParasiteSystem : EntitySystem
 
     private void OnVictimBurstExamine(Entity<VictimBurstComponent> burst, ref ExaminedEvent args)
     {
-        using(args.PushGroup(nameof(VictimBurstComponent)))
+        using (args.PushGroup(nameof(VictimBurstComponent)))
             args.PushMarkup($"[color=red][bold]{Loc.GetString("rmc-xeno-infected-bursted", ("victim", burst))}[/bold][/color]");
     }
 
@@ -303,13 +307,13 @@ public abstract class SharedXenoParasiteSystem : EntitySystem
             return false;
         }
 
-        if(_mobState.IsDead(parasite))
+        if (_mobState.IsDead(parasite))
         {
-            if(popup)
-				_popup.PopupClient(Loc.GetString("rmc-xeno-failed-parasite-dead"), victim, user, PopupType.MediumCaution);
+            if (popup)
+                _popup.PopupClient(Loc.GetString("rmc-xeno-failed-parasite-dead"), victim, user, PopupType.MediumCaution);
 
             return false;
-		}
+        }
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


DrSmugleaf (changes requested):
13 deals with this through the xeno queue, which the infected marine and parasite get put to the front of I think
Does this always happen, or is it only when the player is still in the marine's body? It would probably be better to implement the xeno queue either way
You might have to ask in the #13-qa channel for specifics on how it works though

Dutch-VanDerLinde:
> Thanks for the PR! 13 deals with this through the xeno queue, which the infected marine and parasite get put to the front of I think Does this always happen, or is it only when the player is still in the marine's body? It would probably be better to implement the xeno queue either way You might have to ask in the #13-qa channel for specifics on how it works though

only when the marine is still in the body

Dutch-VanDerLinde:
and i would add xeno queue if i could, but i can't so i made this instead

Sailzington:
My worry is that this incentivizes Marines to play worse, and not follow orders since you would get a guaranteed ghost role, and there is no downside to being captured anymore

Dutch-VanDerLinde:
> My worry is that this incentivizes Marines to play worse, and not follow orders since you would get a guaranteed ghost role, and there is no downside to being captured anymore

the downside to being captured is that you fucking die, it's how it is in CM13

Dutch-VanDerLinde:
if you wanted to play xeno, you would play roundstart xenos

musicmanvr:
> My worry is that this incentivizes Marines to play worse, and not follow orders since you would get a guaranteed ghost role, and there is no downside to being captured anymore

like others have said, you can't play marine after you die, those are the stakes

Sailzington:
The downside to being captured is that you are round removed and have to wait +40 minutes sometimes. Thats why people listen to command and stay behind cades. Now you can just run to the front, not listen to orders because you want to shoot something, get captured and get a free ghost role guaranteed! This is just rewarding players for the wrong reasons…

Rainbeon:
> if you wanted to play xeno, you would play roundstart xenos

If you're not there at roundstart and want to play Xenos (there are plenty, including me), you would spectate and wait for a larva, with this as a mechanic it would be much faster and more convenient for the player on a personal-level to pick marine and just run in and die. Doing so adds an additional xeno to the opposite team and instead of being punished for deliberately killing yourself (for the latejoining Xeno, missing x amount of Larva rolls), you are rewarded with a guaranteed slot on Xenos and collectively worsened the experience for actual Marine players as you've added +1 to Xeno ranks.

There are many other reasons any given player may want to just side-swap to Xeno as well (examples: avoiding fob duty, avoiding Marine-side Hijack, just getting bored of the round in general, and plenty more). These may not be reasons you would think about using, but different players think different things, and many will tend to do what benefits them personally. Just overall this is mechanically encouraging behavior that is detrimental to the round as a whole.

Rainbeon:
> only when the marine is still in the body

As an additional clarifying question, how does this handle Yellow SSD players? In base SS14 the game can still believe the player to be in-body if they were at the time of SSD (a symptom of this issue is the Morgue constantly beeping for some Yellow SSD players, the solution to the annoyance being to just bodybag them).

I ask this because often Marines will just close the game once captured, still in-body and Yellow SSD, instead of /ghosting, and the potential ramifications of this would be the Larva also being Yellow SSD upon bursting. Yellow SSD larva will stay this way forever, effectively making the Larva dead as far as team numbers are concerned, as there is no player-side solution to this occurring.

QuietlyWhisper:
So if you don't get infected you're just shit out of luck because the infected get priority rather than the first dead?

thymol-blue:
Yes, the infected get priority rather than the oldest dead. If the goal is CM-13 parity, I believe Dutch is correct. I think Verm posted some code in the QA chat about it. Basically the priority is burst marine > facehugger that hugged > queue.

The alternative is TGMC queue whereby it is based on a button click to join the larva queue not death timer. That is, the person who clicked the "join larva queue" button the longest time ago will be 1st. They don't have anything about the burst marine or facehugger.

Dutch-VanDerLinde:
Queue won't really matter though? If the bursted marine is always at the top of the queue, it's basically the exact same

deltanedas:
> people listen to command and stay behind cades.

clearly never played the game this is false :trollface:

getting capped 5m after deploying is nasty at least with this you get to do something

Rytil:
I think its fine if they can queue to be a xeno, but rewarding bad gameplay seems counter intuitive. It would also have people who are essentially feeding the xenos more players because they wanted to be xeno, but couldn't pick it since the round already started (or for some reason they couldn't roll it)

deltanedas:
people who are antag rolling 1. can be bwoinked 2. can already just /ghost roundstart and wait for parasite or larva to open up

noisrevni-dev:
This feature is actually make sense, reminds me those zombie games that if you bitten and got full infection, you become the zombie. Almost no difference in terms of mechanic. People also seems complaining of marines going suicide bomber, this feature will lessen that problem because if you suicide bombing, you pretty much round removed until some empty larva slot open or play as parasite more often.

PR #3217, opened 2024-07-23
add LV-624

MACMAN2003:
## About the PR

did my best to manually re-create LV-624 from cm13
## Why / Balance

it wasn't on the repo
the differences from the non-repo version that the main server uses is that there are floodlights, the floors aren't uniformly "CMFloorSteel", and there are podlocks for the garage/research/secure storage
## Media

it's too huge i can't screenshot it all
## Breaking changes

**Changelog**

:cl:
- add: LV-624 is now on the github repository.


Kadeo64:
so does sand temple finally have temple walls, and does containers finally get containers

MACMAN2003:
unfortunately no

deltanedas:
whats wrong with the sussy map that isnt on repo

PursuitInAshes (changes requested):
In the "almayer" map file and prototype, you'll also want to remove the mentions that LV is private with this PR. The mentions that the alamayer map file is private still, should stay. For reference you can find it [here](https://github.com/RMC-14/RMC-14/pull/3089/files) for the location of the files to change.

QuietlyWhisper:
the figurines are kind of randomly scattered around the map on top of tables, there's no real method to placing them so long as they're well spaced apart 

Though yes, Cuir is right, this seems like a downgrade, you should ask Smug for the actual map file like she offered so you can improve what is already done?

QuietlyWhisper (changes requested):
See above

Vermidia:
LV is public on the repo now so I'm closing this.

PR #3218, opened 2024-07-23
Add/Xeno interact/UI blacklist system

0xRainy:
## About the PR

Added a blacklist system to `XenoSystem` that allows placing components in `base_xeno.yml` blacklist that will prevent the xeno from interacting with and/or being shown a UI for those components

## Why / Balance

I think a system like this is more usable than adding checks to item/structure/machines/etc system files individually

## Technical details

The component you choose to add to the blacklist must have a check for either `InteractionAttemptEvent` or `UserOpenActivatableUIAttemptEvent` for it to work.

## Media

## Breaking changes

I tested a whole bunch of things but the components I chose to include in the blacklist apply to a lot of stuff, `ActivatableUIRequiresPower` specifically, so there may be something I missed and a xeno will not be able to interact with something unintentionally. And yes, the dropship nav computer and console on the dropship is unaffected.

I also will be having a look around for areas where a `XenoComponent` check was done to prevent interaction and removing those in a separate PR. If that should be a part of this PR then I can do that.

**Changelog**

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/XenoInteractBlacklistComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/XenoInteractBlacklistComponent.cs
@@ -1,0 +1,12 @@
+using Content.Shared.Whitelist;
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Xenonids;
+
+[RegisterComponent]
+[Access(typeof(XenoSystem))]
+public sealed partial class XenoInteractBlacklistComponent : Component
+{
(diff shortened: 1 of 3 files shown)


QuietlyWhisper:
- #3204 ?

0xRainy:
> * merge it with this system? [Fix xenos being able to open unintended UIs #3204](https://github.com/RMC-14/RMC-14/pull/3204)

Hmm didn't know that was a thing. It looks like we went about solving the issue in two different ways. I'm fine with closing this if you want to use that one.

PR #3228, opened 2024-07-23
rework unique construction

### LINKED ISSUE #3231: hive entity isnt networked
## Description
both roundstart hove and midround from the Open RMC Actions menu doesnt get sent to clients

i tried using PvsOverrideSystem.AddForceSend for each xeno and PvsOverrideSystem.AddGlobalOverride for the spawned hive and it still refused to exist on client so idk
only got synced once i teleported the hive entity onto the same map, maybe nullspace pvs is weird

**Reproduction**
1. become rouny
2. create a new hive and set it
3. its not added to pvs override so client doesnt get sent it

**Screenshots**
server has it

client doesnt

deltanedas:
## About the PR
instead of unique construction being weird and not supporting multiple hives, limits are now stored on the hive entity and can be changed. ~~an example of this being used is hive core, which allows building 8 hive clusters.~~
adds boilerplate for hive cluster, im not sure how it really works so thats for a later pr

also implemented hive core death locking out construction for 5 minutes

## Why / Balance
cm13 also hive weeds eventually being required for nesting :trollface:

fixes #3231

## Technical details
- added `ConstructionLimits` to `XenoHiveComponent` and a bunch of api to the hive system
- added `XenoConstructBuiltEvent` which is raised on constructs, like mapinit but after the hive member is assigned
- added `ModifyHiveLimitsComponent` for hive core to use
- reworked all the unique construction logic
  all constructs require a non-zero limit on the hive, and all constructs have to be specified in the hive's prototype or it logs an error when trying to work with them, this should make random tests fail if you make a structure without doing that
  HiveUniqueConstruction just changes the limit when built and destroyed. essentially a specific-case version of ModifyHiveLimits
- removed some fields from network state when nothing ever changed them from the prototype
- HiveClusterSystem is a stub for implementing later
- added `HiveChangedEvent` for systems to do things, like pvs overriding

also in cm13 adding a template is hardcoded to 400 plasma so made that the default and no longer required. of course its not hardcoded here but it will reduce copy pasta since all our prototypes will probably stay true to cm13

hive is also properly pvs overriden now so clients predict it

## Media
its real

## Breaking changes
idk

**Changelog**
:cl:
- tweak: Destroying the Hive Core now sends an announcement to the hive and disables construction for 5 minutes.


deltanedas:
@DrSmugleaf i cant really make out what the code is trying to do for hive clusters actual purpose so id appreciate a video of how hive clusters spread weeds when built + when they get repaired, whenever you can

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
```diff
@@ -54,7 +66,19 @@ public override void Update(float frameTime)
...
     }
+
+    /// <summary>
+    /// Create a new hive which gets networked to everyone for prediction.
+    /// </summary>
+    public EntityUid CreateHive(string name, EntProtoId? proto = null)
+    {
+        var ent = Spawn(proto ?? "CMXenoHive", MapCoordinates.Nullspace);
+        _metaData.SetEntityName(ent, name);
+        _pvsOverride.AddGlobalOverride(ent);
```
DrSmugleaf, line 81: What shared code needs this? This would make it so every client gets information about the Queen's death and hive tier unlocks

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
```diff
@@ -54,7 +66,19 @@ public override void Update(float frameTime)
                 continue;
 
             var popup = $"The Hive can now support: {string.Join(", ", _announce)}";
-            _xenoAnnounce.AnnounceSameHive(default, popup, hive.AnnounceSound, PopupType.Large);
+            // below is commented out because it used to not do anything, now it works but isnt localized so i think its best to have a dedicated pr look at it
+            //_xenoAnnounce.AnnounceTohive(hiveId, popup, hive.AnnounceSound, PopupType.Large);
```
DrSmugleaf, line 70: It's fine to have it unlocalized for now, I just forgot to fix it
```suggestion
            _xenoAnnounce.AnnounceTohive(hiveId, popup, hive.AnnounceSound, PopupType.Large);
```

Content.Shared/_RMC14/Xenonids/Construction/HiveCoreComponent.cs
```diff
@@ -34,4 +34,11 @@ public sealed partial class HiveCoreComponent : Component
 
     [DataField]
     public TimeSpan NextLesserDroneCooldown = TimeSpan.FromSeconds(125);
+
+    /// <summary>
+    /// How long a new construct can be made after the core is destroyed.
+    /// Only applies to this core's hive for xeno v xeno.
```
DrSmugleaf, line 40: This applies in general doesn't it? Hive core construction cooldown applies in distress signal too I think

Content.Shared/_RMC14/Xenonids/Construction/ModifyHiveLimitsComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+
+namespace Content.Shared._RMC14.Xenonids.Construction;
+
+/// <summary>
+/// When this construct is created, increases the hive's construct limit by some amount.
+/// If it is destroyed the limits are decreased but any old structures stay grandfathered in,
+/// but they cannot be rebuilt if they get destroyed as well.
+/// </summary>
+[RegisterComponent, Access(typeof(ModifyHiveLimitsSystem))]
+public sealed partial class ModifyHiveLimitsComponent : Component
```
DrSmugleaf, line 11: This would probably be better as the hive core handling the "can this be built" event and it checks the number of existing hive clusters in a query for example. That way you don't need to check prototypes, the hive core checks "does the thing to be built have hive cluster component" and if so "how many are already built", and if the event is not handled (no hive core) then the building is blocked.

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
```diff
@@ -384,31 +400,19 @@ private void OnHiveConstructionNodeAddPlasmaDoAfter(Entity<XenoConstructionCompo
         if (_net.IsClient)
             return;
 
-        var hive = CompOrNull<HiveMemberComponent>(target)?.Hive;
-        var spawn = Spawn(node.Spawn, transform.Coordinates);
-        var member = EnsureComp<HiveMemberComponent>(spawn);
-        _hive.SetHive((spawn, member), hive);
-
-        _adminLogs.Add(LogType.RMCXenoOrderConstructionComplete, $"Xeno {ToPrettyString(xeno):xeno} completed construction of {ToPrettyString(target):xeno} which turned into {ToPrettyString(spawn):spawn} at {transform.Coordinates}");
+        // Don't need to check if the construction is limited here since it's checked when spawning the "ghost"
```
DrSmugleaf, line 403: Is there a case where two can be built if two xenos place the ghost down on the same tick?
Might be useful to pull the check out into its own method that is also checked here

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoHiveCoreSystem.cs
```diff
@@ -1,3 +1,30 @@
...
+public abstract class SharedXenoHiveCoreSystem : EntitySystem
+{
+    [Dependency] private readonly SharedXenoAnnounceSystem _announce = default!;
+    [Dependency] private readonly SharedXenoHiveSystem _hive = default!;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<HiveCoreComponent, ComponentShutdown>(OnShutdown);
```
DrSmugleaf, line 15: For this it might be better to use DestructionEventArgs so it doesn't fire when admin-erased or at round-end

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
```diff
@@ -332,10 +346,11 @@ private void OnXenoOrderConstructionDoAfter(Entity<XenoConstructionComponent> xe
...
         var structure = Spawn(args.StructureId, coordinates);
 
-        if (TryComp(xeno, out XenoComponent? xenoComp))
+        // TODO: GetHive once XenoComponent doesn't store Hive
+        if (CompOrNull<XenoComponent>(xeno)?.Hive is {} hive)
         {
-            var member = EnsureComp<HiveMemberComponent>(structure);
-            _hive.SetHive((structure, member), xenoComp.Hive);
+            _hive.SetHive(structure, hive);
+            _hive.ReserveConstruct((hive, Comp<HiveComponent>(hive)), node.Spawn);
```
DrSmugleaf, line 353: Same as above then you don't need reserve

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
```diff
@@ -417,7 +421,7 @@ private void OnActionConstructionChosen(Entity<XenoChooseConstructionActionCompo
             _prototype.HasIndex(args.Choice))
         {
             action.Icon = new SpriteSpecifier.EntityPrototype(args.Choice);
-            Dirty(xeno, action);
+            Dirty(xeno);
```
DrSmugleaf, line 424: Xeno and xeno action are different, one's for the construction action comp and the other is for the action data comp

Content.Shared/_RMC14/Xenonids/Hive/HiveComponent.cs
```diff
@@ -34,9 +34,25 @@ public sealed partial class HiveComponent : Component
...
+    public TimeSpan? NextConstructAllowed;
+
+    /// <summary>
+    /// How many of each hive construction can be built.
+    /// If it reaches 0 it can no longer be built.
+    /// If an entity being constructed is not present it will be logged as an error,
+    /// use 0 to allow it to be constructed later.
+    /// </summary>
+    [DataField(required: true), AutoNetworkedField]
+    public Dictionary<EntProtoId, int> ConstructionLimits = new();
```
DrSmugleaf, line 57: Same as above if you remove this then you can handle this with components and events

DrSmugleaf (changes requested):
Thanks for the PR and the doc comments boss

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
deltanedas: i originally wanted it only overriden to the xenos in the hive but that refused to work, i doubt many people will use malf clients and even then they can know queen died on common radio when she is killed by marines

Content.Shared/_RMC14/Xenonids/Construction/HiveCoreComponent.cs
deltanedas: its on the hive so it wont be broken by xvx, of course still works on distress signal since xvx isnt real :trollface:

Content.Shared/_RMC14/Xenonids/Construction/ModifyHiveLimitsComponent.cs
deltanedas: wouldnt be predicted since a client would need to have all constructs in pvs range, which is why the info is stored on networked hive field

Content.Shared/_RMC14/Xenonids/Hive/HiveComponent.cs
deltanedas: pvs issues...

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoHiveCoreSystem.cs
deltanedas: already shouldnt happen on round end since GetHive checks if the hive is terminating, admins deleting it would brick the hive and i think vv cant edit the limits for sone reason, so you would have no way to fix it

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
deltanedas: starting and ending the ghost doafter both check CanOrderConstructionPopup, so if 2 xenos tried to build ghosts in the same tick an arbitrary one (entity spawn order prob) would succeed and the other fails

should i still add a check anyway to the plasma bit?

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
deltanedas: ok made it handle changing hive not just player attach now force send works

Content.Shared/_RMC14/Xenonids/Construction/ModifyHiveLimitsComponent.cs
DrSmugleaf: At that point you might as well just pvs override the special constructions themselves for xenos

deltanedas: @DrSmugLeaf caveat with your idea is you cant raise events on prototypes so it would have to spawn the thing first then raise event then delete it right after which is bad

deltanedas:
> caveat with your idea is you cant raise events on prototypes so it would have to spawn the thing first then raise event then delete it right after which is bad

@DrSmugleaf this is still true, do you still think it should be changed to work off the entities

DrSmugleaf:
> > caveat with your idea is you cant raise events on prototypes so it would have to spawn the thing first then raise event then delete it right after which is bad
> 
> @DrSmugleaf this is still true, do you still think it should be changed to work off the entities

I don't think you would need to raise an event on the prototype would you? The logic would be get prototype, try get component requires hive core component, if so query to check if one exists, you shouldn't need to spawn anything for those checks

PR #3234, opened 2024-07-23
Add parasites returning to opened eggs

Vermidia:
## About the PR

Using a parasite on an open egg now returns that parasite to the egg (it deletes them). A parasite clicking on an open egg will also return itself to the egg.

Also changed the previous can't infect with dead parasite line to be more inline with the other xeno popups.

## Why / Balance

Implements part of #2092, CM-13 parity

## Technical details

Used AfterInteractUsing and ActivateInWorldEvent (with some logic in the Open method) to check for the parasite component and return the parasite (if it isn't dead) and change the egg back to it's grown state.

There was a double pop-up problem (see video below) but it's fixed now

## Media

https://github.com/user-attachments/assets/ac6edbf5-d148-46ed-8fe0-9c650a005841

## Breaking changes

no?
**Changelog**

:cl:
- add: Parasites can crawl back into open eggs, and drone castes can also put them back in

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
@@ -256,12 +256,6 @@ public sealed class XenoEggSystem : EntitySystem
 			return;
 		}
 
-		if (_mind.TryGetMind(args.Used, out _, out _))
-        {
-            _popup.PopupEntity(Loc.GetString("rmc-xeno-egg-awake-child", ("parasite", args.Used)), args.User, args.User, PopupType.SmallCaution);
-            return;
-        }
-
         if (egg.Comp.State == XenoEggState.Growing || egg.Comp.State == XenoEggState.Grown)
         {
             _popup.PopupEntity(Loc.GetString("rmc-xeno-egg-has-child"), args.User, args.User, PopupType.SmallCaution);
@@ -273,7 +267,13 @@ public sealed class XenoEggSystem : EntitySystem
             return;
         }
 
-        _popup.PopupEntity(Loc.GetString("rmc-xeno-egg-return-user"), args.User, args.User);
+		if (_mind.TryGetMind(args.Used, out _, out _))
+		{
+			_popup.PopupEntity(Loc.GetString("rmc-xeno-egg-awake-child", ("parasite", args.Used)), args.User, args.User, PopupType.SmallCaution);
+			return;
+		}
+
+		_popup.PopupEntity(Loc.GetString("rmc-xeno-egg-return-user"), args.User, args.User);
         _popup.PopupEntity(Loc.GetString("rmc-xeno-egg-return", ("user", args.User), ("parasite", args.Used)), egg, Filter.PvsExcept(args.User), true);
 
         SetEggState(egg, XenoEggState.Grown);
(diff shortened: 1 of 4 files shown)


Dutch-VanDerLinde:
We can't infect the egg!

deltanedas:
handle the event after making popup so it doesnt do it twice
and order the egg system to run before infecting

Vermidia:
Problem is they're different events
The parasite is the one saying it can't infect (Due to afterinteracteevent)
But the egg is the one giving the other messages.

I'm not sure how to stop the former

Vermidia:
Nevermind I fixed it by changing the AfterInteractUsingEvent to just InteractUsingEvent so it runs before AfterInteract

Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
```diff
@@ -229,13 +230,49 @@ private void OnXenoEggAfterInteract(Entity<XenoEggComponent> egg, ref AfterInter
...
         // TODO RMC14 multiple hive support
-        if (!HasComp<XenoComponent>(args.User) || !HasComp<HandsComponent>(args.User))
+        if (!HasComp<XenoParasiteComponent>(args.User) && (!HasComp<XenoComponent>(args.User) || !HasComp<HandsComponent>(args.User)))
             return;
 
         if (Open(egg, args.User, out _))
             args.Handled = true;
     }
 
+    private void OnXenoEggInteractUsing(Entity<XenoEggComponent> egg, ref InteractUsingEvent args)
```
DrSmugleaf, line 240: This will also put them back in and delete them even if they are a player-controlled parasite, 13 doesn't even let you pick those up but for us it might be better to check if there is a player controlling the parasite, and if so to not run this code

Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
```diff
@@ -229,13 +230,49 @@ private void OnXenoEggAfterInteract(Entity<XenoEggComponent> egg, ref AfterInter
     private void OnXenoEggActivateInWorld(Entity<XenoEggComponent> egg, ref ActivateInWorldEvent args)
     {
         // TODO RMC14 multiple hive support
-        if (!HasComp<XenoComponent>(args.User) || !HasComp<HandsComponent>(args.User))
+        if (!HasComp<XenoParasiteComponent>(args.User) && (!HasComp<XenoComponent>(args.User) || !HasComp<HandsComponent>(args.User)))
```
DrSmugleaf, line 233: Is this change necessary or a leftover from something else?

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
Vermidia: Parasites don't have hands so I added it to skip over the early return.

KalimbaMachine:
i have a very specific issue lorewise with calling the parasites "child",
but thats only my opinion.
otherwise its neat.

Vermidia:
It's the CM-13 terminology so I went with it to be consistent

DrSmugleaf (approved):
Thanks for the PR!

Issue #3009, opened 2024-07-12
Squad Preference


XxSWAG-MASTER-OVERLORDxX:
## Description
Add the ability to choose what squad you would want to join before spawning.

tsp01:
I am currently working on this and will probably be done by the end of the weekend.

Issue #2988, opened 2024-07-11
Two alpha SLs.


Verification31:
For some reason, there can be multiple squad leaders. Ive seen this multiple times with alpha.

Tunguso4ka:
"How many Alpha SLs does it take to push the front?"

PR #3237, opened 2024-07-23
Allows squad preference selection

tsp01:
## About the PR
This PR adds a squad preference to character profiles. This preference will be used to determine which squad a player is placed in when they join the game as a squad role. Squad preference will never trump job preference, and a player will only get their chosen squad if said squad does not already have the maximum number of the players job. If no preference is chosen, the player will be sent to the squad with the least number of people with the players job.
resolves: #3009 
resolves: #2988  ~~no it doesnt~~ yes it does.

## Why / Balance
It is a long standing feature from CM13. 

## Technical details
squad priority is an integer ranging from -1 to 3 where 0 through 3 correspond to the squads in alphabetical order. In CMDistressSignalRuleSystem, the function GetSquadJobNumber() was added to return a dictionary of how many people in each squad have the relevant job. NextSquad() has been altered to check if there is a preferred squad, and if there are available positions in that squad for the given job. If either are false, the squad with the fewest number of people in the given job is selected. OnPlayerSpawning() has been altered such that when the first person spawns, the dictionary MaxJobsPerSquad is filled out so that each squad role is the key to either an int or null, where the int is the number of round start slots available divided by 4, and null represents no limit. 

Also to note, as this code adds a new preference to character profiles, this required a database migration.

## Media
Screenshot 2024-07-23 at 10 40 20 AM

## Breaking changes

**Changelog**
:cl: tsp01, DrSmugleaf
- add: Added squad preference in character customization.
- fix: Fixed squads being able to have multiple SLs, SGOs, or weapons specs if one cryos and another one late-joins.


MACMAN2003:
YES!!!!!!

DrSmugleaf:
> also might resolve: #2988 Though on account of I don't know the exact cause of that it also might not. I would assume it had something to do with 'next' being mistakenly set to 0 in NextSquad().

That one happens when a role goes into cryo, the spawning logic doesn't account for cryoing but it should. Not this PR's problem though

tsp01:
> > also might resolve: #2988 Though on account of I don't know the exact cause of that it also might not. I would assume it had something to do with 'next' being mistakenly set to 0 in NextSquad().
> 
> That one happens when a role goes into cryo, the spawning logic doesn't account for cryoing but it should. Not this PR's problem though

Actually thinking about it, if this is due to cryo I think this does actually resolve the issue. Before, all SLs are put into squads in order without regard for if the squad has the max number of SLs, so first person to roll SL goes to alpha, and the fourth person goes to delta. Then if someone cryos, say the charlie SL, an SL slot is opened and after delta comes alpha again so a new SL is put into alpha instead of charlie.

Now, this PR makes it so that all after the example charlie SL cryos, spawning puts the new SL into the only squad with less then the maximum number of SLs(1), ie: charlie.

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -419,7 +422,28 @@ private void OnPlayerSpawning(PlayerSpawningEvent ev)
         var query = QueryActiveRules();
         while (query.MoveNext(out _, out _, out var comp, out _))
         {
-            if (GetSpawner(comp, job) is not { } spawnerInfo)
+            if (comp.MaxJobsPerSquad.Count == 0 && comp is CMDistressSignalRuleComponent)
+            {
+                var allJobs = _stationJob.GetRoundStartJobs(ev.Station!.Value);
```
DrSmugleaf, line 427: This will also error if station is null, nullable warnings need to be handled.
This will also not work for dynamic job slots based on player count later, GetJobs should be used instead but you would need to do this earlier, in the handler for rulePlayerSpawningEvent for example.

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -908,18 +932,53 @@ private Spawners GetSpawners()
...
-        rule.NextSquad[job] = next;
+        while (members.MoveNext(out var uid, out var member))
+        {
+            var name = member.Squad == null ? null : Name(member.Squad.Value);
+            var squadName = "Squad" + name;
+            _jobs.MindTryGetJob(_mind.GetMind(uid), out _, out var prototype);
+            if (prototype! == job)
+            {
+                jobNumPerSquad[squadName]++;
+            }
```
DrSmugleaf, line 954: This should be using rule.SquadIds like abov or EntityUid not concatenating the prototype name of the squad like this.

Content.Server/_RMC14/Rules/CMDistressSignalRuleComponent.cs
```diff
@@ -15,10 +15,13 @@ public sealed partial class CMDistressSignalRuleComponent : Component
...
     [DataField]
-    public Dictionary<EntProtoId, EntityUid> Squads = new();
+    public Dictionary<ProtoId<JobPrototype>, int?> MaxJobsPerSquad = new();
 
     [DataField]
-    public Dictionary<ProtoId<JobPrototype>, int> NextSquad = new();
+    public Dictionary<EntProtoId, EntityUid> Squads = new();
+
+    // [DataField]
+    // public Dictionary<ProtoId<JobPrototype>, int> NextSquad = new();
```
DrSmugleaf, line 24:
```suggestion
    public Dictionary<EntProtoId, EntityUid> Squads = new();
```

Content.Server.Database/Model.cs
```diff
@@ -441,6 +441,7 @@ public class Profile
         public string EyeColor { get; set; } = null!;
         public string SkinColor { get; set; } = null!;
         public int SpawnPriority { get; set; } = 0;
+        public int SquadPriority { get; set; } = 0;
```
DrSmugleaf, line 444: This should just be storing the squad's entity id then you don't need to map back and forth with an integer

DrSmugleaf (changes requested):

UbaserB:
Once this gets merged im going to start playing rmc14 again

PR #3242, opened 2024-07-23
Sorokyne Ports

noctyrnal:
## About the PR

Ports Walls, windows and tiles for Sorokyne Strata

## Why / Balance

Needed for Strata

## Media

- [ X ] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

no cl


noctyrnal:
Might need more tiles in the future, but just PRing as is for now since the walls are also needed for Chance's Claim AFAIK

noctyrnal:
> error on line 35 of window_strata.yml `entity: CMWindowStrata` instead of `entity: RMCWindowStrata` (mentioned in your errors) error line 16 of strata_windows.yml `graph: RMCWindowSolaris` `node: windowFrameSolaris` instead of `graph: RMCWindowStrata` `node: windowFrameStrata` (not mentioned anywhere)

Already marked as draft so I can sort these o7

noctyrnal:
> You're doing great, just lending my eyes lol. I have window PTSD.
> 
> Also ice wall sprite should be the dirty variant probably? I don't see the bluish one used anywhere, but the brownish one is used in both strata and chance's claim for some reason.

According to footage I found on Strata, it seems that the clean ice is actually used rather than the dirty, and that it might just be a mapping error? Though, if thats wrong ill happily flip it to the dirt variant.

PR #3253, opened 2024-07-24
Fix Smartgunner farsight

Vermidia:
## About the PR

Fix smartgunner farsight zoom staying when unequipped. It also won't work or give the action outside of the eye slot now.

## Why / Balance

Discord issues, you could give a whole squad farsight with this.

## Technical details

Added equipped and unequipped events to check for if it's been removed or re-added (with farsight on). When reseting zoom, the EyeComponent is now gotten to reset max zoom to whatever zoom is defined there. Also gave the farsight component a slotflags field so it has to be in certain slots to count as equipped.

## Media

https://github.com/user-attachments/assets/d8463e72-ea74-450f-93bc-348f3432dc9e

(I recorded this before fixing it not working when equipped and already active)

## Breaking changes

**Changelog**

:cl:
- fix: Smartgunner combat-sight giving permanent far sight

### DIFF
--- a/Content.Shared/_RMC14/FarSight/FarSightSystem.cs
+++ b/Content.Shared/_RMC14/FarSight/FarSightSystem.cs
@@ -41,18 +41,7 @@ public sealed class FarSightSystem : EntitySystem
         Dirty(ent);
 
         var user = args.Performer;
-        if (ent.Comp.Enabled)
-        {
-            _eye.SetMaxZoom(user, ent.Comp.Zoom);
-            _eye.SetZoom(user, ent.Comp.Zoom);
-        }
-        else
-        {
-            if (TryComp<EyeComponent>(user, out var eye))
-                _eye.SetMaxZoom(user, eye.Zoom);
-
-            _eye.ResetZoom(user);
-        }
+        SetZoom(ent.Comp.Enabled, user, ent.Comp);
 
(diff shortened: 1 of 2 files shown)


Dutch-VanDerLinde:
does this fix runner farsight glitch too?

Vermidia:
Different components, so no I think

Content.Shared/_RMC14/FarSight/FarSightSystem.cs
```diff
@@ -42,10 +48,47 @@ private void OnFarSightAction(Entity<FarSightItemComponent> ent, ref FarSightAct
...
+            _eye.SetMaxZoom(user, ent.Comp.Zoom);
+            _eye.SetZoom(user, ent.Comp.Zoom);
+        }
+        else
+        {
+            if (TryComp<EyeComponent>(user, out var eye))
+                _eye.SetMaxZoom(user, eye.Zoom);
+
+            _eye.ResetZoom(user);
+        }
```
DrSmugleaf, line 79: This should be its own method so that this and OnFarSightAction don't have to repeat it

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf (approved):

PR #3258, opened 2024-07-25
Add the Boiler, toxic tail stab for Spitter and Boiler, max 1 target tail stab, health analyzer OD amounts

DrSmugleaf:
## Media

**Changelog**
:cl:
- add: Added the tier 3 Boiler xenonid, it has an ability to shoot a glob of corrosive acid gas after a delay, a toxic tail stab, an an acid shroud ability to escape danger. It also has the ability to see further in one direction and glows in the dark so that marines can see it. The neurotoxin glob is not implemented yet. The Boiler's spray acid always stuns.
- add: Added toxic tail stab to the Spitter and Boiler, which injects the target hit with 1 unit of molecular acid, dealing burn and toxin damage over time unless treated. This will show up as an unknown reagent on the health analyzer.
- add: The health analyzer will now tell you if someone is overdosing on a specific chemical.
- fix: Fixed tail stab hitting more targets than intended. It now only hits a maximum of 1 target, down from 3.
- fix: Fixed tail stab having no popups missing a sound.
- fix: Fixed bleeding also removing chemicals from your bloodstream. This will later be added in the form of a dyalisis machine in medbay.


deltanedas:
WERE VERY TALL QUEEN BACK

PR #3321, opened 2024-07-26
Fix bombard range

### LINKED ISSUE #3319: Boiler long-range sight doesn't allow bombard in new range
## Description

When you use long-range sight, you should be able to target tiles with bombard that you can now see because of it. It's how it works in cm-13 but not here.

**Reproduction**

1. Use long range sight
2. Try to target the furtherest away tile from you in your new sight(doesn't work)

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Make it so you can actually target things far away with bombard when using long-range sight (by bumping up the range to 15 like in cm-13)

## Why / Balance

Made long sight useless, fixes #3319

## Technical details

just range changes

## Media

https://github.com/user-attachments/assets/e96caadd-f162-4889-a0e8-f8b2805b208b

## Breaking changes

**Changelog**

:cl:
- fix: Fix bombard range being too small to take advantage of long-range sight

### DIFF
--- a/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
+++ b/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
@@ -398,7 +398,7 @@
       sprite: _RMC14/Actions/xeno_actions.rsi
       state: bombard
     event: !type:XenoBombardActionEvent
-    range: 14
+    range: 10
     checkCanAccess: false
   - type: ActionSharedCooldown
     id: ActionXenoBombard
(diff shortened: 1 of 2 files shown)


Kadeo64:
speedmerge so i can bully marines pretty please

Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
```diff
@@ -398,7 +398,7 @@
       sprite: _RMC14/Actions/xeno_actions.rsi
       state: bombard
     event: !type:XenoBombardActionEvent
-    range: 9
+    range: 14
```
QuietlyWhisper, line 401:
```suggestion
    range: 10
```

QuietlyWhisper (changes requested):
Shouldn't be the entire length of the screen due to the larger viewport, but this is supposed to be range 10 so that works out

QuietlyWhisper (approved):

PR #3330, opened 2024-07-26
Add sniper specialist 'prepare position' ability

Dutch-VanDerLinde:
- requires #3390

## Why / Balance
CM13 content

## Media

https://github.com/user-attachments/assets/83e4423e-4ba5-4909-9542-7fb340ec665c

:cl:
- add: Added the 'prepare position' ability to the sniper specialist, which allows them to go invisible while standing still.

### DIFF
--- a/Resources/Changelog/RMC14.yml
+++ b/Resources/Changelog/RMC14.yml
@@ -1,13 +1,6 @@
 ﻿Entries:
 - author: DrSmugleaf
   changes:
-  - message: Fixed runners being able to pull while leaping.
-    type: Fix
-  id: 40
-  time: '2024-04-13T08:09:14.0000000+00:00'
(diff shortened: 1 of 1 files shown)


Dutch-VanDerLinde:
needs scout specialist PR for the IFF marine component

LankLTE:
I'd also advise using the stealth system from the scout PR- actual transparency rather than the stealth box invisibility. That's how it's handled in 13 and is also probably just harder to see / easier to hide.

Dutch-VanDerLinde:
> I'd also advise using the stealth system from the scout PR- actual transparency rather than the stealth box invisibility. That's how it's handled in 13 and is also probably just harder to see / easier to hide.

yeah gonna do that later (also less taxing on the server because it doesnt use an expensive ass shader)

PR #3349, opened 2024-07-27
Add conversion for SS14 metal to RMC-14 metal

thymol-blue:
## About the PR

Added method to convert SS14 metal to RMC-14 metal

## Why / Balance

I see lots of SS14 metal laying around which is a bummer because that is metal that could be used

## Technical details

This is... also a webedit... so I apologize if it doesn't actually work but...

Added a SS14 to RMC-14 graph and also added the yml for putting it in the construction menu.

## Media

## Breaking changes

**Changelog**

:cl:
- add: Added a way for non-conformant metal to be converted into military-grade steel for usage in barricades and other military operations.


thymol-blue:
This is going to be really dumb but... is there any way we can make it not CRLF with a webedit 😭

QuietlyWhisper:
I'd rather the actual prototypes be fixed

PR #3366, opened 2024-07-28
Skill Pamphlet Functionality

LankLTE:
## About the PR

Adds basic functionality for skill pamphlets. Can be used to give a basic skill in one thing, you get one use, except for some exceptions (policing and intelligence, etc.).

## Why / Balance

13 content. Also necessary for spotter. Still not in the vendors as neither JTAC nor engineering are implemented skills so they'd just be a waste of points. 

## Technical details

Has fields for giving components (useful for spotter, powerloader, etc.) and for giving skills. Can also give unique popups with whitelists (again, useful for spotter). Non-skill pamphlets weren't implemented as their components don't exist yet. 

## Media

https://github.com/user-attachments/assets/383c3c2e-a25f-4307-b619-f80328dc0689

No CL as nothing will change for players.


DrSmugleaf (approved):
ty boss

PR #3381, opened 2024-07-29
Matching parasite sprites

Hyenh:
## About the PR

- Requires #3334
Changed hugged sprite to better match the actual parasite sprite.
## Why / Balance

- fixes #2882
## Media

Before:
image
After:
image
image

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**

:cl:Hyenh

PR #3334, opened 2024-07-26
Parasite is now an item that fits on the mask slot rather than being an overlay sprite

Dutch-VanDerLinde:
## About the PR
fixes #3255

## Media

https://github.com/user-attachments/assets/4247ad12-2258-41e7-9824-04ace20e0cab

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

## Breaking changes

**Changelog**
none

PR #3027, opened 2024-07-13
make wiggles a small host

QuietlyWhisper:
:cl: Whisper
- add: Wiggles is now considered a small host.

PR #3387, opened 2024-07-29
Make mr wiggles and jones a small host

Dutch-VanDerLinde:
- requires #3334
- requires #3381

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/NPCs/animals.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/NPCs/animals.yml
@@ -54,7 +54,7 @@
     prob: 0
 
 - type: entity
-  parent: MobCorgiPuppy
+  parent: [MobCorgiPuppy, RMCSmallHost]
   id: CMMobWiggles
   name: Mister Wiggles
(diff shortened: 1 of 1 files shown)


Tunguso4ka:
[#3027](https://github.com/RMC-14/RMC-14/pull/3027#issuecomment-2227177193)

Dutch-VanDerLinde:
> [#3027](https://github.com/RMC-14/RMC-14/pull/3027#issuecomment-2227177193)

it'll look better once #3334 is merged

QuietlyWhisper:
Smug was against having them as hosts because of cm13 parity and minor balancing reasons but with the two required prs it would be okay in my personal opinion as far as visuals go

TheDoctor1977:
Honestly don’t think this is a great idea, putting CM13 parity aside, there’s a lot of good Xeno RP to be had around unhuggable “pet” mobs like these.

Dutch-VanDerLinde:
xeno rp is nonexistent

Dutch-VanDerLinde:
Oops

DrSmugleaf (changes requested):
I prefer keeping these non-infectable like 13, both so they aren't just treated as another monkey by xenos and so that they don't just get another free host round-start

image

Dutch-VanDerLinde:
Ok

PR #3390, opened 2024-07-29
Scout Weapon Specialist

LankLTE:
## About the PR

Scout Weapon Specialist ported with most of its' gameplay features. 
To my knowledge only missing the alternative ammo types and the scout binoculars, as well as C4. All of which simply aren't implemented yet.

## Why / Balance

Part of the roadmap, porting 13 content.

## Technical details

Probably the most general one is adding a (semi) generic entitystealthsystem, which uses sprite transparency to make the sprite harder to see. EntityTurnInvisible is used to indicate that an entity has the ability to turn invisible, EntityActiveInvisible is used to indicate that it is currently invisible (and the sprite will be set to the opacity field). 
There's also RMCPassiveStealth, which can be given to an entity to passively set it's transparency to the one set. Hardcoded not to work while folded or open (if possible), but that can be changed if it becomes an issue down the line. 
ClothingRequireEquippedComponent was changed to not be hardcoded to say you need a smartgun.
EntityIFFComponent was added to allow an entity to not take friendly fire damage from other entities in the set faction. GunIFFSystem was changed to check for both the projectile's IFF and the entity's. 
SharedAcidSystem was changed to check for UncloakOnHit and do so. 
Everything else is part of ThermalCloak system, as general as it can be but fairly specific due to the fairly specific conditions of cloak (deactivating on acid, etc.) 

## Media

https://github.com/user-attachments/assets/54176203-6331-491e-b68b-1a4f551227d6

## Breaking changes

If guns parent oddly, not all of them may work with friendly-fire immunity. Should be fine for the most part.

**Changelog**

Make sure to take this Changelog template out of the comment block in order for it to show up.
:cl: Lank
- add: Added the Scout Weapon Specialist, and their associated gear.
- add: Added A19 Mags for the M4SPR Custom to the requisition console.


LankLTE:
Marking this as ready for review as, I believe it's finished and ready for review. I'll be filling out the PR description shortly.

Content.Shared/_RMC14/Stealth/RMCPassiveStealthComponent.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Content.Shared._RMC14.Stealth;
+
+/// <summary>
+///     Will slowly lower an entity's opacity
+/// </summary>
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class RMCPassiveStealthComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public float MinOpacity = 0.2f;
```
DrSmugleaf, line 13: On the video it seemed instant but in 13 it is supposed to slowly turn the entity invisible down to 0, at least for the stealth body bag that the scout gets. Like, actually 0. You can probably do the slowly part by storing a timespan of when this was activated and a delay for full invis, then interpolate the invis amount from that.

Content.Shared/_RMC14/Stealth/RMCPassiveStealthSystem.cs
```diff
@@ -0,0 +1,81 @@
...
+        if (ent.Comp.Enabled.Value)
+        {
+            ent.Comp.Enabled = false;
+            RemCompDeferred<EntityActiveInvisibleComponent>(ent.Owner);
+            return;
+        }
+
+        ent.Comp.Enabled = true;
+        var invisibility = EnsureComp<EntityActiveInvisibleComponent>(ent.Owner);
+        invisibility.Opacity = ent.Comp.MinOpacity;
```
DrSmugleaf, line 79: Missing a dirty call here after the data changes
```suggestion
        invisibility.Opacity = ent.Comp.MinOpacity;
        Dirty(ent.Owner, invisibility);
```

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf (approved):
Thanks for the PR!

PR #3400, opened 2024-07-30
prevent pulling a living queen

deltanedas:
## About the PR
you cant pull queen when shes alive

## Why / Balance
no idea about if this is frowned upon / disabled in cm13 but its a really shitty tactic to get queen to a cade **completely silently** so marines get no warning of an upcoming screech

## Technical details
added `NoPullInState` component/system

## Media
it work trust, cant pull when alive
when pulling crit queen on weeds when she wakes up, stops being pulled

## Breaking changes
no

**Changelog**
:cl:
- tweak: The Queen can no longer be pulled when alive.


DrSmugleaf (changes requested):
In 13 she makes a footstep sound when pulled, same should be done here instead of this

PR #3403, opened 2024-07-30
Knockedout fixes + shake stun adjustment

Vermidia:
## About the PR

Current instances where CM-13 would use the paralyze status now properly apply all effects that should be with it (mostly just in parasite system, I forgot the shaking from infection should both mute you and blind you).

Also shaking someone out of a stun from recent infection should now clear up their blindness & muteness too.

Shaking someone out of a stun no longer removes the status, instead subtracting 6 seconds from the status effect timer.

## Why / Balance

Fixes discord issue, CM-13 parity

## Technical details

So we don't have a combined status effect that does stun, blind, and muted at the same time. We do have seperate status effects for each of them however. SharedParasiteSystem now uses the TempBlind status instead of what it was doing before.

Shaking now checks for blind & muted and if they exist and their times are the same removes duration from them. Stuns and knocked down also only remove duration now. It's 6 seconds to each [Copied from CM-13].

Not sure if there's a better way to just make a "knocked out"/Unconscious status that does all of these. 

## Media

https://github.com/user-attachments/assets/7ab9a1fc-0d70-44d4-a349-49451d1ace9b

## Breaking changes

hope not
**Changelog**

:cl:
- tweak: Shaking someone out of a stun now only removes 6 seconds from the stun instead of instantly ending it per help.
- fix: When shaken out of the initial infected state, you lose the blindness & muteness that came with it

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
@@ -8,7 +8,6 @@ using Content.Shared.Damage;
 using Content.Shared.DoAfter;
 using Content.Shared.DragDrop;
 using Content.Shared.Examine;
-using Content.Shared.Eye.Blinding.Systems;
 using Content.Shared.Ghost;
 using Content.Shared.Humanoid;
 using Content.Shared.Interaction;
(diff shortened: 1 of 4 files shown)


Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -203,13 +200,17 @@ private void OnVictimInfectedMapInit(Entity<VictimInfectedComponent> victim, ref
 
     private void OnVictimInfectedRemoved(Entity<VictimInfectedComponent> victim, ref ComponentRemove args)
     {
-        _blindable.UpdateIsBlind(victim.Owner);
+        if (_status.HasStatusEffect(victim, "Muted", null) && _status.HasStatusEffect(victim, "TemporaryBlindness", null))
+        {
+            _status.TryRemoveStatusEffect(victim, "Muted");
+            _status.TryRemoveStatusEffect(victim, "TemporaryBlindness");
+        }
```
DrSmugleaf, line 207: This might cause issues if something else applies a longer muted and temporary blindness status but I think it's fine for now

DrSmugleaf (approved):
Thanks for the PR!

PR #3414, opened 2024-07-30
Fix charge not hitting barricades

Vermidia:
## About the PR

Charge can hit barricades again

## Why / Balance

Broke originally when fixing targets

## Technical details

Can attack ability target now has a bool if it can attack non-marines and if true will return true if the target doesn't have a mobState component.

Doesn't fix them going through barricades with charge unfortunately.

## Media

https://github.com/user-attachments/assets/7def6f14-8178-44ff-b86c-9d16ff8e05ec

## Breaking changes

no
**Changelog**

:cl:
- fix: Crushers can hit barricades with charge


DrSmugleaf (approved):
Thanks for the PR!

Issue #3023, opened 2024-07-13
Parasites should use impregnated sprite instead of dead sprite when they die from infecting


Vermidia:
## Description

When dying after an infection, parasites right now will use a dead sprite, instead of using the impregnanted sprite we have.

Makes it easier to tell if a parasite actually died or if it just infected someone.

**Reproduction**

**Screenshots**

**Additional context**

deltanedas:
likewise it also makes a xeno death announcement when it should either:
- not announce since an infection is good as opposed to a suiciding marine-controlled parasite
- announce they infected someone

Dutch-VanDerLinde:
Fixed

Vermidia:
yep

PR #3419, opened 2024-07-31
Parasites use impregnanted sprite when dying from infecting + infect announcement

Vermidia:
## About the PR

When a parasite dies from infecting instead of the normal announcement you get one where it says they infected a host. Also they now use the impregnanted sprite when they died after infection, with a description so you don't confuse them for sleeping. Also gave sleeping paras a slightly diff sprite.

## Why / Balance

Implements #3023

## Technical details

Mostly made a few checks to check for the ParasiteSpent component. Also moved the impregnanted file to the main parasite folder (and made it's size match everything else)

## Media

https://github.com/user-attachments/assets/9284a05a-4391-4617-8662-2e63c3b6658d

## Breaking changes

I hope not
**Changelog**

:cl:
- tweak: Parasites now get a different sprite and announcement if they die from infecting a host.


DrSmugleaf (approved):
Thanks for the PR!

PR #3422, opened 2024-07-31
Variable-zoom scopes, damage falloff, masterkey damage modifiers

### LINKED ISSUE #3570: M63 attachment bug
## Description
You can keep the effect of the submachinegun brace if you activate it and then switch to the foldings stock giving you both better recoil and you will never drop your gun. 

**Reproduction**
1. Put a submachinegun brace 
2. Activate it
3. Replace it with a folding stock
4. Be unable to put down weapon and have recoil control

**Screenshots**
N/A

**Additional context**
N/A

### LINKED ISSUE #3665: You can deploy bipods on glass shards???
image

Siigiil:
## About the PR
As the title says, this implements multiple zoom levels for scopes, damage falloff for guns and adds the masterkey's damage modifiers. Also fixes a bunch of bugs. See the changelog for details.

Closes #3570
Closes #3665
Closes #3422

## Why / Balance
Because porting.

## Technical details
Falloff is handled by two components: RMCProjectileDamageFalloffComponent and RMCWeaponDamageFalloffComponent.
The former determines the base falloff values and is contained in the projectile. The latter holds the modifier applied by the weapon.

Due to how scuffed max ranges for projectiles are on the live server's tickrate, I instead added a falloff threshold that instantly lowers the projectile's damage as much as it can once it reaches its maximum range.

The current minimum damage is 5% of the original damage. The reason I didn't set it to zero is so we don't have stray shots hit someone and do absolutely nothing. This way, you can see that the projectile does something when it hits you, so it both doesn't look like it's bugged and it doesn't do much damage. Close enough to having a maximum range.

If it's better to add maximum ranges to all projectiles even with how scuffed they are, tell me and I'll do it.

The masterkey in CM fires a worse version of buckshot, the additional projectiles in particular do much less damage. Implemented this here by adding RMCExtraProjectilesDamageModsComponent, which modifies the damage of any additional projectiles fired by the weapon in one shot.

## Upstream file changes
Incorporates the changes from https://github.com/space-wizards/space-station-14/pull/31300, since otherwise falloff and the masterkey damage mods are fucked.

## Media
Falloff:

https://github.com/user-attachments/assets/9714fd44-6f8d-48c6-93e9-7065682a5009

Masterkey:

https://github.com/user-attachments/assets/b19469b4-21ad-4c09-a1e0-25f420ab3ff7

**Changelog**

:cl: Sigil
- add: The M96S integrated scope can now toggle between 2x and 4x magnification. The former allows you to move while aiming.
- add: Damage falloff for bullets added. In broad strokes: The further you are from your target, the less damage your shots will do. Notable weapons and attachments: M63 bullets start rapidly losing damage past 4 tiles, the M4SPR removes falloff from its ammo, M96S ammo has no falloff, buckshot becomes useless at 4 tiles, slugs at 8 and flechettes at 12, active scopes reduce falloff and the silencer increases it.
- tweak: The masterkey now deals massively increased damage to walls, doors and certain other structures. Its damage to other targets has been reduced to its intended values. The primary pellet now deals 55 damage, while the three additional ones deal 20 each.
- tweak: Brought all scope doafters in line with how they are on CM.
- fix: A weapon with a magnetic harness attached no longer becomes immune to gravity if you throw it while something else is in your suit storage.
- fix: Buckshot, slugs and flechettes now do piercing damage instead of blunt.
- fix: The masterkey can no longer be fired when the gun it's attached to is not wielded.
- fix: The bipod can no longer be instantly deployed on glass shards.
- fix: The masterkey can no longer use ammo other than buckshot.
- fix: The arm brace now properly deactivates when swapped out while still active, and doesn't permanently glue the SMG to your arm.


DrSmugleaf (approved):
I need to go over the code but it looks good, might wait on merging because I think the falloff will affect marines too heavily at the moment

DrSmugleaf (approved):
Thanks for the PR!

PR #3465, opened 2024-08-02
Corrosive Acid Upgrading 

Guess-My-Name:
## About the PR

Xenos can now apply corrosive acid onto entities that already had corrosive acid on them, as long as they are applying a stronger type of acid.
Weak acid can be replaced with normal or strong, normal can be replaced with strong.

## Why / Balance

Drone or Sentinel applying weak corrosive acid can slow down entire xeno force, especially during hijack, and it feels like replacing/upgrading already applied acid was a missing feature.
Also, made a discord post [here](https://discord.com/channels/1168210010233376858/1268233005605064716/1268233005605064716) and feedback was positive.

## Technical details

I also added a calculation on how much time should be applied when acid is replaced, so we won't end up in a situation where weak acid was 30 seconds away from melting an entity and it was replaced with strong acid that takes 60 seconds. This is the reason for keeping the extra data in CorrodingComponent.

## Media

https://github.com/user-attachments/assets/e549fae8-7ec0-4a1f-acc6-4157c7118de2

## Breaking changes

**Changelog**

:cl:
- add: Xenos can now apply stronger versions of corrosive acid onto already corroding structures.

### DIFF
--- a/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
+++ b/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
@@ -1,4 +1,4 @@
-﻿- type: entity
+- type: entity
   parent: ActionXenoBase
   id: ActionXenoAcidNormal
   name: Corrosive Acid (100)  # TODO RMC14 proper plasma costs
@@ -13,6 +13,7 @@
       acidId: XenoAcidNormal
       plasmaCost: 100
       time: 150
+      strength: 1
     checkCanInteract: false
 
 - type: entity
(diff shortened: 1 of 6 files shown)


thymol-blue:
I believe you should be able to apply all three types on the same entity, not replace, and their effects stack in CM13

Guess-My-Name:
Well, after last update we got corrosive acid reworked from 'delete after X seconds' into damage over time, and also different types of acid can now be applied onto the same entity, so this entire PR is redundant.

Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
Bobert-Brockson, line 1: If this PR is revived my preference would be XenoAcid{X} being used for the calculation. It reflects the enum pattern used elsewhere in this PR and prevents strength and type decoupling.

PR #3474, opened 2024-08-02
Access lock Booze-Mat

Tunguso4ka:
## About the PR

Access lock Booze-Mat for Kitchen or Colony accesses.

"No more drunk marines, only drunk officers" update.

**Changelog**

:cl:
- fix: Booze-O-Mat is now access locked!


DrSmugleaf (approved):
no more alcoholism (mostly)

DrSmugleaf (commented):
i don't even think there's booze in 13 on the almayer but we revisit this later

Tunguso4ka:
> i don't even think there's booze in 13 on the almayer but we revisit this later

Yeah, you right, Booze-O-Mat mapped only on distress signal maps.
зображення

PR #3478, opened 2024-08-02
Add grenade specialist bag ID lock

Dutch-VanDerLinde:
https://github.com/user-attachments/assets/6c3802ce-838f-4d1e-8333-de92655ea63b

## Breaking changes

:cl:
- add: Grenade specialist's bag now has an ID lock


Dutch-VanDerLinde:
> Clever port implementation, so this allows the user to lock the grenade bag like a crate? Does this allow the user to open it up without unlocking it first?
> 
> The SquadLead access I'm guessing is to try and emulate the 'command level access can remove the lock' feature, dunno if there's a better access level for that. Maybe whatever SOs and above have for the CIC?
> 
> The missing port stuff would just be making it so the user can access it without unlocking, and that other people can't (even SLs) without removing the lock completely. Nonessential, this is a good enough equivalent for grenade specs to not get robbed blind by random marines.
> 
> I'd say definitely remove the SL access and change that to a command one though.

who is "the user"? the weapons specialist that got the kit?

Dutch-VanDerLinde:
or is it whoever equipped the bag

Dutch-VanDerLinde:
> > > Clever port implementation, so this allows the user to lock the grenade bag like a crate? Does this allow the user to open it up without unlocking it first?
> > > The SquadLead access I'm guessing is to try and emulate the 'command level access can remove the lock' feature, dunno if there's a better access level for that. Maybe whatever SOs and above have for the CIC?
> > > The missing port stuff would just be making it so the user can access it without unlocking, and that other people can't (even SLs) without removing the lock completely. Nonessential, this is a good enough equivalent for grenade specs to not get robbed blind by random marines.
> > > I'd say definitely remove the SL access and change that to a command one though.
> > 
> > 
> > who is "the user"? the weapons specialist that got the kit?
> 
> In CM it's whoever locked the bag with their ID. They can access it without unlocking, but it prevents everyone else. This implementation doesn't exactly work like that but it's close enough, and serves a similar purpose. But if they can't open it without unlocking, that makes it harder to grab nades from the satchel or reload from a worn satchel.
> 
> It's minor so don't worry if you can't implement it, but still something to note.

do pill canisters have a similar ID lock?

PR #3492, opened 2024-08-03
Tracker sidebar alerts

ShadowCommander:
## About the PR

Tracker for xenos to follow queen.

Tracker on headsets for marines to follow squad leader and fireteam leader.

## Media

## Breaking changes

**Changelog**
:cl: ShadowCommander, DrSmugleaf
- add: Added Squad Leader tracker for marines and xenonid Queen tracker for xenonids.


ShadowCommander:
I was trying to decide whether to split out the team/hive tracking lists into a new component on the team entity or not.

DogZeroX:
Could we probably have a tracker for hivecore?

ShadowCommander:
Things to look at or TODO:

- [ ] I made the tracker alerts only update once per second, but this causes it to really show that Alerts aren't predicted at all.
- [ ] The changes to SquadSystem.AssignSquad might be able to be moved to TrackerAlertSystem and triggered through more events in SquadSystem.
- [ ] This doesn't check that the maps are the same yet, so it'll point to the coordinates, but it might not be the same map.
- [ ] Input validation on entity selected by the UI. So that malicious packets can't just set the tracker anything.
- [ ] Maybe move Xeno add and remove on mob state change to use RMCTrackerAlertTargetComponent or some new component.
- [ ] Handle squad leader death.
- [ ] Doesn't handle Fireteam leaders. (Only allowing selecting assigned FT leader or SL. Block selecting unassigned FT leaders.) Can probably be done by editing OnGetTrackerAlertEntries.


Also stuff from reading CM code:
```
        // queen_locator dm
        // If hud doesn't exist exit
        // Get type of tracker
            // Queen
            // Hive
            // Leader
            // Tunnel
        // else no tracker
            // Is tracking queen
            // reset tracker to empty
            // If not is tracking queen
                // Track queen
            // return
        // Point to interiors of vehicles
        // If zlevel doesn't match OR distance < 1 tile (adjacent) OR src == tracking
            // trackon-center
        // Else
            // If the self fake zlevel == target fake zlevel (e.g. almayer upper level vs lower level)
                // Directional tracker icon
            // else
                // trackon-center
```
Not sure if we want to do range checks for <1 tile distance just for the center icon state here for performance reasons, but the other stuff looks good.

PR #3495, opened 2024-08-03
CM Flares can now light cigarettes. 

Hyenh:
## About the PR

Flares can now light cigarettes. 
## Why / Balance

It look weird before.
## Technical details

Uses ignitionsource, same as SS14 flare. Will not set welders to explode. Spent flares do not light cigarettes.
## Media

https://github.com/user-attachments/assets/a06c97ba-09cb-42d0-bbd4-82b0d9421a0d

## Breaking changes

**Changelog**

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Tools/flares.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Tools/flares.yml
@@ -42,6 +42,8 @@
     radius: 1
     energy: 5.0
     netsync: false
+  - type: IgnitionSource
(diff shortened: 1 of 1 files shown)


DrSmugleaf (changes requested):
We removed ignition source from these because it was a substantial source of lag on a live server, this would need to be handled in a different way.

Hyenh:
Ah, unfortunate. I'm not sure how I can solve this

deltanedas:
is the ignition source lag atmos related because that could be stripped out since no real atmos

PR #3496, opened 2024-08-03
fix(CMFolderBase): Fix CMFolderBase to work as intended.

CaasGit:
## About the PR
The CMFolderBase was missing a few things that the BoxFolderBase has defined. CMFolderBase can't use BoxFolderBase as its parent due to the sprite layering, so this just pulls in a few missed things that made folders not work as expected and gets them aligned with the upstream folders.

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Make folders work as expected.


DrSmugleaf (approved):
Thanks for the PR!

PR #3501, opened 2024-08-03
Fix pulling mobs that just died from rubberbanding

### LINKED ISSUE #3199: Pulling a character that just died will cause strange behaviour
- pull a crit character
- they die while you're pulling them
- the game thinks you are pulling them, but you are not
- stop pulling
- get teleported back to the dead character's position

Dutch-VanDerLinde:
fixes #3199

we do not need to disable collision when a mob dies because the projectile target behavior has already been added


DrSmugleaf (approved):
Thanks for the PR!

PR #3528, opened 2024-08-04
Fix gas masks to work with air tanks like the medical ones.

BramvanZijp:
## About the PR
Gas masks now allow you to turn on internals just like the medical ones.

## Why / Balance
Because it not working is a bug, plus it helps tone down the currently very oppressive boiler smokes, especially during hijack.
And if it is not a bug, somebody needs to change the fact that oxygen tanks come in the same crate as gas masks when bought in req.

## Technical details
The med gas mask uses the base ss14 gas mask as a parent, which the standard gas mask does too now, allowing you to turn on your internals if you also have an air tank.

## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** __this PR does not require an ingame showcase__

**Changelog**

:cl: BramvanZijp
- fix: Standard gas masks now allow for the use of internals.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
@@ -9,7 +9,7 @@
     hideOnToggle: true
 
 - type: entity
-  parent: CMBaseMask
+  parent: ClothingMaskGas
   id: CMMaskGas
   name: gas mask
   description: A face-covering mask that can be connected to an air supply. Filters harmful gases from the air.


DrSmugleaf (changes requested):
How does this affect Boiler gas? Boiler gas doesn't work through inhalation of surrounding chemicals, atmos simulation isn't run for most things

BramvanZijp:
> How does this affect Boiler gas? Boiler gas doesn't work through inhalation of surrounding chemicals, atmos simulation isn't run for most things

I stand corrected on that then, but they should still work like the other ones for a few reasons, including blocking consumption, etc.

deltanedas:
can confirm that atmos devices arent ticked at all including internals, so if you have 304kpa anesthetic tank it never gets inhaled so stays at 1mpa forever

Dutch-VanDerLinde:
what do gas masks do against boiler gas in cm13

Vermidia:
Nothing :p

Dutch-VanDerLinde:
lol

QuietlyWhisper:
> can confirm that atmos devices arent ticked at all including internals, so if you have 304kpa anesthetic tank it never gets inhaled so stays at 1mpa forever

We are going to need this for surgery, but maybe we should just bypass the usage of atmos systems and make a simple new one instead.. probably a lot easier and performant than working in atmos.

QuietlyWhisper:
closing as stale

PR #3543, opened 2024-08-05
Implements removing ammo from shotguns and grenade launchers

### LINKED ISSUE #2193: You should be able to take out shells from a shotgun

Siigiil:
## About the PR
See title. This lets you remove loaded ammo from the UGL, the masterkey, the M42A2 and the M83. To make this work properly, I made it so removing ammo from RMC guns is done with the activate in world key.

Didn't implement this for revolvers, as we don't have handfuls yet.

Closes #2193

## Why / Balance
Because porting.

## Technical details
The normal alternative activate in world key works by activating a verb on the target. It doesn't send an event to it or anything else I can catch and relay to the item's attachments. So instead I implemented removing ammo with the normal activate in world keybinding, since that actually raises an event I can work with.

BallisticAmmoProviderComponent contains a list of all ammo held in the item and doesn't automatically modify it if you just remove the ammunition from the item. The only way to interact with this list is to raise a specific event which will remove the number of rounds you set in it from the item. So for shotguns and the UGL, I'm having to jump through this hoop instead of just extracting the ammo from its container.

I don't particularly like this solution, but I can't come up with anything less shit.

## Media

**Changelog**

:cl: Sigil
- add: You can now remove ammo from shotguns and grenade launchers. To do so, press the "activate item in world" hotkey — E by default. This hotkey also works for other weapons.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/CMGunSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/CMGunSystem.cs
@@ -2,6 +2,7 @@
 using Content.Shared._RMC14.Marines.Skills;
 using Content.Shared._RMC14.Weapons.Common;
 using Content.Shared._RMC14.Weapons.Ranged.Whitelist;
+using Content.Shared.Containers.ItemSlots;
 using Content.Shared.Hands;
 using Content.Shared.Hands.Components;
 using Content.Shared.Hands.EntitySystems;
@@ -35,6 +36,7 @@ public sealed class CMGunSystem : EntitySystem
     [Dependency] private readonly SharedContainerSystem _container = default!;
     [Dependency] private readonly SharedGunSystem _gun = default!;
     [Dependency] private readonly InventorySystem _inventory = default!;
+    [Dependency] private readonly ItemSlotsSystem _slots = default!;
     [Dependency] private readonly SharedHandsSystem _hands = default!;
     [Dependency] private readonly SharedPhysicsSystem _physics = default!;
     [Dependency] private readonly SharedPopupSystem _popup = default!;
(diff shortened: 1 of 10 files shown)


Content.Shared/_RMC14/Weapons/Ranged/CMGunSystem.cs
```diff
@@ -356,4 +364,44 @@ private void OnRevolverUniqueAction(Entity<RevolverAmmoProviderComponent> gun, r
...
+
+        var ejectedAmmo = container.ContainedEntities[0];
+
+        // For guns with a BallisticAmmoProviderComponent, if you just remove the ammo from its container, the gun system thinks it's still in the gun and you can still shoot it.
+        // So instead I'm having to inflict this shit on our codebase.
+        if (TryComp(gun.Owner, out BallisticAmmoProviderComponent? ammoProviderComponent))
+        {
+            var takeAmmoEvent = new TakeAmmoEvent(1, new List<(EntityUid?, IShootable)>(), Transform(gun.Owner).Coordinates, args.User);
+            RaiseLocalEvent(gun.Owner, takeAmmoEvent);
+
```
metalgearsloth, line 390: Do you mean unspawnedcount or the contained entity or something else? If you're pulling it manually you can just call updateammocount manually iirc.

Siigiil: I mean the `Entities` datafield in BallisticAmmoProviderComponent. It contains the gun's ammo and is not updated if I just remove the ammo from its container. Which makes it so if I just remove the rounds from their container, the gun can still fire them even when they're lying on the floor.

Raising a TakeAmmoEvent is the only way I found of removing ammo from that list, so I can actually pull it out of the gun properly.

UpdateAmmoCount is protected. It also, as far as I can tell, only affects the ammo counter UI.

metalgearsloth: time to add access and fix it cheers on the bug report.

DrSmugleaf (commented):
I'll try to check this for the update after this one ty

DrSmugleaf (approved):
Thanks for the PR!

Issue #3550, opened 2024-08-05
Pocket-sized Guns


Kezu02:
## Description

A player reported that their M63 and M4SPR could fit in their pocket. Other weapons in the area are unaffected, but the ones they were using somehow had the ItemSizeChange comp that probably shrunk it. No idea how it showed up.

**Screenshots**
![image](https://github.com/user-attachments/assets/c0447398-2255-4624-ad35-5dd6dee6b6fb)

Siigiil:
It shows up when guns have size-changing attachments. The bug is because *somehow* the base size is not being set properly. I'm pretty sure I fixed this, but now it's popped up again. I'll try to figure out what's causing this shit.

DaFieFie:
It's definitely attachments related. Still happening.

Dutch-VanDerLinde:
Fixed

PR #3551, opened 2024-08-05
Fixes weapons spontaneously shrinking. Hopefully.

Siigiil:
## About the PR
Another go at getting rid of weapons shrinking when you put size-changing attachmens on them.

I still can't get this to happen in localhost, so I *can't test the fix.* But I *think* the extra checks I added should stop it from continuing to happen.

Unless I'm wrong again, this will deal with #3550. I will close it manually once I can confirm that the bug is gone.

## Why / Balance
Because recurring bug.

## Technical details
What I *think* is happening to cause this is that the item size change system somehow gets an invalid size value, which results in `_sortedSizes.IndexOf(prototype)` returning a -1. The checks I added should prevent that from happening, so guns will at least not *shrink* when you put attachments on them.

I also added ItemSizeChangeComponent to the base attachable holder, so the base size value is initialised when the item spawns.

What I don't understand is how this is only happening on the live server. Somewhere in the US, there's a wizard casting spells in a server room and laughing at me.

## Media

**Changelog**

:cl: Sigil
- fix: Hopefully fixed guns shrinking when you put size-changing attachments on them. If you encounter this bug after this is merged, report it.

### DIFF
--- a/Content.Shared/_RMC14/Item/ItemSizeChangeSystem.cs
+++ b/Content.Shared/_RMC14/Item/ItemSizeChangeSystem.cs
@@ -52,9 +52,12 @@ public sealed partial class ItemSizeChangeSystem : EntitySystem
 
     public void RefreshItemSizeModifiers(Entity<ItemSizeChangeComponent?> item)
     {
-        item.Comp = EnsureComp<ItemSizeChangeComponent>(item.Owner);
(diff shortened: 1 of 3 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #3553, opened 2024-08-05
Remove nesting the dead

### LINKED ISSUE #3512: Xenos are able to drag corpses to walls to put them on despite said corpse not being alive anymore
Title - Drag the corpse sprite to a wall

Vermidia:
## About the PR

You can't nest the dead anymore. Also some hivemind-ification of nesting messages

## Why / Balance

CM-13 parity
Fixes #3512

## Technical details

CanBeNested checks if the mob is dead now.

## Media

## Breaking changes

no
**Changelog**

:cl:
- fix: Xenos can no longer nest the dead.

### DIFF
--- a/Resources/Locale/en-US/_RMC14/xeno/xeno-nest.ftl
+++ b/Resources/Locale/en-US/_RMC14/xeno/xeno-nest.ftl
@@ -1,12 +1,13 @@
-cm-xeno-nest-securing-self = You secrete a thick, vile resin, securing {$target} into the alien nest!
+cm-xeno-nest-securing-self = We secrete a thick, vile resin, securing {$target} into the alien nest!
 cm-xeno-nest-securing-target = {$user} secretes a thick, vile resin, securing you into the alien nest!
 cm-xeno-nest-securing-observer = {$user} secretes a thick, vile resin, securing {$target} into the alien nest!
 
-cm-xeno-nest-pin-self = You pin {$target} into the alien nest, preparing the securing resin.
+cm-xeno-nest-pin-self = We pin {$target} into the alien nest, preparing the securing resin.
 cm-xeno-nest-pin-target = {$user} pins you into the alien nest, preparing the securing resin.
 cm-xeno-nest-pin-observer = {$user} pins {$target} into the alien nest, preparing the securing resin.
 
 cm-xeno-nest-failed = {$target} can't be put into a nest!
 cm-xeno-nest-failed-target-resisting = {$target} is resisting, ground them!
-cm-xeno-nest-failed-cant-there = You can't create a nest there!
+cm-xeno-nest-failed-cant-there = We can't create a nest there!
 cm-xeno-nest-failed-cant-already-there = There is already someone nested there!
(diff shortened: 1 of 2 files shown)


Resources/Locale/en-US/_RMC14/xeno/xeno-nest.ftl
```diff
@@ -1,12 +1,13 @@
...
+cm-xeno-nest-pin-self = We pin {$target} into the alien nest, preparing the securing resin.
 cm-xeno-nest-pin-target = {$user} pins you into the alien nest, preparing the securing resin.
 cm-xeno-nest-pin-observer = {$user} pins {$target} into the alien nest, preparing the securing resin.
 
 cm-xeno-nest-failed = {$target} can't be put into a nest!
 cm-xeno-nest-failed-target-resisting = {$target} is resisting, ground them!
-cm-xeno-nest-failed-cant-there = You can't create a nest there!
+cm-xeno-nest-failed-cant-there = We can't create a nest there!
 cm-xeno-nest-failed-cant-already-there = There is already someone nested there!
+rmc-xeno-nest-failed-dead = This host is dead.
```
Dutch-VanDerLinde, line 13:
```suggestion
rmc-xeno-nest-failed-dead = This host is dead, why would we want to nest it?
```

Rainbeon: The original version is better than this imo, or atleast a we *can not* nest it would be better. There is a very obvious reason for why they would want to nest it (so the Larva doesn't die... if you nest a dead host that falls off the wall, their sister larva will still be born). "Why would we want to do this?" feels fairly taunt-ish in tone on behalf of the game alongside the knowledge that nests deliberately allow larva to be born from dead hosts with the incubation speed increase.

Dutch-VanDerLinde:
it should only disallow uninfected dead hosts i n my opinion

Vermidia:
CM-13 disallowing nesting any dead. I think it's fair enough.

Rainbeon:
Doesn't this just entirely disallow crit captures for? Larva burst is tied to whether the body is rotted, whether the larva is born in time or not depends on whether they get nested afterward or not as a compensation mechanic as it speeds it up to be just slightly quicker than the rot time, from #2995. Now since you can't even nest infected marines, which there is an in-universe reason to do (so your sister inside the host doesn't also die), you just entirely lose the egg investment if they die. 

Often times whether marines get crit or not is entirely out of your control as well (As Marines often FF, unintentionally or on purpose to try and deny caps) and not being able to nest the infected after death removes any counterplay for that.

Vermidia:
Our crit time is generously long compared to CM-13 atm. You can live for awhile crit, so this won't affect much unless the person was already immediately going to die by the time you nest them.

The nesting bonus may be removed and incubation time increased in the future anyway (as they might in CM-13) so heavily crit captures would be unincentivized by those changes later.

DrSmugleaf (approved):
Thanks for the PR!

PR #3566, opened 2024-08-07
Make machete and throwing knife devour usable

Vermidia:
## About the PR

title

## Why / Balance

CM-13 parity. Both do less damage in general while devoured (60% of their original, so machete at 21 and throwing knife at 3) to match bayonet.

## Media

## Breaking changes

**Changelog**

:cl:
- tweak: You can use a machete or throwing knife while devoured


DrSmugleaf (changes requested):
I think in 13 you can unsheathe the machete from the pouch and maybe the throwing knife from the rig, if so you will have to give both the UsableWhileDevoured component as well, like boots do

Vermidia:
Have done so

DrSmugleaf (approved):
Pretty sure it will still have funky priority when using unholster keys since our unholster priority isn't 1:1 with 13's but that's a future PR's problem

PR #3569, opened 2024-08-07
You can now vault while taking damage

### LINKED ISSUE #3352: Vaulting should not be interrupted by damage
## Description

When vaulting onto a table/window frame, you should keep vaulting even when taking damage (as in CM-13, at least with xenos that I tested)

**Reproduction**

1. Start vaulting
2. Get hit and stop vaulting

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Removed a single line from the doafter declaration so you can keep vaulting while getting hit.

## Why / Balance

Fixes #3352, CM-13 parity

## Technical details

Had to modify the upstream file, but 1 line is 1 line so hopefully it's ok...

## Media

## Breaking changes

no?
**Changelog**

:cl:
- fix: Vaulting is no longer cancelled by damage

### DIFF
--- a/Content.Shared/Climbing/Systems/ClimbSystem.cs
+++ b/Content.Shared/Climbing/Systems/ClimbSystem.cs
@@ -223,7 +223,6 @@ public sealed partial class ClimbSystem : VirtualController
             used: entityToMove)
         {
             BreakOnMove = true,
-            BreakOnDamage = true,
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
Ideally we upstream this some day under a data field of some component some day

PR #3612, opened 2024-08-10
Fix being able to use nightvision items from inside your pockets

crystalHex:
Also fixes losing nightvision due to taking a nightvision item out of your pocket even when you have a nightvision item equipped in your eye/mask slot

## About the PR

Nightvision items worked while in pocket slots. In addition, removing a nightvision item from a pocket slot removed your nightvision even when you had another nightvision item equipped and activated.
This PR fixes that

## Why / Balance

Bug fix

## Technical details

Added a few lines of C# to check for item slot before equipping/unequipping/activating a nightvision item

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Fixed being able to use nightvision items in pockets.

### DIFF
--- a/Content.Shared/_RMC14/NightVision/NightVisionItemComponent.cs
+++ b/Content.Shared/_RMC14/NightVision/NightVisionItemComponent.cs
@@ -1,4 +1,5 @@
-﻿using Robust.Shared.GameStates;
+﻿using Content.Shared.Inventory;
+using Robust.Shared.GameStates;
 using Robust.Shared.Prototypes;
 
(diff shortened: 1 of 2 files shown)


Vermidia:
Instead of checking eyes and head directly, you should have the comp specify a slot.

crystalHex:
> Instead of checking eyes and head directly, you should have the comp specify a slot. (like farsight does)

Thanks, added the slot to the component, the system now checks if the equip/unequip slot matches with the slot in the component

DrSmugleaf (approved):
Thanks for the PR!

PR #3626, opened 2024-08-12
Increase ExamineDetail Range from 3 to 8 tiles

Rainbeon:
## About the PR
Increased the range of Character Description inspects from 3 to 8, allowing players to view others' character descriptions from much farther away.

## Why / Balance
This is purely a Quality of Life change to make it easier to see other's character descriptions without having to awkwardly move super close to them, especially given many situations both RP-related and combat-related demand you to be standing in a certain spot away from the character you wish to learn more about.

This does not affect balance at all, as RMC gameplay does not have an element of deception and disguises/identity hiding like standard SS14 does, which is why the range was so low to begin with (as viewing someone in greater detail would always give the same description per-character, you can't fake this, ruining your disguise). You can not hide your identity at all as a marine, and the ability to do wouldn't be important to the gameplay anyway, so being able to see flavor text from afar should cause no balance issues.

## Media
https://github.com/user-attachments/assets/fedbaff2-d071-4197-b18d-067703a4c3d7

**Changelog**

:cl:
- tweak: You can now view player character descriptions from much further away.

### DIFF
--- a/Content.Shared/Examine/ExamineSystemShared.cs
+++ b/Content.Shared/Examine/ExamineSystemShared.cs
@@ -41,7 +41,7 @@ namespace Content.Shared.Examine
         public const float DeadExamineRange = 0.75f;
 
         public const float ExamineRange = 16f;
-        protected const float ExamineDetailsRange = 3f;
+        protected const float ExamineDetailsRange = 8f;
 
         protected const float ExamineBlurrinessMult = 2.5f;
 


dffdff2423:
I think you should PR this upstream to wizden. This would be a nice feature for salamander.

Rainbeon:
> I think you should PR this upstream to wizden. This would be a nice feature for salamander.

I have said in this PR description why the range is so low in upstream (it is exploitable in their gameplay loop) so I am not interested in upstreaming this at this time. If you wish you are free to make an attempt there yourself, but I do not believe it would be accepted there personally.

dffdff2423:
The reason why I mention it is because looking at the git blame that number was picked over 4 years ago and seems to be untouched. I don't think it is as exploitable as you say it is because many people leave their descriptions blank. I might make a PR myself if you are not interested- it really sucks to not be able to read people's flavor text.

DrSmugleaf (approved):
Thanks for the PR!

PR #3628, opened 2024-08-12
Bug fix - the system now announces the death of the queen if it falls into a chasm

aesdfxcgvbh:
## About the PR

There is a bug: if the queen falls into a chasm the system does not announce it.
Now should work properly.

## Why / Balance

bug fix

## Technical details

This is so because the SharedXenoAnnounce system is subscribed to the MobStateChangedEvent but falling into a chasm does not raise it. So I created an event and an event handler for that.
Updated: Now the ChangeMobState changes MobState of the falling entity to Dead right before the entity gets deleted (if it has the MobState component).

## Media

## Breaking changes

the death announce system now has one less bug

**Changelog**

- fix: Fixed SharedXenoAnnounceSystem not working properly when the queen falls into a chasm.


aesdfxcgvbh:
forgot one using

deltanedas:
this will make the announcement if a dead queen gets pitted, it should instead try to set mobstate to dead before deleting which will let it handle it properly

aesdfxcgvbh:
> this will make the announcement if a dead queen gets pitted, it should instead try to set mobstate to dead before deleting which will let it handle it properly

Ok, I'll do it soon.

aesdfxcgvbh:
I think the StartFalling system should check the mobstate of a falling entity instead.

deltanedas:
> > this will make the announcement if a dead queen gets pitted, it should instead try to set mobstate to dead before deleting which will let it handle it properly
> 
> Won't it make more bugs? If somebody falls into a pit or "gets pitted", then they won't be able to talk, to stand, etc.
> Probably it's better to raise MobStateChangedEvent.
> edit: but it will still make the announcement if a dead queen gets pitted. 

i mean do it right before deleting the mob, not just when they start falling

aesdfxcgvbh:
> > > this will make the announcement if a dead queen gets pitted, it should instead try to set mobstate to dead before deleting which will let it handle it properly
> > 
> > 
> > Won't it make more bugs? If somebody falls into a pit or "gets pitted", then they won't be able to talk, to stand, etc.
> > Probably it's better to raise MobStateChangedEvent.
> > edit: but it will still make the announcement if a dead queen gets pitted.
> 
> i mean do it right before deleting the mob, not just when they start falling

Sounds good, I'll try it.

aesdfxcgvbh:
> > > this will make the announcement if a dead queen gets pitted, it should instead try to set mobstate to dead before deleting which will let it handle it properly
> > 
> > 
> > Won't it make more bugs? If somebody falls into a pit or "gets pitted", then they won't be able to talk, to stand, etc.
> > Probably it's better to raise MobStateChangedEvent.
> > edit: but it will still make the announcement if a dead queen gets pitted.
> 
> i mean do it right before deleting the mob, not just when they start falling

Now it should be good.

DrSmugleaf (approved):
Thanks for the PR!

PR #3633, opened 2024-08-12
Make base chair abstract

Tunguso4ka:
## About the PR

- CMSeatBase is now abstract.
- Added migration CMSeatBase -> CMChair

CMSeatBase never was intended to be spawned, it have no construction steps, cant be unanchored.


DrSmugleaf (approved):
Thanks for the PR!

PR #3637, opened 2024-08-12
G&M 28 Breacher

FungiFellow:
## About the PR

I created a heavy duty melee weapon for the purposes of busting through Resin faster than lighter alternatives. Sprites were made by Aleksh
## Why / Balance

Is available to SL, FTL, and CT at 12 points and Rifleman at 15 (This is still a draft and is still being balanced)

Pros
- High Damage
- Magnetic 
- Reusable
- Some AP

Cons
- Is a Melee Weapon
- Ginourmous
- Requires wielding to use
- Movement Speed Penalty
- Slower Swing Speed
- Unable to be used while Devoured

## Media

image
image
image

## Breaking changes

**Changelog**

:cl:
- add: Added the G&M 28 Breacher, a melee weapon specialized for breaking terrain.


Carolyn3114:
250 slash?

FungiFellow:
> 250 slash?

Yes. The current sprites are moreso a woodsplitting maul than big hammer type of maul

Siigiil:
Did you actually run this by smug to see if it would get accepted? We're currently porting things rather than adding our own content.

And this thing having 250 slash is obscenely overpowered. It'll oneshot any human you hit with it and severely fuck up most xenos. Lower the damage and use `DamageMultipliersComponent` to make it do more damage to objects it's meant to breach. See the U7 underbarrel shotgun for the flags used for objects generally affected by breaching weapons.

FungiFellow:
> And this thing having 250 slash is obscenely overpowered. It'll oneshot any human you hit with it and severely fuck up most xenos. Lower the damage and use `DamageMultipliersComponent` to make it do more damage to objects it's meant to breach. 

Will do.

KittenColony:
If your not already, get an app like Github Desktop and clone the repo to your own device, as it seems like you're doing this all on the browser. This lets you do multiple changes in a single commit instead of having 20 'commits' of deleting single assets. It also lets you use tools like Notepad++ or VSC.

Otherwise, looks good visually, 90 damage is still fucking _insane_ though. I do get why it would be like that though.

KittenColony:
I've also spoken with Smug about this, we won't be adding any new custom items to RMC-14 until everything from CM13 is added and in a functional state.

Some more issues with this though, its too similar to just being a better machete. The cost is too low all around to make it worth taking the machete if you want to use it for their shared role, for riflemen it should either be 25+ or not included. I'd also recommend reducing the speed further for both moving and wielding, or potentially making it swing even slower.

Siigiil:
There also already is a breaching hammer we will eventually port. Synths can get it.

FungiFellow:
> I've also spoken with Smug about this, we won't be adding any new custom items to RMC-14 until everything from CM13 is added and in a functional state.
> 
> Some more issues with this though, its too similar to just being a better machete. The cost is too low all around to make it worth taking the machete if you want to use it for their shared role, for riflemen it should either be 25+ or not included. I'd also recommend reducing the speed further for both moving and wielding, or potentially making it swing even slower.

Entirely fair, though this item was still being balanced.

PR #3639, opened 2024-08-13
Add extra damage from cripple strike + visual effect

Vermidia:
## About the PR

Adds the extra 120% multiplier to damage a lurker's cripple strike. Also adds the slow down arrow effect (first seen with crusher) to it.

## Why / Balance

CM-13 parity

## Technical details

Add damage to both activeCripple and Cripple comps for lurker, with the latter set to 7 in the prototype for em (because Cripple strike modifies the hit damage by 120%, but doing that would mean everyone hit would take more damage, instead of just the person hit with cripple strike)

Also renames the stomp effects yml to xeno_slow, and parents off the stomp effect for the cripple effect, just changing the lifetime.

## Media

https://github.com/user-attachments/assets/a170c5bc-8f0b-4875-bee8-8864337c22e7

## Breaking changes

no
**Changelog**

:cl:
- fix: Fixed the Lurker's Crippling Strike not doing 20% extra damage, and added a visual effect on the person hit.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Crippling/XenoCripplingStrikeSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Crippling/XenoCripplingStrikeSystem.cs
@@ -1,4 +1,5 @@
-﻿using Content.Shared._RMC14.Xenonids.Plasma;
+﻿using Content.Shared._RMC14.Armor;
+using Content.Shared._RMC14.Xenonids.Plasma;
 using Content.Shared.Coordinates;
 using Content.Shared.Damage;
 using Content.Shared.Movement.Systems;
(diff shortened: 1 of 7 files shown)


Content.Shared/_RMC14/Xenonids/Crippling/XenoCripplingStrikeSystem.cs
```diff
@@ -61,13 +65,18 @@ private void OnXenoCripplingStrikeHit(Entity<XenoActiveCripplingStrikeComponent>
             victim.ExpireAt = _timing.CurTime + xeno.Comp.SlowDuration;
             victim.SpeedMultiplier = xeno.Comp.SpeedMultiplier;
 
+            _damageable.TryChangeDamage(entity, xeno.Comp.Damage);
```
DrSmugleaf, line 68: Won't armor doubly affect this damage? It will apply separately to the 35 and 7, which might bring it below the threshold under which armor resists more of the damage.
You can use GetMeleeDamageEvent instead to modify the damage of the Lurker's attack, though you might need to check that you only modify it for one target.

DrSmugleaf (changes requested):
Thanks for the PR!
I changed the changelog to indicate that this was an oversight, not something we decided to buff the Lurker with afterwards.
If its missing content from 13 I usually word it as a fix for clarity.

Vermidia:
GetMeleeDamage didn't specify allow me to specify the target so I just used damage modify instead and made the extra damage trigger only once (I don't think any other attack would register before the lurker one, I'd hope).

DrSmugleaf (approved):
Thanks for the PR!

PR #3641, opened 2024-08-13
Crusher Melee AOE

### LINKED ISSUE #3413: Crusher AOE slash
## Description

Crusher's normal melee attack should reach out to this range
and slow the first person hit for 2 seconds (or the person directly in front, either way it slows a single person)
Slow uses the down arrow animation from stomp on that person

**Reproduction**

**Screenshots**

image
The range

**Additional context**

Vermidia:
## About the PR

Crusher's "Brutalizer" ability (as the wiki calls it) has been implemented. The first living target hit in an attack will also have up to 8 living targets be damaged (at 40% of the damage the main target took) by this effect. The crusher will also have her cooldowns reduced depending on how many targets this effect hit (at a minimum of 1.5 secs with 0 brutalize hits, with 0.5 per brutalize target. Defensive Shield only regens by 1.5 per any hit)

## Why / Balance

CM-13 parity, fixes #3413

## Technical details

New system for this ability. It checks in a range after getting the first valid living hit for other valid entities and damages them with 40% of the damage gotten from MeleeHitEvent. As long as a main target is found, the crusher will get -1.5 seconds to her charge and defensive shield abilities. The additional hits give -0.5 each, but only to charge (doing it for defensive shield gave it unlimited uptime in CM-13 so they don't)

It basically hits in a circle around the first target. It does not count any entities that would've been hit in a wideswing as valid targets. (To prevent additional damage from wideswings).

## Media

https://github.com/user-attachments/assets/a5e23609-6482-4b06-b6ce-f7d786ff562d

## Breaking changes

no

**Changelog**

:cl:
- add: Added the Brutalizer passive to the Crusher, which gives her an AOE effect that hits every nearby the first person hit at 40% damage. Attacks now reduce the Crusher's Defensive Shield and Charge abilities' cooldowns, with additional hits reducing Charge's even more.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Brutalize/XenoBrutalizeComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Brutalize/XenoBrutalizeComponent.cs
@@ -1,0 +1,34 @@
+﻿using Content.Shared.Actions;
+using Robust.Shared.GameStates;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared._RMC14.Xenonids.Brutalize;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(XenoBrutalizeSystem))]
+public sealed partial class XenoBrutalizeComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public int MaxTargets = 8;
+
+    [DataField, AutoNetworkedField]
+    public float AOEDamageMult = 0.4f;
+
+    [DataField, AutoNetworkedField]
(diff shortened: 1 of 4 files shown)


Dutch-VanDerLinde (commented):
changelog broken

DrSmugleaf (approved):
Thanks for the PR!
I need to check how we want to adjust regular wide swings ideally before I merge this PR
I think they already split damage between all targets but it should be that and ideally only being able to hit 1 thing at once, with proper prioritization so that hitting a wall doesn't eat your hit against someone

PR #3645, opened 2024-08-13
Glass sheet visual fix

Hyenh:
## About the PR

Adds stack states to glass, reinforced glass, phoron and reinforced phoron sheets.
## Why / Balance

It look weird before.
## Media

image

## Breaking changes

**Changelog**


Resources/Maps/_RMC14/solaris.yml
```diff
@@ -22074,7 +22074,7 @@ entities:
     - type: Transform
       pos: 52.5,-112.5
       parent: 1
-- proto: CMSheetGlass25
```
DrSmugleaf, line 22077: Is this change intentional?
You should be able to keep the 25 stack

DrSmugleaf (changes requested):

Resources/Maps/_RMC14/solaris.yml
Hyenh: The prototypes are 10 30 and 50. I changed it to be like the metal sheets but didn't think too much about if the old prototypes were used in maps

DrSmugleaf: 13 has some weird numbers sometimes, best to return the 25 stack and the one on Solaris

DrSmugleaf (approved):

PR #3649, opened 2024-08-14
Add boiler neurotoxin

### LINKED ISSUE #325: Boiler abilities
- [x] #364 
- [x] #365 
- [x] #366 
- [x] #367 
- [x] #368 
- [x] #407
- [x] #408
- [x] #393
- [x] #376 
- [x] #409

### LINKED ISSUE #408: Toggle gas type

Vermidia:
## About the PR

the title. See below for more details

## Why / Balance

Closes #408 and closes #325
CM-13 parityyy
## Technical details

Well, this is going to be a bit

So to start off - what it doesn't do:
Dizziness uses drunkness so it's not quite the same as cm-13s
No Eye Blurr
No Welder vision
No Hallucinations (someone else can have fun trying to get these in)
No Deafness or liver damage
No Agony effect and thus no fake damage (maybe when we get pain)

The stamina damage and all the other slew of effects are mostly there though, with most being status effects and stumbling using the throw system like fling and punch. Neurotoxin is injected via the NeurotoxinInjectorComponent, which has been added long with the neuro gas.

The BombardToggleGasEvent has been renamed to be more general, as acid shroud now also uses it and I added a system to make sure the action gets toggled correctly too.

Values are probably all over the place. Damage numbers and stuff have been kept the same as CM-13 even though they might be faster there (to match acid gas rn). So right now it takes 10 seconds in the gas to start instant dying.

The neurogas expands 1 tile farther out to match cm-13, though I think it's supposed to spread slightly slower (5 instead of 6 which is acid's spread speed). I couldn't find a way to do that so they spread at the same speed.
## Media

https://github.com/user-attachments/assets/b2579a4d-f57a-45bb-b7c2-14479e466fc4

(Not 100% accurate to the current timings)

## Breaking changes

**Changelog**

:cl:
- add: Added Boiler's neurotoxin gas, which is more oriented for capturing and causes a slew of chaotic effects if it builds up enough.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/NeurotoxinSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/NeurotoxinSystem.cs
@@ -1,0 +1,20 @@
+﻿using Content.Server.Chat.Systems;
+using Content.Shared._RMC14.Xenonids.Neurotoxin;
+
+namespace Content.Server._RMC14.Xenonids;
+
+public sealed class NeurotoxinSystem : SharedNeurotoxinSystem
+{
+    [Dependency] ChatSystem _chat = default!;
+    public override void Initialize()
+    {
+        base.Initialize();
+
+        SubscribeLocalEvent<NeurotoxinComponent, NeurotoxinEmoteEvent>(OnEmote);
+    }
+
+    public void OnEmote(Entity<NeurotoxinComponent> victim, ref NeurotoxinEmoteEvent args)
+    {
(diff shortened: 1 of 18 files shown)


DrSmugleaf (approved):
Thanks for the PR!
The effects look correct 
For gas spread speed, we are gonna need to wait for a future system that can do spread at different speeds properly
Need to apply it to weeds too since their spread is like 3x faster than intended

PR #3650, opened 2024-08-14
Adjusted some xeno pull speeds

Vermidia:
## About the PR

Adjusted the xeno pull speeds based on what we had and what was defined in CM-13.
Boiler now has the default pull speed of 0.75 (not sure where the 0.5 was from)
Warrior now pulls slightly slower (0.625, since the value in CM was 2 and for Queen is was 3)
Runner now pulls slightly faster (0.875, the value in CM was -0.5)

## Why / Balance

CM-13 parity

## Technical details

no
## Media

## Breaking changes

**Changelog**

:cl:
- fix: Fixed some xeno pull speeds. Runner pulls faster, boiler pulls the same as most xenos, and warrior pulls a little slower then they used to.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/boiler.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/boiler.yml
@@ -81,7 +81,5 @@
     radius: 1.8
     softness: 1
     autoRot: true
-  - type: SlowOnPull
-    multiplier: 0.5
   - type: RMCXenoDamageVisuals
     prefix: boiler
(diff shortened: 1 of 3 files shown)


DrSmugleaf (changes requested):
You might need to check this in-game in 13 locally, comparing the move delay doesn't always work and the actual speed when timed over 10 seconds is different

Vermidia:
Changed them again. Did tests in cm-13 comparing the darg speed of runner and warrior with the normal xeno pull speed and pull mult vs any changes their castes made and came out to about 0.86 speed for runner and 0.56 for warrior.

DrSmugleaf (approved):
Thanks for the PR!

PR #3661, opened 2024-08-14
Add Plasteel barricades

Vermidia:
## About the PR

Adds plasteel barricades. They have twice as much health as metal ones, but take long to set up (8 secs) and use 6 plasteel.

## Why / Balance

CM-13 pa-ri-ty

## Technical details

no

## Media

barricade

https://github.com/user-attachments/assets/74e353f8-a837-4c2c-8a76-688c3a28d0de

## Breaking changes

I changed the sprite states for the folding barricade so it's obvious which is which. This probably doesn't break anything?
**Changelog**

:cl:
- add: Added plasteel barricades, craftable with 6 plasteel. They have more health than normal barricades but take longer to set up.


DrSmugleaf (approved):
Thanks for the PR!

PR #3666, opened 2024-08-15
Add slow variations of building animations

### LINKED ISSUE #3472: Resin Building Animation for Lesser Drones is too fast
## Description
The new animation that plays with Resin Building does not currently account for the Lesser Drone's slower building speed, causing the animation to play and then disappear before completion of the structure

**Reproduction**
1) Build anything as a Lesser Drone
2) Structure will disappear ~60% of the way through, and then reappear on completion

Vermidia:
## About the PR

Made the queen and lesser drone building animation line up with their build speed

## Why / Balance

Fixes #3472, its mostly fine now

## Technical details

New visualizer system and event for triggering it. It tracks the expected frame of the animation and sets/unsets auto animate depending on it.

Had to cut the building anim's blank space as the animation is scaled based on total frames and would end up with the last bit being blanks.

## Media

https://github.com/user-attachments/assets/dd366015-0693-4ca1-8e1d-4f0ec98e3aa5

## Breaking changes

no?

**Changelog**

nah


Content.Shared/_RMC14/Xenonids/Construction/XenoConstructionComponent.cs
```diff
@@ -42,4 +42,7 @@ public sealed partial class XenoConstructionComponent : Component
     {
         Params = AudioParams.Default.WithVolume(-10f)
     };
+
+    [DataField, AutoNetworkedField]
+    public bool IsSlowAnimation = false;
```
Dutch-VanDerLinde, line 47: this is like the worst implematation ever, i was gonna make construction use prorotypes and have an animation field

hardcoding slow animations to 4 seconds is bad

Dutch-VanDerLinde (changes requested):

DrSmugleaf (changes requested):
This could be done by manually setting the animation time on the client in a visualizer system, and setting auto animate to false, I think we have one example of this in rmc-14 specific code I'll try to link it when I get back

Content.Shared/_RMC14/Xenonids/Construction/XenoConstructionComponent.cs
Vermidia: I did feel liek there was a better way... I'll look at what smug is talking about

Vermidia:
https://github.com/user-attachments/assets/76dd422e-53c2-4603-90ba-1da261a353d0

I got dynamic animations to semi-work, by splitting all the frames out so they're seperate RSIs. Weirdest problem is, if I don't divide the framecounter by 6 (???) it uh, is too slow? There's also now a problem of the effect not disappearing after the doafter quickly enough since I got rid of the fixed timed despawns. When I check this later I'll probably just make the doafter after delete it...

If anyone has any idea why the visualizer is working weirdly, please tell me. I did originally use the original files in their sprite sheet form, but changing the animation length just...did nothing >_>.

DrSmugleaf:
You don't need to split out the rsi states, look at what RequisitionsSystem does, you will need to set the animation frame depending on how much time is left to construct something

Vermidia:
I have done the thing but correctly this time! Works like a charm.

https://github.com/user-attachments/assets/0df26188-e2fa-4ee6-abc5-b73a6948eadd

DrSmugleaf (changes requested):
Are the changes to the delays and states intentional? Does it not work without those changes?

Vermidia:
Yes, it's intentional to it scales with a 1 second animation (the fastest it could go)
The cutting of the blank animation spaces in the sprites was for the same reason.

DrSmugleaf (approved):
Thanks for the PR!

PR #3669, opened 2024-08-15
Fix hypo not being inject only

Vermidia:
## About the PR

Title. It's not supposed to be able to draw from containers.

## Why / Balance

CM-13 parity

## Technical details

no

## Media

## Breaking changes

**Changelog**

:cl:
- fix: Fixed hyposprays being able to draw from containers. Use a bottle on it to refill it.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
@@ -16,6 +16,9 @@
     fillBaseName: hypospray
     emptySpriteName: hypospray
     changeColor: true
+  - type: Hypospray
+    onlyAffectsMobs: false
+    injectOnly: true
   - type: Item
     size: Small
     sprite: _RMC14/Objects/Medical/hypospray.rsi
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
@@ -21,7 +21,6 @@
     sprite: _RMC14/Objects/Medical/hypospray.rsi
   - type: Hypospray
     transferAmount: 5
-    onlyAffectsMobs: false
     injectOnly: true
     injectSound: /Audio/_RMC14/Medical/hypospray.ogg
   - type: Appearance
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
```diff
@@ -21,6 +21,8 @@
     sprite: _RMC14/Objects/Medical/hypospray.rsi
   - type: Hypospray
     transferAmount: 5
+    onlyAffectsMobs: false
+    injectOnly: true
```
DummeHytteOst, line 24: `    onlyAffectsMobs: false
    injectOnly: true`
won't this make it so you can't refill the hypospray at all, while still allowing it to inject into containers?

the hypospray currently can't 'draw' from a container, but can be filled by a container, since it doesn't have the 'InjectorToggleMode', which is what allows something like syringes, droppers and such, to switch between draw and inject. 

if you want it to just stop being able to inject into containers then remove line 25 'injectoonly: true' and change line 24 'onlyaffectsmobs: false' to 'onlyaffectmobs: true'

i could very well be wrong, since i  am still very new to coding, but that is how i had understod it

Zap527:
I know that CM13 parity is the end goal of all of this, but I always saw hyposprays being able to draw from containers as a QOL update over anything else. I feel like this is just making a doctor's job slightly less wieldy.

Resources/Prototypes/_RMC14/Entities/Objects/Medical/hypospray.yml
Vermidia: Injecting into containers is fine. It's the fact you can draw from them that isn't.

Vermidia:
Sorry, but this means bloodpacks are the best med storage if they can draw from them. Hypospray component inherantly uses drawable so no can do.

Vermidia:
Ok so it turns out it is only supposed to affect mobs but we don't have the vial system for it yet, so I'm just not making it able to draw still.

DrSmugleaf (changes requested):
Without the vial system at the moment this will make it unable to draw from anything right?
Should probably wait for the vial system if so

Vermidia:
You can use a bottle on it. Hypos are refillable.

DrSmugleaf (approved):
Thanks for the PR!

PR #3673, opened 2024-08-15
Port melee executions from upstream

Dutch-VanDerLinde:
- Ported https://github.com/space-wizards/space-station-14/pull/30104 from upstream

## About the PR
executions

## Why / Balance
Dramatic way to kill yourself, ported upstream

## Technical details
Compatible with the suicide changes, marines cannot kill themselves by using the suicide command

## Media

https://github.com/user-attachments/assets/db0743c8-a9c7-4356-a694-2032e9533a4d

## Breaking changes
None

🆑 Scribbles0, Plykiya, nikthechampiongr

- add: You can now execute incapacitated/cuffed entities or yourself with certain sharp weapons. Executions require a do-after, deal 9x the damage of a normal attack, and ignore damage resistances.


Vermidia:
We merged upstream, I don't think this is needed now?

QuietlyWhisper:
I talked with smug about this and we don't want to add it for the headaches it can probably cause staff, for the only benefit of having a niche way to kill yourself in niche situations, basic suicide was also disabled for the same reasons.

Also because upstream merge.

PR #3684, opened 2024-08-16
Add invulnerability time system + apply it to barricades

Vermidia:
## About the PR

Title. It makes hits not connect if it receives a certain amount of hits in a certain time.

## Why / Balance

Make xeno deathball less deathball-y because 20 xenos attack a single barricade is...not great

## Technical details

New system! Checks AttackedEvent and GettingAttackAttemptEvent to cancel and count attacks. Update loop is used to check the queue and dequeue expired attacks.

Current values is 3 hits per 1 second. It can be adjusted and added to other things, but for now it's only on barricades and only checking if xenos attack em.

## Media

https://github.com/user-attachments/assets/5fa11423-820c-4fab-b5ce-00924a6729ef

## Breaking changes

no
**Changelog**

:cl:
- fix: Barricades now have a short invulnerability time if they take too many hits from xenos in a small period. Same for resin and marines.

### DIFF
--- a/Content.Shared/_RMC14/Barricade/InvulnerabilityTimeSystem.cs
+++ b/Content.Shared/_RMC14/Barricade/InvulnerabilityTimeSystem.cs
@@ -24,17 +24,9 @@ public sealed class InvulnerabilityTimeSystem : EntitySystem
 
     private void OnAttack(Entity<InvulnerabilityTimeComponent> ent, ref AttackedEvent args)
     {
-//        if (_net.IsClient)
-//           return;
-
         if (_whitelist.IsWhitelistPass(ent.Comp.Whitelist, args.User) && _whitelist.IsBlacklistFailOrNull(ent.Comp.Blacklist, args.User))
         {
-            if (ent.Comp.CurrentHits == 0)
-            {
-                ent.Comp.WearOffTime = _timing.CurTime + ent.Comp.InvulnerabilityTime;
-            }
-
-            ent.Comp.CurrentHits++;
+            ent.Comp.AttacksToExpire.Enqueue(_timing.CurTime + ent.Comp.InvulnerabilityTime);
             Dirty(ent, ent.Comp);
         }
     }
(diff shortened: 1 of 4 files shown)


Content.Shared/_RMC14/Barricade/InvulnerabilityTimeSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+using Robust.Shared.Timing;
+using Robust.Shared.Network;
+using Robust.Shared.Player;
+
+namespace Content.Shared._RMC14.Barricade;
+
+public sealed class InvulnerabilityTimeSystem : EntitySystem
+{
+    [Dependency] EntityWhitelistSystem _whitelist = default!;
+    [Dependency] IGameTiming _timing = default!;
```
DrSmugleaf, line 14:
```suggestion
    [Dependency] private readonly EntityWhitelistSystem _whitelist = default!;
    [Dependency] private readonly IGameTiming _timing = default!;
```

Content.Shared/_RMC14/Barricade/InvulnerabilityTimeSystem.cs
```diff
@@ -0,0 +1,69 @@
...
+
+    public override void Update(float frameTime)
+    {
+        if (_net.IsClient)
+            return;
+
+        var time = _timing.CurTime;
+        var query = EntityQueryEnumerator<InvulnerabilityTimeComponent>();
+
+        while (query.MoveNext(out var uid, out var timeComp))
```
DrSmugleaf, line 55: You don't need to tick this in an update method, you can peek and dequeue on attack attempt instead, letting each attack check if the previous ones have expired their timer

Resources/Prototypes/_RMC14/Entities/Structures/Walls/barricade_metal.yml
```diff
@@ -41,6 +41,11 @@
             visible: false
   - type: DamageMultiplierFlags
     flags: Breaching
+  - type: InvulnerabilityTime
+    whitelist:
+      components:
+      - Xeno
+    hitsToInvulnerability: 3
```
DrSmugleaf, line 48: We should probably apply this to resin structures while we are at it as well

DrSmugleaf (changes requested):

Dutch-VanDerLinde (commented):
changelog is broken

DrSmugleaf (approved):
Thanks for the PR!

DrSmugleaf:
Shoutout to the two people thumbing down this PR as if their opinion matters, when being able to do this has been a bug from day 1

Carolyn3114:
WHAT DO YOU MEAN I CAN'T DELETE BARRICADES WITH MY ENTIRE XENO ARMY?

PR #3700, opened 2024-08-17
Tweak bursting

Vermidia:
## About the PR

There is now 60 seconds before a larva autobursts. It now also doesn't eat into the original burst time, and is instead added onto it.

It properly sets the stage to stage 6 when the larva spawns. The larva can now move to instantly burst instead of waiting out the timer.

## Why / Balance

CM-13 parity.

## Technical details

Evo system now checks for a burster component and doesn't give points if found. This burster component is also checked for movement to burst the target, and is removed upon bursting.

InputMoveEvent on the burster component will burst the victim instantly. Rotting will also make autobursting happen.

Some cleanup was done to remove redundancy. Some other notes:
There is a seperate countdown for a larva autobursting and when a player takes them. The first is about 60 seconds and the second is 20. Leaving it at 60 for now until that's implemented in (probably with queue).

Also prevent evolving while in a container, for obvious reasons.

## Media

https://github.com/user-attachments/assets/e18f5d8c-0fa8-44ad-ae3e-68126d5fefe5

## Breaking changes

**Changelog**

:cl:
- fix: Larva auto-burst time has been upped to 60 (from 20) seconds
- fix: Larva can now move to burst immediately from their host

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Evolution/XenoEvolutionSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Evolution/XenoEvolutionSystem.cs
@@ -3,7 +3,6 @@ using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.Xenonids.Announce;
 using Content.Shared._RMC14.Xenonids.Egg;
 using Content.Shared._RMC14.Xenonids.Hive;
-using Content.Shared._RMC14.Xenonids.Parasite;
 using Content.Shared.Actions;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Climbing.Components;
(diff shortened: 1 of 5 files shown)


Dutch-VanDerLinde (approved):

Vermidia:
Removed the evo point blocker - tested in cm-13 and they seem to gain points fine while in a victim.

Dutch-VanDerLinde:
> Removed the evo point blocker - tested in cm-13 and they seem to gain points fine while in a victim.

can they evolve while in a victim

Vermidia:
Ahh, yeah they can't. They need to be on a turf. I believe being in a container counters here.

DrSmugleaf (approved):
Thanks for the PR!

PR #3711, opened 2024-08-18
Savannah Update: remove booze dispenser, vendors

Tunguso4ka:
## About the PR

- Adds CO, XO, SO, Pilots vendors
- Adds We-Ya portable vendor to CL room
- Replaces Booze Dispenser with Soda Fountain
- Adds Booze-O-Mat to CIC
- Added dropship fabricator


DrSmugleaf (approved):
Thanks for the PR!

DrSmugleaf (changes requested):

PR #3740, opened 2024-08-19
Infiltrator Boots

FungiFellow:
## About the PR

Made a pair of silent boots (sneakers) for WS Scout.
## Why / Balance

Scouts are outed by their footsteps being hella loud. These shoes mute the wearers footsteps. CM13 lacks footstep noises so it could be considered parity?
## Media

## Breaking changes

**Changelog**

:cl:
- add: Infiltrator Boots are now part of WS Scout Kit


FungiFellow:
> Some people have said 13 doesn't have footstep sounds, if that is the case I'd just make this mute footsteps entirely imho

If I get that confirmed, I'll make it mute the noise.

DrSmugleaf (approved):
Thanks for the PR!

PR #3743, opened 2024-08-19
Small Scout Update

LankLTE:
## About the PR

Lower the opacity on scout by 0.01, makes cloaking hide external markings (ie. wings, antennae, frills, tails), and gives the A19 magazine a unique sprite. Stealth may need to be adjusted later but for now I think making it less visible is a good idea. 
Also makes spit not ever decloak because apparently it doesn't in 13.

## Why / Balance

First 2 changes were with the intention of better stealth. Scout stands out like a sore thumb even while invisible and being a nonhuman species with bright coloration made it even worse. You can already hide your entire body with a coif + armor but the markings were unremovable, this should make every species on the same level as humans stealth-wise. 
A19 magazine is 13 parity, I didn't realize it had a sprite. 

## Media

image
image

## Breaking changes

**Changelog**

:cl: Lank
- tweak: Scout now hides external markings (wings, frills, tails, etc.) while cloaked, and is slightly less visible overall.
- tweak: Scout's A19 magazines now have a unique sprite. 
- fix: Scout is now correctly not decloaked by acid spit.


LankLTE:
I don't think this is particularly going to solve any of the issues that makes scout not great (at least in my experience) but it should help significantly. #3740 is also probably a good idea but I'm not sure how 13 handled that.

Dutch-VanDerLinde:
> I don't think this is particularly going to solve any of the issues that makes scout not great (at least in my experience) but it should help significantly. #3740 is also probably a good idea but I'm not sure how 13 handled that.

13 just doesnt have footstep sounds pretty sure

Verification31:
Question, does this remove hairstyles as well? The floorlength bedhead I use is pretty large, so I think it might affect my scout gameplay.

LankLTE:
> Question, does this remove hairstyles as well? The floorlength bedhead I use is pretty large, so I think it might affect my scout gameplay.

Yes, though you can also just turn on tucking on your helmet to disable that currently

UnicornOnLSD:
just to make sure, the picture with invisibility on has upped gamma to actually demonstrate the changes, because if not they're just more visible

LankLTE:
> just to make sure, the picture with invisibility on has upped gamma to actually demonstrate the changes, because if not they're just more visible

Yes, the visibility in the example was upped from 0.03 to 0.5 so that you can actually see the changes well.

DrSmugleaf (approved):
Thanks for the PR!

PR #3745, opened 2024-08-19
Made Penlights able to be worn on the Head, Mask or Neck

FungiFellow:
## Why / Balance

Convienience/QoL, in 13 you could place it on your second ear slot.
## Technical details

Basic .yml edit
## Media

## Breaking changes

**Changelog**

:cl:
- add: Penlights can be worn in the Head, Mask or Neck slot


DrSmugleaf (changes requested):
It would be better to add a second ear slot like 13 I think, I don't think it'd break anything

PR #3751, opened 2024-08-20
update roletime requirements for high experience requiring roles

QuietlyWhisper:
# About the PR
- Increases requirements in high level roles, notably whitelisted roles and command/command-adjacent
- Should help lower the amount of people playing 1-2 rounds and then being allowed to play higher requirement roles easily
- Ensures players have at minimum played the roles they will be leading in some cases
- For reference, the average round is 2-3 hours.
- current whitelisted players are not affected.
## CO
image
## SEA
image
## SO
image
## Comtech
image
## FTL
image
## SGO
image
## SL
image
## WS
image
## CMO
image
## CMP
image
## MT
2 hours squad roles, as to not prank new players into getting lost as their first experience.
## MP
20 hours squad roles, due to their high requirement and strong equipment.

**Changelog**

Make sure to take this Changelog template out of the comment block in order for it to show up.
:cl: Whisper

- tweak: Command, MP and some squad roles roles now require more experience to play.


LankLTE:
I figure I'll at least throw my hat in the ring and say that as someone who was personally planning on applying for SEA, 30 hours of MP requirement seems super overkill and I don't really get the point? It doesn't seem super relevant to teaching and I already know several people who are turned away from applying for SEA because of the existing MP requirement. 
Ditto for 50 hours of MP for CO, and command since those are often contested roles, but I guess CO is supposed to be well-established players so requiring 200 hours as specific roles is more acceptable.

Vermidia:
Small thing: Getting WS and SG even once is... kinda hard with how many people want it. I'm not sure that should be in SL requirements.

Ariphylor:
I'm going to share a bit of input as someone who's already applied for SEA.

30 hours of MP is overkill. You're not getting any extra experience from playing an extra however many rounds of MP. SEA has vast roletimers because they should know the in and outs of each department. So.. engineering, CMSquad, Command, all of those are reasonably high.

What extra experience is gained from 30 hours of MP? How to better harass rifleman? 

All I'm saying is you get all the core experience you need from MP in like, three or four rounds.

FungiFellow:
This feels like far too much for every changed role considering the fact several require a whitelist already and then there's the chance you don't even get the role due to 4 other guys minimum going for it

lanse12:
I feel like WS and SG should stay the same

P1ksqu:
> Small thing: Getting WS and SG even once is... kinda hard with how many people want it. I'm not sure that should be in SL requirements.

I second this, WS and SG are two of the rarest roles to get, due to how few aviable spots there is in a round (4 each, with no way of getting more)
It would simply make it extremely painfull to try to unlock anything above SL due to how many peoples are trying to get these two specific Jobs.

deltanedas:
some of these should also need xeno playtime imo, know your enemy

LankLTE:
I know you voiced on the (now closed?) discord thread that you have no intention of changing this, but the CO requirements still seem very unreasonable. 50 hours of MP was already a lot, and you then moved 20 of it to two roles there are only one of per round, both of which are contested and need to be rolled for. You say this is so players have “at minimum played the roles they will be leading in some cases” but this isn’t a minimum, this is expecting 10-20 rounds each as some of the hardest roles to get in the game. 
People shouldn’t have to play for months to a year to get a role unlocked *on top* of applying for a whitelist- I know you said this is what you wanted but I frankly disagree.

Rainbeon:
I dislike the WS requirement on SL, it is extremely hard to roll WS and will only continue to get harder as it gets more cool toys, and I don't think experiencing a round as one does all too much to help SL players, moreso just arbitrarily lock good SL players behind the set-WS-high-and-pray rat race, it makes WS-high rolling mandatory which *also* cuts into the chances of the many many people that... actually want to play WS

To me it just makes the 2 problems of immense WS queue clutter and general lack of SL players even worse for zero gain

dffdff2423:
I assume that the goal of this PR is to increase the quality of players in leadership roles. I do not think this PR will achieve that goal. From my experience the role timer is meaningless for player quality. Instead, I think this will have the opposite effect because most high quality players have lives outside of the game. In addition, I don't think increasing the time for whitelisted roles is a good approach. The whitelist is at the will of the admins and they should exercise their digression and not whitelist LRP players and remove whitelist from people who cannot meet our RP expectations. With a few exceptions (most notably SO) I thought that the role timers were in a good state before this proposal. 

I think a better solution could be either: all game bans are accompanied by some kind of leadership role-ban, or a "salamander style" whitelist for leadership roles that is automatically accepted if you meet certain requirements. However, I think that any of these changes are not really needed as the overwhelming majority of players in leadership positions are fine right now and the bad eggs could easily be dealt with IC if we are given proper tools (mainly the Marine Law changes PR).

Rainbeon:
Isn't the CMCommand requirement for CO superfluous, as you require 30 hours of XO anyway? By fulfilling the 30 hours XO requirement you also fulfill the 20 hours Command requirement automatically, as 30 > 20

QuietlyWhisper:
> Isn't the CMCommand requirement for CO superfluous, as you require 30 hours of XO anyway? By fulfilling the 30 hours XO requirement you also fulfill the 20 hours Command requirement automatically, as 30 > 20

Technically it is yes, it used to be 20 hours of SO

BramvanZijp:
I still think that the warden playtime is an excessive requirement. Like its not a role that has a ton of overlap with CO in a way CMP does not, so at that point its just bloat and an un-neccesary extra grind. And as mentioned, CMP will need 5 hours anyway.

DrSmugleaf (approved):

PR #3772, opened 2024-08-21
Makes corrosive acid eventually dissipate on barricades

Vermidia:
## About the PR

Title.

## Why / Balance

CM-13 parity. Eventually the acid just disappears

## Technical details

Added a Cvar (set at 45) that controls when acid will expire off barricades, which is set in the server system and checked by the shared system. 45 is how long it lasts in cm-13 (9 ticks times 5 seconds).

## Media

## Breaking changes

no?
**Changelog**

:cl:
- fix: Fixed corrosive ccid on barricades not expiring after 45 seconds.


DrSmugleaf (approved):
Thanks for the PR!

PR #3790, opened 2024-08-22
fix(CMFax): Make the Fax use the new default paper.

CaasGit:
## About the PR
The upstream FaxMachine component has options to set the paper entity now thus allowing the default paper to be CMPaper when you Open File to print something from a fax machine.

**Changelog**

:cl:
- fix: Fax Machines now print local files using the CMPaper sprite.


DrSmugleaf (approved):
Thanks for the PR!

PR #3791, opened 2024-08-22
Add SEA Equipment and Gun racks

KalimbaMachine:
## About the PR

Adds the SEA vendors with minor additions that deviate from parity: 
 *Adds MP Security Glasses

Removed Alpha Headset from SEATech Vendor
(Removed the Officer's Jacket)

## Media

Updated pictures showing limited quantity:
seaequip11
seaequip12
seaweapon11

## Requirements

**Changelog**

:cl:
- add: Added SEA Equipment and Gun Racks!


QuietlyWhisper (changes requested):
These are the wrong types of vendors, this pr will not be merged half finished, but thank you for doing the fills.

Also please remove the officer jacket as it's not part of their uniform for the sake of recognizability.

Sec glasses can stay.

KalimbaMachine:
> These are the wrong types of vendors, this pr will not be merged half finished, but thank you for doing the fills.
> 
> Also please remove the officer jacket as it's not part of their uniform for the sake of recognizability.
> 
> Sec glasses can stay.

do you want the kind that have limited selection? or what kind should i be using here?
How can i improve it.

QuietlyWhisper:
> do you want the kind that have limited selection? or what kind should i be using here? How can i improve it.

It should be the limited selection yes, all SEA players can access these vendors and get equipment from them, they work the same as the rifleman armor vendor when done correctly.
see `Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/rifleman.yml` for how to implement it correctly

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/senior_enlisted_advisor.yml
```diff
@@ -0,0 +1,106 @@
...
+  - type: CMAutomatedVendor
+    sections:
+    - name: Personal Primary
+      choices: { CMPrimary: 1 }
+      entries:
+      - id: WeaponRifleM54C
+    - name: Ammunition
+      entries:
+      - id: CMMagazineRifleM54C
+        amount: 10
```
DrSmugleaf, line 99: This costs points, it's not a fixed amount
```suggestion
        points: 9
```

DrSmugleaf (approved):
Thanks for the PR!

PR #3823, opened 2024-08-24
Fix crusher stomp damage range

### LINKED ISSUE #3793: Crusher Stomp Behavior Parity
## Description

Crusher's Stomp ability does not behave with parity. Opening this issue to list the differences and determine if this is intended, and if anything should be done.

crusher_powers.dm has the stomp ability:
- targeting mobs on the same tile as it (in SS13, this should only happen if they were either resting or knocked down by another stun) to deal 65 chest damage
- targeting mobs within the 2 tile range, it applies superslow (5x5)

Currently in RMC14 it instead deals that massive 65 damage to every target in a 5x5, as well as the slow.

A closer parity to how it's intended to function would be checking for prone mobs that are within the Crusher's collision when casting and applying the damage to them. In 13 this is unlikely to ever be more than 1 but the code doesn't restrict this so theoretically it could hit multiple, and that'd let the ability be more useful with RMC's pixel movement.
**Reproduction**

**Screenshots**

**Additional context**

Vermidia:
## Why / Balance

Fixes #3793. ~~I know the issue mentioned maybe only effecting those that are prone due to impossibility of them normally being on the crusher, but I'm treating it like I'm treating the crusher's melee attack and just using the range by itself, even if it's normally hard to hit.~~ It both checks the short distance and only damages if they are prone.

## Technical details

Added a shortrange and shortlookup pretty much, and check that for the damage range.

## Media

https://github.com/user-attachments/assets/94969a18-c952-46a7-85e9-8f1f4b5d8c74

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Crusher stomp damaging marines more than half a tile away from the crusher that aren't laying down

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
@@ -24,9 +24,9 @@ public sealed class XenoStompSystem : EntitySystem
     [Dependency] private readonly EntityLookupSystem _entityLookup = default!;
     [Dependency] private readonly SharedStunSystem _stun = default!;
     [Dependency] private readonly SharedAudioSystem _audio = default!;
-	[Dependency] private readonly StandingStateSystem _standing = default!;
+    [Dependency] private readonly StandingStateSystem _standing = default!;
 
-	public override void Initialize()
+    public override void Initialize()
     {
         base.Initialize();
 
@@ -34,7 +34,6 @@ public sealed class XenoStompSystem : EntitySystem
     }
 
     private readonly HashSet<Entity<MarineComponent>> _receivers = new();
-    private readonly HashSet<Entity<MarineComponent>> _shortReceivers = new();
 
     private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompActionEvent args)
     {
@@ -56,7 +55,6 @@ public sealed class XenoStompSystem : EntitySystem
             return;
 
         _receivers.Clear();
-        _shortReceivers.Clear();
         _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.Range, _receivers);
 
         if (_net.IsServer)
--- a/Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
@@ -31,6 +31,7 @@ public sealed class XenoStompSystem : EntitySystem
     }
 
     private readonly HashSet<Entity<MarineComponent>> _receivers = new();
+    private readonly HashSet<Entity<MarineComponent>> _shortReceivers = new();
 
     private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompActionEvent args)
     {
@@ -52,7 +53,9 @@ public sealed class XenoStompSystem : EntitySystem
             return;
 
         _receivers.Clear();
+        _shortReceivers.Clear();
         _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.Range, _receivers);
+        _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.ShortRange, _shortReceivers);
 
         if (_net.IsServer)
             _audio.PlayPvs(xeno.Comp.Sound, xeno);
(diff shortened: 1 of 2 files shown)


Rainbeon:
Ideally this gets merged at the same time as Crusher's AOE basic attack so the caste doesn't die balance wise in the time between

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -67,16 +70,16 @@ private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompAct
                 SpawnAttachedTo(xeno.Comp.Effect, receiver.Owner.ToCoordinates());
         }
 
-        foreach (var receiver in _receivers)
+        foreach (var shortReceiver in _shortReceivers)
```
DrSmugleaf, line 73: For this I think it would be better to check standing state is down, I read your reasoning but it's already more likely to be able to hit multiple downed mobs here because of pixel movement, I think this would make it too strong

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -52,7 +53,9 @@ private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompAct
             return;
 
         _receivers.Clear();
+        _shortReceivers.Clear();
         _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.Range, _receivers);
+        _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.ShortRange, _shortReceivers);
```
DrSmugleaf, line 58: You should be able to get the transform component when looping over receivers and then get the distance, instead of doing two lookups

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -22,15 +24,17 @@ public sealed class XenoStompSystem : EntitySystem
     [Dependency] private readonly EntityLookupSystem _entityLookup = default!;
     [Dependency] private readonly SharedStunSystem _stun = default!;
     [Dependency] private readonly SharedAudioSystem _audio = default!;
+	[Dependency] private readonly StandingStateSystem _standing = default!;
```
DrSmugleaf, line 27:
```suggestion
    [Dependency] private readonly StandingStateSystem _standing = default!;
```

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -22,15 +24,17 @@ public sealed class XenoStompSystem : EntitySystem
...
 
-    public override void Initialize()
+	public override void Initialize()
```
DrSmugleaf, line 29:
```suggestion
    public override void Initialize()
```

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -22,15 +24,17 @@ public sealed class XenoStompSystem : EntitySystem
...
     {
         base.Initialize();
 
         SubscribeLocalEvent<XenoStompComponent, XenoStompActionEvent>(OnXenoStompAction);
     }
 
     private readonly HashSet<Entity<MarineComponent>> _receivers = new();
+    private readonly HashSet<Entity<MarineComponent>> _shortReceivers = new();
```
DrSmugleaf, line 37: Don't need this anymore
```suggestion
```

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -52,6 +56,7 @@ private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompAct
             return;
 
         _receivers.Clear();
+        _shortReceivers.Clear();
```
DrSmugleaf, line 59: Same here
```suggestion
```

Content.Shared/_RMC14/Xenonids/Stomp/SharedXenoStompSystem.cs
```diff
@@ -65,22 +70,19 @@ private void OnXenoStompAction(Entity<XenoStompComponent> xeno, ref XenoStompAct
...
+                if (damage?.GetTotal() > FixedPoint2.Zero)
+                {
+                    var filter = Filter.Pvs(receiver, entityManager: EntityManager).RemoveWhereAttachedEntity(o => o == xeno.Owner);
+                    _colorFlash.RaiseEffect(Color.Red, new List<EntityUid> { receiver }, filter);
+                }
             }
         }
-
-        if (_net.IsServer)
-            SpawnAttachedTo(xeno.Comp.SelfEffect, xeno.Owner.ToCoordinates());
```
DrSmugleaf, line 84: I presume this is an intended change?

DrSmugleaf (approved):
Thanks for the PR!

PR #3838, opened 2024-08-25
Add action for using the whistle on the action bar

crystalHex:
## About the PR

Adds an action for using the whistle on the action bar

## Why / Balance

Mostly quality of life
Not sure if CM13 has an action for whistling, but the source code seems to indicate it does

## Technical details

Duplicated the Whistle component and system to avoid having to modify upstream files too much.
Re-parented the RMCWhistle to not inherit the upstream Whistle component.

Added EmitSoundOnActionComponent and SoundActionEvent to upstream files, plus a function in SharedEmitSoundSystem to play a sound once said event gets raised.

## Requirements

## Breaking changes

**Changelog**

:cl:
- tweak: Whistles can now be used from the action bar.

### DIFF
--- a/Content.Shared/Sound/Components/EmitSoundOnActionComponent.cs
+++ b/Content.Shared/Sound/Components/EmitSoundOnActionComponent.cs
@@ -1,0 +1,22 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared.Sound.Components;
+
+/// <summary>
+/// Simple sound emitter that emits sound on InstantAction
+/// </summary>
+[RegisterComponent]
+public sealed partial class EmitSoundOnActionComponent : BaseEmitSoundComponent
+{
+    /// <summary>
+    ///     Whether or not to mark an interaction as handled after playing the sound. Useful if this component is
+    ///     used to play sound for some other component with on-use functionality
+    /// </summary>
+    /// <remarks>
+    ///     If false, you should be confident that the interaction will also be handled by some other system, as
+    ///     otherwise this might enable sound spamming, as use-delays are only initiated if the interaction was
+    ///     handled.
+    /// </remarks>
+    [DataField("handle")]
+    public bool Handle = true;
+}
--- a/Content.Shared/Sound/Components/EmitSoundOnActionComponent.cs
+++ b/Content.Shared/Sound/Components/EmitSoundOnActionComponent.cs
@@ -17,6 +17,6 @@ public sealed partial class EmitSoundOnActionComponent : BaseEmitSoundComponent
     ///     otherwise this might enable sound spamming, as use-delays are only initiated if the interaction was
     ///     handled.
     /// </remarks>
-    [DataField("handle")]
+    [DataField]
     public bool Handle = true;
 }
--- a/Content.Shared/Sound/SharedEmitSoundSystem.cs
+++ b/Content.Shared/Sound/SharedEmitSoundSystem.cs
@@ -102,11 +102,11 @@ public abstract class SharedEmitSoundSystem : EntitySystem
             args.Handled = true;
     }
 
-    private void OnEmitSoundOnAction(EntityUid uid, EmitSoundOnActionComponent component, SoundActionEvent args)
+    private void OnEmitSoundOnAction(Entity<EmitSoundOnActionComponent> ent, ref SoundActionEvent args)
     {
-        TryEmitSound(uid, component, args.Performer);
+        TryEmitSound(ent, ent, args.Performer);
 
-        if (component.Handle)
+        if (ent.Comp.Handle)
             args.Handled = true;
     }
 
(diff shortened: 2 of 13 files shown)


DrSmugleaf (changes requested):
Thanks for the PR!
I need to look this over to see if we can reduce code duplication, we should be able to create a new event and component that just calls whistle system.

Content.Shared/Whistle/WhistleComponent.cs
```diff
@@ -9,7 +9,7 @@ namespace Content.Shared.Whistle;
 /// Spawn attached entity for entities in range with <see cref="HumanoidAppearanceComponent"/>.
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-public sealed partial class WhistleComponent : Component
```
DrSmugleaf, line 12: You shouldn't need to inherit this component, you can just get the whistle component when handling events for rmc whistle component

Content.Shared/Sound/SharedEmitSoundSystem.cs
```diff
@@ -99,6 +102,14 @@ private void OnEmitSoundOnUseInHand(EntityUid uid, EmitSoundOnUseComponent compo
             args.Handled = true;
     }
 
+    private void OnEmitSoundOnAction(EntityUid uid, EmitSoundOnActionComponent component, SoundActionEvent args)
+    {
+        TryEmitSound(uid, component, args.Performer);
+
+        if (component.Handle)
+            args.Handled = true;
+    }
```
DrSmugleaf, line 111: This format of subscribing to an event is preferred, it associates the comp with the entity and is always handled by ref
```suggestion
    private void OnEmitSoundOnAction(Entity<EmitSoundOnActionComponent> ent, ref SoundActionEvent args)
    {
        TryEmitSound(ent, ent, args.Performer);

        if (ent.Comp.Handle)
            args.Handled = true;
    }
```

Content.Shared/_RMC14/Whistle/RMCWhistleSystem.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        TryWhistle(uid, comp, args.Performer);
+        args.Handled = true;
+    }
+
+    public void OnUseInHand(EntityUid uid, RMCWhistleComponent comp, UseInHandEvent args)
+    {
+        TryWhistle(uid, comp, args.User);
+        args.Handled = true;
+    }
```
DrSmugleaf, line 48: Same here for these, they can be made into Entity<RMCWhistleComponent> ref handlers

Resources/Prototypes/_RMC14/Entities/Objects/Devices/whistle.yml
```diff
@@ -1,19 +1,25 @@
...
     sound:
       collection: TrenchWhistle
+  - type: EmitSoundOnAction
+    sound:
+      collection: TrenchWhistle
   - type: UseDelay
     delay: 10
-  - type: Whistle
+  - type: RMCWhistle
     distance: 15 # Let your enemies hear the sound of death approaching
```
DrSmugleaf, line 25: You don't need to modify upstream's whistle, we only use the downstream one with ID RMCWhistle

Content.Shared/Sound/Components/EmitSoundOnActionComponent.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// <summary>
+    ///     Whether or not to mark an interaction as handled after playing the sound. Useful if this component is
+    ///     used to play sound for some other component with on-use functionality
+    /// </summary>
+    /// <remarks>
+    ///     If false, you should be confident that the interaction will also be handled by some other system, as
+    ///     otherwise this might enable sound spamming, as use-delays are only initiated if the interaction was
+    ///     handled.
+    /// </remarks>
+    [DataField("handle")]
```
DrSmugleaf, line 20: If not specified the yaml tag will default to the camelcase version of the field's name
```suggestion
    [DataField]
```

DrSmugleaf (changes requested):
Any added code needs to be moved to RMC14 specific files and systems if possible, under the _RMC14 folder of Content.Shared

crystalHex:
Thank you for the review! I applied the changes and will keep them in mind for future PRs.

Content.Shared/Whistle/WhistleComponent.cs
crystalHex: Reverted the change of the component now

Content.Shared/_RMC14/Whistle/RMCWhistleSystem.cs
crystalHex: Modified functions to be entity ref handlers

Resources/Prototypes/_RMC14/Entities/Objects/Devices/whistle.yml
crystalHex: Added the base Whistle component back to the prototype, the RMCWhistle component is now solely responsible for adding the action and for synchronizing the delays between in-hand and action usage

crystalHex:
Moved all the RMC-specific code to the _RMC14 folders, did have to change the access level of TryEmitSound to be able to access it from outside the SharedEmitSoundSystem without code duplication.

DrSmugleaf (approved):
Thanks for the PR!

PR #3842, opened 2024-08-26
Add vials to hyposprays + medical stuff

Vermidia:
## About the PR

Adds vials, vial pouches, and vial-powered hypos. Also fixed a few thing (vial pouch and syringe not being full, and using the wrong protos)

## Why / Balance

CM-13 parity

## Technical details

So...RMCHypospray is a shared system with a client and server version, and a shared system. Client just handles item status, thats it. It mainly works through itemslots.

The shared system handles updating the appearance, which is mainly through taking parts of the updateAppearance from solutionContainer stuff and feeding it stuff from the inserted vial. Shared also handled the after interact (Server handles the doafter) and toggling parts.

Server also handles the interact using. Which is kinda the most eh part of the code, due to the fact it manually reimplemented part of transferring solutions (this mostly uses transfersolutionsystem so not too much) and the injectorsystem and spikablesystem, since without modifying those directly it won't get the solution of the vial. The doafter of the hypo is just...hyposprays one pretty much, with slight modifications. The doafter from tactical reloading is also here, it just inserts the vial as the shared system should have already ejected the previous vial into hand.

CMRefillableSolutionSystem had to be tweaked to work with the vials too, since they can, be refilled while in a hypo. Had to add an event so afterwards the Hypo system could correctly update the visuals.

There's a visualizer system just to set the hypo being filled or empty - I couldn't find a generic thing for this.

There's probably something I missed talking about here, since it's pretty big.

## Media

https://github.com/user-attachments/assets/c3d15be0-7cbb-4933-8055-e09ac8cd16ea

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Hyposprays now have vials, which can be tactically reloaded by using a hypo on one, or you can swap them in with a vial in hand.
- fix: You can now toggle a Hypospray's transfer amount.
- fix: Hyposprays can't inject into beakers
- fix: Hyposprays from We-Ya Meds now come with a vial of tricordizine. You can use the vial or the hypo with the vial inserted on a We-Ya Med to refill it.
- add: Vial pouches can now be selected by corpsman as a pouch option. Comes with 6 vials
- add: Droppers are now available in the We-Ya Med Plus
- fix: You can add lids to beakers, large beakers, and high capacity beakers


DrSmugleaf (approved):
Thanks for the PR!

PR #3919, opened 2024-09-02
fix(healthscanner): Add a basic component to allow for scanning.

CaasGit:
## About the PR
Instead of relying on the CMStasisBag component while checking to see if the scanner will look into a container, I've added a basic component to add on any container that should be looked in, this adds it on CMBodyBag which is the parent of CMStasisBag. This also allows for future additions like CMMorgue or other items that could contain someone who needs to be health scanned.

**Changelog**

:cl:
- fix: Allow for body bags to be health scanned.


DrSmugleaf (approved):
Thanks for the PR!

PR #3979, opened 2024-09-10
Fix rouny speed

Vermidia:
## About the PR

Title. Now this might seem a bit weird since I boosted it before, but hear me out:

Everytime I test rounies in cm their pull speed vs normal speed compared to like, every caste is MUCH much worse. I wanted to put it on move delay or tile delay being weird and just increase pull speed as seems intended, but that might've made them too good. So I adjusted the numbers so cm-13 and rmc match, and this amounts to rounies actually having a really bad pull speed compared to other xenos. It comes out to 30% when other xenos are 75%. There could be a bunch of reasons in cm why this turns out this way but that is the number I always get.

## Why / Balance

Parity and balance.

 If 30% is too much it can always be raised back to the xeno default of 75%.

## Requirements

https://github.com/user-attachments/assets/39fa8130-2d2f-48cd-938e-c1afdb6b4e4c

https://github.com/user-attachments/assets/12e6bee5-3b76-4acc-a896-88d4c0442e1c

Comparison to cm-13 with new speed

## Breaking changes

**Changelog**

:cl:
- fix: Fix Runner pull speed being much faster than it should be

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/runner.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/runner.yml
@@ -57,7 +57,7 @@
   - type: XenoDevour
   - type: XenoHide
   - type: SlowOnPull
-    multiplier: 0.86
+    multiplier: 0.3
   - type: XenoLeap
     delay: 0
     knockdownTime: 2


DrSmugleaf (approved):
Thanks for the PR!

PR #4013, opened 2024-09-11
You can now take over parasite eggs with a right click verb.

### LINKED ISSUE #3926: Players should be able to become parasites by clicking on an egg as a ghost
CM13 parity

There is a 3 minute cooldown when you ghost so you cant spam parasite spawns

Dutch-VanDerLinde:
CM13 parity

resolves #3926

https://github.com/user-attachments/assets/b657210d-ede3-46e5-9005-7e218918140c

:cl:
- add: You can now become a parasite by using a right click verb on an egg.


Content.Server/_RMC14/Xenonids/Egg/XenonidEggSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+        if (_net.IsClient)
+            return;
+
+        var user = args.Actor;
+
+        if (!TryComp(user, out GhostComponent? ghostComp))
+            return;
+
+        // Must have been dead for 3 minutes
+        if ((int)_gameTiming.RealTime.Subtract(ghostComp.TimeOfDeath).TotalSeconds < 180)
```
DrSmugleaf, line 72: RealTime is for UI and out of game simulation, you want CurTime
You should also be able to do this instead so you don't need to get the seconds manually and cast it to an int
```suggestion
        if (_gameTiming.RealTime.Subtract(ghostComp.TimeOfDeath) < TimeSpan.FromMinutes(3))
```

Content.Server/_RMC14/Xenonids/Egg/XenonidEggSystem.cs
```diff
@@ -0,0 +1,89 @@
...
+using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
+using Robust.Shared.Serialization;
+using Robust.Shared.Timing;
+using Content.Server.Ghost.Roles;
+using Content.Shared._RMC14.Admin;
+
+namespace Content.Server._RMC14.Xenonids.Egg;
+
+public sealed class XenonidEggSystem : EntitySystem
```
DrSmugleaf, line 41: Should probably call it XenoEggRoleSystem if you don't want to inherit the shared one and split it, just so there's not a xeno egg system and a xenonid egg system

Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
```diff
@@ -519,4 +526,44 @@ public override void Update(float frameTime)
...
+
+    private void OnGetVerbs(Entity<XenoEggComponent> ent, ref GetVerbsEvent<ActivationVerb> args)
+    {
+        var uid = args.User;
+
+        // if it doesn't have an actor and we can't reach it then don't add the verb
+        if (!EntityManager.TryGetComponent(uid, out ActorComponent? actor))
+            return;
+
+        if (!EntityManager.TryGetComponent(uid, out GhostComponent? ghostComp))
```
DrSmugleaf, line 538: Entity systems have a short-hand format for component methods
```suggestion
        if (!TryComp(uid, out ActorComponent? actor))
            return;

        if (!TryComp(uid, out GhostComponent? ghostComp))
```

DrSmugleaf (changes requested):
Thanks for the PR!

Dutch-VanDerLinde:
Addressed review

DrSmugleaf (approved):

PR #4018, opened 2024-09-12
Make the humanoid bullet hitbox slightly smaller

DrSmugleaf:
## About the PR

To be revisited with hostile factions maybe, but this basically makes it harder to shoot someone behind you too, see screenshot for the longest distance where that happens
Alternative was make guns shoot further in front of you, but that then affects combat vs melee units too

## Media

Content Client_U9RhWrSiX0

Bigger circle is normal collision, smaller circle is bullet collision
image

## Requirements

## Breaking changes

**Changelog**
:cl:
- tweak: Made the humanoid bullet hitbox slightly smaller (0.35 > 0.25), making humanoids slightly harder to hit with bullets, specially when one is far behind you. You will still shoot someone if they are too close behind you, to prevent stacking. This change doesn't affect xenonid projectiles.


Kadeo64:
anti-ff PR real

Rainbeon:
Does this affect Spitters and Sentinels as well?

DrSmugleaf:
> Does this affect Spitters and Sentinels as well?

In retrospect it does, need to think about what to do about that because I think they are treated like regular projectiles in 13 too
At worst we could make xeno projectiles collision different

Kadeo64:
> At worst we could make xeno projectiles collision use different groups

that seems like the most obvious thing to do unless you want the spit projectile's hitbox to be larger (and thus get clipped on geometry more often) although idk if that will affect performance at all

PR #4050, opened 2024-09-14
Revise most Xeno action descriptions

Rainbeon:
## About the PR
Most Xeno action descriptions either don't give the player much information, or just flat out lie to the player. This PR fixes these cases and also makes some of the wording more consistent.

## Why / Balance
**Drone/Hivelord**
- Fixed Give Plasma's description, prior it didn't mention xenonids or use punctuation.

**Defender**
- Charge: Formerly implied that it was both tile targeted and able to hit multiple enemies, wording updated to describe its player-targeted and single-target nature. 
- Toggle Crest Defense: Formerly stated it only reduced projectile damage, but it also decreases melee damage as well. Given the two terms are directly opposed to each other it wasn't intuitive. Things that Crest actually doesn't reduce are indirect damage sources, such as barbed wire on cades.
- Fortify: Same as above, but now mentions you become impassible, this is a rmc14-unique mechanic due to use not having bodyblocking, so wasn't present in the original 13 description. Also mentions now directional damage reduction.

**Warrior**
- Overall de-specified marines, as it was the only Xeno to do this. Updated each ability with its actual effects as many did not tell the full effects. Made Lunge more clear that it can be intercepted.

**Crusher**
- Charge/Shield: Now mentions its CD refunding properties from its passive, important to know as there is no way to currently view these less noticeable passive effects in-game.
- Stomp: Mentions its disarming properties, as well as damaging enemies you stand on.

**Sentinel**
- Clarified that the projectiles do not deal damage, as it is one of the most common misconceptions across all Xeno castes, due to targets glowing red when you hit them still, and not having any way to actually tell what a Marine's HP is.

**Spitter**
- Clarifies the effects of Charge Spit by mentioning its DoT effect
- Clarifies the effects of Spray Acid on cades and enemies afflicted by Charge Spit DoT
- Mentions Charged Spit's movement speed increase

**Boiler**
- Adds Neurotoxic Gas to Bombard's description, since it was forgotten about when Neuro was added.
- Acid Shroud: Clarifies that what you produce is the same gas as bombard, and that you can make a cloud of either kind.

**Lurker**
- Invisibility now tells the player the duration of the cloak, as saying "a short time" is both not very useful for planning, and also wrong (30 seconds is a fairly long time)
- Now mentions Invisibility's movement speed increasing properties, which are slightly hard to notice in-game.

## Requirements

**Changelog**

:cl:
- tweak: Changed most Xeno ability descriptions to give more complete and correct information.

### DIFF
--- a/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
+++ b/Resources/Prototypes/_RMC14/Actions/Xeno/xeno_offense_actions.yml
@@ -127,7 +127,7 @@
   id: ActionXenoHeadbutt
   parent: ActionXenoBase
   name: Headbutt (10) # TODO RMC14 proper plasma costs
-  description: Charge in a direction and push back any enemies in your way.
+  description: Charge towards the targeted enemy and push back the first enemy hit.
   components:
   - type: EntityTargetAction
     itemIconStyle: NoItem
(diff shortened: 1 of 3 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4052, opened 2024-09-15
Make landing signs a dynamic structure

Dutch-VanDerLinde:
So im able to put them on the right spot on LV-624


DrSmugleaf (approved):
Thanks for the PR!

PR #4071, opened 2024-09-17
make window frames deconstructable with a wrench

### LINKED ISSUE #4070: window frames should be deconstructed by a wrench, not a welder
cm13 parity

Dutch-VanDerLinde:
resolves #4070
:cl:
- tweak: Window frames are now deconstructed with a wrench rather than a welder


DrSmugleaf (approved):
Thanks for the PR!

PR #4074, opened 2024-09-18
Squad icons can only be seen if you're wearing a headset, shade marine icons

Dutch-VanDerLinde:
marine icons are now shaded, you need a headset to see them

https://github.com/user-attachments/assets/f86ff910-d0c2-47fe-897a-057066c4ca52


DrSmugleaf (approved):
Thanks for the PR!

PR #4082, opened 2024-09-19
Smartgun falloff update

Vermidia:
## About the PR

Updated smartgun stuff to cm-13s newest changes. So falloff starts at 7 tiles, and the gun has a max range of 12. And removed farsight.

## Why / Balance

Parity, fix. Smartgunners shouldn't be sniping xenos 20 tiles away with full damage bullets.

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Smartgun has falloff start at 7 tiles and loses 4 damage a tile
- remove: ML66A head mounted sight no longer gives farsight

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/smart_gun.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/smart_gun.yml
@@ -18,7 +18,7 @@
     - CMCartridgeSmartGun10x30mm
 
 - type: entity
-  parent: BaseBullet
+  parent: CMBulletBase
   id: CMBulletSmartGun10x30mm
   name: bullet (10x30mm)
   noSpawn: true
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/smart_gun.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/smart_gun.yml
@@ -34,3 +34,11 @@
         Piercing: 30
   - type: CMArmorPiercing
     amount: 0
+  - type: RMCProjectileDamageFalloff
+    thresholds:
+    - range: 12
+      falloff: 9999
+      ignoreModifiers: true
+    - range: 7
+      falloff: 4
+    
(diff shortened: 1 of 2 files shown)


Dutch-VanDerLinde:
this just makes the smartgun ass

Vermidia:
It had constant falloff before, making it only good at close range. This is paritying the recent changes made in cm-13 where it was changed because to actually get the most damage melee SG was best.

MACMAN2003:
the (rmc) smartgun doesn't have any falloff right now

Vermidia:
Thats a bug lol. It parented from base bullet and was forgotten in the update.

Dutch-VanDerLinde:
Whats the default falloff values?

Vermidia:
Falloff starts out at 0 range and loses 1 damage per tile travelled. Max range is like 22 when falloff becomes 9999.

MACMAN2003:
mfw something that makes the game fun was actually a bug

DrSmugleaf:
> i feel like you'd need more in the "why" section than just "parity" cause this affects balance and cm13 is balanced differently because they have tile movement.

Parity is still parity even when you don't like the change, it being tile movement or not is irrelevant when every single gun in the game has correct damage fall-off values except for the smart gun

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Ammunition/Cartridges/smart_gun.yml
```diff
@@ -34,3 +34,11 @@
         Piercing: 30
   - type: CMArmorPiercing
     amount: 0
+  - type: RMCProjectileDamageFalloff
+    thresholds:
+    - range: 12
+      falloff: 9999
+      ignoreModifiers: true
+    - range: 7
+      falloff: 4
```
DrSmugleaf, line 43: Are these the updated values from 13's most recent PR?

Vermidia: Yep. The max range was 12 before the PR, and it had the normal bullet stats otherwise. It was only the effective range that was changed to 7 and falloff from it to 4.

DrSmugleaf: I'll merge this probably with either the IO update or motion detectors if we get them before then, I want to add job slot scaling in the same update that removes autobalancing and I don't want that to screw with the numbers

EpicJoston:
I agree with fall off change but with free movement and how fast Xenoids can get to you I disagree with removing far sight. I think that utility can still be given to SGO without it being too powerful in RMC, even if the change breaks "parity" with CM.

DrSmugleaf (approved):
Gonna be a sec until motion detectors but I can try working on them after shotgun handfuls

PR #4114, opened 2024-09-21
fix landing signs being pullable

Dutch-VanDerLinde:
so dynamic structures are anchorable normally, we want it so you can place the landing sign anywhere, not just snapgrid so we remove dynamic structure parent and turn off anchored transform

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Misc/landing_signs.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Misc/landing_signs.yml
@@ -8,7 +8,7 @@
   - type: Transform
     anchored: false
   - type: Physics
-    bodyType: Dynamic
+    bodyType: Static
   - type: InteractionOutline
   - type: Sprite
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
ty boss

Resources/Prototypes/_RMC14/Entities/Structures/Misc/landing_signs.yml
```diff
@@ -1,10 +1,14 @@
...
+  parent: CMBaseStructureCorrodible
   abstract: true
   id: RMCBaseLandingSign
   name: sign
   description: A sign.
   components:
+  - type: Transform
+    anchored: false
+  - type: Physics
+    bodyType: Dynamic
```
DrSmugleaf, line 11: Actually this will still make it pullable, you want Static
```suggestion
    bodyType: Static
```

DrSmugleaf (changes requested):

PR #4120, opened 2024-09-22
fix super reinforced wall

Dutch-VanDerLinde:
it had no name so it appeared as:
image


Resources/Prototypes/_RMC14/Entities/Structures/Walls/walls.yml
```diff
@@ -90,7 +90,9 @@
 
 - type: entity
   parent: CMBaseWall
+  name: super reinforced wall
```
DrSmugleaf, line 93: From 13 the name is just reinforced wall apparently
```suggestion
  name: reinforced wall
```

DrSmugleaf (dismissed):
Thanks for the PR!

PR #4129, opened 2024-09-23
Remove fire delay from guns when you switch hands

### LINKED ISSUE #4088: Gun delay is too high
In CM, it's almost instant to equip and begin shooting a gun, but here it takes like 3 seconds

Dutch-VanDerLinde:
In CM13 you can dual wield like this:

https://github.com/user-attachments/assets/7ef65909-b47a-4148-9c05-a7ffddb23a51

currently, this isn't possible, because there's a cooldown

resolves #4088

I disabled the "resetOnHandSelected" datafield in this PR
also gave a melee sound for the gun

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/cm_base_gun.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/cm_base_gun.yml
@@ -22,6 +22,8 @@
     damage:
       types:
         Blunt: 5
+    soundHit:
+      collection: GenericHit
   - type: EmitSoundOnPickup
     sound:
       collection: RMCEquipGun
@@ -30,3 +32,4 @@
       collection: RMCRustleGun
   - type: Gun
     meleeCooldownOnShoot: false
+    resetOnHandSelected: false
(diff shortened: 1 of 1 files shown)


DogZeroX:
What if you just make it a mechanic instead of relying on wacky x spam so you can add penalty to or something

Dutch-VanDerLinde:
> What if you just make it a mechanic instead of relying on wacky x spam so you can add penalty to or something
I dont understand

DogZeroX:
Add an action or button to dual-wield instead of having to swap hand spam

Vermidia:
You can see in the shot that cm already makes both guns fire if you have them in both hands. I believe it comes with a heavy hit to accuracy and or scatter.

DogZeroX:
I can't see that, I just see the hands quick-swapping which looks silly

MACMAN2003:
THERE WAS A DELAY THIS WHOLE TIME?!! I THOUGHT IT WAS JUST PING

Dutch-VanDerLinde:
> You can see in the shot that cm already makes both guns fire if you have them in both hands. I believe it comes with a heavy hit to accuracy and or scatter.

theres still not supposed to be a delay anyway, messes up with picking up dropped guns

Kadeo64:
MAX PAYNE

DrSmugleaf (approved):
Thanks for the PR!

k0r4y:
https://github.com/user-attachments/assets/1162a7e3-1ff2-485e-8c7f-2ba90326c1e6

just putting this here, not sure if this is an intended mechanic of this change, seems a bit overpowered with shotguns

Vermidia:
Yeah we need fire group delays, so you can't fire another shotgun after firing a shotgun. It's mainly specifically for to prevent juggling.

PR #4197, opened 2024-09-30
Fix xeno hud and medhud appearing while a scout is invisible

### LINKED ISSUE #4181: scouts need to uncloak if they're knocked down
mainly so marines can actually see dead scouts and also so xenos can see their captures

Dutch-VanDerLinde:
image

resolves #4181

:cl:
- fix: Fixed xeno hud icons being visible while a scout is invisible.
- fix: Fixed scout specialists not becoming visible on death.

### DIFF
--- a/Content.Client/StatusIcon/StatusIconSystem.cs
+++ b/Content.Client/StatusIcon/StatusIconSystem.cs
@@ -1,3 +1,4 @@
+using Content.Shared._RMC14.Stealth;
 using Content.Shared.CCVar;
 using Content.Shared.Ghost;
 using Content.Shared.StatusIcon;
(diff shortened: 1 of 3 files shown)


DrSmugleaf:
> > Does the stealth cancel when dying like in 13? Otherwise this might make the scout too hard to find for marines currently
> 
> I can add that but when does stealth cancel? only on dying? when they're stunned?

Crit doesn't cancel it, death does, warrior lunge doesn't, queen neuro spit doesn't, so seems to be just death

Dutch-VanDerLinde:
> Does the stealth cancel when dying like in 13? Otherwise this might make the scout too hard to find for marines currently

Added the stealth cancel on death

DrSmugleaf (approved):
Thanks for the PR!

PR #4223, opened 2024-10-02
hive member unification and many fixes

deltanedas:
## About the PR
- removed hive member from Xeno / XenoEgg / XenoProjectile
  the thing for when a marine is infected should have its own field still i think, it would be weird a marine getting infected suddenly being a full fledged member of the hive (only xeno agent patent pending antag should!)
- added proper hive member api to SharedXenoHiveSystem
- moved hive creation to server since nothing called it from shared anyway
- fixed a bunch of subtle bugs that wouldve appeared with multiple hives

## Why / Balance
5 different things with Hive member and C-F tier apis for using them
:bug: :hammer:

## Technical details
the hive member api: `GetHive / SetHive / SetSameHive / FromSameHive / IsMember`
morbillion things changed to use it
very nice to use no adding .Value everywhere it just works (only have to sometimes use (hive, hive.Comp) because Entity trollage but it could be fixed in the future)

added `HiveChangedEvent` to not have night vision tightly coupled to xenos hive changing

bugs squished:
- if you don't have a hive tracker would display like there's no queen, now you don't get a tracker since you don't have a hive, naughty rogue xeno
- previously 2 rogue xenos couldnt use abilities on eachother because it did dumb hive == hive check, now it gets IsSameHive for friendly fire prevention. rogue xenos can now use lunge etc on eachother probably
- probably more i forgor
- using rmc actions to create a new hive didnt set the xeno's hive to it, now it does so you don't have to do that manually
- fix unique construction ignoring the other construct's hive i assume that in xeno v xeno each hive can still have their own core and its not just ping racing
- some tiny micro optimisations, mostly from using entityqueries
- rogue xenos can shoot/lunge/headbutt eachother
- rogue xenos trying to use the watch action get nothing instead of an empty list (this is actually parity too i checked the dm)

## Media
:trollface:

## Requirements

## Breaking changes
removed `Hive` field and system's api from `XenoComponent` and `XenoEggComponent`, they use `HiveMemberComponent` and `SharedXenoHiveSystem` now
`SetSameHive` in hive system takes src, dest instead of dest, src

**Changelog**
no cl no fun

### DIFF
--- a/Content.Server/_RMC14/Admin/RMCAdminEui.cs
+++ b/Content.Server/_RMC14/Admin/RMCAdminEui.cs
@@ -1,6 +1,7 @@
 ﻿using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Content.Server._RMC14.Rules;
+using Content.Server._RMC14.Xenonids.Hive;
 using Content.Server.Administration;
 using Content.Server.Administration.Managers;
 using Content.Server.EUI;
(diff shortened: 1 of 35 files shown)


Kadeo64:
greeno events are REAL

DrSmugleaf (approved):
Thanks for the PR!

PR #4243, opened 2024-10-04
Fix IFF text always displaying on examine regardless of it actually has IFF

Rainbeon:
## About the PR
The examine event now checks if IFF is actually enabled before saying it is, also added the examine event handling to GunAttachableIFF for smartscopes since otherwise it'd not tell the player that the gun is now IFF. Localized IFF examine text as well.

## Why / Balance
The examine event would lie to you and just say the gun ignores friendlies regardless of if it does or not, as all guns have GunIFFComponent, but not all of them are actually enabled. Also, localization good (i think). 

Prevents new players on update day from unwittingly mass teamkilling because their gun lied to them.

## Media
Video Demonstration
https://github.com/user-attachments/assets/3db929d9-ba43-49f1-8f34-b54834556e48

## Requirements

## Breaking changes

**Changelog**
- fix: Guns no longer lie to you in their examine text about having IFF


DrSmugleaf (approved):
ty

PR #4248, opened 2024-10-04
fix lessers not having a hive

deltanedas:
## About the PR
when hive core is built it tries to set hive to the construction node (SetHive makes this a no-op, added logging now to find this error in the future)

now it correctly sets the hive so lessers are not rogue


Kadeo64:
bug of the year 2024 award rmc14 edition

PR #4294, opened 2024-10-06
add xeno announcement for when drone tax ends

deltanedas:
## About the PR
`It has been long enough, we will no longer accumulate evolution points.`

## Why / Balance
- saves someone having to either set a timer or wait patiently until evo stars going backwards to tell everyone tax is over
- if you arent aware of it, it might seem like a bug that you are *losing* evo. now it will be much more obvious

this isnt a cm13 thing so **no parity here**, just good tax fun

## Technical details
naughty system bool to track announcement being made

## Requirements

## Breaking changes
no

**Changelog**
:cl:
- add: Xenos now get an announcement when they stop accumulating evo points past the maximum (15 minutes in).

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Evolution/XenoEvolutionSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Evolution/XenoEvolutionSystem.cs
@@ -57,6 +57,8 @@ public sealed class XenoEvolutionSystem : EntitySystem
     private TimeSpan _evolutionPointsRequireOvipositorAfter;
     private TimeSpan _evolutionAccumulatePointsBefore;
 
+    private bool _didAccumulationAnnouncement;
+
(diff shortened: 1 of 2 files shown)


DrSmugleaf (changes requested):
This is the same time as the T3 unlock announcement which is what people use in 13 to time it, could still have it but it might be a bit much in the chat, or be loud sound-wise if both play at the same time

deltanedas:
can make it silent or just be a popup then to avoid sound spam

deltanedas:
actual sound is null by default already

DrSmugleaf:
> actually sound is null by default already @DrSmugleaf do you want it to just be a popup

That would make both popups clip right?

DrSmugleaf:
Reopen if you come back to this

deltanedas:
t3 popups have existed for a while so

PR #4298, opened 2024-10-06
let xenos melee rogue xenos + let xenos break weeds

deltanedas:
## About the PR
xenos can now fight eachother if they are not in the same hive

weeds now have a hive so xenos can break other hives' weeds

distress signal assigns the hive after loading the planet map, xeno v xeno would need something smarter

## Why / Balance
- there was a todo
- they already can with abilities so

## Technical details
if the target has XenoFriendly now it also checks if they are in the same hive
so you can bonk rogue xenos and for xeno v xeno you can break the other hives weeds and whatever else

## Media
figure out which xeno is rogue 99% impossible

## Requirements

## Breaking changes
no

**Changelog**
no cl no fun

### DIFF
--- a/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
+++ b/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
@@ -1182,7 +1182,7 @@ public sealed class CMDistressSignalRuleSystem : GameRuleSystem<CMDistressSignal
     public void SetWeedsHive(EntityUid hive)
     {
         var query = EntityQueryEnumerator<XenoWeedsComponent>();
-        while (query.MoveNext(out var weeds, out _)
+        while (query.MoveNext(out var weeds, out _))
         {
             _hive.SetHive(weeds, hive);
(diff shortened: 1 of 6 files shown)


DrSmugleaf (changes requested):
Xeno's aren't supposed to be able to hit their own weeds. If this is for nesting, that's fixed by letting them unnest in a different PR
I wouldn't really worry about performance for that, the xeno weeds component itself is heavier

deltanedas:
ok ill give hive member to weeds and keep it as XenoFriendly

DrSmugleaf:
This also will need to account for round-start weeds

deltanedas:
would it be fine to just assign the distress call hive to every xenofriendly entity after the planet is loaded

DrSmugleaf:
> would it be fine to just assign the distress call hive to every xenofriendly entity after the planet is loaded

Sure

deltanedas:
seems to work as expected on shivas at least

DrSmugleaf (approved):

PR #4345, opened 2024-10-08
Xeno Healing Parity (almost, still slightly underpowered)

Rainbeon:
## About the PR
Xeno Healing in RMC14 has roughly halved compared to 13's healing, and it's been because it actually has been halved in the formula (which does not exist in 13), this boosts up healing to near-13 levels. Of note, this still seems to be slightly slower than 13's in some cases. Of note Garden Drone's 1.05x healing recovery is also missing, for when that is implemented.

## Why / Balance
CM13 Parity

## Technical details
Removes random division by 2

## Media
Quick Video Demo (healing is still lower than what it should be for this scenario, but much much closer)
https://github.com/user-attachments/assets/da6b8390-b9e4-4c88-8582-30cb4246b9cd

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Xeno Health Regeneration being half of what it should be in 13.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/XenoSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/XenoSystem.cs
@@ -265,7 +265,7 @@ public sealed class XenoSystem : EntitySystem
         var passiveHeal = threshold.Value / 65 + xeno.Comp.FlatHealing;
         var recovery = (CompOrNull<XenoRecoveryPheromonesComponent>(xeno)?.Multiplier ?? 0) / 2;
         var recoveryHeal = (threshold.Value / 65) * (recovery / 2);
-        return (passiveHeal + recoveryHeal) * multiplier / 2;
+        return (passiveHeal + recoveryHeal) * multiplier; // TODO RMC14 add Strain based multiplier for Gardener Drone
     }
 
     public void HealDamage(Entity<DamageableComponent?> xeno, FixedPoint2 amount)


crystalHex:
I've currently implemented the gardener's increased health regen by just multiplying the default multipliers by 1.05 and adding them to its prototype: 
image

Rainbeon:
> I've currently implemented the gardener's increased health regen by just multiplying the default multipliers by 1.05 and adding them to its prototype

How 13 handles this Gardener only has the 1.05x modifier (5%) and then just multiplies the whole formula by it at the end rather than maintaining 3 different values, which would generally be more elegant and scalable (as a healing mult of 0.3465 is a bit strange)

Kadeo64:
so if whisper/smug actually reads the comments on this PR i want to ask why this was done
is it because you guys expected xenos to be OP when marines had no content and no strategy? like a pre-emptive playtest nerf?

Rainbeon:
> so if whisper/smug actually reads the comments on this PR i want to ask why this was done is it because you guys expected xenos to be OP when marines had no content and no strategy? like a pre-emptive playtest nerf?

This PR came from a call with Smug actually and I did ask her, from the sounds of it it was unintentional and this wasn't really looked into in much detail because Xenos were doing really well for the time. 

I brought up while we were playing cm13 that I felt like there was a disparity between our HP recovery, and she didn't know why it was, and eventually I just discovered the difference myself.

Kadeo64:
> This PR came from a call with Smug actually and I did ask her, from the sounds of it it was unintentional and this wasn't really looked into in much detail because Xenos were doing really well for the time.

Huh, wonder how it happened then. I did always think xenos healed a little bit slow, with me sitting in a room without pheromones for 5 minutes to get all my health back.

Vermidia:
I assume it might've been due to how ss13 time works, in that sometimes updates happen only every other tick apparently (so every other second). But this is kinda all over the place and varies for so many things so you usually just have to test it manually.

DrSmugleaf (changes requested):
Actually I remember now what Verm said is correct, I added / 2 because life ticks every 2 seconds in dm, and we tick it every second, so something else must be incorrect

DrSmugleaf (approved):

Rainbeon:
For the onlookers, CM13 seems to actually tick it once per second, but only updates the display every 2 seconds. Damage healing in the same scenario displayed below (Post-HEDP Explosion on a Lurker, Resting on Weeds next to Warden Prae Recovery Pheromones)

CM13 (54 hp/2 seconds)
https://github.com/user-attachments/assets/c2fee1ed-f06d-4c9c-8901-40d0257c9b13

This PR (51 hp/2 seconds)
https://github.com/user-attachments/assets/d3a15c61-02f8-477c-8a5c-f4d2308705fd

Ours is still slightly underpowered because our burn heal is not as strong (but our brute heal is slightly stronger), though the net sum of the 2 is still slightly lower than 13's.

PR #4401, opened 2024-10-10
Fix RMC admin actions transforming into gardener instead of base drone

crystalHex:
## About the PR

Attempting to transform into a drone through the RMC admin actions menu transformed the target into the gardener drone instead of the regular drone. This PR fixes that.

## Why / Balance

Bug fix

## Technical details

Commented out an unfinished prototype

## Requirements

## Breaking changes

**Changelog**


QuietlyWhisper:
Why does the same bug not happen with the warden strain?

Rainbeon:
I don't think we need Changelog here as the Gardener Drone isn't released yet, so no player-facing change is made here

Dutch-VanDerLinde:
> I don't think we need Changelog here as the Gardener Drone isn't released yet, so no player-facing change is made here

gardenerer was mreged right?

Rainbeon:
> > I don't think we need Changelog here as the Gardener Drone isn't released yet, so no player-facing change is made here
> 
> gardenerer was mreged right?

It's merged but not live so normal players don't experience this and thus have no need to be notified of it being fixed, it's similar in effect to a bug being fixed in the original before its merged

crystalHex:
So uh, turns out this was entirely on me. I added an unfinished prototype for the healer to test out lesser resin fruit (and to make sure whoever ports it has the fruit ability working correctly), but didn't add it as a viable strain for the drone (due to it not being finished). Commenting it out also fixes the issue.

PR #4454, opened 2024-10-13
Change suicide delay to 2 seconds

Dutch-VanDerLinde:
- https://github.com/cmss13-devs/cmss13/pull/7294

:cl:
- tweak: Changed gun suicide do-after delay from 4 to 2 seconds.

### DIFF
--- a/Content.Shared/_RMC14/Suicide/RMCSuicideComponent.cs
+++ b/Content.Shared/_RMC14/Suicide/RMCSuicideComponent.cs
@@ -12,7 +12,7 @@ public sealed partial class RMCSuicideComponent : Component
     public DamageSpecifier Damage = new();
 
     [DataField, AutoNetworkedField]
-    public TimeSpan Delay = TimeSpan.FromSeconds(4);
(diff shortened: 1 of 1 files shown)


MACMAN2003:
outside the scope of this PR but i feel like a confirmation box would be required before this is merged, primarily because 2 seconds is not a lot of time to cancel the action.

deltanedas:
confirmation box would make it too long to kill yourself to avoid capture, especially if you get tunnel vision

crystalHex:
As long as this remains in the rules (see screenshot) and as long as the keybind for suiciding is the same as a mundane action, e.g. vaulting a window frame, there really should be a confirmation window.

image

Dutch-VanDerLinde:
> As long as this remains in the rules (see screenshot) and as long as the keybind for suiciding is the same as a mundane action, e.g. vaulting a window frame, there really should be a confirmation window.
> 
> image

An admin isnt gonna ban you for accidentally suiciding yourself

DrSmugleaf (approved):

PR #4489, opened 2024-10-16
Makes xenos jitter on evolve

Dutch-VanDerLinde:
https://github.com/user-attachments/assets/7a124440-2912-4d2d-a86d-9ec598c4be69

cm13


DrSmugleaf (dismissed):
Thanks for the PR!

PR #4494, opened 2024-10-16
Fix Shotgun + Grenade Launcher ammo counters not updating when loaded

Rainbeon:
## About the PR
Fixes the ammo counter for Grenade Launchers, Shotguns, and other non-magazine fed ballistic weapons to update when ammo is manually inserted.

## Why / Balance
Bug fix, it doesn't do anything at all currently when being loaded, only when being taken out or shot.

## Technical details
Grenade Launchers are fixed by a simple UpdateAmmoCounter call

Shotguns had to be more involved, since Handfuls completely deprive the client of information regarding how much ammo is loaded until after the ammo is already inserted, due to returning if the client reaches it. The ammo counter is entirely client sided so the server does nothing, and you can't make the loading un-client-sided without introducing entirely different bugs (it causes the handful to disappear when loading, spam swapping your weapon hand if loading too quickly)

To circumvent this I added an int field named ArtificialIncrease to UpdateAmmoCounter that allows us to spoof ammo being loaded since its all in the client, even if it hasn't been loaded yet. This is helpful in this specific use case as we know the amount of ammo we load from a handful will always be exactly one, so we can just tell it we are loading one shell and it works very well, while still letting you load the shotgun fast.

## Media
Video Demonstration

https://github.com/user-attachments/assets/775eb26a-7dc7-4ca4-b82a-290f6f56a9d9

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Shotguns and Grenade Launchers now correctly display ammo when being loaded.

### DIFF
--- a/Content.Client/Weapons/Ranged/Systems/GunSystem.AmmoCounter.cs
+++ b/Content.Client/Weapons/Ranged/Systems/GunSystem.AmmoCounter.cs
@@ -45,20 +45,21 @@ public sealed partial class GunSystem
         UpdateAmmoCount(uid, component);
     }
 
-    private void UpdateAmmoCount(EntityUid uid, AmmoCounterComponent component)
+    private void UpdateAmmoCount(EntityUid uid, AmmoCounterComponent component, int artificialIncrease = 0)
     {
         if (component.Control == null)
             return;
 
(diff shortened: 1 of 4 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4507, opened 2024-10-17
Add evacuation start and cancel sound

Dutch-VanDerLinde:
cm13, more IMMERSION


Resources/Audio/_RMC14/Announcements/ARES/attributions.yml
```diff
@@ -27,3 +27,8 @@
   license: "CC-BY-SA-3.0"
   copyright: "cmss13"
   source: "https://github.com/cmss13-devs/cmss13/21e6447cc08aea502f671c819fdbcecbb85e6028master/sound/AI/hijack.ogg"
+
+- files: ["evacuate.ogg"]
+  license: "CC-BY-SA-3.0"
+  copyright: "cmss13"
+  source: "https://github.com/cmss13-devs/cmss13/blob/21e6447cc08aea502f671c819fdbcecbb85e6028/sound/AI/evacuate.ogg"
```
DrSmugleaf, line 34: We can't use this sound. That's the reason it wasn't included with the power evac update. The alarm sound is the same one used in the movies.

DrSmugleaf (changes requested):

PR #4518, opened 2024-10-18
make fence have 50 HP

Dutch-VanDerLinde:
image

in cm13 it has 50 hp

:cl:
fix: Fixed fence health being twice the intended amount.


DrSmugleaf (approved):
Thanks for the PR!

PR #4529, opened 2024-10-19
Improve Runner/Parasite/Lurker Leap Hit Registration at Close Range

### LINKED ISSUE #4486: Parasite Parity: Parasites should be able to leap-infect Marines on the same tile as themselves
## Description
When Parasites use Leap, Marines on the same tile as themselves are able to be infected, Parasites do not get body-blocked in 13 so this is something that can explicitly happen, CM13 Parity. Video included at the bottom.

Ours just glitch around a bit and do nothing, like so:

https://github.com/user-attachments/assets/c2ffae32-85fe-48f6-b915-09eb0b66fb85

**Videos**
CM13 Video Demonstration
https://github.com/user-attachments/assets/4e50fea0-eab4-4891-b9ae-32717af27e7b

Rainbeon:
## About the PR
1) When starting your leap, you now check if you are already contacting other valid entities for leap targeting. This entirely fixes close-range leap hit registration since the way leap hit registration worked before was that it only applied the effect if you started contacting a new entity mid-leap. If you were slightly too close or they walked into your sprite slightly beforehand they became immune to being leaped.

2) Moved the Leap Hit Audio from playing with predicted to server-sided Pvs instead, to remove false positives and false negatives. The leap hit sound is supposed to be confirmation you've hit your target, but when playing predicted it would often play when it's not supposed to (you missed from the server's eyes) and could fool you into thinking you revealed yourself (Lurker).

## Why / Balance
Hit Registration improvements good, especially good for high ping as being too close despite what your client shows you is no longer an issue.

This is also pure CM13 Parity, as Parasites, Runners, and Lurkers can share the same tile as a target and still knock them down with their Leap ability. (Runners and Lurkers can occupy the same tile by standing on top of someone who is on the ground, and then them getting up).

https://github.com/user-attachments/assets/c18e71d2-a3c9-4df3-8c8a-84eb437aa56e

Resolves #4486 

## Technical details
Moved the code from OnXenoLeapingDoHit() to a new helper function named ApplyLeapingHitEffects(), allowing it to be re-used without code duplication and so we can return a bool on successful/failed target validation. 

Called ApplyLeapingHitEffects() to OnXenoLeapDoAfter() which is only called when the contact, gathered from _physics.GetContactingEntities(), is not a Xeno of the same hive (minorly clipping a Xeno when starting your leap leading to nothing would suck), if it can already find a target when you begin to jump, it returns.

Additionally, changed the leap hit sound from PlayPredicted to PlayPvs with a _net.IsServer check so the sound only plays when the server thinks you hit something, because the client is really bad at telling what you actually did without predicted abilities.

## Media
Video Demonstration
https://github.com/user-attachments/assets/bf3164f2-4dbb-4f72-a502-57bbbfd983a2

## Requirements

**Changelog**

:cl:
- fix: Runner, Lurker, and Parasite Leaps will now register hits on Marines that are too close or sharing the same tile as them.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
@@ -137,61 +137,21 @@ public sealed class XenoLeapSystem : EntitySystem
 
         _physics.ApplyLinearImpulse(xeno, impulse, body: physics);
         _physics.SetBodyStatus(xeno, physics, BodyStatus.InAir);
+
+        //Handle close-range or same-tile leaps
+        foreach (var ent in _physics.GetContactingEntities(xeno.Owner, physics))
+        {
+            if (_hive.FromSameHive(xeno.Owner, ent))
+                continue;
+
+            if (ApplyLeapingHitEffects((xeno, leaping), ent))
+                return;
+        }
     }
 
     private void OnXenoLeapingDoHit(Entity<XenoLeapingComponent> xeno, ref StartCollideEvent args)
     {
-        var other = args.OtherEntity;
-        if (xeno.Comp.KnockedDown)
-            return;
-
-        if (!HasComp<MobStateComponent>(other) || _mobState.IsIncapacitated(other))
-            return;
-
-        if (_standing.IsDown(other))
-            return;
-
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4542, opened 2024-10-20
New M54c and Heavy sprites

QuietlyWhisper:
:cl: Whisper
- add: Updated the sprites of the M54C and M54CE2 (In development)

image

image


VictorJob:
As it's a resprite, I think it would be good to change the shapes a little more, making some things clearer, like this
image
I dare to suggest also putting bullpup on the gun too
image

QuietlyWhisper:
It's inspired by the vector design
image

DrSmugleaf (approved):
Thanks for the PR!

PR #4564, opened 2024-10-22
Admin alert for faxes includes the receiver

QuietlyWhisper:
Needed to help differentiate urgency in fax messages

image


DrSmugleaf (approved):
Thanks for the PR!

PR #4592, opened 2024-10-25
Port Bipod 90 degree turning radius

### LINKED ISSUE #4590: Bipod Parity: Bipods should only auto-dismount when firing at a 90 degree angle from where you were when you started, not 45 degrees
## Description
Bipod is supposed to have a full 180 degree "cone" it can fire in without dismounting, rather than the current 90 degree cone. This does not apply to scopes, just bipods.

Scopes may direction lock independently of the Bipod's locked direction. For example, you may direction lock a Bipod North, and a Scope East. Shooting straight North will only break the Scope, while shooting straight East (exactly 90 degrees away from North) will break the Bipod mount.

**Video**
https://github.com/user-attachments/assets/2b0d97f4-73ec-414b-ad71-311c0a6bbd77

Rainbeon:
## About the PR
Ports the correct firing cone for the Bipod from CM13, increased from 45 degrees to a full 90 degrees in each direction, giving you a firing cone of 180 degrees total before your Bipod dismounts itself. The checks for this are similar but entirely separate from the other checks for scopes, which are correct at 45 degrees, and you can have your scope aimed one direction and your bipod aimed another direction without them all incorrectly cancelling each other.

## Why / Balance
CM Parity. 

Resolves #4590 

## Technical details
Adds CheckUserBreakOnFullRotate() to AttachableToggleableSystem. This functions similar to CheckUserBreakOnRotate() takes in the player's and target's position as arguments, and uses these arguments to calculate a difference in angle between your initial mount angle and the angle that you just shot. There is some funky looking math to account for angle wrap-around, because angles represent a circle, you have to add +540, modulo 360, and then take away 180 afterward to get this number. If the difference in angle is 90 degrees or more in either direction, the Bipod will dismount.

Adds a field BreakOnFullRotate to AttachableTogglableComponent, and assigns it to the Bipod in YML.

Adds a new component AttachableSideLockedComponent to pair with BreakOnFullRotate, it functions similarly to AttachableDirectionLockedComponent and its pair BreakOnRotate. They have to be separate despite how similar the code is because their functions are separate and have no gameplay relation to each other, they act independently.

In SharedGunSystem, GunShotEvent now passes the coordinates of the user and the gun shot, so that event listeners can use the data. In this case, the Bipod needs to know this information to determine the World Angle of the gunshot, which it has no way of knowing otherwise.

## Media
Video Demonstration
https://github.com/user-attachments/assets/4b1e9913-c999-4be9-8edc-11f47b632ef1

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: The Bipod can now correctly turn 90 degrees in either direction before automatically dismounting, up from 45 degrees.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
@@ -381,7 +381,7 @@ public abstract partial class SharedGunSystem : EntitySystem
 
         // Shoot confirmed - sounds also played here in case it's invalid (e.g. cartridge already spent).
         var projectiles = Shoot(gunUid, gun, ev.Ammo, fromCoordinates, toCoordinates.Value, out var userImpulse, user, throwItems: attemptEv.ThrowItems, predictedProjectiles, userSession);
-        var shotEv = new GunShotEvent(user, ev.Ammo);
+        var shotEv = new GunShotEvent(user, ev.Ammo, fromCoordinates, toCoordinates.Value);
         RaiseLocalEvent(gunUid, ref shotEv);
 
         if (userImpulse && TryComp<PhysicsComponent>(user, out var userPhysics))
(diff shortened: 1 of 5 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4598, opened 2024-10-26
fix yml errors on master

### LINKED ISSUE #4596: noSpawn needs removing
it's causing a linter error
image

QuietlyWhisper:
- fix #4596

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Jobs/Other/senior_enlisted_advisor.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Jobs/Other/senior_enlisted_advisor.yml
@@ -87,7 +87,7 @@
   id: CMGuidebookRoleSEA
   name: Senior Enlisted Advisor
   description: Mentor for the Marines!
-  noSpawn: false
+  categories: [ HideSpawnMenu ]
   components:
   - type: Sprite
     state: senior_enlisted_advisor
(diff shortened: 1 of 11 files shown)


Resources/Prototypes/_RMC14/Roles/Jobs/Other/senior_enlisted_advisor.yml
```diff
@@ -87,7 +87,7 @@
   id: CMGuidebookRoleSEA
   name: Senior Enlisted Advisor
   description: Mentor for the Marines!
-  noSpawn: false
+  categories: [ HideSpawnMenu ]
```
deltanedas, line 90: as i commented in the issue, its nospawn false not true, so set categories to []

deltanedas (changes requested):

Resources/Prototypes/_RMC14/Roles/Jobs/Other/senior_enlisted_advisor.yml
QuietlyWhisper: This shouldn't be spawnable, as it's a guidebook entity I created, is that not the correct usage?

DrSmugleaf (approved):
Thanks for the PR!

PR #4631, opened 2024-10-30
Stumps

CyberTropic:
## About the PR

Added stumps to trees that had existing stump graphics for them and also created one for the alien tree. These stumps inherit from bushes, so they're purely visual entities that don't block movement and can be easily destroyed with melee. 
Adds suffix to trees so you can see in the entity spawner which ones are RMC and which ones create logs (yuck)
ALSO, stumps are used in CM13, but it's used for foliage clearing mechanic which we don't have yet, so that sprite was ported as well but not used.

## Why / Balance

Visual change. Trees simply blinking out of existence after being shot looked weird.

## Media

Screenshot 2024-10-30 102208

https://github.com/user-attachments/assets/ae1c3a96-5e4a-406e-9cc1-ee005dba49ef

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Trees leave behind stumps when destroyed.


vero5123:
would be cool if trees fell :P

DrSmugleaf (approved):
Thanks for the PR!

PR #4634, opened 2024-10-30
SEA Vendor Update

PursuitInAshes:
## About the PR
The SEA uniform vendor now has the option for the service jacket, ballistics vest, medkit pouch.
The SEA Weapon Vendor now has the reddot sight.

## Technical details
YAML Updating.

## Requirements

## Breaking changes
N/A

**Changelog**
:cl:
- tweak: The SEA vendors have been updated to include new items and equipment: Service Jackets, medkit pouch, reddot sight.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/senior_enlisted_advisor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/senior_enlisted_advisor.yml
@@ -104,4 +104,3 @@
       - id: RMCAttachmentS5RedDotSight
       - id: RMCAttachmentS6ReflexSight
       - id: RMCAttachmentRailFlashlight
-      - id: RMCAttachmentBipod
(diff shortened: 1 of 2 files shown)


Vermidia:
Almayer is supposed to have training nades in the firing range. That's why there isn't any in the vendor.

PursuitInAshes:
> Almayer is supposed to have training nades in the firing range. That's why there isn't any in the vendor.

Aye. I believed that was the case, but even then I do believe it best that SEAs have a consistent way of getting training grenades.

PursuitInAshes:
The Ubuntu fail seems unrelated.

QuietlyWhisper (changes requested):
SEA are meant to be given a MK1 rifle (not the basic m54c) which is not implemented yet, which is why the vendor doesn't contain a bipod.

The SEA if they would like a different weapon should be issued one by requisitions, though per SOP they would have to keep those weapons stored in their office anyway, unless the SEA is deploying. There's no reason they should be assigned whatever weapon they want otherwise.

If an SEA is "stealing" weapons they probably shouldn't be a whitelisted SEA. These roles are expected to follow marine law and SOP the same.

QuietlyWhisper (approved):
Thanks

PR #4660, opened 2024-11-02
Downed players can't be pulled over tables/window frames, restores Runner window travel, fix the game not compiling

Rainbeon:
## About the PR
Made being knocked down not allowing you to pass through tables, window frames, flaps, etc.

Restored Runner/Larva/Parasite running over window frames as the former was making this way too broken before, but it now works properly.

## Why / Balance
CM13 Parity. Needed for Runner traveling over window frames. Removes currently existing Runner table cheese.

Also I think prevents a Marine-only out of bounds bug for New Varadero (there are flaps that exit the map) in advance.

## Technical details
Added a new arg "changeCollision" to Down() in StandingStateSystem that is set to false by default, setting it to true lets you pass objects like before, but by default your collision will not change, as in CM13.

Added changeCollision: true to everything that needs it (I tested every .Down() call myself), that being Fireman Carry (so you can carry over objects when you are able to), Xeno nesting (prevents weird stuff happening with nesting over tables), and Buckle system. Everything else works correctly being set to false.

Restores MidImpassible to window frames now that the collision mask doesn't change

Removed erroneous FormatRipper inclusion in DropshipWeaponSystem so the game compiles. (Smug said to include this)

## Media
[Video Demonstration including edge cases](https://cdn.discordapp.com/attachments/1168210011823026270/1302077952577634345/2024-11-01_21-00-16.mp4?ex=6726ce17&is=67257c97&hm=858fd2ed1adb0cb4f68188e40b20f42a2ace180ad5bd82d4740ee7cbf1e0d78f&)

## Requirements

## Breaking changes
Changes default behavior for .Down(), but all cases should be okay as I've set the problem ones to the original behavior.

**Changelog**

:cl:
- tweak: Mobs that are knocked down can no longer be dragged through tables, window frames, and flaps without manually vaulting them.
- tweak: Runners, Larvae, and Parasites can now travel on top of window frames without vaulting once again.

### DIFF
--- a/Content.Server/_RMC14/Dropship/Weapon/DropshipWeaponSystem.cs
+++ b/Content.Server/_RMC14/Dropship/Weapon/DropshipWeaponSystem.cs
@@ -1,6 +1,5 @@
 using Content.Shared._RMC14.Dropship.Weapon;
 using Content.Shared.Decals;
-using JetBrains.FormatRipper.Elf;
 using Robust.Server.GameObjects;
 using Robust.Shared.Player;
 using System.Numerics;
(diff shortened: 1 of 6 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4663, opened 2024-11-02
Add Information Tags

Vermidia:
## About the PR

Title.

## Why / Balance

Parity!

## Technical details

TakeableTags Comp, pretty simple. Once a body is dead and either rotted or the user has police skill 2, they can take an info tag from a person using a verb (on the body or strip menu). The comp itself just adds a bit of examine text.

TakeableTags defines a proto which itself has the InformationTagsComp, which also has special text on examine but more importantly stores a list with name, job, and bloodtype (all O- until we get actual bloodtypes). Information tags can be used on other information tags to join them into one item while preserving the list.

The RMCMemorialComp also accepts the Info Tags and adds them to a list of fallen, discarding everything but the name.

All of this is done in a new DogtagsSystem. List modifiying is done server side only to prevent weirdness.

There's like a condition for diff factions can't take enemy tags in HvH - I didn't add that here.

## Media

https://github.com/user-attachments/assets/c215ef66-8d64-491d-b677-2ae8f5e79201

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added Information Tags. When a body can no longer be defibbed or is dead and you have police skill 2 or more, you can take an information tag from somebody. You can combine information tags together and use them on the slab of victory to engrave the names of the fallen.


DrSmugleaf (approved):
holy shit god bless ty

PR #4693, opened 2024-11-03
Xeno Projectiles now correctly aim at your cursor

### LINKED ISSUE #4692: Xeno Spits aren't firing at the crosshair correctly
## Description
Single shot spits at most angles in live games will not actually go to where your crosshair is aiming, consistently deviating from the crosshair in the same way each time. Instead what is actually happening is they are attempting to fire at whatever entity your crosshair is over, including non-targets such as catwalks and floor weeds.

**Video**
https://github.com/user-attachments/assets/a3f65e0d-727f-46df-aea5-e61c62faec49

Rainbeon:
## About the PR
Fixed Xeno Projectiles firing at random objects on the floor like weeds, catwalks, and other entities, instead of where you are aiming.

## Why / Balance
This caused your projectiles to aim at random things like weeds on the ground, instead of where your mouse is aiming, now it will always aim where you tell it to, like how guns do.

Resolves #4692 

## Technical details
Nuke TryShootAt() in favor of just using TryShoot(), converts all calls of the former into TryShoot().

## Media
https://github.com/user-attachments/assets/78ecf11c-a2d8-49c2-bfdf-6dff7bc5fee5

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Xeno Spit Projectiles and Runner Bone Chips not shooting where you are actually aiming.


DrSmugleaf (approved):
Thanks for the PR!

PR #4694, opened 2024-11-03
Fix no melee damage sound not working

Dutch-VanDerLinde:
before, melee weapons would make the hit sound even if the damage didn't do anything
this was broken because melee weapons used the modifiedDamage instead of the applied damage, which is never zero

this should be fixed because it can be confusing to hear the hit sound when you aren't doing any damage, especially with things like xeno claw sharpness where you don't know that you're doing damage unless you look in the code


DrSmugleaf (approved):
Thanks for the PR!

PR #4703, opened 2024-11-04
Un-predict Shove, Tail Stab, Paralyzing Slash, and Crippling Strike popups

Rainbeon:
## About the PR
Makes the pop-ups for paralyzing slash, tail stab, tackling, and crippling strike run only on the server to prevent mis-predictions.

## Why / Balance
Melee/Shove prediction is broken and the client can't be trusted with giving you information. Having a mis-predicted popup can cause very negative consequences (mis-predicted shoving on Runner can cause you to drop a grab, mis-predicted Para Slash can cause you to not go for another that will actually work or play more passive as you wait for the stun, only for it to never come out). The cooldown for Tail Stab is already not predicted and it also frequently lies so did the same for it with the sound and the popup

## Technical details
Convert the 4 relevant PopupClient()s to a server-only PopupEntity().
Convert Tail Stab's sound from PlayPredicted() to a server-only PlayPvs().

## Media
Shove, Paralyzing Slash, Crippling Strike

https://github.com/user-attachments/assets/5b6fda94-c85f-40ca-a3b9-2e4b73f2a2a1

Tail Stab

https://github.com/user-attachments/assets/99a89e44-681b-4c91-a2f1-ec05ad3ac7a9

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Popups for Shoving, Tail Stab, Sentinel Paralyzing Slash, and Lurker Crippling Strike are no longer predicted. This means the message will no longer mispredict and lie to you about their effects happening, but will appear later than they used to on successes.
- The sound for Tail Stab is no longer predicted, the correct hit or miss sound will now always play, but may play delayed depending on your ping.

### DIFF
--- a/Content.Shared/_RMC14/Tackle/TackleSystem.cs
+++ b/Content.Shared/_RMC14/Tackle/TackleSystem.cs
@@ -51,7 +51,9 @@ public sealed class TackleSystem : EntitySystem
         if (recently.Current < tackle.Threshold)
         {
             _adminLog.Add(LogType.RMCTackle, $"{ToPrettyString(user)} tried to tackle {ToPrettyString(target)}.");
-            _popup.PopupClient(Loc.GetString("cm-tackle-try-self", ("target", target.Owner)), user, user);
+
+            if (_net.IsServer)
+                _popup.PopupEntity(Loc.GetString("cm-tackle-try-self", ("target", target.Owner)), user, user);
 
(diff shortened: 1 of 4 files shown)


Rainbeon:
Un-predicted Tail Stab as well now

Kadeo64:
finally the game no longer fucking lies to me when i'm playing lurker

DrSmugleaf (approved):
Thanks for the PR!

PR #4747, opened 2024-11-07
Add IMP Ammo Rack

Vermidia:
## About the PR

Title.

## Why / Balance

parity, be an ammo carrier or something.

## Technical details

YML. Speed is completely guessed - move delay in cm is 0.4, someone else will need to test how much it actually slows someone down. Currently made it 0.7 between heavy and medium slowdowns. It does slow you while in hand aswell, and I don't see anything that would skip backpack logic so it does the doafter too.

Clothing storage visual sprites ported but unused because that'd probably needs a system or something. CM doesn't technically use any but the empty one but the full one looks better so I went with it.

## Media

https://github.com/user-attachments/assets/25b66b80-c507-438a-9c07-c236ccadb8f0

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Add the IMP Ammo Rack, available from req in small quantities. It can carry 3 ammo boxes, at the cost of a backslot and slowdown no matter how you carry it.


DrSmugleaf (approved):
Thanks for the PR!

PR #4752, opened 2024-11-07
Accuracy, PBs, dual-wielding modifiers

### LINKED ISSUE #702: Order: Focus

Siigiil:
## About the PR
This implements the accuracy mechanic, point-blank shots and the correct modifiers for dual-wielding weapons.

Most projectiles have an accuracy value. When they would hit a creature, modifiers — such as the target's evasion — are applied and a roll of 1 to 100 is made. If the roll is equal to or less than the final accuracy score, the shot hits. Otherwise, it passes through the target and does not damage it.

The hit chance is calculated in the following order:
- The projectile's base accuracy is multiplied by the firing weapon's multiplier.
- The focus order flat bonus is applied. This is +1.5 per leadership level. Yes, this is shit.
- Any falloff from range is subtracted and the focus order distance bonus is applied. This is +0.35 per leadership level per tile. Yes, this is *also* shit.
- -15 if there is no direct line of sight between the target and where the projectile was fired from.
- The target's FF evasion is subtracted if it and the shooter are friendlies. The taser's projectiles ignore FF evasion.
- The target's evasion is subtracted.
- If at this point accuracy is less than 5, it is set to 5.
- A random number between 1 and 100 is rolled. If this number is equal to or less than the final accuracy score, the projectile hits.

Evasion is calculated as follows, in no particular order:
- Creatures have a base evasion value. This is currently 0 for all of them except the lesser and normal drones, which have 10 and 20 respectively.
- FF evasion has a base value of 15 for all creatures.
- Small creatures — facehuggers and larvae — get a +10.
- Big creatures — boilers, carriers, crushers, hivelords, praes, aboms, queens and ravs — get a -10.
- Creatures affected by the movement order get +5 per the order issuer's leadership level.
- Prone creatures and resting xenos get a -15.
- The scout cloak and the sniper's ghillie suit both give +1000 to FF evasion when active.

A "PB" is a point-blank shot. If you click directly on a creature while in melee range of it with most guns, you will PB it instead of making a normal attack.
All projectiles fired in a PB are guaranteed to hit. They only hit the target you clicked on, even if there are others between you. If the target is prone, it takes +20% damage from the PB.

Dual-wielding no longer reduces the damage of rifles and SMGs. It now reduces the weapons' accuracy multiplier by -0.6.. It also increases scatter by 16 degrees. This applies to all weapons that can be dual-wielded together.

Closes #702

## Why / Balance
Porting.

## Technical details
Not sure how to make CM-style dual-wielding without prediction fucking with it, so for now I just implemented CM-style modifiers for it.

## Media

https://github.com/user-attachments/assets/fc0ffe00-2c8b-429e-9c5a-6c686bd2ab1d

## Requirements

## Mechanical differences from CM
No accuracy and scatter penalties for movement. Implementing this would be a bit scuffed, especially with it applying for half a second after you stop moving. Additionally, in CM it only actually applies to one thing that is already bad without it: onehanding larger guns without appropriate attachments. If we really want to further disincentivise this, we can just jack up the unwielded penalties for those guns.

No randomisation in the dual-wielding accuracy penalty. Originally, it rolled a random number between 5 and 7. I instead just took 6 as the average. The random rolls are pointless and this lets me apply this modifier when the weapon is picked up instead of on every shot.

No randomisation of projectile accuracy. Same reason as for damage variance - it really doesn't change the number much and is just a pointless dieroll made for every single shot.

No proper burst PB implementation, it functions the same as normal PBs here, as the main changes to burst PBs on CM are reliant on the autotargeting that we don't have here.

You can PB on full-auto here. Since full-auto has the same rate of fire as semi-auto, the only real effect of this is that you don't have to spam click. No idea why CM doesn't have this.

**Changelog**

:cl: Sigil
- add: The accuracy mechanic has been implemented. Projectiles can now miss creatures and pass through them without doing damage. Accuracy depends on the specific projectile and the firing weapon's multiplier. The latter can be seen by examining any given weapon. Accuracy decays rapidly past a certain distance.
- add: Evasion implemented. It is subtracted from any impacting projectiles' accuracy. The spec invisibility items now grant a massive bonus to evasion against friendly fire. Tasers ignore FF evasion.
- add: A new attachment is now publicly available: the red-dot sight. It improves the accuracy multiplier of the weapon it's attached to. The exact values can be seen by examining it. Some other attachments also affect accuracy. Examine them to see the effects.
- add: THIS IS A PUBLIC SERVICE ANNOUNCEMENT: THE FOCUS ORDER IS BAD. DO NOT USE IT. The focus order is now available to any marines able to use orders. It slightly improves the accuracy of all projectiles fired while under its effect.
- add: PBs have been implemented. A "PB" is a point-blank shot. You need to click directly on the target creature in melee range of you to make a PB. A PB always hits the target, ignoring its evasion, the shot's accuracy and any creatures between the shooter and the target. If the target is prone, it takes +20% damage from the PB. Note that you can't PB yourself on full-auto and will instead shoot normally.
- tweak: Dual-wielding weapons in the same weapon group now worsens their scatter and accuracy. Dual-wielding weapons no longer affects their damage.
- tweak: The mobility order buffs the evasion of any marines under its effect.
- fix: The laser sight is now available in the SG gear rack.


Carolyn3114:
HORRRY SHIIIIT

DrSmugleaf (approved):
thank fucking god

DrSmugleaf (changes requested):
Regular drone is supposed to have 20 evasion
Other than that, in 14 you can disable screenshake which brings back the dual smg problem since that allows you to reliably pb, which is impossible in 13 with 2 smgs
You can also shoot yourself way faster than you can in 13, you can do it at regular firerate here which is a lot more delayed in 14, we probably don't want being able to shoot yourself at all since its pixel movement

Siigiil:
> Regular drone is supposed to have 20 evasion

Fixed.


> Other than that, in 14 you can disable screenshake which brings back the dual smg problem since that allows you to reliably pb, which is impossible in 13 with 2 smgs

It's doable in CM. The screenshake *exists*, but it really isn't much of an issue unless you're running around with both stocks extended for some reason. You still need to stay directly adjacent to the xeno, while here the dual SMG thing worked at a slight to moderate distance as well.

Can make it impossible to PB while dual-wielding specific weapon groups if you're *sure* you want me to.


> You can also shoot yourself way faster than you can in 13, you can do it at regular firerate here which is a lot more delayed in 14, we probably don't want being able to shoot yourself at all since its pixel movement

Fixed. Made it so you can't shoot yourself while on full-auto.

Though if someone is *actually* swinging their mouse through the centre of the viewport while actively firing on full-auto, the resulting self-PBs will probably save multiple nearby marines from getting lead poisoning.

DrSmugleaf (dismissed):

Siigiil:
Force pushed because rebase. No changes from reviewed point other than fixing the conflicts.

PR #4755, opened 2024-11-08
Lurker attack reset

Rainbeon:
## About the PR
Ports Lurker's melee attack reset when activating Crippling Strike.

## Why / Balance
CM13 Parity.

## Technical details
Adds a Component named MeleeResetComponent, currently only used in XenoCripplingStrikeSystem.

Subscribes SharedRMCMeleeWeaponSystem to all 3 attack events, running before the upstream version.
MeleeResetInit() is a function called externally on startup (request of Smug), it readies your melee's attack immediately once called, and stores the original time you would've been able to next melee for the purpose of disarm (you are not allowed to use your reset for tackling).

TryMeleeReset() removes the comp so that the melee reset is consumed when attacking, and if a disarm is attempted, it prevents you from using your reset on disarming by reverting the reset timer.

## Media
https://github.com/user-attachments/assets/128e2361-68e7-468c-9003-e4ad7741bcf6

https://github.com/user-attachments/assets/77963e98-d04f-4584-88a7-b64d99626a12

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Ported Lurker's Crippling Strike Attack Reset from CM13. Whenever the Crippling Strike ability is used, you can melee attack again instantly, regardless of when you last meleed. This lets you melee twice in quick succession.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Melee/MeleeResetComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Melee/MeleeResetComponent.cs
@@ -1,0 +1,11 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Weapons.Melee;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(SharedRMCMeleeWeaponSystem))]
+public sealed partial class MeleeResetComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public TimeSpan OriginalTime;
+}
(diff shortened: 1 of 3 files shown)


Rainbeon:
For some reason Heavy Attacks specifically refuse to animate with this. 
If SharedRMCMeleeWeaponSystem calls !_timing.IsFirstTimePredicted, then MeleeWeaponSystem.Effects will fail its same check regardless of how much time has passed.

The attack will work perfectly normally, but DoLunge() will think it's been predicted before and refuse to play. This specifically affects Heavy Attacks, and not Light Attacks. Oddly, only DoLunge() will fail its call, but ClientHeavyAttack (which comes before this) will not.

Rainbeon:
Okay somehow the original problem solved itself at some point so we don't even need to care about the last part, it now works fully

PR #4757, opened 2024-11-08
revert color change of Almayer channel

QuietlyWhisper:
## About the PR

Revert the changes to the Almayer channel color, as a result of [accessibility feedback](https://discord.com/channels/1168210010233376858/1302544720874045520)* 
(*after the chatbox was shaded to be darker)

**Changelog**

:cl: Whisper

- tweak: Reverted color change of Almayer channel.


DrSmugleaf (changes requested):
If this is for color blindness then a new setting needs to be made for it, we are not changing the color back specially after changing the chat color to be darker

PR #4766, opened 2024-11-08
Fix larva bursting animation looping

### LINKED ISSUE #4673: Burst animation can repeat after death
## Description

yeah.... no idea how

**Reproduction**

**Screenshots**

**Additional context**

Dutch-VanDerLinde:
resolves #4673

old code set the appearance data of the victim infected component as victim infected component was getting deleted, which caused issues

made it use a similar system to ``XenoEggSystem`` and its states

makes TryBurst run on both server and client instead of just server


Dutch-VanDerLinde:
I have no idea why but if an AI larva bursts instead of a player one it doesn't update the bursting sprite

DrSmugleaf (approved):
Thanks for the PR!

PR #4776, opened 2024-11-09
Fixes UGL unloading

Siigiil:
## About the PR
Makes it so you can extract grenades from the UGL when it's not attached to a gun. Makes it so you can't extract ammo from breechloaders when the breech is closed.

## Why / Balance
Bug.

## Technical details
The MOU and several other guns have a similar issue where you can't extract the ammo they spawn with through RMCAmmoEject. This is because BallisticAmmoProvider is a pile of shit. I can probably work around it without editing upstream files, but it'll be obnoxious because I'll need to figure out how to get it to not have weird interactions with shell handfuls. Leaving it for later.

## Requirements

## Breaking changes
None.

**Changelog**

:cl: Sigil
- fix: You can now extract grenades from the UGL even when it's not attached to a gun.
- fix: You can no longer extract grenades from the UGL when the breech is closed.


DrSmugleaf (approved):
Thanks for the PR!

PR #4781, opened 2024-11-09
Improve Warrior Lunge Hit Reg, unpredict Fling + Lunge CC, Fling CD

Rainbeon:
## About the PR

* Improved Warrior's Lunge hit registration at close range, using similar code to #4529. Checks if the target you clicked on is close enough when starting the Lunge so that you don't just uselessly fly through them.

* Unpredicted Lunge stun. It will now appear to happen slightly later, but will never incorrectly show the stun visual. Prior to this it was VERY problematic, as if Lunge mis-predicted its stun then the person you used it on would appear to be knocked down for the rest of the game.

* Unpredicted basically the entirety of Fling for the following reasons/purpose:

1. The stun of Fling made the target appear to just fall over for half a second if mis-predicted, giving incorrect feedback to the Warrior player about hitting their ability.

2. The cooldown of Fling on high ping previously could mispredict itself to go on cooldown, and then not refund it when the server declines the fling. This made it to where your Fling cooldown could just be consumed entirely by nothing.

All of these mis-prediction issues are VERY common, even on incredibly low levels of ping.

## Media
Video Demonstration on various levels of lag (~average West Coast/EU connection for first half, ~375ms of lag for second half to showcase edge cases of insane lag)

https://cdn.discordapp.com/attachments/1168210011823026271/1304872256026972212/2024-11-09_13-06-01.mp4?ex=6730f87d&is=672fa6fd&hm=52bb37556f962a8b4dce90f60ac1349f7fb67a3dedcc47c06e94bb5409a4002f&

## Requirements

- [X] I have tested all added content and changes.

## Technical details
Make Lunge/Fling stuns and Fling's cooldown to only run on the server, not the client, using _net.IsServer checks.

Moved code in OnXenoLungeHit() to ApplyLungeHitEffects(), and calls it from there, so that this code can be re-used.

When using the Lunge ability, it now checks for targets very close to it using _physics.GetContactingEntities(), similar to how Runner, Lurker, and Parasite leaps now do. Unlike them, it only checks for the specific target, due to being a click-targetted ability.

## Breaking changes

**Changelog**

:cl:
- fix: Warrior Lunge will now register hits on Marines that are too close or sharing the same tile as them, instead of making the Warrior uselessly fly through them.
- fix: Warrior Lunge's knockdown and stun is no longer predicted. It will no longer lie about hitting when it actually missed, and no longer cause Marines to appear on the floor for the rest of the game when it fails, but will appear to happen slightly later on a successful hit.
- fix: Warrior Fling is no longer predicted. It will no longer consume it's cooldown on a mis-predicted hit, and will not appear to knock Marines down when the ability actually failed.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Fling/XenoFlingSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Fling/XenoFlingSystem.cs
@@ -43,10 +43,12 @@ public sealed class XenoFlingSystem : EntitySystem
         if (attempt.Cancelled)
             return;
 
-        args.Handled = true;
-
         if (_net.IsServer)
+        {
+            args.Handled = true;
             _audio.PlayPvs(xeno.Comp.Sound, xeno);
+        }
+            
 
         var targetId = args.Target;
         _rmcPulling.TryStopUserPullIfPulling(xeno, targetId);
@@ -64,10 +66,12 @@ public sealed class XenoFlingSystem : EntitySystem
         var length = diff.Length();
         diff *= xeno.Comp.Range / 3 / length;
 
-        _stun.TryParalyze(targetId, xeno.Comp.ParalyzeTime, true);
-        _throwing.TryThrow(targetId, diff, 10);
-
         if (_net.IsServer)
+        {
+            _stun.TryParalyze(targetId, xeno.Comp.ParalyzeTime, true);
+            _throwing.TryThrow(targetId, diff, 10);
+
             SpawnAttachedTo(xeno.Comp.Effect, targetId.ToCoordinates());
+        } 
     }
 }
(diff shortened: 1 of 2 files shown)


Dutch-VanDerLinde:
does this make the warrior lunge not miss like CMSS13?

Rainbeon:
> does this make the warrior lunge not miss like CMSS13?

No it just makes it hit at close range

Also I'm pretty sure Warrior Lunge can miss, because I've definitely missed it in the recent past

Dutch-VanDerLinde:
> > does this make the warrior lunge not miss like CMSS13?
> 
> No it just makes it hit at close range
> 
> Also I'm pretty sure Warrior Lunge can miss in 13, because I've definitely missed it in the recent past

Yeah, i think the reason why it 'misses' here is because i was too close, which this pr fixes

Binbagg:
goat

lanse12:
> > does this make the warrior lunge not miss like CMSS13?
> 
> No it just makes it hit at close range
> 
> Also I'm pretty sure Warrior Lunge can miss in 13, because I've definitely missed it in the recent past

you can miss it in cm13, but only if the target walks out of the range or goes behind a wall
here on the other hand marines can sidestep it and you'll go past them probably hitting the marine behind them instead

DrSmugleaf (approved):
Thanks for the PR!

PR #4788, opened 2024-11-10
Medical reactives parity change

N0lim:
## About the PR
Massive medicine reagents parity change. Overdoses are worse. Ethylredoxrazine removes ethanol.

## Why / Balance
Parity 

## Technical details
ChemHealEyeDamage have Amount field, if value is positive then deals damage to eye

## Media
Tricordrazine overdose blindness and Imidazoline heal

https://github.com/user-attachments/assets/20cdc90b-a781-4f55-88b4-767435d3bb77

Dylovene overdose sleepness

https://github.com/user-attachments/assets/d271c1ef-123e-411b-b3c2-4189af8d3319

Epinephrine overdose hallucinations and Dylovene hallucination removement

https://github.com/user-attachments/assets/64f4e4bd-74bb-4d40-9f9d-f55b837cced9

## Requirements

**Changelog**

:cl:
- tweak: Massive medicine reagents parity change. Medicines treat better. Overdoses are worse. Ethylredoxrazine removes ethanol.

### DIFF
--- a/Resources/Prototypes/_RMC14/Reagents/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Reagents/medicine.yml
@@ -133,7 +133,7 @@
   overdose: 15
   metabolisms:
     Medicine:
-      metabolismRate: 0.025
+      metabolismRate: 0.03
       effects:
       - !type:EqualHealthChange
         damage:
--- a/Resources/Prototypes/_RMC14/Reagents/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Reagents/medicine.yml
@@ -343,9 +343,9 @@
         conditions:
         - !type:RecentlyDefibrillated
         damage:
-        - Brute: -40
-        - Burn: -40
-        - Toxin: -40
+        - Brute: -20
+        - Burn: -20
+        - Toxin: -20
       - !type:AdjustReagent
         conditions:
         - !type:RecentlyDefibrillated
--- a/Resources/Prototypes/_RMC14/Reagents/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Reagents/medicine.yml
@@ -311,7 +311,12 @@
         - Brute: -16
         - Burn: -16
         - Toxin: -16
-        - Asphyxiation: -16
+      - !type:HealthChange
+        conditions:
+        - !type:RecentlyDefibrillated
+        damage:
+          types:
+            Asphyxiation: -16
       - !type:AdjustReagent
         conditions:
         - !type:RecentlyDefibrillated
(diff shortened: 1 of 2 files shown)


Siigiil:
> Accelerating the metabolism of medications.

CM life ticks happen every two seconds, as opposed to once every second here. The current metabolism values are correct.

N0lim:
Okay, I need some time for reading code of CM13
I will write notes and calculations below

Rainbeon:
> CM life ticks happen every two seconds, as opposed to once every second here. The current metabolism values are correct.

I think this may be worth investigating further in-game to tell concretely. We had previously thought Xeno life ticks were *also* every 2 seconds but they actually weren't, or weren't in the way we thought (which is why we un-halfed Xeno healing already), could very well be the case here too.

N0lim:
@Siigiil 
According to the code, the default reagent metabolism is 0.1 per second, doesn't this mean that we can get values ​​without dividing by 2?
https://github.com/cmss13-devs/cmss13/blob/055c4127d76facec463411169bd2bc513980b0e1/code/__DEFINES/chemistry.dm#L38

Vermidia:
Some notes:
Epi does have a low overdose effect: It does toxin damage, wiki is lying as usual
Chems in CM drain 0.2 every 2 seconds testing in game

Siigiil:
> @Siigiil According to the code, the default reagent metabolism is 0.1 per second, doesn't this mean that we can get values ​​without dividing by 2? https://github.com/cmss13-devs/cmss13/blob/055c4127d76facec463411169bd2bc513980b0e1/code/__DEFINES/chemistry.dm#L38

That macro is just the default value. It's 0.1 per second, yes. Which is the same as our current value for applicable chems. So yes, when taking numbers from the code, you can divide the amount metabolised by the number of seconds and get the correct number.

The values you took are from the wiki and are listed per life tick, not per second, because when I was reworking that page they were listed like that in the code.

N0lim:
Huh, dexalin has an OXG greater than 3, which means that it removes all airloss, what is the point of dexalin plus then, is it better to remove lexorin?

N0lim:
Mmm, I understand, potency is half the value
https://github.com/cmss13-devs/cmss13/blob/0108b016bf98ca77b8698d7bcccf81b56848a492/code/modules/reagents/Chemistry-Reagents.dm#L181

DrSmugleaf:
> Huh, dexalin has an OXG greater than 3, which means that it removes all airloss, what is the point of dexalin plus then, is it better to remove lexorin?

The actual potency is halved before being processed

Vermidia:
Ignore, all PRs are failing that rn

PR #4795, opened 2024-11-11
Fix chemistry chem metabolisms + parity hunger

Vermidia:
## About the PR

Changed all the basic chems metabolisms to match their effects and colors on CM, if they had any, and removed any upstream fluff that wasn't nessary.

Also changed humanoids to use hunger values from CM, and modified nutriment to match CM in how much it gives.

## Why / Balance

Parity.

## Technical details

The most suspect is the nutrition drain of iron. I saw that the code scales satiate with metabolism rate, so not only is it scaled with our max hunger, it's also x10 to match the metabolism rate. Testing in CM the conversion matches for blood and nutrition.

All damage rates are 1/2 like the medicines so they can process over 0.1 secs correctly.

## Media

color
(Note iron and lithium fixed later, got their colors confused with their fire colors)

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed metabolism rate and properties of basic dispenser chems. Most dispenser chems are shown as unknown on the health analyzer.
- fix: Iron correctly heals more blood but also requires nutrition and drains it as it processes.
- fix: Nutriment heals blood if you're not hungry
- fix: Modified Prepared MRES and components to intended values

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Food/food.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Food/food.yml
@@ -6,16 +6,23 @@
   desc: reagent-desc-nutriment
   physicalDesc: reagent-physical-desc-opaque
   flavor: nutriment
-  color: "#24591F"
+  color: "#664330"
   metabolisms:
     Food:
+      metabolismRate: 0.1
       effects:
       - !type:SatiateHunger
-  plantMetabolism:
-  - !type:PlantAdjustNutrition
-    amount: 1.5
-  - !type:PlantAdjustHealth
-    amount: 0.75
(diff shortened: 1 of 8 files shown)


Vermidia:
I think I need to look at the conversion rate more, doesn't match up with the medicines right

Vermidia:
Confirmed the values seem to be correct for damage blood etc.

Vermidia:
Actually, going to try unscaling the hunger stuff and just, changing our nutrition stuff to match CM.

DrSmugleaf (approved):
Thanks for the PR!

PR #4810, opened 2024-11-13
Change Ignored Prototypes remove upstream chem combinations

Aurallianz:
## About the PR
Removed upstream chemical combinations; this may cause issues; please report them.

## Why / Balance
Admin Issues

**Changelog**
:cl:
- remove: Removed upstream chemical combinations

### DIFF
--- a/Resources/IgnoredPrototypes/cm_ignoredPrototypes.yml
+++ b/Resources/IgnoredPrototypes/cm_ignoredPrototypes.yml
@@ -2,3 +2,13 @@
 - /Prototypes/Traits/neutral.yml
 - /Prototypes/Traits/disabilities.yml
 - /Prototypes/game_presets.yml
+- /Prototypes/Recipes/Reactions/medicine.yml
+- /Prototypes/Recipes/Reactions/botany.yml
+- /Prototypes/Recipes/Reactions/chemicals.yml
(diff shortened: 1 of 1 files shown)


Rainbeon:
Mess Tech needs the drinks.yml reactions as many of them are shared between upstream and CM13

Aurallianz:
This section should be removed. Most of our drinks are not upstream drinks. They are custom RMC drinks, so someone must add custom options for RMC if they need to be included, we should not be using the upstream ones.

QuietlyWhisper:
Probably not a good idea unless someone wants to recreate the ENTIRE baystation chef content because I'm fairly certain they're just using their upstream's content for that

QuietlyWhisper:
TLDR this will result in an extremely long delay for mess tech release if we don't make a compromise

Aurallianz:
We could keep the drinks; the only issue is that many of those recipes don't work anyway, and if they did, we should probably copy them out and put them in a separate rmc-controlled drinks.yml file. I didn't want to go through and remove each recipe individually and then check to see if it worked, mostly because I do not like screwing around with upstream files.

Aurallianz:
I was also working on mess-tech, but someone else took it over. I'm not sure who it was, but they should definitely be auditing this list then.

PR #4832, opened 2024-11-15
Fix Bipods "randomly" dismounting (Deletes unused Lock Comps when untoggling)

Rainbeon:
## About the PR
Fixes the Bipod sometimes un-bipodding itself on use when it shouldn't. The reason this occurred is that the side locking component was not removed when manually toggling.

This applies to all attachments with lock toggles but most of them are separately hard-coded for some reason (such as scopes) so it isn't noticeable on most.

## Technical details
Creates a function RemoveUnusedLocks() that is called when manually un-toggling an attachment. It removes the attachment from each locked component's list, and if the list is empty afterward, it will correctly remove the component.

## Media

https://github.com/user-attachments/assets/ce9ad35a-a45e-4e3d-9aa5-277dafaa8c28

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Bipods using the wrong deployment direction when being deployed after a manual un-deploy. Bipods should no longer seem to "randomly" undeploy when being fired.

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
+++ b/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
@@ -568,6 +568,33 @@ public sealed class AttachableToggleableSystem : EntitySystem
         Dirty(attachable);
     }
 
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4833, opened 2024-11-15
Fix Client Crash on attachment forced-untoggle + Code Cleanup

Rainbeon:
## About the PR
Fixed a client crash that occurs when being forced to untoggle an attachment by a fail condition (such as shooting the wrong way). 

## Why
This got missed in the last PR as the F5 build has had a bullet-prediction related DebugAssert for the Bipod when turning around while shooting ever since that was added. This crash was found when manually building instead.

## Technical details
Swaps the foreach check in all of the force un-toggles to a for loop that runs in reverse, as Toggle() is actively removing from the list it is iterating over. Foreach() doesn't really work unless it's read-only, which this isn't.

Removed the unneeded RemCompDeferred from each of the CheckUserBreakOns as Toggle() is now the one that handles all component removal.

## Media

https://github.com/user-attachments/assets/14c4b224-c529-49e1-b7fe-080585a33305

## Requirements

## Breaking changes

**Changelog**

No CL

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
+++ b/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
@@ -298,8 +298,9 @@ public sealed class AttachableToggleableSystem : EntitySystem
         if (!args.HasDirectionalMovement)
             return;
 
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
This probably fixes some of the attachment errors I was seeing in-game ty

PR #4843, opened 2024-11-16
Make marines in lockers count as dead, remove xenos seeing through lockers

Dutch-VanDerLinde:
## About the PR
CM13 parity
Marines in lockers don't count for the round end condition anymore, remove ability for xenos to see through containers in hijack

:cl:
- tweak: Xenos can no longer see through containers during hijack.
- tweak: Marines in lockers are now considered as escaped and no longer count towards the xeno major condition.


DrSmugleaf (approved):
Thanks for the PR!

PR #4844, opened 2024-11-16
Make sniper and scout cloaks turn off when the user is the last human left

### LINKED ISSUE #4785: sniper and scout cloaks should disable when the user is the last person left
CM13 parity

Dutch-VanDerLinde:
## About the PR
CM13 parity, resolves #4785

Makes the cloak and ghillie suit turn off and have a 2 hour delay when the user is the last human


Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -763,6 +768,7 @@ private void CheckRoundShouldEnd()
 
             var marines = EntityQueryEnumerator<ActorComponent, MarineComponent, MobStateComponent, TransformComponent>();
             var marinesAlive = false;
+            var marineList = new List<EntityUid>();
```
DrSmugleaf, line 771: This should be a field on the system that gets cleared before iterating the query, for performance

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf (approved):

PR #4879, opened 2024-11-19
Bodybags down you on enter

### LINKED ISSUE #4732: you shouldnt be able to take any actions while in stasis bag
such as taking off your armor
!? real

parity

Vermidia:
## About the PR

Title.

## Why / Balance

Fixes #4732. Not 100% parity. CM doesn't actually force you down when in a bodybag, it just preserves whatever state you were in. Which is a little silly, as it does mean you can just take everything off white "laying down" in a bodybag (how else would you fit)

I think it makes more sense this way, and makes a lil animation coming out of the bag to stand which is neat. To preserve some of CMs handling you don't drop items being put in the bag.

## Technical details

Copied some bits from the buckle code to work with a new comp that downs on entering a container. Pretty simple.

Also made the comp block use attempts of items with the comp. Could be removed maybe, it just prevents some weird stuff like using a nade while resting and such (since resting you're supposed to drop everything but we aren't doing it here)

## Media

https://github.com/user-attachments/assets/d93271a6-e6e8-4148-aded-64d7e32b5c9c

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Bodybags and Stasisbags letting you perform some actions by having them down you on enter (you still don't your drop held items)

### DIFF
--- a/Content.Shared/_RMC14/Standing/DownOnEnterComponent.cs
+++ b/Content.Shared/_RMC14/Standing/DownOnEnterComponent.cs
@@ -1,0 +1,6 @@
+﻿using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Standing;
+
+[RegisterComponent, NetworkedComponent]
+public sealed partial class DownOnEnterComponent : Component;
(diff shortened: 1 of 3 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4880, opened 2024-11-20
Add the M59A prototype rifle (CO's protag gun)

MACMAN2003:
## About the PR

Added the CO's protag gun, alongside the functionality for a gun to have an ID lock and/or an IFF toggle.
Also gave the smart gun the IFF toggle.
## Why / Balance

Parity
I tried to make the stats as close to the CM13 protag gun as possible
## Technical details

I don't truly know how it works, it just does. I will still try my best to give a summary.

Changes to existing stuff:
- GunIFFSystem was given a new public method "SetIFFState" that changes the "Enabled" bool of a given GunIFFComponent.
- RMCSelectiveFireSystem was given a new public method "SetModifiers" that swaps out the "Modifiers" dictionary with a provided one, and then refreshes the fire modes.

New Stuff:
- GunIDLockComponent & System. The component can be added to any gun to make it "owned" by a specific person, only they can fire it. It has an option to "unlock" the gun so that anyone can fire it.
- IFFToggleComponent & System. The component can be added to any gun with IFF, allowing it to be toggled. There are three bools that control how the system operates, "RequireIDLock", "ChangeStats", and "StatChangeOnIFFEnabled" which should not be changed during runtime. ~~The last one I only just now realized is redundant.~~ The former is for synergy with the GunIDLockComponent, requiring authorization from the ID lock before allowing IFF to be toggled.
Regular guns with IFF that allow toggling should not be concerned with changing stats.
## Media

[youtube video because 30mb is too much for github](https://youtu.be/2f0tAAh1634)

## Requirements

## Breaking changes

I do not have any idea if I broke something or not.
**Changelog**

:cl:
- add: Added the CO's M59A prototype rifle.
- add: Gave the ML66A smart gun the option to disable its IFF.


Boaz1111:
test fails are unrelated

Dutch-VanDerLinde:
Probably should add the IFF toggle to the smart gun

DrSmugleaf (approved):
Thanks for the PR!

PR #4893, opened 2024-11-20
Make parasite only die after it falls off from the host

Dutch-VanDerLinde:
## About the PR
Parity

Parasites only begin the infection process after they fall off from the mask
also makes it so the death sound only plays AFTER they detach

https://github.com/user-attachments/assets/f808527a-e7d1-45ed-bb29-35eb1244874e


DrSmugleaf (dismissed):
Does this still correctly ghost the parasite player on hug?

Dutch-VanDerLinde:
> Does this still correctly ghost the parasite player on hug?

yep

PR #4897, opened 2024-11-21
Make pulls rotate the puller

Dutch-VanDerLinde:
## About the PR
Makes the pull effects despawn shorter ( they don't last that long in CM, and if you spam pull it can get clouded)
Make entities auto rotate when they're pulling an object, it rotates to face towards the object

## Why / Balance
Parity , more immersion

## Media

https://github.com/user-attachments/assets/9153311e-f091-42c4-84db-e4794a7aceb4

## Requirements

### DIFF
--- a/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
+++ b/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
@@ -381,26 +381,28 @@ public sealed class RMCPullingSystem : EntitySystem
             _pulling.TryStopPull(uid, pullable);
         }
 
-        var pullerQuery = EntityQueryEnumerator<PullerComponent, TransformComponent, CanFiremanCarryComponent>();
-        while (pullerQuery.MoveNext(out var uid, out var puller, out var xform, out var firemanCarry))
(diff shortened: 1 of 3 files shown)


Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
```diff
@@ -365,5 +380,27 @@ public override void Update(float frameTime)
 
             _pulling.TryStopPull(uid, pullable);
         }
+
+        var pullerQuery = EntityQueryEnumerator<PullerComponent, TransformComponent, CanFiremanCarryComponent>();
```
DrSmugleaf, line 384: This will iterate every entity able to pull even if they are not doing so currently, it would be better for performance to iterate something like <BeingPulledComponent, PullableComponent>, get what they are being pulled by, and run the code on that

DrSmugleaf (changes requested):
Thanks for the PR!

Dutch-VanDerLinde:
made it use a new system that's less performance taxing

DrSmugleaf (approved):
Thanks for the PR!

PR #4898, opened 2024-11-21
SADAR Assisted Reload

Rainbeon:
## About the PR
Ports the SADAR's assisted reload from CM13. When wielding the M5-ATL, another marine can quickly load a rocket into your weapon if they are standing behind you. This takes 3 seconds, as opposed to the standard 6 second self-load.

## Technical details
Creates 3 components, an Ammo, Weapon, and Receiver component. The Ammo component stores the amount of time it takes to load (for flexibility), the Receiver component is attached to the player wielding the rocket and lets the player be interacted with, and it stores the Uid of the relevant weapon. The Weapon component adds and removes the Receiver component to the player wielding and unwielding it.

CMGunSystem handles everything about assisted reloading besides actually inserting the ammunition (which can't be detached from SharedGunSystem). Event handlers handle the components and interactions between the in-hand ammo and the receiver.

TryAssistedReload() is called to attempt an assisted reload with an assisted reload eligible item. It will fail if the required components are not present, the reload fails, or if IsBehindTarget() fails, which grabs the angle of the 2 players, and compares them with the target's facing direction to ensure the loader is within 90 degrees of their backside.

There is an additional whitelisting check that would otherwise be handled by the reloading process itself, in order to give feedback to the player if they attempt to insert the ammo into the wrong type of weapon. As is, this should never trigger, as only the M5-ATL has the components for assisted reload currently, but is established now for the future if other such weapons were to be made, especially for other HvH factions.

One upstream file is minorly changed: SharedGunSystem.Ballistic.cs. All that was changed about this is that the behavior for ammo inserting was made a generic function TryAmmoInsert(), rather than being hard coded into a single event handler. It has also been changed to output a bool, true on success, false on fail, so that Assisted Reload may receive feedback on whether it worked or not. These changes do not affect standard reloading in any way, there should be no behavioral difference between this and live.

## Media

Loader POV

https://github.com/user-attachments/assets/d0500449-0e95-45d8-bb94-6e6467062797

Shooter POV

https://github.com/user-attachments/assets/6aaceccf-09a7-4729-a72c-ee4120ebf9f2

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added the Assisted Reload mechanic. You can now perform an assisted reload on a SADAR (M5-ATL) by clicking on a player that is actively wielding a SADAR while you are behind them. This will load the SADAR in just 3 seconds, as opposed to the 6 second self-load time.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.Ballistic.cs
@@ -91,30 +91,36 @@ public abstract partial class SharedGunSystem
         if (args.Handled)
             return;
 
-        if (_whitelistSystem.IsWhitelistFailOrNull(component.Whitelist, args.Used))
-            return;
+        TryAmmoInsert(uid, component, args.Used, args.User, args.Target, component.InsertDelay);
+        args.Handled = true;
+    }
+
+    public bool TryAmmoInsert(EntityUid uid, BallisticAmmoProviderComponent component, EntityUid ammo, EntityUid loader, EntityUid weapon, double insertDelay)
+    {
+        if (_whitelistSystem.IsWhitelistFailOrNull(component.Whitelist, ammo))
+            return false;
 
         //Prevent primed grenades or other primed ordanance from being loaded into weapons.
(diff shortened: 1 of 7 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4900, opened 2024-11-22
Adjusts the speed of everything.

Vermidia:
## About the PR

Title.

## Why / Balance

So after a bunch of digging and a bunch of math, I was able to easily make a formula and spreadsheet for calculating pull speed slowdowns and speed without need for testing. Why is this important?

Because BYOND only processes every decisecond so smaller adjustments to delays get ignored. It can also make it hard to measure the actual speeds, as you'll see below pull speeds in general have lowered greatly (except marines) while general speeds have risen slightly. The sprint numbers are very similiar to what you can get testing it out, it's just free from byonds tickrate.

The formula is optional for getting pull speeds as in game I have tested grabbing a mob and checking pull delay on the puller, and it matches up. so the way I get speed also matches and alot of things are actually really slow at pulling.
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed speeds across the board. Pull speeds in general have fallen greatly. Marines also now get slowed pulling bodybags. Xeno walking speeds are more accurate aswell. Ability speeds have been adjusted to stay accurate to the changes.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Crest/XenoCrestComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Crest/XenoCrestComponent.cs
@@ -14,7 +14,7 @@ public sealed partial class XenoCrestComponent : Component
     public int Armor = 5;
 
     [DataField, AutoNetworkedField]
-    public float SpeedMultiplier = 0.70f;
+    public float SpeedMultiplier = 0.706f;
 
     [DataField, AutoNetworkedField]
     public string[] ImmuneToStatuses = { "KnockedDown" };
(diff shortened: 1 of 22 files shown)


DrSmugleaf (changes requested):
From testing the Lurker alone, pull speed is incorrect
Not sure about the rest

Rainbeon:
These are vastly incorrect, these are way too slow and I can see why, since you have merged Marine and XenoLight pull.

The reason we have manually measured out of all these pull speeds is *because* automatic conversions fail due to rmc14 specific mechanics. This is because *objects with mass resist pulling, and objects with higher mass resist pulling greater than lower mass*. Xenos have more mass than marines, hence the need to make XenoLight and Marine pull speed modifiers entirely separate in the first place. Whatever automatic conversion is going on here is not accounting for this pull friction.

Vermidia:
100% fair I didn't consider mass so I undid all the pull speed changes. This does make some xenos much faster so some will probably need pull stuff adjusted. Will have to work on that later.

PR #4912, opened 2024-11-22
Fix xenos not rotating when pulling an object

Dutch-VanDerLinde:
Gets fireman query, only continues if the fireman component actually exists


DrSmugleaf (approved):
Thanks for the PR!

PR #4917, opened 2024-11-23
Full shove input parity

Rainbeon:
## About the PR
Missing a shove now puts your shove on a 0.6 second cooldown, instead of full
Missing a shove now shows a visual effect
You can no longer attempt to shove far away objects

## Technical details
The client now makes a check when shoving that will cancel the shove before sending it to the server if the target is too far away.

The cooldown of a shove on a missed shove is set to 0.6 seconds.

## Media

https://github.com/user-attachments/assets/6c9a0329-0f58-4469-b4ec-58943154311a

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Re-enabled tackle cooldown reduction on a miss. Missing a tackle will now put it on a 0.6 second cooldown, instead of a full-length cooldown.
- fix: Fixed no visual effect occuring on a tackle miss.
- fix: Fixed being able to try to tackle a very far away object that you would otherwise always fail to tackle.


DrSmugleaf (approved):
Thanks for the PR!

PR #4931, opened 2024-11-24
Fix humanoids being too slow + fix xeno slow amounts

Vermidia:
## About the PR

Title.

## Why / Balance

Should be parity. Humans in CM are as fast as a lurker wearing shoes armorless due to how it works. Slows on them however are much harsher (1 delay for slow, 4 for superslow).

This means that slow would bring them down to boiler speed, and superslow makes them slower than queen. The percents should be about 0.66 for slow and 0.33 for super slow.

Xeno slowdown from ammo is about right already I think (it's never going to be perfect due to how speed is handled in CM rn) so leaving it alone.

All these values should be easy to test (on account of being whole numbers) in CM. I don't trust myself with a stopwatch to get it exactly right though.

Edit: Pull speeds and armor speeds were updated based on some testing information.

## Media

https://github.com/user-attachments/assets/7b442b24-f393-46e8-b980-27b721054eab

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed Humanoids being slower than intended when unarmored. Armor speeds have mostly not changed.
- fix: Slowing effects caused by Xenos should slow marines more (they weren't slowing enough before).
- fix: Fixed IMP ammo rack slowing marines down more than intended.
- fix: Bodybags and stasis bags slow on drag.
- fix: Fix bonechips slowing the wrong amount.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Species/base.yml
@@ -188,18 +188,22 @@
     baseSprintSpeed: 5
   - type: SlowOnPull
     slowdowns:
-    - multiplier: 0.7575
+    - multiplier: 0.745
       whitelist:
         components:
         - Marine
-    - multiplier: 0.8725
+    - multiplier: 0.855
       whitelist:
         components:
         - XenoLight
-    - multiplier: 0.59
+    - multiplier: 0.5775
       whitelist:
         components:
         - XenoHeavy
+    - multiplier: 0.645
+      whitelist:
+        tags:
+        - Bodybag
   - type: MobThresholds
     thresholds:
       0: Alive
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Species/base.yml
@@ -200,10 +200,10 @@
       whitelist:
         components:
         - XenoHeavy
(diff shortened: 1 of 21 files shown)


Rainbeon:
Did you test marine armor speeds alongside this? Because changing their base speeds would affect that too

Vermidia:
I haven't since they already don't quite match up. Light armor should give you a speed equivalent to drone speed, medium and heavy would both give you boiler level speed.
Currently:
Light: Slightly higher than drone level
Medium: Tiny bit higher than boiler level
Heavy: Slightly less than boiler level
(though boiler this is taking into account boiler speed being about 3.5 and not 3.33, which may or may not be closer)

Rainbeon:
I approve of these speed values and modifiers

Resources/Prototypes/_RMC14/Entities/Mobs/Species/base.yml
```diff
@@ -183,20 +183,27 @@
...
         - XenoLight
-    - multiplier: 0.59
+    - multiplier: 0.5775
       whitelist:
         components:
         - XenoHeavy
+    - multiplier: 0.645
+      whitelist:
+        tags:
+        - Bodybag
```
DrSmugleaf, line 206: Commenting this out for now since we can't rollerbed stasis bags yet

```suggestion
//    - multiplier: 0.645
//      whitelist:
//        tags:
//        - Bodybag
```

DrSmugleaf (approved):

PR #4940, opened 2024-11-25
Map checker

TsjipTsjip:
## About the PR

Adds a configurable GitHub workflow that scans RMC14's map files for prototypes that should not be mapped at any time. This is a sub-project that was made in rust cause it's fast. See [the test job](https://github.com/TsjipTsjip/RMC-14/actions/runs/12015204911/job/33492764821) for an example without caches. With caches (which only need to be generated once per update to the source files), the workflow is another 23 seconds faster.

## Why / Balance

Cause a Marine High Command stamp ended up on Kutjevo and *somehow* there was still three CMP stamps mapped on Shiva's which were not abused yet.

## Technical details

All code lives in `.github/map_checker`. The workflow file lives in `.github/workflows`. There's a README.md file in `.github/map_checker` which explains how the `matchers.yml` file can be used to configure this workflow fully. The workflow file itself should generally not be touched unless the binary itself needs updating.

## Media

No.

## Requirements

## Breaking changes

God I hope not.

**Changelog**

No CL, no joy.


TsjipTsjip:
Test fails are not related.

DrSmugleaf (approved):
Thanks for the PR!

PR #4944, opened 2024-11-26
Makes the SG unable to make point-blank shots

Siigiil:
## About the PR
Didn't notice that the SG inherited from the base rifle when making the accuracy PR.

## Why / Balance
Bug.

## Requirements

**Changelog**

:cl: Sigil
- fix: Fixed the smartgun being able to PB.


Vermidia:
I did do that. The PB couldn't hit a marine though so I dont think it was a true PB.

GoodWheatley:
> I did do that. The PB couldn't hit a marine though so I dont think it was a true PB.

PBs respect IFF regardless, so it wouldn’t hit a marine.

Siigiil:
> > I did do that. The PB couldn't hit a marine though so I dont think it was a true PB.
> 
> PBs respect IFF regardless, so it wouldn’t hit a marine.

They don't and it would. Which is intentional.

DrSmugleaf (approved):

PR #4949, opened 2024-11-26
Make correspondent microphone be not an instrument

TsjipTsjip:
## About the PR

New microphone prototype without its instrument components. Correspondent now spawns with this one instead.

## Why / Balance

Cause people generally only use it to be disruptive, and the cases where having MIDI's blared at your speakers is actually good roleplay or useful are very very limited in frequency.

Requested by @PolarTundraa .

## Requirements

**Changelog**

:cl:
- remove: Correspondent microphone can no longer be used to play MIDI files.


Kadeo64:
RIP but also; deserved. MIDIs are a gameplay issue since they can cause a lot of lag on some systems.

I just wish there were other ways to get MIDI instruments, maybe through req, for the occasional scenario when having an instrument would be nice.

DrSmugleaf (approved):

PR #4961, opened 2024-11-27
Add sound for picking up and hanging up a telephone

Dutch-VanDerLinde:
Adds the sound from CM13

https://github.com/user-attachments/assets/2c0ebb16-ff2b-49de-ac44-a96ba81c2c79


DrSmugleaf (approved):
Thanks for the PR!

PR #4971, opened 2024-11-28
Downed Drag Speed Fix + Sound on Successful Tackle

Rainbeon:
## About the PR
Fixed being downed causing Xenos (and Marines) to drag you faster.

Made it so the knockdown sound always plays upon successfully re-applying a tackle, even if the target is already knocked over.

CM13 Parity for both.

## Technical details
Commented out friction modifier when knocked, not needed here.

Made it so the knocked down sound will play (server-side, to prevent prediction errors) if standingState.Standing is false.

## Media

https://github.com/user-attachments/assets/bbf6dd7f-a498-4858-a5f1-6f1ac101e7d7

## Requirements

**Changelog**

:cl:
- fix: Fixed being knocked down causing you to be dragged faster by ~25%. This means Xenos can not drag away Marines as quickly, and Marines can not drag critical or dead Marines as fast either without the assistance of a Rollerbed.
- tweak: The sound for a knock down will now always apply upon a successful tackle, even if the target is already knocked down.

### DIFF
--- a/Content.Shared/Stunnable/SharedStunSystem.cs
+++ b/Content.Shared/Stunnable/SharedStunSystem.cs
@@ -276,7 +276,7 @@ public abstract class SharedStunSystem : EntitySystem
 
     private void OnKnockedTileFriction(EntityUid uid, KnockedDownComponent component, ref TileFrictionEvent args)
     {
-        args.Modifier *= KnockDownModifier;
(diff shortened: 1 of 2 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #4972, opened 2024-11-28
Add gauze and scar markings to Vulpkanin

PolarTundraa:
## About the PR
Made Vulpkanin compatible with gauze/scar markings

## Why 

More customisation!

## Technical details
Just added Vulpkanin to the whitelist for existing gauze/scar markings

## Media
image

## Requirements

:cl:
- add: Allowed Vulpkanin to use gauze/scar markings

### DIFF
--- a/Resources/Prototypes/_RMC14/Species/vulpkanin.yml
+++ b/Resources/Prototypes/_RMC14/Species/vulpkanin.yml
@@ -75,9 +75,6 @@
       points: 1
       required: true
       defaultMarkings: [ VulpEar ]
-    Overlay:
-      points: 2
-      required: false
 
 - type: humanoidBaseSprite
   id: MobVulpkaninHead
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/_RMC14/Species/vulpkanin.yml
```diff
@@ -75,9 +75,6 @@
       points: 1
       required: true
       defaultMarkings: [ VulpEar ]
-    Overlay:
-      points: 2
-      required: false
```
DrSmugleaf, line 80: Was removing this intended?

DrSmugleaf (changes requested):

Resources/Prototypes/_RMC14/Species/vulpkanin.yml
PolarTundraa: Yes, they have 2 existing overlay markings that would set a limit on the amount of bandage markings you could use

no other species (as far as I could tell, anyway) had a limit on overlay markings, so I removed that field entirely to prevent there being a limit on the amount of gauze/bandage markings you can apply

DrSmugleaf (approved):

PR #5027, opened 2024-12-02
Fix button system

dalush:
Now only entities with a SurvivorComponent or MarineComponent can press buttons

## About the PR
 Content.Shared/_RMC14/Doors/CMDoorSystem.cs

## Why / Balance
 bug fixes

## Technical details
 added check for SurvivorComponent and MarineComponent


DrSmugleaf (changes requested):
Is this to fix monkeys using the buttons, or something else? That should only happen if an admin ghost roles one, but either way they are access locked
This would prevent future ERT spawns from trying to press a button at all

PR #5028, opened 2024-12-02
Compact and Tactical Nailgun

CyberTropic:
## About the PR

Added compact and tactical nailguns. These don't fire at all unlike the smg weapon nailgun that survivors get, so they're listed under tools. They use the same magazine for repairs.
Added tactical nailgun and nails to vendors, storage whitelists. Added nailgun ammo to vendors.
Welder cannot repair nailgun repairable barricades that are too damaged (25% of full health in parity)

## Why / Balance

Parity

## Media

image
image
image

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Added compact and tactical nailguns. Tactical nailgun comes with the essential engineer set. These must be used to repair a barricade if it is too damaged, as welders won't be able to anymore.


DrSmugleaf (changes requested):
If we are adding these to CT loadouts I would rather we port the mechanic where highly damaged cades can only be repaired by nailguns and not welders too.
The threshold for that seems to be less than 25% max health

DrSmugleaf (approved):

PR #5045, opened 2024-12-03
Fix auto punctuation breaking with sanitization

Dutch-VanDerLinde:
If you say "lol" with auto punctuation enabled, normally it will come out as "*laughs*", but instead it will come out as "Lol."

this is a simple pr, just makes the smilieys santize run before the message gets auto punctuated

:cl:
- fix: Fixed auto punctuation causing messages like "Lol" to not sanitize


DrSmugleaf (approved):
Thanks for the PR!

PR #5057, opened 2024-12-05
Nerf Frenzy Pulling boost by ~50%

Rainbeon:
## About the PR
If you are a xeno with frenzy and pulling, your frenzy boost gets halved. Far closer to CM13 Parity.

## Why / Balance
It Brings Lurker frenzy pulling into Parity primarily Lurker was travelling at 2.47 TPS whe pulling with T4 Queen Frenzy when it was supposed to be travelling at ~2.23 TPS. Making the modifier 0.03 instead of 0.06 specifically when pulling reaches this breakpoint. Other Xenos may not reach their exact breakpoints but this is the one everyone was attacking me over so this is the one that I used.

## Technical details
Add pulling speed modifier field, set to 0.03, used when pulling, as opposed to the standard 0.06.

SharedXenoPheromoneSystem now calculates speed based on whether or not you are pulling. 2 new events were added to this system to refresh your movement speed modifiers when you start and stop pulling, specifically happening after RMCPullingSystem has handled the events so that the PullingSlowedComponent can be correctly added or removed.

## Media

https://github.com/user-attachments/assets/03d9ebde-2c3b-4a22-999d-5e60a11e226d

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Halved the speed bonus Frenzy gives while pulling to more closely match CM13. This means overall pulling speeds are reduced by ~10% total.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Pheromones/SharedXenoPheromonesSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Pheromones/SharedXenoPheromonesSystem.cs
@@ -1,4 +1,5 @@
 ﻿using System.Linq;
+using Content.Shared._RMC14.Pulling;
 using Content.Shared._RMC14.Xenonids.Plasma;
 using Content.Shared.Actions;
 using Content.Shared.Coordinates;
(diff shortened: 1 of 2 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5074, opened 2024-12-07
Overwatch console now displays marine pronouns

snebl:
## About the PR

The overwatch console now shows the pronouns of each marine under their name.

(The name becoming left-aligned when the marine has no helmet was already a thing, I opted not to change that since its easier to read anyway.)

## Why / Balance

There is no way to examine a marine from their helmet cameras (unless your character is close enough to do so normally) and so you can't find their pronouns during overwatch. Staff officers would either have to always use they/them or make an assumption based on the marine's hair style or color pattern which is not ideal.

With this change it's now much harder to accidentally misgender someone on the ground from lower orbit.

## Technical details

Added a single locale file that `OverwatchConsoleBui.cs` now uses.

## Media

{58AF463E-A209-4312-8996-D7E3634C9958}

## Requirements

## Breaking changes

**Changelog**

:cl: snebl
- add: The overwatch console now displays the pronouns of each marine.

### DIFF
--- a/Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
+++ b/Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
@@ -233,14 +233,14 @@ public sealed class OverwatchConsoleBui : BoundUserInterface
                 if (marine.Camera == default)
                 {
                     var watchLabel = new RichTextLabel();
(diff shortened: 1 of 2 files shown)


snebl:
Do not merge yet, just realized there is a bug where the name column does not line up with the other columns now.

snebl:
branch is COOKED sorry i will make a new PR

PR #5079, opened 2024-12-07
Overwatch console now displays marine pronouns

snebl:
## About the PR

The overwatch console now shows the pronouns of each marine under their name.

(The name becoming left-aligned when the marine has no helmet was already a thing, I opted not to change that since its easier to read anyway.)

## Why / Balance

There is no way to examine a marine from their helmet cameras (unless your character is close enough to do so normally) and so you can't find their pronouns during overwatch. Staff officers would either have to always use they/them or make an assumption based on the marine's hair style or color pattern which is not ideal.

With this change it's now much harder to accidentally misgender someone on the ground from lower orbit.

## Technical details

Added a single locale file that `OverwatchConsoleBui.cs` now uses.

Also makes the vertical margin thickness on the squad table 0 (it looked like that was how it was supposed to be for the buttons but it was overwritten a few lines down, this change is mostly imperceptible but stops the names from misaligning from the statuses on the table)

## Media

{083830FD-B317-4021-92BC-D8E59534ED63}

## Requirements

## Breaking changes

**Changelog**

:cl: snebl
- add: The overwatch console now displays the pronouns of each marine.

### DIFF
--- a/Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
+++ b/Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
@@ -20,6 +20,7 @@ namespace Content.Client._RMC14.Overwatch;
 public sealed class OverwatchConsoleBui : BoundUserInterface
 {
     [Dependency] private readonly IPrototypeManager _prototypes = default!;
+    [Dependency] private readonly IEntityManager _entity = default!;
 
     private const string GreenColor = "#229132";
     private const string RedColor = "#A42625";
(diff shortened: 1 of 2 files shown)


GoodWheatley:
The better solution here is that you should just be able to examine things through helmet cameras, which is how 13 overwatch works.

BramvanZijp:
> The better solution here is that you should just be able to examine things through helmet cameras, which is how 13 overwatch works.

I agree, this is a partial solution to a problem that can be solved in its entirety in a much better way (Allowing examine through OW), as opposed to what feels like a temporary/ducttape solution.

snebl:
@GoodWheatley 
good point, but if this PR is merged it could always be reverted after your proposed implementation is added.

@BramvanZijp 
the point of a duct tape solution is to temporarily solve a problem in wait of a better solution. There isn't any sort of tradeoff here.

mmails:
Good fix but redundant as examining stuff through helmet cams is parity

Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
```diff
@@ -250,16 +251,15 @@ void SendSquadMessage()
                 if (marine.Camera == default)
                 {
                     var watchLabel = new RichTextLabel();
-                    watchLabel.SetMarkupPermissive($"[color={YellowColor}]{name} (NO HELMET)[/color]");
+                    watchLabel.SetMarkupPermissive($"[color={YellowColor}]{name} (NO HELMET)\n{Loc.GetString("rmc-overwatch-marine-pronouns", ("target", _entity.GetEntity(marine.Id)))}[/color]");
```
DrSmugleaf, line 254: I'd rather make examining work through cameras first and then add this later if overwatch still uses the wrong pronouns
For that, ExamineSystemShared.cs line 123 needs to be changed to use the overwatched entity instead of the examiner if in overwatch
Same applies for xenos

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Client/_RMC14/Overwatch/OverwatchConsoleBui.cs
snebl: > make examining work through cameras first and then add this later

So should I make a new PR and leave this open for later?

PR #5125, opened 2024-12-13
Fix pump action shotgun sound

### LINKED ISSUE #5120: Shotgun pump/XM88 lever sound effect will not play if you press the action button too quickly
## Description
As described in the title the sound effect just fails to play if you press the action button rapidly

**Reproduction**
shoot the pump shotgun or xm88 constantly while spamming the spacebar/unique action keybind, they will pump correctly and be made able to fire but the sound effect will not play

https://github.com/user-attachments/assets/225a9fa7-528d-44fa-afa5-928d11d6f2de

doesn't affect gameplay really other than immersion factors and knowing with certainty when exactly you have pumped the weapon

Dutch-VanDerLinde:
## About the PR
This unpredicts the sound, but fixes a bug where if you spam the unique action button, the sound sometimes wouldnt play
resolves #5120

:cl:
- fix: Fixed the shotgun pump or XM88 cycle sound not playing when you spam the unique action button.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
@@ -5,11 +5,13 @@ using Content.Shared.Weapons.Ranged.Events;
 using Content.Shared.Weapons.Ranged.Systems;
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Containers;
+using Robust.Shared.Network;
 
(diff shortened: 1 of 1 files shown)


DrSmugleaf (commented):
I need to check why this is happening to see if we can predict it properly, it's probably something to do with prediction

DrSmugleaf:
I got this fixed in a local branch for the update ty, the problem was with how gun prediction was not causing use delay and pump status to be reset correctly

PR #5148, opened 2024-12-15
Fix movement compensation and wield slowdown values for guns & armor

Vermidia:
## About the PR

Title.

## Why / Balance

parity, keeping with what we do with the armors. The original comp speeds were based on marines going at 4.5 speed.

## Technical details

Wield comp speed for walk/sprint was given the same values (going to do all the walk speeds would take a bit)
for the guns:
light slowdown: 0.725 (same as light)
rifle slowdown: 0.66 (same as medium)
other values: 0.5 (In CM this would be about move delay 4 because all of them are about 1.75, so queen speed.)

for the armor:
light compensation: 0.273 (1 - light armor slowdown)
med compensation: 0.34 (1 - medium armor slowdown)
very heady compensation: 0.429 (This one might be wrong, it's only used for pmc armors and no actual armor slowdowns by this much, it should math out to a marine with no shoes + light armor speed as compensation, probably)

Note: currently messing with med armor and rifle slowdown, as it's too fast and too slow for some armor combos.

feel free to check over them
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed various wield compensation and wield speeds. They should be about the same, though some might be slightly faster/slower

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/base_armor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/base_armor.yml
@@ -202,8 +202,8 @@
       - PillCanister
       - CMSurgicalCase
   - type: WieldSlowdownCompensation
-    walk: 0.273
-    sprint: 0.273
+    walk: 0.275
+    sprint: 0.275
   - type: RMCNameItemOnVend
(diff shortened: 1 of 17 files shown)


Rainbeon:
Implemented fully with system reworks in #5150

PR #5158, opened 2024-12-16
Add steelcrest + defender fixes

Vermidia:
## About the PR

Title. Defender can now attack fortified, but only structures, but can use tail "slam" (just renamed for now) freely on anyone.
Mostly bug fixing otherwise.

Also gave burrow a desc, it didn't have one.

## Why / Balance

Parity!

## Technical details

FortifyComp got a few fields so the fixture changing is opt out and whether they're allowed to headbutt or not, explosion calc was changed to match how it's done in other places. Now checks melee attack so it can lower damage dealt for steelcrest and checking attack was split into its own function since we only want to cancel on mobs.

Headbutt now has it's AP, and knockback and damage change when crested/fortified. Clarified a bunch of descs.

## Media

https://github.com/user-attachments/assets/03fc76ce-c2d0-471a-8f3d-d7761839cd14

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed defender's fortify and toggle crest sharing a cooldown.
- fix: Fixed headbutt not doing reduced damage and having less range with crest lowered, and not having lower knockback normally.
- fix: Fixed headbutt not having slight armor piercing.
- fix: Fixed defender's explosion armor from fortify being applied differently than other similar explosion armor increasing abilities. (Damage from explosions may be slightly higher).
- fix: Defenders can now attack structures while fortified. They can attack mobs using their tail "slam" (tail stab but renamed).
- add: Added steelcrest defender. They can move and headbutt while fortified at the cost of less armor from it.


DrSmugleaf (approved):
Thanks for the PR!

PR #5167, opened 2024-12-16
Actually fix duplicate m5-atl for real

MACMAN2003:
## About the PR

Fixes the duplicate m5-atl that demo specs get, without implementing any extra bugs, hopefully.
## Why / Balance

Bugfix
## Technical details

adds a boolean to the ItemCamouflageComponent that if true will prevent the storage transfer between the spawned entity and the replaced entity. It is false by default so it doesn't interfere with loadouts.
## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Demolition specialists no longer get two M5-ATLs


DrSmugleaf (approved):
Thanks for the PR!

PR #5168, opened 2024-12-17
Fix xeno abilities that move a target not breaking all pulls

Vermidia:
## About the PR

Title.

## Why / Balance

Parity, all the moves I tested that move you just break anyones pull on you in CM. Fixes a few discord bugs

## Technical details

New func in the RMC pulling system to try to call to break any pulls on a target, put it everywhere that's relevant. Also a seperate function to run both this and the previous function. In CM when an ability moves you it breaks every pull on your and every pull your doing. For the fake movement stuff it only breaks pulling from others as it's like your moving.

Also went through and fixed up a few remaining distance calc things.

## Media

https://github.com/user-attachments/assets/5ac92cb6-9c79-41fd-a767-c9bacce95baf

## Requirements

## Breaking changes

**Changelog**

:cl:
- fix: Fixed xeno abilities that move a target not breaking pulls.

### DIFF
--- a/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
+++ b/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
@@ -283,6 +283,17 @@ public sealed class RMCPullingSystem : EntitySystem
         _pulling.TryStopPull(puller.Pulling.Value, pullable, user);
     }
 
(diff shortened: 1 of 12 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5170, opened 2024-12-17
Long Ears

Farrellka-dev:
## About the PR

I removed the old long ears of people and added new ones, more pleasant to look at.
If you want, I can drop it in _RMC14.

## Why / Balance

I want to see these elf ears from my PR wizards here: https://github.com/space-wizards/space-station-14/pull/33572

## Technical details

I just added more variations of long ears. Just pictures, just meta, just prototypes and translations. Nothing special.

## Media

It was before me:
image
An example of one of my ears:
image
More ears here:
image

## Requirements

## Breaking changes

**Changelog**

:cl: Farrellka
- remove: Removed the old long ears for human!
- add: Added more long ears for human!


Resources/Prototypes/Entities/Mobs/Customization/Markings/ears.yml
```diff
@@ -1,10 +1,65 @@
 - type: marking
-  id: HumanLongEars
+  id: LongEarsStandard
```
QuietlyWhisper, line 2: I recommend not changing this ID as it will break any characters already using the marking.

QuietlyWhisper (changes requested):
I'm fine with this change under the condition that it is present as explicitly "long ears" and not elf ears. Whatever player's headcanon is, is fine by me but it is not part of RMC lore.

Alekshhh:
this PR has been denied by true dwarf patriots

Farrellka-dev:
I need to check my PRs more often. 

> I'm fine with this change under the condition that it is present as explicitly "long ears" and not elf ears. Whatever the player's headcanon is, is fine by me but it is not part of RMC lore.

Okay, it's not a problem at all, I don't call it elf ears in the game, it's just the name of the PR.

Resources/Prototypes/Entities/Mobs/Customization/Markings/ears.yml
Farrellka-dev: > I recommend not changing this ID as it will remove the marking from players already using it, but it's not necessary requirement

I don't think those who use these ears will want to play with them when the best alternative is available.

QuietlyWhisper (approved):

DrSmugleaf (approved):

PR #5175, opened 2024-12-19
added Xeno Energy Alert

Kolakolanikolay:
Completely by analogy with plasma alarm, I made the same thing, only for the energy of xeno (for kamikaze runner and warden Praetorian).
Now they will be able to clearly track the amount of their energy.
The plasma system was also slightly modified. So that the alert does not appear when the target does not have plasma (plasmaMax==0). 
It will be useful for the same acid runner. Now he will have a single alert associated with his energy reserve.


Rainbeon:
1) These are called Alerts, not Alarms

2) Test fail is related, please fix and make sure it runs

3) Please include a video demonstration of it working

Kolakolanikolay:
fixed, I also attach a video demonstration

https://github.com/user-attachments/assets/32464d85-b21e-4184-8f64-1aa866086484

Vermidia:
If possible could you have a way to have custom names and descriptions? Aciders energy is called acid reserves, Warden's health reserves etc, they usually aren't called energy by abilities and such.

Kolakolanikolay:
It sounds logical. The only solution I found would be to create a new alert for each case. Added.

Kolakolanikolay:
https://github.com/user-attachments/assets/858c436f-1686-4491-957e-38cbe91f8e24

DrSmugleaf (approved):
Thanks for the PR!

PR #5179, opened 2024-12-20
Larva obligated to Remain in the Hive and not perform Recon

FungiFellow:
## About the PR

Added Rules from 13 Regarding larva, see attached image for details
## Why / Balance

Parity
## Media

image

## Requirements

## Breaking changes

**Changelog**

:cl:
- add: Ported Rules Restricting Larva's allowed zones


FungiFellow:
Test Fails are Unrelated to content included in the PR.

QuietlyWhisper:
This is not enforceable without constantly checking replays of each round, which can only be done after the up to two or three hour round is over. This is a waste of resources.

It is the fault of the xenonid player and not the survivor if they seek out survivors and die because of it.

Additional solutions such as *roundstart larva taking damage off weeds is entirely possible and would be significantly more effective.
 (*Either up to first drop which we have an event for, or after a time limit.)

PR #5187, opened 2024-12-21
Fix warden praetorian heal popup

Dutch-VanDerLinde:
When a warden praetorian healed you, it would say "You are healed by (yourself)" instead of "You are healed by (praetorian that healed you)", this fixes that


DrSmugleaf (approved):
Thanks for the PR!

PR #5190, opened 2024-12-21
Make xeno jobs appear as xeno instead of naked character in the editor

Dutch-VanDerLinde:
Just adds the ``JobEntity`` field to the queen and selectable xeno jobs, like a cyborg and station AI


Dutch-VanDerLinde:
so it's not ported to upstream yet

PR #5214, opened 2024-12-23
Make Resin Whisperer Remote Build LoS A LOT more permissive

Rainbeon:
## About the PR
As long as you have Line of Sight to any part of the tile (either center or any of 4 corners/center of any 4 flat edges), you are allowed to remote build

## Why / Balance
CM13 is VERY permissive due to how sight works there (you can actually build without "direct" line of sight because your LoS wraps around corners by 45 degrees for whatever reason), so this makes it as close as possible to that. This is still more restrictive but it's the best we can get intuitively.

## Technical details
Check 9 different points on the tile to see if you can see it, if so, allow you to build (assuming other checks don't fail). First checks the center of a tile, then spirals outward in a clockwise motion.

## Media

https://github.com/user-attachments/assets/23f0ee03-0a86-417f-ac82-ff66582d7d0e

## Requirements

**Changelog**

:cl:
- fix: Fixed the Hivelord Resin Whisperer's Remote Build being too restrictive on Line of Sight. You now only need to see any tiny part of a tile to remote build on it, similar to CM13.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Construction/ResinWhisper/ResinWhispererSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Construction/ResinWhisper/ResinWhispererSystem.cs
@@ -7,6 +7,8 @@ using Content.Shared.Examine;
 using Content.Shared.Interaction;
 using Content.Shared.Popups;
 using Content.Shared.Verbs;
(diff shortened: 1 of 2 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5233, opened 2024-12-25
Fix every Xeno Ghost Name bug

Rainbeon:
## About the PR
Fixes various bugs relating to Xeno Ghost Names.

## Technical details
Converted the handler for PlayerAttachedEvent into one for MindAddedMessage. PlayerAttachedEvent has inconsistent behavior in interactions with ghost roles, in that it will run after the mind is added if it is a fresh ghost role, but run BEFORE the mind is added if the entity has been a ghost role in the past, which caused it to be unable to perform any actions relating to the mind on subsequent ghost roles (causing you to steal the previous name). It functions exactly the same otherwise, just more consistently.

Stopped handling name changes in hard-coded places and instead condensed it into one handling inside OnRefreshNameModifiers() so that it catches name changes generically, allowing it to catch more easily overlooked changes like the Queen maturing.

Made XenoRoleSystem handle the RefreshNameModifiersEvent before SharedXenoNameSystem, as otherwise XenoNameSystem has no clue what Rank you are when assigning your ghost name. This has to be a before call and not an after call, because the handler in shared can not interact with the server-only handler in XenoRoleSystem.

## Media
Main Fix Demonstration on Discord, channel link below

https://discord.com/channels/1168210010233376858/1168210011823026271/1321408086602158081

Additional Distress Signal round initialization test

https://github.com/user-attachments/assets/9a76b9c6-0412-4e93-949b-99c3bfcc77f6

## Requirements

**Changelog**

:cl:
- fix: Taking a Xeno ghost role that previously belonged to someone else no longer causes your ghost to commit identity theft.
- fix: Fixed the Queen's ghost staying immature forever. It will now update correctly to change into the playtime rank when she matures.

### DIFF
--- a/Content.Server/_RMC14/Name/XenoNameSystem.cs
+++ b/Content.Server/_RMC14/Name/XenoNameSystem.cs
@@ -1,7 +1,6 @@
 ﻿using Content.Server.GameTicking;
 using Content.Shared._RMC14.Xenonids.Name;
 using Content.Shared.GameTicking;
-using Content.Shared.Mind;
 using Content.Shared.NameModifier.EntitySystems;
 using Robust.Shared.Player;
 using Robust.Shared.Random;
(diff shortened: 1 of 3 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5235, opened 2024-12-25
Solution transfer prediction fix

Inconnu1337:
## Why / Balance

Just bugfix
## Technical details

added check in SolutionTransferSystem
## Requirements

**Changelog**

:cl: Inconnu
- fix: Fixed the display of solution transfer.

### DIFF
--- a/Content.Shared/Chemistry/EntitySystems/SolutionTransferSystem.cs
+++ b/Content.Shared/Chemistry/EntitySystems/SolutionTransferSystem.cs
@@ -8,6 +8,7 @@ using Content.Shared.Popups;
 using Content.Shared.Verbs;
 using Robust.Shared.Network;
 using Robust.Shared.Player;
+using Robust.Shared.Timing;
 
 namespace Content.Shared.Chemistry.EntitySystems;
 
@@ -22,6 +23,7 @@ public sealed class SolutionTransferSystem : EntitySystem
     [Dependency] private readonly SharedSolutionContainerSystem _solution = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
 
+    [Dependency] private readonly IGameTiming _gameTiming = default!;
     /// <summary>
     ///     Default transfer amounts for the set-transfer verb.
     /// </summary>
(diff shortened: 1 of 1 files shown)


Content.Shared/Chemistry/EntitySystems/SolutionTransferSystem.cs
```diff
@@ -101,7 +103,7 @@ private void AddSetTransferVerbs(Entity<SolutionTransferComponent> ent, ref GetV
 
     private void OnAfterInteract(Entity<SolutionTransferComponent> ent, ref AfterInteractEvent args)
     {
-        if (!args.CanReach || args.Target is not {} target)
+        if (!args.CanReach || args.Target is not {} target || !_gameTiming.IsFirstTimePredicted)
```
DrSmugleaf, line 106: This usually indicates that state handling isn't being done correctly, in this case in solution code, I'll look into it in a sec

DrSmugleaf (commented):
Thanks for the PR!

Content.Shared/Chemistry/EntitySystems/SolutionTransferSystem.cs
Inconnu1337: Events on the client are called one by one until it synchronizes with the server and here this visual bug happens due to missing predict

It's some engine trick I can't figure out, but the solution I suggested in this pool works correctly

DrSmugleaf: Fixed in https://github.com/RMC-14/RMC-14/pull/5357/files#diff-c675638a225de556eb434b1b2c0c1320b51487717865092b2f7601c65a5aacd5
That also fixes it for chemical dispensers etc

DrSmugleaf:
See https://github.com/RMC-14/RMC-14/pull/5235#discussion_r1920944171 and https://github.com/RMC-14/RMC-14/pull/5357/files#diff-c675638a225de556eb434b1b2c0c1320b51487717865092b2f7601c65a5aacd5

PR #5246, opened 2024-12-27
Digital Clocks

CyberTropic:
## About the PR

Ports digital clock and cat clock. Creates an RMCClockSystem to format our own time display.
Digital clocks were recently added to cm13, and we don't really have a good way to know time. Now the XO can just look at the clock on the almayer table (to be mapped) instead of guessing when the briefing is.
The cat clock is in the medbay waiting room.

## Why / Balance

Parity!
https://github.com/cmss13-devs/cmss13/pull/7844

## Media

clock1
clock2

## Requirements

**Changelog**

:cl:
- add: Added digital clocks that can tell round time!


DrSmugleaf (approved):
Thanks for the PR!

PR #5271, opened 2024-12-30
Parity age requirements

AverageNotDoingAnythingEnjoyer:
## About the PR

Added age requirements for ship-side roles and SL

## Why / Balance

It's parity

## Technical details

Added basic requirement of 20 years for ship-side roles and added job-specific age requirements [from here](https://github.com/cmss13-devs/cmss13/blob/e533b3bb907963ef789a52e21dda811cd3cb8302/code/modules/gear_presets/uscm_ship.dm#L5)
Also made adjustments to age reqs. for MP [ported from here](https://github.com/cmss13-devs/cmss13/blob/e533b3bb907963ef789a52e21dda811cd3cb8302/code/modules/gear_presets/uscm_police.dm#L5) and introduced role-specific age requirements for SL [from here](https://github.com/cmss13-devs/cmss13/blob/e533b3bb907963ef789a52e21dda811cd3cb8302/code/modules/gear_presets/uscm.dm#L438)
Surprisingly there's no specific age requirements for Medbay, so all of them have the req. of 20 years.

## Media

Screenshot 2024-12-30 at 14 23 38
Screenshot 2024-12-30 at 14 23 56
Screenshot 2024-12-30 at 14 24 13
Screenshot 2024-12-30 at 14 24 26
Screenshot 2024-12-30 at 14 24 40
Screenshot 2024-12-30 at 14 24 54

## Requirements

**Changelog**

:cl: J C Denton
- tweak: All ship-side roles and Squad Leader role are now age-restricted.


AverageNotDoingAnythingEnjoyer:
Tested at second commit, forgot to push it before creating this PR

QuietlyWhisper (commented):
Think this would fit in better when/if we have character per role functions, so that you can roll for many roles using all your chosen characters at once, meaning the role you get is associated with the character rather than the other way around

This way you'll still be able to roll for all age requirement roles without having to be ancient

Adding it as is seems unproductive

AverageNotDoingAnythingEnjoyer:
> Think this would fit in better when/if we have character per role functions, so that you can roll for many roles using all your chosen characters at once, meaning the role you get is associated with the character rather than the other way around
> 
> This way you'll still be able to roll for all age requirement roles without having to be ancient
> 
> Adding it as is seems unproductive

Waiting for https://github.com/space-wizards/space-station-14/pull/36493 to come back to this.

Utmanarn:
Question, does this take in the species specific ages? Certain species have differing aging from humans and thus have different ages where they'd be eligable for certain roles. For example, moths have a shorter lifespan but mature faster.

AverageNotDoingAnythingEnjoyer:
That's a good question that needs testing

But if I recall correctly any current RMC species can set their age to whatever they want to, no? Reasoning being the age in-case of non-human species is adjusted in-lore to their real one

Might be wrong ofc

Utmanarn:
> But if I recall correctly any current RMC species can set their age to whatever they want to, no? Reasoning being the age in-case of non-human species is adjusted in-lore to their real one

This may be the case. Not entirely sure how we work with lore and all that.

Dutch-VanDerLinde:
> Question, does this take in the species specific ages? Certain species have differing aging from humans and thus have different ages where they'd be eligable for certain roles. For example, moths have a shorter lifespan but mature faster.

The age you pick in the character editor is in human years. So you'd convert a 20 (human years) year old moth to moth years separately.

None of the species have low lifespans to where this would even matter though. Moths USED to have  lifespan of 6-8 years (Wtf) but it was changed to 59.7 or something.

PR #5278, opened 2024-12-31
Tunnels cant be placed on the alymer

blueDev2:
## About the PR

CM13 Pairity
## Media

image

## Requirements

**Changelog**
:cl:
- fix: Fixed most Almayer tiles being able to have Burrower tunnels placed on them.


MetalSage:
Maybe we should do a check to see if the card has an AlmayerComponent?

blueDev2:
> Maybe we should do a check to see if the card has an AlmayerComponent?

Non Alymayer Tiles also can be untunnelable.

DrSmugleaf (approved):

PR #5290, opened 2025-01-02
 Fix missing names for tacwrap variations

SeamLesss:
## About the PR
Previously, every tacwrap color and variation all had the name of "tactical wrap" with no prefix to indicate the color or squad they represent. This is a very simple fix of adding the color or squad the clothing represents to it's name.

## Why / Balance
The headbands, which act very similarly to the tacwrap, all have an extra word to indicate their color or squad. It should be obvious that the tacwrap should do the same.

## Media
Before:
image

After:
image

## Requirements


DrSmugleaf (approved):
Thanks for the PR!

PR #5291, opened 2025-01-02
Adjust XM88  Hit stack stats

DummeHytteOst:
## About the PR
changes the stats to parity
## Why / Balance
Parity

## Requirements

**Changelog**
:cl:
- fix: Fixed the XM88's damage increase on hit being higher than intended (50% > 20%) and the duration until its combo stacks expire being higher than intended (4 > 2).

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksComponent.cs
@@ -15,7 +15,7 @@ public sealed partial class GunStacksComponent : Component
     public int MaxAP = 50;
 
     [DataField, AutoNetworkedField]
-    public FixedPoint2 DamageIncrease = FixedPoint2.New(0.5);
+    public FixedPoint2 DamageIncrease = FixedPoint2.New(0.2);
 
     [DataField, AutoNetworkedField]
     public float SetFireRate = 1.4285f;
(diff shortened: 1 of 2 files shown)


Vermidia:
Curious about where the fire delay numbers came from. on CM the unboosted delay is 12.5 and boosted is 9, which matches up with the previous fire rate stats.

DummeHytteOst:
> Curious about where the fire delay numbers came from. on CM the unboosted delay is 12.5 and boosted is 9, which matches up with the previous fire rate stats.

i see where i went wrong, but those firerates still doesn't match up with what we had before tho.

1/(12.5/10) = 0.8 base
1/(9/10) = 0.9 for boosted

Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksComponent.cs
```diff
@@ -15,16 +15,16 @@ public sealed partial class GunStacksComponent : Component
     public int MaxAP = 50;
 
     [DataField, AutoNetworkedField]
-    public FixedPoint2 DamageIncrease = FixedPoint2.New(0.5);
+    public FixedPoint2 DamageIncrease = FixedPoint2.New(0.2);
 
     [DataField, AutoNetworkedField]
-    public float SetFireRate = 1.4285f;
+    public float SetFireRate = 1.11f;
```
DrSmugleaf, line 21: Is this tested in-game? You need to take into account buff_fire_reduc being substracted from world time for when to fire next but we don't have that as a mechanic next

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksComponent.cs
DummeHytteOst: i had missed buff_fire_reduc, should be good now, assuming world.time - buff_fire_reduc(2) makes it so you can shoot every 7 tick, which would make the previous firerate right

DummeHytteOst:
should be fine now

DrSmugleaf (approved):

PR #5292, opened 2025-01-02
Layer xenos under other mobs

DummeHytteOst:
## About the PR
Changed xenos layer to be under other mobs, and also increased humanoids priority in xenonightvision
## Why / Balance
Marines have for a while now abused the fact that they can hide under xenos, which prevents xenos from using click abilities, tackle and basic attacks on them, and even seeing that there is someone under them, i think we all can agree that this isn't ideal, marines also appear under dead xenos while xenos are in nightvision mode,  this PR aims to fix all these problems and give xenos a chance to fight marines who are directly on them.
## Media

https://github.com/user-attachments/assets/14147bc0-4523-4c70-8fc2-5d0654598170

https://github.com/user-attachments/assets/173274fa-cc40-4571-85a8-67819a5d96d3

## Requirements

**Changelog**
:cl:
- tweak: Xenos now appear under other mobs.


DrSmugleaf (changes requested):
Doing this is going to be a problem when many xenos stack on a single marine
For cades and walls we fixed that by giving them invincibility frames that act as if only around 3 xenos can hit them at once, but for marines we can't do that
I don't know how you would reconcile the two better than we have it right now

Vermidia:
Ideally it'd be like CM where you can click a tile itself and it'll just act on a thing standing on said tile.
So here probably an area of 1x1 where the xeno clicked and the ability would auto-target if a valid sprite wasn't clicked to the next possible target on the tile and so on and so forth. Same for tackles or click attacks.

DummeHytteOst:
what if we just keep the nightvision priority part? that way xenos at least can see marines, but not click them, this will only be visible to xenos, and prevent marines from hiding under xeno corpses.
.
.
.
.
### 1. response to @Vermidia 
> Ideally it'd be like CM where you can click a tile itself and it'll just act on a thing standing on said tile. So here probably an area of 1x1 where the xeno clicked and the ability would auto-target if a valid sprite wasn't clicked to the next possible target on the tile and so on and so forth. Same for tackles or click attacks.
>
i do not see how this would differ in terms of balance, other than making the system more complicated, while still allowing marines to hide under xenos visually.
### 2. reponse and plea to @DrSmugleaf 
> Doing this is going to be a problem when many xenos stack on a single marine For cades and walls we fixed that by giving them invincibility frames that act as if only around 3 xenos can hit them at once, but for marines we can't do that I don't know how you would reconcile the two better than we have it right now

currently the mostly melee & CQC focused side is being punished for being in CQC range, which feels odds, i among others see the change from my PR as a positive, as it fixes problems that arise in rounds a lot more often.

 i do hope a test merge of this PR in its current state could be considered, and see how it actually will play out.

Vermidia:
The tile thing means you don't need to sprite click specifically, just clicking the ground nearest enough to a target is good enough. This would make tackling and some xeno abilities easier to use.

Dicerson:
yes but that doesn't deal with the more fundamental issue SMugleaf is concerned with.

I 100% understand the concern. Alot of the numbers and other statistics in this game are pulled straight from 13, and in 13 those numbers are designated/balanced around the fact that bodyblocking exists and marines can only get hit by a couple xenos at a time in a worst case. The result is that even without this issue, things like marines and cades die way too quickly when a doomstack of xenos rolls over them than they should.

This problem was solved for Cades by limiting them to 5 hits every few seconds, which is still more than what 13 could possibly have (realisticaly like 3, one for each xeno on the 3 adjacent tiles on one side of the cadeline).

For marines, the issue of spritestacking is a sort of "soft" fix for this issue similar to the cade's hitlimit; xenos are encouraged/forced to spread out and maintain something of a spread just so they don't get in each other's way. This, for the most part, works. Xenos figure out very quickly that we need to spread out a bunch, which is also good strategy in general because explosives exist.

The problem is what happens when Doomstacking isn't in the equation. When marines are protecting themselves by standing directly on top of xenos and making it physically impossible to actually attack them with any targeted ability (such as our only reliable method for capping, the Tackle). There's also a bug around that causes corpses to render on top of sprites sometimes, and marines abuse this too for similar reasons. The issue also results in the particularly large castes struggling to actually tackle at all even without their target trying to abuse this, simply because the majority of their tackle range is "taken up" by their sprite (Hivelords, with a large sprite and the shortest dinosaur arms, suffer this problem the worst of all xenos in the game) making it extraordinarily difficult for them to tackle even at a baseline.

Making target abilities have something of an "autoaim" comes not only with the issue of deciding which target to hit, but also doesn't solve the issue Smugleaf has; which is xenos doomstacking and rolling over marines with near instant kills. Now, some contend that this is not a problem because doomstacking is just plain bad- it makes you too vulnerable to explosives. UGL just got nerfed, so right now only SADAR and Grenadier can actually counter a "doomstack" effectively. But when flamethrower comes out, that will hands down be THE way to destroy a Doomstack and render the strategy utterly useless.

But even before all of this; I think my proposed solution is the best compromise we can get. By making it so each xeno's sprite renders underneath marines (for them only; sort of like how the hide ability always shows on top even though its actually "under" things for marines when its hiding) you solve the small-scale problem (marines standing on top of individual xenos in 1v1 scenarios to render themselves untargetable and xenos struggling to tackle-drag a marine simply because they're too fat) without affecting the larger one (stacks of xenos getting in each other's way) and thus remove the majority of the issues people have with the rendering without actually resulting in doomstacking.

Binbagg:
Really don't believe this is going to be such a big of a problem, I don't think marines should have the chance to escape BECAUSE there is more xenos for example when being dragged through some doorway and you got tackle blocked, this undermines the idea behind the ratio cause a marine should be punished for being underneath 5 xenos

DummeHytteOst:
@DrSmugleaf would this be considered once body block is here?

PR #5293, opened 2025-01-02
Fix xeno explosion dmg coefficient 

DummeHytteOst:
## About the PR
Removes a 2x explosion multiplier xenos had, which was leftovers and didn't get removed when we got grenades stuns

## Why / Balance
brings explosion damage closer to parity.
## Requirements

**Changelog**
:cl:
- fix: Fixed xenos taking too much damage from certain explosions.


Vermidia:
https://github.com/cmss13-devs/cmss13/blob/04b313463e6645f7a68ffc62e7a8976a10cbc002/code/datums/combat_personalized.dm#L102
Explosion damage being x2 for xenos is parity, it's not the damage mult that's the problem it's how explosions are calc'd that is.

PR #5299, opened 2025-01-03
Use squad color for chat names

QuietlyWhisper:
## About the PR

When a squad member speaks, it uses their squad color for their chat name.

## Media

image

## Technical details

- *Modifies upstream file in a non-intrusive way, but still modifies it and will need to be fixed if this upstream code is refactored. (But only the part where name color is applied.)
- Checks for squad component, then pulls the squad color and uses that instead of the randomly assigned color.
- Tested and works with squad transfers (tested with RMC admin ui but not in game mechanic), and any new squads with a color.
- Did not test the lack of a squad color if a squad is found, and I think the code doesn't account for a squad not having a color.

## Requirements

**Changelog**

:cl: Whisper
- add: Marines in squads now speak with their squad color.

### DIFF
--- a/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
+++ b/Content.Client/UserInterface/Systems/Chat/ChatUIController.cs
@@ -17,6 +17,7 @@ using Content.Client.Stylesheets;
 using Content.Client.UserInterface.Screens;
 using Content.Client.UserInterface.Systems.Chat.Widgets;
 using Content.Client.UserInterface.Systems.Gameplay;
+using Content.Shared._RMC14.Marines.Squads;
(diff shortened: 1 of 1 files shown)


QuietlyWhisper:
Test fail is related
image

Dutch-VanDerLinde:
> Test fail is related image

it's because only ``SquadSystem`` can access ``SquadMemberComponent``
it should probably get the squad prototype that the squad member is apart of instead of using the background color from the component

chavonadelal:
we are waiting with impatience 🥺

DrSmugleaf:
Replaced by https://github.com/RMC-14/RMC-14/pull/6201

PR #5301, opened 2025-01-04
Ability Reprites + Prae changes and fixes

Vermidia:
## About the PR

New ability icon sprites for xenos by Aleksh! Observer and alert for map changed aswell.

Prae vanguard cleave and prae warden aid have been split into 2 abilities instead of 1 ability and 1 toggle ability. Same amount of buttons, shared cooldown so they can't be used at the same time as before.

Also prae rejuve removes ammo slows now.

## Why / Balance

New art, fixes.

For the button split, having them be seperate buttons instead of a toggle makes them easier to work with, less clicks / keys to get to the ability you want. Having a toggle when it only effects a single ability is inbuilt clunkiness imo. It should make warden/vanguard a lil easier to use while not really changing how their abilities work. I would do queen but that'd require alot more modification.

## Technical details

Changed the aid and cleave states to be passed by the event so it can switch (so I don't have to rewrite a bunch of code).

## Media

image

## Requirements

**Changelog**

:cl: Vermidia, Aleksh
- add: New xeno ability icons by Aleksh! Related icons have been changed
- tweak: Vanguard Cleave fling & root and Warden Aid heal & rejuvenate are now separate abilities instead of being switched with a toggle. They have a shared cooldown so should work the same as before.
- fix: Fixed Warden rejuvenate not removing slows from ammo.


DrSmugleaf (approved):
Thanks for the PR!

PR #5320, opened 2025-01-06
max survivor limit increase to 8

Binbagg:
## About the PR
survivors maximum limit from 6 to 8

i also think survivor per marines is meant to be 10 but might be much

## Why / Balance
parity

## Requirements

**Changelog**

:cl:
- tweak: Max survivors limit

### DIFF
--- a/Content.Shared/_RMC14/CCVar/RMCCVars.cs
+++ b/Content.Shared/_RMC14/CCVar/RMCCVars.cs
@@ -236,7 +236,7 @@ public sealed class RMCCVars : CVars
         CVarDef.Create("rmc.survivors_minimum", 2, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCSurvivorsMaximum =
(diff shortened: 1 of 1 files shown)


DrSmugleaf:
This was changed intentionally from 8 to 6 since even though the code in 13 says that the limit is 8, that is rarely the case, same with the ratio
Until we can find an explanation for that so we can add what we are missing that won't be changed

Kadeo64:
> This was changed intentionally from 8 to 6 since even though the code in 13 says that the limit is 8, that is rarely if ever the case if you spectate many 13 rounds at round-start, same with the ratio Until we can find an explanation for that so we can add what we are missing that won't be changed

is it a case of people just.. not playing surv on cm13? (even with all the extra gamer loot from what i've heard survs get wrecked on most maps)

might also be tied to the map, like, maybe the map has flags to + or - survivors and eight is just the absolute maximum?

Frogge6645:
This will likely organically change once FORECON is added because there are more than 6 unique roles for FORCON.

PR #5328, opened 2025-01-08
Misc shells, m4spr, mataba parity fixes

DummeHytteOst:
## About the PR
Adjusted the stats of some projectiles to be more in line with parity.

~removed the stuns from the normal mateba .454, since it  doesn't have stuns in cm13, its the HI version that has stuns.~
image

## Why / Balance
Parity

## Requirements

**Changelog**
:cl:
- fix: Fixed M4SPR A19 incendiary rounds having 95 accuracy instead of 105.
- fix: Fixed incendiary slugs having 20 AP instead of 5, and regular buckshot having 0 AP instead of 5.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
@@ -127,6 +127,9 @@
         Piercing: 60
   - type: CMArmorPiercing
     amount: 20
+  - type: RMCStunOnHit
+    maxRange: 4.5
+    stunTime: 1
   - type: RMCProjectileAccuracy
     accuracy: 90
 
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
@@ -127,9 +127,6 @@
         Piercing: 60
   - type: CMArmorPiercing
     amount: 20
-  - type: RMCStunOnHit
-    maxRange: 4.5
-    stunTime: 1
   - type: RMCProjectileAccuracy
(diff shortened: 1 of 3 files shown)


Rainbeon:
This is incorrect, you have to double cm13's to get the correct stun time values since they handle it weird, what we have now is correct

DummeHytteOst:
> This is incorrect, you have to double cm13's to get the correct stun time values since they handle it weird, what we have now is correct

yeah, wasn't aware their time values were that funky, removed the unrelevant changes, and kept the misc parity stuff

Rainbeon:
Can you make the changelog a bit more specific please? The changes seem pretty simple to explain so may as well so people don't have to surf github to see

UnicornOnLSD:
isn't the current mateba bullets stunning for the reason that we do not have the current different ammo types for mateba ?

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
```diff
@@ -127,9 +127,6 @@
         Piercing: 60
   - type: CMArmorPiercing
     amount: 20
-  - type: RMCStunOnHit
-    maxRange: 4.5
-    stunTime: 1
```
DrSmugleaf, line 132: We don't have high-impact speedloaders yet, but all of the ones that the belt with are high-impact, so this is fine staying until then.

DrSmugleaf (changes requested):
Thanks for the PR!

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Revolvers/mateba.yml
DummeHytteOst: alright, reverted the mateba change

DummeHytteOst:
requested changes done, should be ready now.

DrSmugleaf (approved):

PR #5329, opened 2025-01-08
New defense icons + updates on changes

Vermidia:
## About the PR

Title! The Icon is now diff depending on a xeno's current armor

## Why / Balance

Technically not parity (the armor icon is always static in CM, not even green fire seems to mess with it for some reason) but better for visuals + the status text for armor would reflect these things even if the alert did not. Sprites were prettied up and I made it so instead of it just being 100 = armor comp level, 100 = 55, where 55 is the max diff colored shield sprite, for over 50 armor. This is to make the diff armor levels for diff xenos more distinct at a glance, and the increments of 5 allow small armor changes to show up.

## Technical details

CMArmor system new func for updating the xeno alert, make it a public function and any curr xeno armor changing thing calls it to update it. Was having some problems with green fire not updating right but it fixed itself sooo.

## Media

https://github.com/user-attachments/assets/8a8ef7c3-bb37-4989-aad5-6d8d771bcc04

## Requirements

**Changelog**

:cl:
- add: Added new xeno armor alert sprites, by Aleksh!
- fix: Fixed the armor alert not updating with different armor changing effects. It also now has a different icon depending on a xenos base armor.


DrSmugleaf (approved):
Thanks for the PR!

PR #5330, opened 2025-01-09
Fix XM88's Gun Stacks System

Rainbeon:
## About the PR
Rewrote XM88's stack system to properly update damage and reset stacks correctly, as well as added ftl localization.

## Technical details
Split GunStacksComponent into itself + GunStacksActiveComponent. The active version tracks the shots hit and the expire time, this needed to be done so that an Update() loop can be implemented without being expensive to run to properly reset the stacks after 2 seconds.

Removed the condition that resets your stacks when you miss, it is only supposed to reset after the timer runs out.

Added helper functions to manage the stack reset process, as well as updating the gun stats.

The code no longer tries to refresh the gun damage multiplier of the bullet, and instead now correctly updates the gun.

## Media
Video Demonstration available on Discord at https://discord.com/channels/1168210010233376858/1168210011823026270/1326725713113518194

## Requirements

**Changelog**

:cl:
- fix: Fixed the XM88 not increasing its damage on a successful hit.
- fix: Fixed the XM88 getting stuck at maximum damage for extended periods of time.
- fix: Fixed XM88 stacks not resetting once the timer runs out.
- fix: Fixed XM88 stacks resetting when hitting something that is not an alive Xeno, it now only resets after the timer runs out.
- fix: Fixed the XM88 not consistently giving a popup that its stacks have reset.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksActiveComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/Stacks/GunStacksActiveComponent.cs
@@ -1,0 +1,18 @@
+using Robust.Shared.GameStates;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom;
+
+namespace Content.Shared._RMC14.Weapons.Ranged.Stacks;
+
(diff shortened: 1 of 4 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5334, opened 2025-01-09
Fix idle outgoing telephone sound not playing

Dutch-VanDerLinde:
Before, the idle telephone sound would constantly play the dialing sound, instead of playing the idle outgoing noise after 3 seconds

The reason this happened is because ``RotaryPhoneDialingComponent`` wasn't present on the receiving telephone, but an if statement continued if it didn't have ``RotaryPhoneDialingComponent``

this removes the check and fixes another one


DrSmugleaf (approved):
god bless

PR #5351, opened 2025-01-11
Fix coif/balaclava toggling not showing snouts correctly

Dutch-VanDerLinde:
## About the PR
Since the foldable system removes the humanoid layers whenever you toggle the coif while wearing, it won't show the snout

this updates it so any layer that isn't present on the ``HideLayerClothingComponent`` is shown

Also makes it check for layer hidden clothing after folding in both paths so it updates at the correct time


DrSmugleaf (approved):
Thanks for the PR!

PR #5366, opened 2025-01-12
Update Mp armor description and warden armor stats

Dutch-VanDerLinde:
## About the PR
warden armor has the same stats as normal MP armor in 13
also changes the M2 armor description to parity

:cl:
- fix: Fixed warden armor not having the same stats as M2 MP armor.


DrSmugleaf (approved):
Thanks for the PR!

PR #5369, opened 2025-01-12
Setup Publish

JerryImMouse:
https://docs.spacestation14.com/en/server-hosting/setting-up-robust-cdn.html#setting-up-publishing


JerryImMouse:
shit, my bad

PR #5378, opened 2025-01-13
Fix voicemail sometimes appearing when it shouldn't

Dutch-VanDerLinde:
When the receiving end hangs up on the dialing end and then another line calls the receiving end, then the voicemail SFX will play on the dialing end because the dialing end still thinks the receiving phone is 'ringing', even though they already hung up

this PR sets the ``Other`` data field to null whenever the phone hangs up, so this issue shouldn't occur


DrSmugleaf (approved):
Thanks for the PR!

PR #5383, opened 2025-01-13
Remove NVV comp from dead xenos

DummeHytteOst:
## About the PR
removes the nightVisionVisible component from dead xenos.
keeps the death icon visibile so xenos can still identify that there is a dead xeno there.

## Why / Balance
fixes the issues of marines appearing under dead xenos while in nightvision mode.
fixes the issue of barricades being under dead xenos while in nightvision mode.

## Technical details
added OnMobConditionChanged to SharedNightVisionSystem .
OnMobConditionChanged gets called when MobStateChangedEvent happens.

returns if they are not dead or doesn't have the Xeno component
`if (!_mobState.IsDead(ent) || !HasComp<XenoComponent>(ent))
            return;`
then it removes and dirty the entity
`RemCompDeferred<RMCNightVisionVisibleComponent>(ent);
        Dirty(ent);`
## Media

https://github.com/user-attachments/assets/477510b8-ef9b-46e4-8bf5-e37361b2a343

## Requirements

**Changelog**
:cl:
- fix: Dead Xenonids no longer appear above barricades and mobs.

### DIFF
--- a/Content.Shared/_RMC14/NightVision/SharedNightVisionSystem.cs
+++ b/Content.Shared/_RMC14/NightVision/SharedNightVisionSystem.cs
@@ -1,11 +1,14 @@
-﻿using Content.Shared._RMC14.Marines.Skills;
+using Content.Shared._RMC14.Marines.Skills;
 using Content.Shared._RMC14.Scoping;
 using Content.Shared._RMC14.Visor;
+using Content.Shared._RMC14.Xenonids;
 using Content.Shared.Actions;
 using Content.Shared.Alert;
 using Content.Shared.IgnitionSource;
 using Content.Shared.Inventory;
 using Content.Shared.Inventory.Events;
+using Content.Shared.Mobs;
+using Content.Shared.Mobs.Systems;
 using Content.Shared.Popups;
 using Content.Shared.Rounding;
 using Content.Shared.Toggleable;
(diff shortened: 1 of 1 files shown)


DummeHytteOst:
>This will make someone permanently not show up in night vision even if they get revived later

i didn't really think that would be an issue, since it only removes the comp from dead xenos, who can't be revived anyway, but i didn't consider it for admin use.

>Either way this is not how it works, you can see dead bodies as well so that you can play around marines trying to rescue them

you would still be able to see every dead body other than xenos, for xenos only the death icon would show, i asked around and there doesn't seem to be any reason for xenos to see dead xenos in nightvision, you would still be able to see the death icon of dead xenos.
> If this is for bodies blocking cades, modifying night vision is not the way to do it, it would be better to change the layering on cades to do that, but that doesn't address other things like items overlaying over cades

its specifically for bodies in nightvision mode, anything with nightVisionVisible will render on top of basically everything no matter the layering, i choose to remove the comp instead of changing the priority, since even with a lower priority the bodies would still appear over barricades and stuff in xeno nightvision.

i am unsure how else to do this with our current systems, i was thinkin of trying to add nightVisionVisible to cades, and lower the opacity/transparency to 0 or near 0, and lower the nightVisionVisbile priority of dead bodies, but i doubt that would work.

Rainbeon:
This is just for xenos, as far as I see it (given the Xeno check)

I'd still just go with reducing the priority and transparency of dead xenos though.

QuietlyWhisper:
There is an item that revives dead xenonids in normal gameplay that we don't have yet iirc

Also not accounting for any gameplay issue that would require a revival from an admin

DummeHytteOst:
closing for now.

PR #5386, opened 2025-01-13
Add the ability to sleep on bedrolls.

S1ss3l:
Added the ability to sleep on bedroll. It add to the immersion and well, it's a bed, you should be able to sleep on it, no ?
___
**Requirements**

**Changelog**

:cl:
- fix: It is now possible to sleep on bedroll


DrSmugleaf (commented):
As funny as it would be to have Bravo sleeping in FOB I don't think the admins are ready for briefing and FOB to be marines sleeping and snoring

Lyacs:
In a way it would give work to MP without having to deal with rulebreakers  :D

S1ss3l:
My take is that we're an MRP server, we shouldn't limit RP opportunity based on the fact that they could be abused. But I'm no maintainer, this PR is in your hands !

QuietlyWhisper:
Without it having mitigations to stop people annoying each other yes I agree with smug, this doesn't add anything to the game rp or otherwise

PR #5402, opened 2025-01-16
Sticky and Fast resin + construction fixes

Vermidia:
## About the PR

Adds xeno sticky and fast resin, + a few construction changes.

## Why / Balance

Parity.

## Technical details

Modifications to the gardener resin surge code to add hive checks, and spawn sticky resin when a target tile doesn't meet any previous criteria.

SharedXenoWeedSystem had modifications to support hive membership for speedups and slowdowns, and for new slows/speedups caused by sticky and fast resin. Sticky resin slows negate any slowdown that would be caused by the weeds below, effectively overriding em. Fast resin and weed speedup stack.

Slows have been tested and have a unarmored and armored to make speeds closer to CM. Fast resin speedup was based on hivelord speedup. In CM the speedup is 0.8 * MOVE_DELAY and weed speed up is 0.95 * MOVEDELAY. Because of this I did not change weed speed up since it would basically be 0 with the same criteria (hivelord speed is 3.2, it goes to about 3.04 which could be a speed increase of about 11%. For now this PR will not mess with that.

Some modifications to the construction system to support a new component: One allows for a structure to modify build delay (sticky and fast are constructed in half the normal build time). Some more effects were added and theres some other misc fixes.

~~Note that hivecore still spawns hiveless weeds which will properly slow down xenos of not the same hive now. So that should probably be fixed.~~ Should be fixed now.

## Media

https://github.com/user-attachments/assets/f69ade8e-af14-44db-a504-be110b1a96aa

https://github.com/user-attachments/assets/c3751cd1-1e13-404f-a1b8-7ce7899065b3

(Fast resin not stacking with weed speedup has been fixed as well as any other similar issues)

## Requirements

**Changelog**

:cl:
- add: Added sticky and fast resin. Building castes can make these faster than walls and doors, and they're quite cheap. Sticky resin slows non-hivemembers on it more than hive weeds. Fast resin speeds up hivemember xenos on it.
- fix: Fixed how much weeds slow marines down. Generally slow on normal weeds for armored marines is slightly less now, but the penalty is heavier unarmored. Hardy and Hive weeds now also have stronger slows compared to normal weeds.
- fix: Fixed Resin Whisperer nodes not costing plasma. Placing a node now also has an animation.
- fix: Fixed Hive Core spawning hiveless hive weeds.


Vermidia:
Should be mostly done now.

DrSmugleaf:
> Note that hivecore still spawns hiveless weeds which will properly slow down xenos of not the same hive now. So that should probably be fixed.

Is this still a problem? If so can't merge this yet

Vermidia:
No longer, had to add a new event because just setting the hive on MapInitEvent didn't work (they spawned before the hive core's hive could be set) but it works now.

Content.Shared/_RMC14/Xenonids/Construction/FloorResin/StickyResinSurgeBlockerComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Xenonids.Construction.FloorResin;
+/// <summary>
+/// Blocks resin surge from spawning sticky resin ontop of it
+/// </summary>
+
+[RegisterComponent, NetworkedComponent]
+
+public sealed partial class StickyResinSurgeBlockerComponent : Component;
```
DrSmugleaf, line 10:
```suggestion
namespace Content.Shared._RMC14.Xenonids.Construction.FloorResin;

/// <summary>
/// Blocks resin surge from spawning sticky resin ontop of it
/// </summary>
[RegisterComponent, NetworkedComponent]
public sealed partial class StickyResinSurgeBlockerComponent : Component;
```

DrSmugleaf (dismissed):
Thanks for the PR!

PR #5406, opened 2025-01-16
Stopping metabolism in corpses

### LINKED ISSUE #5231: Chemicals shouldn't metabolize in dead people
"noo my epi ran out before i could defib"
fear not, friendly marine, for parity is here to save you :finnadie:

N0lim:
## About the PR
Stopping metabolism in corpses

## Why / Balance
Resolves https://github.com/RMC-14/RMC-14/issues/5231

## Media

https://github.com/user-attachments/assets/984f7f63-8ae1-4bad-95b9-1616d398146e

## Requirements

**Changelog**

:cl:
- fix: metabolism in corpses


N0lim:
Now CMEpinephrine decreases over time in corpse because "worksOnTheDead: true" and don't work without it, fix would be soon

N0lim:
Wait a minute... Epi is removed after each defibrillation, but only heals if the mob is alive???
https://github.com/cmss13-devs/cmss13/blob/4e28767c66cfddf0deb012b5fbd69afd2cb96779/code/game/objects/items/devices/defibrillator.dm#L270
https://github.com/cmss13-devs/cmss13/blob/58987f9edb3edd0738d03bef0d2c704868224b73/code/modules/reagents/chemistry_properties/prop_positive.dm#L539

N0lim:
That's really sounds like a bug
In my understanding, epi should either heal and be removed, or not heal and not be removed

N0lim:
Hmm, now epi used only for 300+ damage healing and defibrillation doesn't apply heart damage, so first case should be better I guess

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -182,22 +183,24 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
...
+                    // if it's possible for them to be dead, and they are,
+                    // then we shouldn't process any effects or remove reagents
+                    if (TryComp<MobStateComponent>(solutionEntityUid.Value, out var state))
+                    {
+                        if (!(HasComp<CMRecentlyDefibrillatedComponent>(solutionEntityUid.Value) && proto.ID == "CMEpinephrine")) // TODO change to PROPERTY_ELECTROGENETIC)
+                        {
+                            if (!proto.WorksOnTheDead && _mobStateSystem.IsDead(solutionEntityUid.Value, state))
+                                continue;
+                        }
+                    }
```
DrSmugleaf, line 195: I don't think this needs to be moved up since the 3 lines below are not changing any state

Content.Server/Body/Systems/MetabolizerSystem.cs
```diff
@@ -182,22 +183,24 @@ private void TryMetabolize(Entity<MetabolizerComponent, OrganComponent?, Solutio
                     if (!proto.Metabolisms.TryGetValue(group.Id, out var entry))
                         continue;
 
+                    // if it's possible for them to be dead, and they are,
+                    // then we shouldn't process any effects or remove reagents
+                    if (TryComp<MobStateComponent>(solutionEntityUid.Value, out var state))
+                    {
+                        if (!(HasComp<CMRecentlyDefibrillatedComponent>(solutionEntityUid.Value) && proto.ID == "CMEpinephrine")) // TODO change to PROPERTY_ELECTROGENETIC)
```
DrSmugleaf, line 190: This should be a data field on the CMEpinephrine reagent, not an id check

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Server/Body/Systems/MetabolizerSystem.cs
N0lim: They change, more precisely, MostToRemove determines how much substance will be removed and moving it to the previous one if with continue causes the reagent to be removed in the corpse

Content.Server/Body/Systems/MetabolizerSystem.cs
N0lim: Done

DrSmugleaf (changes requested):
Forgot to note this down on the PR, but we probably need the dialysis machine ported before we do this, otherwise ODs will be extremely punishing

DrSmugleaf (approved):

PR #5415, opened 2025-01-17
Fix armor variants not being patreon named

Dutch-VanDerLinde:
Title

**Changelog**
:cl:
fix: Fixed armor variants not being patreon named.


DrSmugleaf (approved):
ty boss

PR #5416, opened 2025-01-17
Finally fix burst visuals 

Dutch-VanDerLinde:
## About the PR
Finally updated the system to use the appearance system instead of the egg system
now has its own visualizer

IM FREE

https://github.com/user-attachments/assets/a545d71e-725b-4360-8c38-8e8580161f39

https://github.com/user-attachments/assets/d36ba3a8-8f65-41d8-a9a3-315acc974494

:cl:
- fix: Fixed burst animation looping.


DrSmugleaf (approved):
end of an era

PR #5434, opened 2025-01-20
Prevent placing multiple xeno weed sources on the same tile

blueDev2:
## About the PR

Bufix
## Requirements

**Changelog**
:cl: blueDev2
- fix: Prevent placing multiple xeno weed sources on the same tile

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
@@ -113,6 +113,7 @@
     layers:
     - state: weed0
     - state: constructionnode
+  - type: XenoConstruct
   - type: Destructible
     thresholds:
     - trigger:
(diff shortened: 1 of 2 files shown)


Rainbeon:
Does this prevent Resin Whisperer Hivelord from refreshing nodes or, because it needs to be able to do that still pretty sure

blueDev2:
> Does this prevent Resin Whisperer Hivelord from refreshing nodes or, because it needs to be able to do that still pretty sure

It refreshes, it adds the node, then delete the all nodes other than the one just added.

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
```diff
@@ -113,6 +113,7 @@
     layers:
     - state: weed0
     - state: constructionnode
+  - type: XenoConstruct
```
DrSmugleaf, line 116: This will make it impossible to plant fruit, eggs or build resin I believe

DrSmugleaf (changes requested):

blueDev2:
I've been a bit burned out from coding recently, probably won't finish this in any reasonable time

Issue #5164, opened 2024-12-16
Health Analyzer window shouldn't close if you put it away


Skarletto:
heavy medic qol use 2 hands instead of waste 1 on the analyzer, also incredibly important for when organ/limb damage are added so you dont have to remember which of the 11 limbs you have to splint

https://github.com/user-attachments/assets/371b2749-4260-40b4-a4bc-17b9cde989f3

PR #5439, opened 2025-01-21
Fixed: CMHealthAnalyzer closed the window while it was not in hand

N0lim:
## About the PR
Fixed: CMHealthAnalyzer closed the window while it was not in hand

## Why / Balance
Resolves https://github.com/RMC-14/RMC-14/issues/5164

## Requirements

**Changelog**

:cl:
- fix: health analyzer closed the window while it was not in hand


DrSmugleaf (changes requested):
This will make it auto-update even if it is not in hand which isn't intended

N0lim:
Fixed

PR #5444, opened 2025-01-22
Added Functionality for Constructing Spam cans from the construction menu and refilling them from magazines.

Dicerson:
Added Functionality for Constructing Spam cans from the construction menu and refilling them from magazines.

## About the PR
Added functionality to Spam Cans so they can be refilled with magazines by toggling them with an Alt-click and then interacting with them as normal with a magazine, including appropriate popups for fail states. Added ability to construct Spam Cans out of cardboard.

## Why / Balance
Parity

## Technical details
 Lots of little things inside of BulletBoxSystem, BulletBoxComponent, magazine_box_crafting.yml (which was also moved to the appropriate Recipe folder), and bullet_boxes.yml. Mostly the alternative verb function for the toggle and conditional logic handling the popups and inverting the loading mechanic for refilling.

## Requirements

**Changelog**
:cl: Dicerson
add: Spam cans are now available to be built in the construction menu, for 1 Cardboard per can.
add: Alt-Clicking on a spam can with a valid magazine in hand now load all bullets from the magazine into the can after a 5 second doafter (Try HAR Magazines!)


Dicerson:
I can't get the changelog formatting to actually show up in the pr... t.t

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
```diff
@@ -35,18 +40,40 @@ private void OnExamined(Entity<BulletBoxComponent> ent, ref ExaminedEvent args)
         }
     }
 
+    private void OnGetAlternativeVerbs(Entity<BulletBoxComponent> ent, ref GetVerbsEvent<AlternativeVerb> args)
```
DrSmugleaf, line 43: It would be better to have left click always draw and alt click always insert, instead of having alt click switch the mode, as most people will want to draw from an ammo box, so that they do not have to check
Otherwise this way is not even displayed clearly to them until they try to do it

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
```diff
@@ -35,18 +40,40 @@ private void OnExamined(Entity<BulletBoxComponent> ent, ref ExaminedEvent args)
...
+    {
+        var user = args.User;
+        var verb = new AlternativeVerb()
+        {
+            Act = () => { ent.Comp.TransferToBox = !ent.Comp.TransferToBox;
+                if (_net.IsServer)
+                {
+                    _popup.PopupEntity(Loc.GetString("rmc-bullet-box-refilling-" + ent.Comp.TransferToBox.ToString()), ent, user);
```
DrSmugleaf, line 51: The popup to display should be two LocId fields on the component, not concatenating it like this

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
```diff
@@ -35,18 +40,40 @@ private void OnExamined(Entity<BulletBoxComponent> ent, ref ExaminedEvent args)
...
         args.Handled = true;
-        if (!CanTransferPopup(ent, args.User, ref used))
+        var user = args.User;
+        if (!CanTransferPopup(ent, user, ref used))
             return;
 
         var ev = new BulletBoxTransferDoAfterEvent();
-        var doAfter = new DoAfterArgs(EntityManager, args.User, ent.Comp.Delay, ev, ent, ent, args.Used)
+        var delay = ent.Comp.DelayTransferFromBox;
+        if(ent.Comp.TransferToBox){
```
DrSmugleaf, line 73:
```suggestion
        if (ent.Comp.TransferToBox)
        {
```

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
```diff
@@ -85,24 +128,42 @@ private bool CanTransferPopup(Entity<BulletBoxComponent> box, EntityUid user, re
...
+            {
+                popup = Loc.GetString("rmc-bullet-box-box-full");
+            }
+        }
+        if(popup is not null)
+        {
+            if (_net.IsServer)
+            {
+                _popup.PopupEntity(popup, box, user);
+            }
```
DrSmugleaf, line 164: This needs to be popup client, not popup entity inside a net is server check, otherwise it won't be predicted on the client
This also makes it no longer a medium caution popup

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
```diff
@@ -63,20 +90,36 @@ private void OnTransferDoAfter(Entity<BulletBoxComponent> ent, ref BulletBoxTran
...
 
-        _popup.PopupClient(Loc.GetString("rmc-bullet-box-transfer-done", ("amount", transfer), ("used", used)), user);
+            transfer = Math.Min(transfer, used.Comp2.Count);
+            _gun.SetBallisticUnspawned((used, used.Comp2), used.Comp2.UnspawnedCount - transfer);
+            ent.Comp.Amount += transfer;
+        }
+        if (_net.IsServer)
+        {
+            _popup.PopupEntity(Loc.GetString("rmc-bullet-box-transfer-done", ("amount", transfer), ("used", ent)), ent, user);
+        }
```
DrSmugleaf, line 121: Same here, this needs to be popup client, not net is server popup entity, or it won't be predicted

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
Dicerson: You know I agreed and tried to do this initially but was advised against it by Blu when I consulted him XD

I'll see about making it function the way that I had originally intended!

Content.Shared/_RMC14/BulletBox/BulletBoxSystem.cs
Dicerson: PopupClient didn't work for reasons I did not understand, it very consistently resulted in the wrong popup. When I asked for help from Blu he believed it was due to some kind of prediction issue, so we then tried popup Entity and that resulted in getting the correct popup but a dozen times- which led to the net is server.

Dicerson: I do have an idea though that might fix that- especially since I'm about to change the functionality to what what I'd originally thought about

DrSmugleaf (approved):
Thanks for the PR!

PR #5475, opened 2025-01-28
Fix armor modifiers + boiler gas again

Vermidia:
## About the PR

So, we we're doing x4 in the armor calculation, which is correct for bio armor, but not for melee which is x5 (more detail in this file https://github.com/cmss13-devs/cmss13/blob/53834d3f1132f519dcdaa0722ac1787a2e552674/code/datums/combat_personalized.dm#L41)

So...boiler gas. After alot of testing I found out that:
1. I got the topical damage wrong, it's not too high, it's actually not high enough! the RNG is about 30-40 damage because I failed to see the amount variable inherited at 2, not 1. Neuro is at 1 though so I don't need to care about it.
2. There is some rare, weird case where I'm not sure if it's a CM bug where gas sometimes targets the same limb as unarmored damage, but it will fail to deal topical damage very rarely in the gas.
3. Damage is lower overtime the more limbs reach max damage mainly because armor gets calculated more than once when propagating to a new limb.

Otherwise the gas in general is a lil easier to escape due to the pain system mostly. So I upped the damage, ~~but shrunk the bounding box ever so slightly so it's a tad easier to escape, maybe. You still take damage fine in the middle of the gas so there doesn't seem to be any weird gap problems.~~

## Why / Balance

It's parity, I swear. I did alot of testing for this and this is what I got.

## Technical details

YML + new comp to hold the ranged and melee modifiers. As far as I can tell most things all use the ranged mods and specifically melee armor uses the melee mod. Fire etc are their own things that could be added later, if needed.

## Requirements

**Changelog**

:cl:
- fix: Fixed boiler gas armor effecting damage being 1/2 as much as its supposed to be
- fix: Fixed marines armor reducing melee damage slightly more than it should have.


DrSmugleaf (dismissed):
Need conflicts fixed and then this can be merged, check that its correct with the new armor value split

DrSmugleaf (approved):
Thanks for the PR!

PR #5477, opened 2025-01-28
Cleanup all slowing effects + minor fixes

Vermidia:
## About the PR

Cleaned up all instances I could find of slowdown/root effects in RMC and put them all under one system.

## Why / Balance

Parity and cleanup. This is needed as under a superslow a normal slow's effects won't register. Also allows them to be visible on the xeno HUD instead of just spawned as an effect.

## Technical details

New System, RMCSlowSystem with 3 components for the diff slows, and a new RMCSpeciesSlowModifier comp. The latter is so we can have xenos and marines get slowed diff amounts by the same effects, as in CM. There's also a duration mult because xenos recover from slows faster, and thus like stuns the time needs to by x 2/3s, except in the case where the effect is a "xeno_slow" because those are their own separate effect in CM so they run out the full duration anyway.

Every comp that was a standalone slow from an ability now all goes through this system and uses these comps, except in specific cases:
Pounce Rooting and Charging still use their own root as this kinda root isn't healable by a warden.
Acider Slowdown on For The Hive in CM applies a superslow and then kinda goes through hurdles to make sure it only removes the superslow from the effect, I'm just leaving that part of that comp honestly.
Anything not here I missed and need to go and mess with. This should make it easier to make alerts etc in the future for them.

This new system also has 2 diff components for the "freeze" and "slow" visuals which are now applied automatically in the right conditions (being under a slow, or being stunned w/o knockdown / rooted. They aren't spawned and are instead part of the xeno HUD as only xenos can see them in CM, also increases their visibility here. Note in CM xenos can't see generic slows but can see any immobilization whatsoever. Didn't do the prior detail because just having them see every status slow is much easier and because of how they work would have the same effect anyway.

Other stuff:
Stomp didn't have a slow, it now does a superslow of 1 sec to all it hits [knew the number and effect was in CM somewhere and finally found it]
tried to cleanup any abilities that checked plasma / can attack targeted that could easily be replaced with XenoPlasmaAction and XenoOffensiveAction.

## Media

https://github.com/user-attachments/assets/a702b4b0-fe85-4373-80bf-1cb43e8ac308

https://github.com/user-attachments/assets/600156cb-5616-4c60-950a-b360613bc6c1

## Requirements

**Changelog**

:cl:
- fix: Fixed xeno slowdowns from different abilities stacking.
- fix: Fixed stronger slows stacking with smaller slows
- fix: Fixed stomp not causing any slowdown
- fix: Fixed slowing and stunning visual effects being visible to marines
- fix: Fixed slows being applied to xenos in most cases not having 2/3rds of the duration
- fix: Fixed charge not slowing down marines in the area of impact.
- fix: Fixed tail sweep having a lower range than all other melee abilities (1.25 > 1.5). It and blitz can no longer hit through walls.


DrSmugleaf (approved):
Thanks for the PR!

PR #5387, opened 2025-01-14
Add  meson goggles

Dutch-VanDerLinde:
## About the PR
CM13 parity
add mesons and NVG filter

https://github.com/user-attachments/assets/b6b271a8-db9d-424c-9ea5-fb643c58e2bf

https://github.com/user-attachments/assets/25f5ec50-0781-4fd5-adb4-c28cb904347c

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

:cl:
- add: Added meson functionality to the weapon specialist NVGs. Meson goggles allow you to see a room's layout.

PR #5484, opened 2025-01-28
Add night vision film grain

Dutch-VanDerLinde:
Cool effect from CM13

Split from #5387 because they weren't really related, see there for a video

:cl:
- tweak: Night vision visors now have their film grain overlay


DrSmugleaf (approved):
Thanks for the PR!

PR #5491, opened 2025-01-29
Instant chair deconstruction

### LINKED ISSUE #5451: Deconstructing chairs should be instant
They're instant or at least faster to deconstruct in CM13

Arendian:
## About the PR
Chair deconstruction doafter removed.
Chair deconstruction tool changed from screwdriver to wrench.
Resolves #5451 

## Requirements

**Changelog**

:cl: Dygon
- tweak: Chair are now instantly deconstructed with a wrench, instead of a screwdriver.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Furniture/chairs.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Furniture/chairs.yml
@@ -22,7 +22,6 @@
     anchored: false
   - type: Physics
     bodyType: Static
-  - type: Anchorable
   - type: Construction
     graph: CMSeat
     node: chair
@@ -59,7 +58,6 @@
     anchored: true
   - type: Physics
     bodyType: Static
-  - type: Anchorable
   - type: Construction
     graph: CMSeat
     node: chairComfy
--- a/Resources/Prototypes/_RMC14/Recipes/Construction/Graphs/furniture/seats.yml
+++ b/Resources/Prototypes/_RMC14/Recipes/Construction/Graphs/furniture/seats.yml
@@ -42,7 +42,7 @@
               prototype: CMSheetMetal1
           steps:
             - tool: Screwing
-              doAfter: 1
+              doAfter: 0
 
     - node: stool
       entity: RMCStool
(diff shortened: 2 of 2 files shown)


DrSmugleaf (changes requested):
Thanks for the PR!
While you are changing this, it's supposed to be done with a wrench, not screwdriver
You can set tool to Anchoring for this

Arendian:
> Thanks for the PR! While you are changing this, it's supposed to be done with a wrench, not screwdriver You can set tool to Anchoring for this

This does mean that I'll have to remove the anchorable component (it shows the doafter for anchoring but it gets cancelled since the chair is deconstructed). Is unanchorable chairs also the intention, or should I change the anchoring tool for chairs to another one (screwdriver?)?

DrSmugleaf:
> This does mean that I'll have to remove the anchorable component from all chairs (it shows the doafter for unanchoring but it gets cancelled because the chair is deconstructed before it finishes). Is unanchorable chairs also the intention, or should I change the anchoring tool for chairs to another one (screwdriver?)?
> 
> I also noticed that all chairs deconstruct to 1 metal, but comfy chairs cost 2 metal and office chairs 5 metal to create. I was wondering which of the following options you prefer:
> 
> 1. It stays as it is and sometimes metal just gets deleted if someone makes a comfy or an office chair, and deconstructs it afterwards.(I don't think this happens a lot anyways)
> 2. The amount of metal from deconstruction becomes the same and the amount of salvageable metal on the Almayer gets increased by quite a bit, since almost all chairs on the Almayer are comfy/office ones.
> 3. The construction cost of all chairs is set to 1 metal, so no metal gets deleted and the total amount of metal on the Almayer also stays the same

All chairs are supposed to deconstruct into 1 metal even if they cost more, so 1 would be best
Removing the anchoring component is fine

Arendian:
> All chairs are supposed to deconstruct into 1 metal even if they cost more, so 1 would be best Removing the anchoring component is fine

Arendian:
Woops I clicked the wrong button

DrSmugleaf (approved):
Thanks for the PR!

PR #5496, opened 2025-01-30
Evacuation can now be authorized while on red or delta alert.

### LINKED ISSUE #4651: Initiate Evac on Hijack Start
## Description

You are supposed to be able to initiate evacuation procedures upon the dropship hijack message, not AFTER it has already crashed into the Almayer.

Arendian:
## About the PR

Resolves #4651 
Evacuation can now be authorized while on red or delta alert.
Pumps still only start working after the dropship has crashed into the Almayer.

## Technical details

The EvacuationProgressComponent is now also given upon auhorizing the evac, but a seperate boolean check will prevent pumps activating until the dropship crashes.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Evacuation can now be authorized while on red or delta alert.


GoodWheatley:
Commenting here that evac doesn’t need to be authorized for pumps to work, they’ll work regardless of evac being active or not. It’s actually common for a CM XO to say they’ll only authorize evac when fuel is half/fully pumped.

Arendian:
> Commenting here that evac doesn’t need to be authorized for pumps to work, they’ll work regardless of evac being active or not. It’s actually common for a CM XO to say they’ll only authorize evac when fuel is half/fully pumped.

Thanks for the input! I removed the authorization authorization requirement for fueling since it's not parity.

PR #5505, opened 2025-02-01
increase volume of dropship landing and startup

QuietlyWhisper:
Tweaks the audio balancing of these sounds and increases their volume

**Changelog**

:cl: Whisper

- tweak: Increased the volume of dropship startup and landing.


DrSmugleaf (approved):
Thanks for the PR!

PR #5507, opened 2025-02-01
Give CE their shotgun and M54C Mk2 case

Dutch-VanDerLinde:
## About the PR
Ports content, CE has a shotgun and M54C mk2 case in their vendor
shotgun case contains the shotgun, 3 handfuls of random shells

M54C case contains the gun and 4 mags

image
image


DrSmugleaf (approved):
Thanks for the PR!

PR #5508, opened 2025-02-02
Nightmare Map Inserts

CyberTropic:
## About the PR

Creating a system for randomized map inserts AKA Nightmare inserts in cm13.
Created SE checkpoint insert on Solaris and Communications insert on Kutjevo.

This system doesn't cover ALL the ways of having map inserts, as there are some inserts in cm13 that are co-dependent.

## Why / Balance

Parity

## Technical details

Creates mapInsert spawners entity that points to an insert map.
On mapinit, we optionally clear all the existing entities where the inserted grid will go, we update the areas with the areas from the insert map, and finally we merge the grids together.

## Media

Map insert on its own:
image
_Note: item spawners on map inserts work without issues_

No insert:
noinsert
Insert:
mapinsert

**Areas**
No insert:
image
Insert:
image
_Note: the whole security area exists only on the insert and doesn't exist otherwise_

Another example on Kutjevo, with 35% spawn probability:
kutInsert
image

## Requirements

**Changelog**

:cl:
- add: Added Nightmare map insert system.
- add: Added Solaris SE Checkpoint insert with 70% probability.
- add: Added Kutjevo Communications insert with 35% probability.


FeelFlows:
> Does Tacmap's sprite change to show this?

It doesn't in 13 and it's not intended to, the nightmare system is supposed to be a surprise

Content.Server/_RMC14/MapInsert/MapInsertSystem.cs
```diff
@@ -0,0 +1,179 @@
...
+using Robust.Shared.Network;
+using Robust.Shared.Physics;
+using Robust.Shared.Physics.Components;
+using Robust.Shared.Physics.Systems;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Random;
+
+namespace Content.Server._RMC14.MapInsert;
+
+public sealed class MapInsertSystem : SharedMapInsertSystem
```
DrSmugleaf, line 18: You should be able to put this in Shared and have a single system there

Resources/Prototypes/_RMC14/Entities/Objects/map_inserts.yml
```diff
@@ -0,0 +1,35 @@
...
+    
+# Kutjevo
+- type: entity
+  parent: RMCMapInsertBase
+  id: RMCMapInsertKutjevoCommunications
+  name: SE Checkpoint
+  suffix: Insert Kutjevo
+  components:
+  - type: MapInsert
+    spawn: "/Maps/_RMC14/inserts/kutjevo/communications.yml"
```
DrSmugleaf, line 32: inserts needs to be uppercase Inserts

DrSmugleaf (changes requested):
this is epic

Content.Server/_RMC14/MapInsert/MapInsertSystem.cs
CyberTropic: I originally had it in Shared, but I moved the code to Server because the GridFixtureSystem is only on the server. Should I just clean it up so that the whole thing is a server system only?

DrSmugleaf: It would be best to make an RMCGridFixtureSystem in Shared/Client/Server that allow you to use the Merge method from Shared, but if you don't want to do that putting the system in servedr for now is fine too

CyberTropic: Moved it all to server for now then

DrSmugleaf (approved):

PR #5512, opened 2025-02-02
Add radio send sound

Dutch-VanDerLinde:
## About the PR
Adds the sound from CM13 when you send a radio message

https://github.com/user-attachments/assets/b2654559-d778-46f2-8ac6-f58a92ec9b48


noctyrnal:
you might actually be the goat for this

Content.Server/Radio/EntitySystems/HeadsetSystem.cs
```diff
@@ -53,6 +61,7 @@ private void OnSpeak(EntityUid uid, WearingHeadsetComponent component, EntitySpo
             && keys.Channels.Contains(args.Channel.ID))
         {
             _radio.SendRadioMessage(uid, args.Message, args.Channel, component.Headset);
+            _audio.PlayPvs(_radioSound, uid); // RMC14
```
DrSmugleaf, line 64: Can't remember 100% if this is supposed to play for others near you or not but I don't think it does, would need to check but for some reason starting 2 clients at once in byond locally is broken for me

DrSmugleaf (changes requested):
insane tech

chavonadelal:
just wondering if you can hear this sound when someone else is talking on the radio? (not standing next to you, but as a notification that someone said something on the radio)

Dutch-VanDerLinde:
> insane tech

I THINK it plays another sound if you aren't the one talking on the radio, as I just played a CM13 round


Can't find this sound in the files tho

Dutch-VanDerLinde:
> insane tech

image

according to this it plays in local as long as the person who hears it is a human or synth

DrSmugleaf (approved):

PR #5513, opened 2025-02-02
Fix weed killer effects

Dutch-VanDerLinde:
Makes the weed killer sound and camera shakes actually work


DrSmugleaf (approved):
Thanks for the PR!

PR #5528, opened 2025-02-06
Fix hypo tactical reload not working

Vermidia:
## About the PR

Title.

## Why / Balance

Bug.

## Technical details

At some point inserting something into an itemslot that's empty required you to be able to drop said item, from your hands? This just unsets the user so we don't have to check that so it actually just grabs the item (because it's not going to be in your hands tacreloadings whole point is that)

## Requirements

**Changelog**

:cl:
- fix: Fixed hypospray tactical reloads not working in most cases


DrSmugleaf (approved):
Thanks for the PR!

PR #5530, opened 2025-02-06
Remove M54C MK1 camo

Dutch-VanDerLinde:
## About the PR
It's supposed to have only desert camo, it's like this in CM13

A less-stocked older version of the M54C, so camo wasn't mass produced for it

image
image
image


Carolyn3114:
same for HAR, right?

DrSmugleaf (approved):
Thanks for the PR!

PR #5531, opened 2025-02-06
fix mou53 being too loud

QuietlyWhisper:
**Changelog**

:cl: Whisper

- fix: Fixed MOU53 being too loud, the sound was not mono.


DrSmugleaf (approved):
Thanks for the PR!

PR #5542, opened 2025-02-07
Fix welder backpacks not refueling welders by just clicking on them

Dutch-VanDerLinde:
## About the PR
now it's more intuitive to refuel welders with a backpack welder tank
cm13 has it like this

## Media

https://github.com/user-attachments/assets/5944a1af-22a6-4d6a-9585-77acf1070c67

## Technical details
Removes the upstream event and replaces it with a ``InteractUsingEvent`` that runs before storage

## Requirements

:cl:
- fix: Fixed welder packs not refueling by just clicking on them.

### DIFF
--- a/Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
+++ b/Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
@@ -47,7 +47,7 @@ public sealed class RMCRepairableSystem : EntitySystem
         SubscribeLocalEvent<NailgunRepairableComponent, InteractUsingEvent>(OnNailgunRepairableInteractUsing);
         SubscribeLocalEvent<NailgunRepairableComponent, RMCNailgunRepairableDoAfterEvent>(OnNailgunRepairableDoAfter);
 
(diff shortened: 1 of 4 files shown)


Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
```diff
@@ -319,4 +324,40 @@ private float GetRepairValue(Entity<NailgunRepairableComponent> repairable,
 
         return repairValue;
     }
+
+    private void OnWelderInteractUsing(Entity<ReagentTankComponent> ent, ref InteractUsingEvent args)
```
DrSmugleaf, line 328: This should give you a warning if you try to do it while the welder is on
`That was close! However, you realized you had the welder on and prevented disaster.`

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
Dutch-VanDerLinde: Done

DrSmugleaf (approved):

PR #5548, opened 2025-02-08
Fix Acid Traps detection around open corners

Rainbeon:
## About the PR
Acid Traps now correctly trigger when walking into their radius around an open corner specifically. They will NOT trigger through solid walls still.

## Technical details
Check if the resin hole can see directly first, then check if any tile straight to the north, east, south, and west can see the target. If nothing can see the player, the trap does not trigger. If the offset tile is inside a wall, the trigger automatically fails. Each offset vision check is 1.5f to maintain desired behavior + the vision range check does not run regardless if you are not in range to trigger it.

## Media

https://github.com/user-attachments/assets/b0e8b35a-3731-4c62-8436-ab732ea81453

## Requirements

**Changelog**

:cl:
- fix: Fixed Acid Resin Traps not triggering around open corners. (Traps will still not trigger if they are entirely obstructed)


DrSmugleaf (approved):
ty boss

KingOnBudget:
sobbing rn

PR #5577, opened 2025-02-10
takeOne field for vendor sections

Ian321:
## About the PR

This PR adds a "takeOne" field to vendors.

## Why / Balance

There where a bunch of "takeAll" + "choices: 1" that can lead to confusion, as the title would say to take everything but you where only able to take one.

## Technical details

CMVendorSection and CMVendorUserComponent now have a new "TakeOne" field that tracks singular mandatory selections.

## Media

Screenshot From 2025-02-10 20-55-57

## Requirements

**Changelog**

No cl.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
@@ -115,12 +115,10 @@
       - id: RMCGlassesAviators
       - id: RMCM5Bayonet
     - name: Specialisation Kit
-      choices: { CMBundle: 1 }
+      takeOne: CMBundle
       entries:
       - id: CMVendorBundleCombatTechnicianEssentials
-        recommended: true
       - id: CMVendorBundleSquadMedicalEssentials
-        recommended: true
     - name: Belt
       choices: { CMBelt: 1 }
       entries:
@@ -232,8 +230,7 @@
     - CMExecutiveOfficer
     sections:
     - name: Captain's Primary
-      choices: { CMPrimary: 1 }
-      takeAll: CMStandard
+      takeOne: CMPrimary
       entries:
       - id: RMCM54CMK1CaseAP
         name: M54C MK1 assault rifle
@@ -253,7 +250,7 @@
       - id: RMCBoxShotgunFlechette
         points: 20
     - name: Specilisation Kit
-      choices: { CMBundle: 1 }
+      takeOne: CMBundle
       entries:
       - id: CMVendorBundleSquadMedicalEssentials
       - id: CMVendorBundleCombatTechnicianEssentials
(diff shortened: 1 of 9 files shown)


Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/intel.yml
```diff
@@ -125,9 +125,9 @@
       - id: CMMRE
     - name: Armor
       choices: { CMArmor: 1 }
-      takeAll: RMCIntelArmor
       entries:
       - id: RMCArmorXM4
+        recommended: true
```
Dutch-VanDerLinde, line 130: this is not parity
``takeAll`` marks all the options as 'mandatory', which is how the intel armor section is marked in cm13

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
```diff
@@ -115,12 +115,10 @@
       - id: RMCGlassesAviators
       - id: RMCM5Bayonet
     - name: Specialisation Kit
-      choices: { CMBundle: 1 }
+      takeOne: CMBundle
       entries:
       - id: CMVendorBundleCombatTechnicianEssentials
-        recommended: true
```
Dutch-VanDerLinde, line 121: you should replace these with a commented out # mandatory: true

these options are supposed to be mandatory

Dutch-VanDerLinde (changes requested):

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/intel.yml
Ian321: So should the IO HAVE to take the XM4 or give them the choice for the others? If the former, ill move it up into "Standard Equipment" and delete the armor selector. If the latter ill do the same but keep the other two in the armor selector.

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
Ian321: isn't that already signaled via the takeOne? Should everything that uses takeAll also have "# mandatory: true"?

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/intel.yml
Ian321: wait I read that wrong

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
Dutch-VanDerLinde: yea, that works
didn't see that take one would mark it as mandatory

was saying this because you can mark item entries individually as mandatory in CM13

DrSmugleaf (approved):

PR #5583, opened 2025-02-11
Survivors can no longer interact with intel

QuietlyWhisper:
## About the PR

- Survivors can no longer mechanically interact with intel objectives, resulting in no accidental rulebreaks / ban bait.
- Survivors are forbidden to use any methods of picking up / moving intel objectives if discovered, by roleplay standards (excluding if supported by marines).

## Why / Balance

- Prevents early game intel gathering for free by survivors, makes sure intel roles have more to do / actually engage with the intel mechanics.
- Roleplay Standards.

## Requirements

**Changelog**

:cl: Whisper

- tweak: Survivors can no longer interact with intel objectives.
- tweak: Rules change; Survivors may not secure intel objectives without marine support.

### DIFF
--- a/Content.Shared/_RMC14/Intel/IntelSystem.cs
+++ b/Content.Shared/_RMC14/Intel/IntelSystem.cs
@@ -242,7 +242,7 @@ public sealed class IntelSystem : EntitySystem
     {
         var user = args.User;
 
-        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
+        if (HasComp<SurvivorComponent>(user))
         {
             _popup.PopupClient($"You have no need to read the {Name(ent)}.", ent, user);
             return;
@@ -260,7 +260,7 @@ public sealed class IntelSystem : EntitySystem
         ContainerGettingInsertedAttemptEvent args)
     {
         var user = args.Container.Owner;
-        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
+        if (HasComp<SurvivorComponent>(user))
         {
             args.Cancel();
             _popup.PopupClient($"You have no use for the {Name(ent)}.", ent, user);
@@ -272,7 +272,7 @@ public sealed class IntelSystem : EntitySystem
     private void OnIntelPullAttempt(Entity<IntelRetrieveItemObjectiveComponent> ent, ref PullAttemptEvent args)
     {
         var user = args.PullerUid;
-        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
+        if (HasComp<SurvivorComponent>(user))
         {
             args.Cancelled = true;
             _popup.PopupClient($"You have no use for the {Name(ent)}.", user, user);
--- a/Content.Shared/_RMC14/Intel/IntelSystem.cs
+++ b/Content.Shared/_RMC14/Intel/IntelSystem.cs
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Immutable;
 using System.Diagnostics.CodeAnalysis;
+using System.Net;
 using Content.Shared._RMC14.Areas;
 using Content.Shared._RMC14.ARES;
 using Content.Shared._RMC14.CCVar;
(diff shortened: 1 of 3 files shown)


Ian321:
Maybe use the "IntelRescueSurvivorObjectiveComponent" instead of "SurvivorComponent"? That way survivors that visited the warship can pick up intel (cause that component gets deleted).

OmerF-afk:
> Maybe use the "IntelRescueSurvivorObjectiveComponent" instead of "SurvivorComponent"? That way survivors that visited the warship can pick up intel (cause that component gets deleted).

It's Intentional I believe since you can be recruited into marine ranks after surviving and getting to USS Almayer

Ian321:
Also as a side. I understand this from a gameplay perspective, but realistically someone in that situation would most likely pick up all their data and get it somewhere safe/where it can be found.

One more alternative way to handle this would be to make the intel items appear as normal paper to survivors until they had contact with marines.

AverageNotDoingAnythingEnjoyer:
I kind of don't get the reasoning behind the "Survivor cant interact with intel" in the first place.

Yeah, sure, they shouldn't but there's already a mechanical punishment (their death) if they spend too much time on it. So a survivor having a few papers on them is not that big of an issue, most will not interact with it anyway, since it takes away precious space and survs dont have any document pouches for it to make any sense.

Ian321:
I think that the biggest issue is that anyone can read it at the moment and thus gather points even before the marines deploy. Maybe instead create a PR that requires the intel skill to read.

AverageNotDoingAnythingEnjoyer:
> I think that the biggest issue is that anyone can read it at the moment and thus gather points even before the marines deploy. Maybe instead create a PR that requires the intel skill to read.

I thought you're supposed to enter every single read intel in intel computer for it to count...
But yeah, that is a good idea too

AverageNotDoingAnythingEnjoyer:
In any case, how does CM13 deal with it?

Dutch-VanDerLinde:
> In any case, how does CM13 deal with it?

It doesn't actually
You can read intel no matter your skill, it just takes longer to read if you have lower or no skill

Ian321:
> > I think that the biggest issue is that anyone can read it at the moment and thus gather points even before the marines deploy. Maybe instead create a PR that requires the intel skill to read.
> 
> I thought you're supposed to enter every single read intel in intel computer for it to count...
> But yeah, that is a good idea too

You get points for reading and some more for storing documents. You type em into the computer to be able to read bigger documents (and to get a clue where to find em).

Content.Shared/_RMC14/Intel/IntelSystem.cs
```diff
@@ -234,13 +241,44 @@ private void OnRequiresRemove<T>(Entity<IntelRequiresComponent> ent, ref T args)
     private void OnReadUseInHand(Entity<IntelReadObjectiveComponent> ent, ref UseInHandEvent args)
     {
         var user = args.User;
+
+        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
```
DrSmugleaf, line 245:
```suggestion
        if (HasComp<SurvivorComponent>(user))
```

Content.Shared/_RMC14/Intel/IntelSystem.cs
```diff
@@ -234,13 +241,44 @@ private void OnRequiresRemove<T>(Entity<IntelRequiresComponent> ent, ref T args)
...
         if (_doAfter.TryStartDoAfter(doAfter))
             _popup.PopupClient($"You start reading the {Name(ent)}", ent, user);
     }
 
+    private void OnHandPickUp(EntityUid ent,
+        IntelRetrieveItemObjectiveComponent component,
+        ContainerGettingInsertedAttemptEvent args)
+    {
+        var user = args.Container.Owner;
+        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
```
DrSmugleaf, line 263:
```suggestion
        if (HasComp<SurvivorComponent>(user))
```

Content.Shared/_RMC14/Intel/IntelSystem.cs
```diff
@@ -234,13 +241,44 @@ private void OnRequiresRemove<T>(Entity<IntelRequiresComponent> ent, ref T args)
...
+            _popup.PopupClient($"You have no use for the {Name(ent)}.", ent, user);
+            return;
+        }
+
+    }
+
+    private void OnIntelPullAttempt(Entity<IntelRetrieveItemObjectiveComponent> ent, ref PullAttemptEvent args)
+    {
+        var user = args.PullerUid;
+        if (_ent.GetComponentOrNull<SurvivorComponent>(user) != null)
```
DrSmugleaf, line 275:
```suggestion
        if (HasComp<SurvivorComponent>(user))
```

DrSmugleaf (dismissed):

PR #5594, opened 2025-02-11
allow saved survivors to pick up documents

Ian321:
## About the PR

As discussed in #5583 I feel like survivors should be allowed to pick up documents AFTER they had visited the warship.

## Why / Balance

Allows survivors to collect intel AFTER being rescued.
A survivor that wants to help the marines should be able to do so without this limitation.

## Technical details

Instead of checking for the SurvivorComponent, we check for the IntelRescueSurvivorObjectiveComponent.

## Media

https://github.com/user-attachments/assets/37cda1e0-d8bc-478d-9cbb-96999ea38105

## Requirements

**Changelog**

:cl:
- tweak: Survivors can handle intel after being rescued.

### DIFF
--- a/Content.Shared/_RMC14/Intel/IntelSystem.cs
+++ b/Content.Shared/_RMC14/Intel/IntelSystem.cs
@@ -242,7 +242,7 @@ public sealed class IntelSystem : EntitySystem
 
         if (HasComp<IntelRescueSurvivorObjectiveComponent>(user))
         {
-            _popup.PopupClient($"You have no need to read the {Name(ent)}.", ent, user);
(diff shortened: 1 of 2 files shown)


QuietlyWhisper (dismissed):
Assuming this works as intended and Smug also thinks it's good I think this is a better implementation of the original concept

QuietlyWhisper (commented):
You could probably add onto the popup at the end "Focus on getting out alive first." to imply it's not a permanent restriction anymore

QuietlyWhisper (approved):
I approve just needs an updated video of it working and smug's approval as well

DrSmugleaf (approved):
Thanks for the PR!

PR #5602, opened 2025-02-12
Make NVG film grain slower

Dutch-VanDerLinde:
## About the PR
Players are complaining about the intensity of the film grain
This makes the grain a lot slower

:cl:
- fix: Made NVG film grain less intense


DrSmugleaf (changes requested):
This doesn't (shouldn't?) make it slower, it just changes the "random" value that the grain is multiplied by

DrSmugleaf:
Reopen if you come back to this

PR #5617, opened 2025-02-13
Fix evacuation doors being pryable while locked and kill anyone left outside the docking area after a lifeboat departs

### LINKED ISSUE #4412: you can pry evac doors after it launches to go into spess

### LINKED ISSUE #5586: Xenos able to pry open lifeboat doors
## Description
Xenos are currently able to pry the lifeboat doors open before hijack. The doors are indestructible and impossible to open by marines until hijack starts, and this has been abused in several rounds when a larva got loose aboard the ship since the doors close automatically.

**Reproduction**
1. Spawn as any xeno able to pry doors
2. Pry open one of the locked lifeboat docking hatches
3. You are now in an area that is inaccessible by marines

Arendian:
## About the PR

Evacuation doors now can't be pryed anymore while they are locked. This means that xeno's can't pry pod or lifeboat doors anymore after their launch sequence has been initiated or when hijack hasn't started yet.

When a lifeboat jumps into FTL space it now spawns a new entity on every tile it occupied that deals 5000 damage on collision with another entity.

## Why / Balance

Resolves #5586 
Resolves #4412

## Technical details

Added an event listener that cancels pry attempts if an evacuation door is locked.

A new event is called before FTL is started. 
A new event listener stores every tile an entity occupies before it goes into FTL if the FTLing entity has the RMCSpawnEntityOnFTLStartComponent.
After the entity goes into FTL space, on every tile it previously occupied a new entity is spawned(the default one kills anything near it).

## Media

https://github.com/user-attachments/assets/9d9f7b91-377b-46ad-9d2f-d3661dbfdab4

## Requirements

**Changelog**

:cl: Dygon
- fix: Evacuation doors(pods, lifeboats) are now unable to be pryed open while locked.
- fix: Being outside the docking area when a lifeboat has left will now instantly kill you.

### DIFF
--- a/Content.Server/Shuttles/Systems/ShuttleSystem.FasterThanLight.cs
+++ b/Content.Server/Shuttles/Systems/ShuttleSystem.FasterThanLight.cs
@@ -1,6 +1,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Numerics;
+using Content.Server._RMC14.Shuttles;
 using Content.Server.Shuttles.Components;
 using Content.Server.Shuttles.Events;
 using Content.Server.Station.Events;
@@ -393,6 +394,10 @@ public sealed partial class ShuttleSystem
         var ftlMap = EnsureFTLMap();
         var body = _physicsQuery.GetComponent(entity);
         var shuttleCenter = grid.LocalAABB.Center;
+
+        // RMC14
+        var beforeFTL = new BeforeFTLStartedEvent(uid);
+        RaiseLocalEvent(uid, ref beforeFTL);
 
         // Leave audio at the old spot
         // Just so we don't clip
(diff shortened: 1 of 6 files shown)


DrSmugleaf (changes requested):
Problem with this is that currently marines can get locked outside, so this would lead to a situation where xenos can't get to them and the round end stalls
13 solves this by killing anyone stuck in-between the doors or outside after the lifeboat launches

Arendian:
> Problem with this is that currently marines can get locked outside, so this would lead to a situation where xenos can't get to them and the round end stalls 13 solves this by killing anyone stuck in-between the doors or outside after the lifeboat launches

Lifeboats now spawn a kill area after they go into FTL that lasts for 30 seconds, I also added a video demonstrating it.

DrSmugleaf (approved):
Thanks for the PR!

PR #5637, opened 2025-02-15
Fix intel headband and squad beret

Dutch-VanDerLinde:
## About the PR
Bugfix

:cl:
- fix: Fixed intelligence headband and squad beret not spawning


DrSmugleaf (approved):
Thanks for the PR!

PR #5677, opened 2025-02-20
make booze dispenser invincible

QuietlyWhisper:
## About the PR

Fixed booze dispenser being destructible, which was not intended behavior.
They're apparently supposed to be part of vendors being tipped over which makes sense to me.

**Changelog**

:cl: Whisper

- fix: Fixed booze dispensers being destructible.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Science/Dispensers/booze.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Science/Dispensers/booze.yml
@@ -21,6 +21,7 @@
       - Xeno
   - type: ApcPowerReceiver
     needsPower: false
+  - type: Godmode # todo rmc14 we really need the technology to remove single components because copying an entire parent is actually torturous and prone to errors in the future
 
 - type: reagentDispenserInventory
   id: CMDispenserBoozeInventory


Dutch-VanDerLinde:
Why can't you just remove destructible from the parent

I'm pretty sure none of the vendors are supposed o be destructible

QuietlyWhisper:
> Why can't you just remove destructible from the parent
> I'm pretty sure none of the vendors are supposed o be destructible

there's like 5 nested parents two of which have their own damageable stuff that needs to be removed to the point where I asked smug if we can get component removal technology because this is stupid as fuck and you shouldn't need to do all that to remove components

Dutch-VanDerLinde:
> > Why can't you just remove destructible from the parent
> 
> > I'm pretty sure none of the vendors are supposed o be destructible
> 
> 
> 
> there's like 5 nested parents two(?) of which have their own damageable stuff that needs to be removed to the point where I asked smug if we can get component removal technology because this is stupid as fuck and you shouldn't need to do all that to remove components
> 
> 
> 
> but maybe I made a mistake when I was checking

Wait, this isn't the correct booze vendor

This is the booze vendor which is the bartenders dispenser chemistry thing

The thing that is the real vendor is the booze-o-mat, that one only has 1 destructible parent

Dutch-VanDerLinde:
500 HP destructible even has a note "TODO vendor tipping"

Dutch-VanDerLinde:
> is there two different dispensers then? because this is a dispenser with bottles as well

Yes, the CM made vendor
Not the upstream booze dispenser

Dutch-VanDerLinde:
If you destroy the upstream esque booze dispenser it'll just drop the booze on the floor it also mainly uses upstream bottles which can't be Molotov

QuietlyWhisper:
I see the problem now

AverageNotDoingAnythingEnjoyer:
So uh...tipping vendors PR got closed, might merge this as hotfix, but don't forget to add a comment that it's not a permanent change.

PR #5689, opened 2025-02-21
fix more tests

Ian321:
## About the PR

This PR fixes some more tests.

## Why / Balance

More test fixes.

## Technical details

- Reset and ignore the FailAndStartPresetTest as we do not have nukeops.
- Adjusted the wiggleroom in PullingTest, as pulling and movement are different in RMC.
- Ran the DB migration script because someone removed a duplicate underscore from a column.
- Made "MapInsertComponent" a Spawner, cause it is one.
- Removed "parent: 2" from map inserts, as they referenced the wrong/nonexistent parent.

I have yet to find the one that causes the client/server pair to die.

## Requirements

**Changelog**

No CL


DrSmugleaf (approved):
Thanks for the PR!

PR #5708, opened 2025-02-23
Add early evolution point boost

Vermidia:
## About the PR

Title. For the first 15 minutes of the game, xenos gain 2 instead of 1 point every 2 seconds (so 1 per second here)

## Why / Balance

parity!

## Technical details

New var for the time, component now has an early point boost comp in it, if it's before the timeup than they get the boost.

## Requirements

**Changelog**

:cl:
- fix: Fixed xenos not getting double non-hivecore evo points before 15 minutes.


DrSmugleaf (approved):
Thanks for the PR!

PR #5714, opened 2025-02-23
Acids-The-Map is no more

AverageNotDoingAnythingEnjoyer:
## About the PR

Made Indestructible podlocks non-corrodible

## Why / Balance

I don't think they should be corrodible. 
Maps use these to mark fully unaccessible zones, so Xenos being able to access them shouldn't happen.
Example is podlocks at the corners of the map which lead nowhere but can still be acided.

## Technical details

Just made indestructible prototype noncorrodible and made the destructible podlocks corrodible since they have noncorrodible one as their parent.

## Media

https://github.com/user-attachments/assets/19afab97-afec-4e14-8fc7-52b1efe9f861

## Requirements

**Changelog**

:cl: J C Denton
- fix: Indestructible podlocks are now unacidable too.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Doors/Shutters/poddoor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Doors/Shutters/poddoor.yml
@@ -43,6 +43,8 @@
   - type: StaticPrice
     price: 280
   - type: RMCPodDoor
+  - type: Corrodible
+    isCorrodible: false
 
 - type: entity
   parent: RMCShutterBaseIndestructible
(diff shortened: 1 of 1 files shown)


AverageNotDoingAnythingEnjoyer:
> they're supposed to be corrodible, if I had to guess the actual issue is the mapped ones you're referencing specifically are not supposed to be corrodible

Alright, I'll create a separate prototype for unacidable podlocks then, if that's alright?

AverageNotDoingAnythingEnjoyer:
I think with the current amount of podlocks in spawn menu it's just redundant to make the same prototypes but with 1 parameter changed so closing this for now.

PR #5745, opened 2025-02-26
Ceremonial Sabre proto

Vermidia:
## About the PR

Title.

## Why / Balance

Since we have a resprite sabre and scabbard already, adds the ceremonial sabre and scabbard protos. Like CM, it's a reskinned machete and the scabbard only fits on the belt.

## Technical details

YML mostly.

## Requirements

**Changelog**

no cl admeme only for now...


mmails:
iirc scabbard might also fit on the back slot not sure I haven't tried it in cm

Vermidia:
I checked and it only has sprites and slots for the waist. Fun fact is that the sabre by itself does fit in the suitslot of some of the royal marine things.

DrSmugleaf (approved):
Thanks for the PR!

PR #5767, opened 2025-02-28
Fix neuro and ammo knockback being too slow.

Vermidia:
## About the PR

Title.

## Why / Balance

They are very fast in CM, it's not very long from shot to throw ending, so tried to match it better than just putting it on min speed.

## Technical details

Change the speed numbers, make the lil less forceful so we don't overshoot so much. Also removed the can move part from stuns,, CM doesn't check if you can.

## Requirements

**Changelog**

:cl:
- fix: Fixed knockback from ammo stuns and neurotoxin forced movement being too slow
- fix: Fixed knockback from ammo stuns not applying in certain cases.


DrSmugleaf (approved):
Thanks for the PR!

PR #5797, opened 2025-03-03
Fix raffles being 10 seconds too long

Arendian:
## About the PR

Raffles now have the proper timer again.

## Requirements

**Changelog**

No CL


QuietlyWhisper:
What does this actually do

Arendian:
> What does this actually do

In the parasite spawn delay PR I clamped the value to min 10 seconds, this then got added to the original delay. When I tested in dev it showed as the correct 10 second regular raffle duration, but since it has quicklottery turned on by default it was actually 20 seconds.(which is currently the value on live). This PR makes that minimum value added 0 instead of 10.

DrSmugleaf (approved):

PR #5807, opened 2025-03-03
Fix tackle stacks getting reset on every knockdown, fix infected not staying down

Vermidia:
## About the PR

Title.

## Why / Balance

Bug and Parity. Knocked down runs *everytime* the status runs, while in CM it needs to run only when the state changes for real, whiched the downed event does. And also can probably handle more cases. Also infected staying down is parity, they don't get stacks removed this way. Fixes #5807

## Technical details

Knocked down event -> DownedEvent
Check for not having infected comp to remove stacks.

## Requirements

**Changelog**

:cl:
- fix: Fixed infected hosts losing tacklestacks and not consistently being able to keep them tackled down.
- fix: Fixed tacklestacks getting removed if you would get knocked down while already knocked down.

### DIFF
--- a/Content.Shared/_RMC14/Tackle/TackleSystem.cs
+++ b/Content.Shared/_RMC14/Tackle/TackleSystem.cs
@@ -1,5 +1,6 @@
 using Content.Shared._RMC14.Xenonids.Parasite;
 using Content.Shared.Administration.Logs;
+using Content.Shared.Buckle.Components;
 using Content.Shared.CombatMode;
 using Content.Shared.Damage.Systems;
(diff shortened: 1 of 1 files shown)


DrSmugleaf (changes requested):
Problem with downed event is that it would (might?) apply to things like laying on roller beds, which is technically parity even if stupid, but here roller beds do not break in a single hit from a xeno

Vermidia:
I could like, make it not apply if someone's buckled? Since rollerbeds count as buckling.

DrSmugleaf (approved):

PR #5842, opened 2025-03-07
build speed of queen fixed

N0lim:
## About the PR
build speed delay of queen is 2 seconds now (the same as drone)

## Why / Balance
Parity
https://github.com/cmss13-devs/cmss13/blob/c20d21622048d99106e10ed40c4bc2cb1a2ebd56/code/modules/mob/living/carbon/xenomorph/castes/Queen.dm#L20
https://github.com/cmss13-devs/cmss13/blob/c20d21622048d99106e10ed40c4bc2cb1a2ebd56/code/__DEFINES/xeno.dm#L77

## Requirements

**Changelog**

:cl:
- fix: Fixed Queen building speed being half the intended amount.


DrSmugleaf (approved):
Can fix the range later

PR #5850, opened 2025-03-07
Acid spray extinguishing tile fires

Tyzemol:
## About the PR
Xeno spray now extinguishes tile fires.

## Why / Balance
parityy

## Technical details
Now, here is the reason i open it as a draft, the code is done, it works, but...
No matter how many times i re read CM13 code for this, i see the extinguishing power of every acid type
and i see how that value is applied, but then i see that _green fire specifically_ takes tripple the damage from acid sprays, i have not played CM13, i have no clue why it has to take tripple the damage, but the way it works out in current RMC14 is boiler can completely remove freshly placed green fire in one spray, meanwhile regular orange fire and blue fire still have most of their duration left.
Meanwhile weakest spitter spray removes meager 6 seconds of orange fire duration, out of 60 total.

I hardly believe that fancy green fire is supposed to be weaker than orange fire so i assume i am missing some critical information, either RMC14 just doesnt have full tile fire parity, and orange fire is not supposed to have same burn duration as green, or there is other more complex reason.

So i will leave maintainers to decide what do i even do with this.
## Media
https://github.com/user-attachments/assets/e0f5e751-ccf9-4bf0-b7a2-98e4c9a62554

## Requirements

**Changelog**
:cl:
- add: Xeno sprays now extinguish tile fires.

### DIFF
--- a/Content.Shared/_RMC14/Atmos/ExtinguishFireComponent.cs
+++ b/Content.Shared/_RMC14/Atmos/ExtinguishFireComponent.cs
@@ -12,4 +12,7 @@ public sealed partial class ExtinguishFireComponent : Component
 
     [DataField, AutoNetworkedField]
     public CollisionGroup Collision = CollisionGroup.MobLayer | CollisionGroup.MobMask;
+
+    [DataField, AutoNetworkedField]
+    public float? ExtinguishingPower;
 }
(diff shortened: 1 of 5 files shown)


Vermidia:
Green fire in CM extinguisher faster through spray acid and pats, it's basically supposed to be easy to put out iirc.

PR #5874, opened 2025-03-11
Mortar operator and loader pamphlets now change your job title and job icon

Dutch-VanDerLinde:
## About the PR
Now only riflemen can use them
Adds the icon and special title like CM13

image
image
image

## Technical details
Adds a new system for getting squad dogtag titles ``GetMarineSquadNameEvent``
Adds ``MapBlipIconOverrideComponent`` for overriding tacmap icons

## Media

:cl:
- add: You now get a special job icon and title when using a loader or mortar operator pamphlet.
- fix: Fixed roles other than riflemen being able to use the loader or mortar operator pamphlet.


DrSmugleaf (approved):
Thanks for the PR!

This will need to either modify the actual job or just what shows up in a future implementation of an in-game crew manifest (so that it doesn't actually change the playtime tracker, the latter is probably better)

PR #5875, opened 2025-03-12
Remove intel squad stripes

Dutch-VanDerLinde:
## About the PR
From CM13

Looks better on intel rifleman


DrSmugleaf (changes requested):
Would it be possile to keep goggles but not armor and helmet?

DrSmugleaf (approved):
Thanks for the PR!

PR #5913, opened 2025-03-16
Fix intelligence officer ID card label

Dutch-VanDerLinde:
## About the PR
Fixes this:
image

After:
image

## Requirements

:cl:
- fix: Fixed intel officer ID job title just saying 'Intel', it will now display 'Intelligence Officer' as intended.


DrSmugleaf (changes requested):
Having to do this means that the role isn't getting read properly during squad assign
Proper fix would be moving the code in AssignSquadComponent MapInitEvent into an update loop that removes the component I think

Dutch-VanDerLinde:
> Having to do this means that the role isn't getting read properly during squad assign Proper fix would be moving the code in AssignSquadComponent MapInitEvent into an update loop that removes the component I think

could this be added as-is? squad name issue still appears if you transfer an intel officer
for example if you transfer IO to delta nd then back to intel it'll apepar as Intel Intelligence Officer which is weird

DrSmugleaf (approved):

PR #5930, opened 2025-03-16
Fix Morgue/Crematorium capacity

stranthbackup:
Remove Morgue and Crematorium capacity.
Parity.

## About the PR

Remove Morgue and Crematorium Cap
## Why / Balance

Parity
## Technical details

Removed Capacity flag from Storage component of the Crematorium and Morgue.
## Requirements

**Changelog**
:cl:
- fix: Fixed the crematorium and morgue having a maximum capacity of 1. They now have a maximum capacity of 30.


Resources/Prototypes/Entities/Structures/Storage/morgue.yml
```diff
@@ -38,7 +38,7 @@
   - type: EntityStorage
     isCollidableWhenOpen: true
     showContents: false
-    capacity: 1
+    capacity: 999
```
DrSmugleaf, line 41: For this it's probably better to have a limit even if very high so we don't get ugly surprises with someone putting 1000 glass shards in it
30 is around the same as crates
```suggestion
    capacity: 30
```

Resources/Prototypes/Entities/Structures/Storage/morgue.yml
```diff
@@ -115,7 +115,7 @@
...
```
DrSmugleaf, line 118:
```suggestion
    capacity: 30
```

DrSmugleaf (changes requested):
Thanks for the PR!

PR #5941, opened 2025-03-18
Fix bipod not slowing on break + fix drop not breaking it

### LINKED ISSUE #5918: Bipod needs to undeploy on drop + slow
## Description

On dropping a weapon with a bipod, it slows the person who dropped it by 4 seconds, and superslows for 2 (this is the status effect numbers already doubled.

Noting this here, not sure if I'm going to work on it rn.

Note the slow should apply when undeploying through anything but pressing the undeploy button aswell, so if you look away.

**Reproduction**

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Title.

## Why / Balance

Parity. Bipod breaks on drop in CM. And it slows you down if you don't toggle it before any of the forced breaking stuff happens.

Technically unequipping doesn't breaking it but requipping does (shrug) so I just put it on unequip aswell anyway.

Fixes #5918

Bipod requiring anchoring isn't parity, but it's silly that it doesn't honestly. A loose cade isn't very stable. Dragging one around in CM would be harder as it slows you down alot more (we don't have generally slower pull speed for alot of objects atm) and tiles block a lot more. It just makes more sense to check anchored. Technically the facing direction also needs to match which we don't have that being check either.

## Technical details

DroppedEvent is relayed to the attachment holder now.

New fields for breaking on drop, and slowing on break. Slow atm is hardcoded for timings because I don't really think anything else has the behavior but if it does with diff numbers it won't be hard to add in.

## Media

https://github.com/user-attachments/assets/31ccb18b-abb3-4d33-8a07-9e0aa906251f

## Requirements

**Changelog**

:cl:
- fix: Fixed bipod not breaking on drop.
- fix: Fixed bipod not slowing you down if toggled by anything but the action. Manually undeploying the bipod with the action still won't slow you down.
- fix: Fixed bipod being instant deployable on non-anchored barricades.

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Components/AttachableToggleableComponent.cs
+++ b/Content.Shared/_RMC14/Attachable/Components/AttachableToggleableComponent.cs
@@ -47,6 +47,18 @@ public sealed partial class AttachableToggleableComponent : Component
     public bool BreakOnFullRotate = false;
 
     /// <summary>
+    /// If set to true, the attachment will deactivate upon being dropped.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool BreakOnDrop = false;
+
+    /// <summary>
+    /// If set to true, the attachment will slow the user upon being toggled due to any source but toggling it manually.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public bool SlowOnBreak = false;
+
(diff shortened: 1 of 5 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #5943, opened 2025-03-18
Fix acid melt times

Vermidia:
## About the PR

Title.

## Why / Balance

parity. Ours was too long. The mistake was assuming items take 9 ticks to melt, they only take 3~ in parity. But in reality is 8 and 2 because the first tick is basically skipped to set up time stuffs and it deletes when it hits 0. So all acid times are much much faster.

Walls however now take longer because they count as turfs. Might be other things that need that but I put it on base wall for now.

## Technical details

New Datafield for melt time mult, otherwise it's just YML. Also because of the -1 tick thing barricades were lasting 5 seconds too long.

## Requirements

**Changelog**

:cl:
- fix: Fixed acid taking 4.5x as long as it should be to melt most things. Walls melting lasted 1.5x as long as they should have.
- fix: Fixed acid on barricades lasting 5 seconds too long.


DrSmugleaf (approved):
Thanks for the PR!

PR #5944, opened 2025-03-18
Split Brute Armor into three.

stranthbackup:
## About the PR

Changed CMArmorComponent and CMArmorSystem to facilitate Bullet Armor. Xenos now have XenoArmor in their prototypes, this is still the same as old Armor. Humanoids now have Melee and Bullet armor values. All relevant YMLs have been updated, as well as a couple of systems that enable some Xeno Castes to gain more armor. All jackets and uniforms no longer have resistance versus bullets. Unless you're SPP for some reason, ya dirty commie!
## Why / Balance

Parity!
## Technical details

Armor is now split into XenoArmor, Melee and Bullet. All Castes and Armor Variants have been updated to reflect this change.
## Requirements

**Changelog**
:cl:
- fix: Fixed armors using the same value when resisting bullet and melee damage. This makes the M3-VL and M3-EOD heavy armor better at resisting bullet damage for example.

### DIFF
--- a/Content.Shared/_RMC14/Armor/CMArmorSystem.cs
+++ b/Content.Shared/_RMC14/Armor/CMArmorSystem.cs
@@ -2,6 +2,7 @@ using System.Linq;
 using Content.Shared._RMC14.Medical.Surgery;
 using Content.Shared._RMC14.Medical.Surgery.Steps;
 using Content.Shared._RMC14.Projectiles;
+using Content.Shared._RMC14.Weapons.Ranged;
 using Content.Shared._RMC14.Xenonids;
 using Content.Shared._RMC14.Xenonids.Projectile;
 using Content.Shared._RMC14.Xenonids.Projectile.Spit.Slowing;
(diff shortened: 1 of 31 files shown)


Content.Shared/_RMC14/Armor/CMArmorSystem.cs
```diff
@@ -227,14 +260,28 @@ private void ModifyDamage(EntityUid ent, ref DamageModifyEvent args)
...
                 }
             }
         }
 
         args.Damage = new DamageSpecifier(args.Damage);
-        Resist(args.Damage, ev.Armor, ArmorGroup);
-        Resist(args.Damage, ev.Bio, BioGroup);
+        if (!HasComp<XenoComponent>(ent))
+        {
+            if (HasComp<RMCBulletComponent>(args.Tool))
```
DrSmugleaf, line 271: Can't projectile component be checked instead of a new component?

Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/spp.yml
```diff
@@ -89,7 +94,8 @@
     walkModifier: 0.62
     sprintModifier: 0.62
   - type: CMArmor
-    armor: 30 # CLOTHING_ARMOR_HIGHPLUS (bullets), CLOTHING_ARMOR_MEDIUMHIGH (melee); choosing an intermediate value until more damage types are implemented
+    melee: 25 # CLOTHING_ARMOR_MEDIUMHIGH (melee)
+    bullet: # CLOTHING_ARMOR_HIGHPLUS (bullets)
```
DrSmugleaf, line 98:
```suggestion
    bullet: 35 # CLOTHING_ARMOR_HIGHPLUS (bullets) 
```

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Armor/CMArmorSystem.cs
stranthbackup: I can check again since I forget if I was having an issue with spit projectiles being reduced by bullet armor. This component just allowed me to make a simpler check

stranthbackup: Yea, XenoBaseProjectile also has the Projectile component. I can make it check for the Tools ProjectileComponent and not the XenoProjectile component if you would rather have it like that

Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/spp.yml
stranthbackup: Oof that would have been b a d. Thanks for catching that!

Content.Shared/_RMC14/Armor/CMArmorSystem.cs
DrSmugleaf: Fine like this then

DrSmugleaf (approved):

PR #5958, opened 2025-03-19
Aimed shot and AMR for sniper specialist

### LINKED ISSUE #5551: AMR Specialist

Arendian:
## About the PR

 I ported over aimed shot and the XM43E1 anti-materiel rifle from CM13.
 Sniper loadouts now contain a spotter kit.
Requisitions and vendors now have flak and XM43E1 marksman ammo.

Below you'll find the effects of the ammo types, if you notice anything that's not parity or still missing please point it out.

**M96:**
- Marksman ammo now deals 70 damage with a normal shot and 210 damage with an aimed shot.
- Flak ammo now deals 55 damage against the hit target, 55(110 for xenonids) to entites in a 1 tile radius, and half of that to any entities in a 2 tile radius.
- Flak's Aimed shot does either the basic effect if it  hits any entity that's not the target, or deals 110 damage and superslows the target.
- Incendiary ammo now does 60 damage and has double the fire intensity compared to what it had before this PR.
- Incendiary's aimed shot gives the hit target an additional 10 firestacks and slightly blurs their screen for a few seconds.
- All projectiles shot using aimed shot are homing.
- Aimed shot has a base aim time of 1.25 seconds, this is increased by 0.05 seconds per tile of distance to the target.
- Aimed shot has a cooldown of 2.5 seconds, this cooldown starts the moment you start aiming.
If the sniper laser is turned on, the aim time is multiplied by 0.6; if the target is spotted, the aim time is multiplied by 0.5; and if both, the aim time is multiplied by 0.35.

**XM43E1**
- Marksman ammo's normal shot deals 125 damage with 75 AP.
- Ammo penetrates entities,damage and range is reduced with every entity hit:
  -  Any soft target the same size or smaller than a T2 xenonid multiplies the projectile damage by 0.8 and reduces the range by 3 tiles per hit.
  -  T3 xenonids multiply the projectile damage by 0.6 and reduce the range by 6 tiles per hit.
  - Walls and other obstacles multiply the projectile damage by 0.4 and reduce the range by 9 tiles per hit.
- An aimed shot using the XM43E1 will get a different effect based on what type of target is hit on top of the base projectile damage:
  - Anything smaller than a T1 xenonid will take 80% of the original damage as additional damage.
  - Anything the size of a T1 xenonid will take 60% of the original damage dealt and 10% of their current health as additional damage.
  - Anything the size of a T2 xenonid will take 20% of the original damage and 20% of their current health as additonal damage.
  - Anything the size of a T3 xenonid will take no bonus damage from the projectile, but will take 30% of their current health as bonus damage.

  - If the amount of damage dealt is high enough, T2 and lower xenonids will be knocked back and stunned for a short amount of time.; T3 xenonids will be stunned for a short time, interrupting all current action; and humanoids will take the full damage as stamina damage.
  - All bonus damage will be multiplied by 0,.5 on the first shot, 0.75 on the second shot on the same target and 1.0(so no reduction) on the third shot on the same target. Using an aimed shot on a different target will reset this streak.
- All projectiles shot using aimed shot are homing.
- Aimed shot has a base aim time of 2 seconds, this is increased by 0.05 seconds per tile of distance to the target.
- Aimed shot has a cooldown of 4.5 seconds, this cooldown starts the moment you start aiming.

## Why / Balance

Resolves #5551 

I doubled the firestacks of the incendiary bullets to make their damage over time the same as in CM13.

Current balance issue that probably need to be addressed before this is merged:
- Superslow on xenonids in RMC14 is currently only 25% , in CM13 superslow roughly slows somewhere in between 50%-60% based on caste. This makes flak ammo's aimed shot effect almost unnoticeable until changed.

## Technical details

The  TargetingSystem and component manage the duration of any entities targeting another entity. An event is raised when targeting is started, finished or cancelled.

The  FocusingSystem and component update focus stacks of an entity that is targeting something every time an aimed shot is performed. Every time an aimed shot is performed at the same target the focus stack and a bonus damage multiplier connected to it are increased.

The  HomingProjectileSystem and component update the velocity of a projectile every tick so that it's always on course towards it's target. The component is removed right before the target is hit.

The AimedProjectileSystem and component manage any damage modifiers and effects a projectile should get if shot using an aimed shot.

The AreaDamageSystem and component manage dealing damage in an area around a hit entity. It does an entitylookup in an area set on the component.

The StoppingPowerSystem and component manage any effects applied to a target based on how much damage is dealt to them in a single hit. Big xenonids require a higher amount of stopping power before effects apply to them.

The SpottingSystem and component mark an entity as spotted when the spotting action is used. When someone uses aimed shot at a spotted entity, the aim duration is reduced.

The AimedShotSystem and component shoot a projectile after targeting successfully finishes. After a certain amount of time specified on the component, the gun will shoot at the target.

The PenetratingProjectileSystem manages how much damage and range a projectile loses if it hits another entity and if it can penetrate it.

The TargetingOverlay draws a line between any targeting entities and their targets.

The BlindSystem enables the BlurOverlay  while a target is blinded.

## Media

https://github.com/user-attachments/assets/b8ab1656-d6c1-4a3a-9aec-c19d90988388

## Requirements

**Changelog**

:cl: Dygon
- add: Sniper rifles can now be used to perform aimed shots, projectiles fired with an aimed shot will have additional effects.
- add: Sniper specialists' loadouts now include a spotter kit, which contains the gear needed to transform any rifleman into a fully-fledged spotter.
- add: Sniper specialists can now obtain the XM43E1 anti-materiel rifle from their vendor. This rifle fires projectiles powerful enough to penetrate any obstacle in its path.
- Removed: Sniper ammo no longer has its placeholder 2x damage modifier.


mmails:
Does this finally complete the sniper specialist

Ariphylor:
N.. NO WAY..
ITS _ACTUALLY_ HERE?
AND ITS NOT JUST AIMED SHOT.. COMPLETE AMMO TYPES, AMR WS KIT..? _THIS_ **SINGLE** PR WILL CHANGE THE BALANCE OF THE GAME!!
Jokes aside, this is peak. Hope it gets merged. Also yes I believe this is all Sniper WS content, aside from changes needed mentioned in the PR by the author. Good work!

Arendian:
I added a video and technical details.

Content.Shared/_RMC14/Projectiles/Aimed/AimedProjectileComponent.cs
```diff
@@ -0,0 +1,13 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Projectiles.Aimed;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class AimedProjectileComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public EntityUid Target;
```
DrSmugleaf, line 9: Aimed shot projectiles are also supposed to track their target, so that you can't just sidestep to dodge them
I can try to code that but it would basically be setting the linear velocity of the projectile in an update loop to aim towards the target

DrSmugleaf (commented):
Thanks for the PR!

Content.Shared/_RMC14/Projectiles/Aimed/AimedProjectileComponent.cs
Arendian: I separated that functionality from the aimed projectile with the HomingProjectileComponent so that it can be used for other stuff than aimed shot too. It does what you describe in an update loop.

DrSmugleaf: That's fine then all good

DrSmugleaf (approved):
Thanks for the PR!

PR #5959, opened 2025-03-19
Pilot lights were not hot. Added heat to Pilot lights!

### LINKED ISSUE #5545: Paper does not take damage from fire
Paper does not take damage from fire stacks (tested), resulting in the fire stacks expiring and the paper just indefinitely on fire.

stranthbackup:
## About the PR

Added heat to pilot lights, now Marines can recklessly light their cigarettes on fire.
Added intensity to Paper so that it now doesn't burn forever. Goodbye eternal flame. Fixes #5545 

## Why / Balance

Cool Factor + Parity.
## Technical details

SharedFlamerSystem now checks if Igniter is active and throws an IsHotEvent matching the state of the igniter.
Added an intensity value of 10 to paper.yml so it can actually take damage. Damage is slow enough to be able to recover if needed.
## Requirements

**Changelog**


DrSmugleaf (approved):
Thanks for the PR!

PR #5962, opened 2025-03-20
Adds light variant of Commanding Officer Armor.

Arcticular1:
## About the PR
Added a light weight variation of Commanding Officer Armor.

## Why / Balance
A requested variation of CO armor. The armor is purely cosmetic changes. 
This PR is to give Commanding Officers more cosmetic options in terms of their kit, It a light weight armor option that is akin to the smart gun combat harness, rather than the current armor which is heavy in appearance. 
## Technical details
RSI, YML.
## Media
image
image

## Requirements

**Changelog**

:cl:
- add: Added M3-L pattern commanding officer armor.
- add: Added M3-L pattern commanding officer armor to CO's vendor.


Resources/Prototypes/_RMC14/Entities/Clothing/OuterClothing/marine_armor.yml
```diff
@@ -347,6 +347,31 @@
...
+    sprite: _RMC14/Objects/Clothing/OuterClothing/Armor/m3/co-light/jungle.rsi
+  - type: CMArmor
+    armor: 30
+    bio: 20
+    explosionArmor: 15
+  - type: SmartGunArmor
+  - type: Storage
+    maxItemSize: Small
+    grid:
+    - 0,0,5,1 # 3 slots
```
QuietlyWhisper, line 366: You can drop a ton of lines by inheriting from the CO armor and only modify the sprites
```suggestion
  parent: RMCArmorM3CO
  id: RMCArmorM3LCO
  name: M3-L pattern commanding officer armor
  description: A variant of M3 Pattern Commanding Officer Armor. A Robust, well-polished suit of armor for the Commanding Officer. Custom-made to fit its owner with special straps to operate a smartgun. Show those Marines who's really in charge.
  components:
  - type: Sprite
    sprite: _RMC14/Objects/Clothing/OuterClothing/Armor/m3/co-light/jungle.rsi
```

QuietlyWhisper (changes requested):
This is fine as a cosmetic but it probably shouldn't gaslight COs into thinking it actually is light armor

Better naming and describing examples would be like "alternative" or just a "variant"

PolarTundraa:
isn't this sprite a little misleading in that it looks like a vest while still protecting extremities when we get limb damage?

QuietlyWhisper:
> isn't this sprite a little misleading in that it looks like a vest while still protecting extremities when we get limb damage?

Also that, it might be better to just have an actual CO light armor but that would require design discussion with smug if it's not just a copy of regular light armor

Arcticular1:
> isn't this sprite a little misleading in that it looks like a vest while still protecting extremities when we get limb damage?

that is true, i didn't think of that, if needed i can get some armor on the arms (when wound med comes out i will 100% make sure that it isnt misleading)

Arcticular1:
> > isn't this sprite a little misleading in that it looks like a vest while still protecting extremities when we get limb damage?
> 
> that is true, i didn't think of that, if needed i can get some armor on the arms (when wound med comes out i will 100% make sure that it isnt misleading)

Also to add onto this, i did a little bit of digging and im not exactly sure how limb damage works.. because such as chief MP armor and as well as M4 armor, alot of these do not have sprites for arm armors.

Arcticular1:
> > isn't this sprite a little misleading in that it looks like a vest while still protecting extremities when we get limb damage?
> 
> Also that, it might be better to just have an actual CO light armor but that would require design discussion with smug if it's not just a copy of regular light armor

I would 100% prefer this to be a lighter variant. I'm not sure how balancing would go since the CO's current armor is already at a light speed modifier, anymore than that and i would believe it would be a balancing concern. If a consensus is made on it though, i would happily work on it!

kleinerstation13:
I will be closing this as the PR author has left the RMC community and will likely not continue to work on this PR in the future.

PR #5965, opened 2025-03-20
Use the new layer hiding to fix the gas mask snout bug

### LINKED ISSUE #5173: Gas Mask does not properly hide snouts
## Description

RMC Gas Mask's HideLayerComponent doesn't work. Other RMC masks (coif, sterile mask, rebreather) still properly hide snouts, but the gas mask doesn't.

**Reproduction**

Reproducible with any snouted species, but steps to reproduce for lizards:
1. Temporarily delete RMC-14/Resources/Textures/_RMC14/Objects/Clothing/Mask/gas.rsi/equipped-MASK-reptilian.png
2. Load up the game, make a lizard character, and put on a gas mask.
3. Snout clips through the mask.

Note that this doesn't happen if reproduced with the coif, or a sterile mask, or a rebreather; the default masks applying will smush the snout off of the lizard head, but the snout won't clip through. This is because the other masks still have functioning HideLayerComponents.

It is of course possible to work around this by making custom gas mask sprites for varying snout sizes, and this is preferable aesthetically anyway. But this is still a bug and should get fixed.

**Screenshots**
2024-12-16_23_16_52-Rounys_Marine_Corps_14
(A rodentia with a large bat snout clipping through the default human gas mask.)

**Additional context**

Mostly creating this issue as a reference point for a known issue in my next PR to add Rodentia. I have a solution to this, but it may break the gas mask's suit storage compatibility. If I can make a fix that doesn't break anything else, I'll follow-up with another PR to fix this.

paige404:
## About the PR
Fixes #5173 by using the new functionality added upstream to the HideLayerClothingComponent

## Why / Balance

Ugly bug on some sprites.

## Technical details

Uses the new `layers` Dictionary in `HideLayerClothingComponent` (introduced upstream here: https://github.com/space-wizards/space-station-14/pull/34080) to specify that a mask should only cover the snout when equipped on the mask slot. Without this change, it is possible that the gas mask will start hiding snouts when equipped to the suitstorage slot of an engineering vest. So, this both fixes the core of a bug but also prevents a new bug from being introduced by the fix upstream.

This SHOULD NOT BE MERGED until the upstream PR it depends on has been brought down here. It WILL break without those changes (which is why the YAML linter doesn't like it right now).

## Requirements

**Changelog**

:cl:
- fix: Particularly large-snouted members of their species should now be able to fit their noses into their gas masks.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
@@ -304,3 +304,4 @@
     sprite: _RMC14/Objects/Clothing/Mask/Keffiyeh/keffiyeh_red.rsi
   - type: Clothing
     sprite: _RMC14/Objects/Clothing/Mask/Keffiyeh/keffiyeh_red.rsi
+    
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/Masks/masks.yml
@@ -4,8 +4,8 @@
   abstract: true
   components:
   - type: HideLayerClothing
-    slots:
-    - Snout
+    layers:
+      Snout: Mask
     hideOnToggle: true
 
 - type: entity
@@ -68,7 +68,7 @@
     - Mask
   - type: Item
     size: Small
-    
+
 - type: entity
   parent: CMBaseMask
   id: RMCMaskKutjevoRespirator


paige404:
This has become a little more complicated. Problems in the upstream PR have led to it being reverted, per https://github.com/space-wizards/space-station-14/issues/36230. This will need to wait on a fix.

FairlySadPanda:
👋 This was only reverted for the April Fool's branch. The only bug is what happens if you have a critter with certain equipment on their loadout (e.g. a welding mask) - a test attempts to spawn the critter and the code gets sad because it's not able to properly hide the sprite layers.

The solution is really simple I think but I'm not currently prioritizing the code. But as long as you don't have non-humanoids wearing welding masks in their loadouts it's probably fine :)

paige404:
Okay, this should be good to go now.

DrSmugleaf (approved):

PR #5968, opened 2025-03-20
OT Chems (Currently non-functional)

stranthbackup:
## About the PR

This likely needs an entire Solution / Reagent refactor while also taking into account that Research needs the same Solution / Reagent refactor. As stated by Whisper, it is likely we get research chem before OT.

Added more reagent ymls for different chems.
Added more reagents as a base for future OT updates.
Added more chemical reactions so that these can get made.
Added descriptions for some of these, they are by no means done.
## Why / Balance

Parity
OT Chems don't have a current draft PR, maybe having this here will spur some inspiration in other folks to add functionality if I can't.
## Technical details

Ported OT chems with no functionality. (Yet?)
## Requirements

**Changelog**


stranthbackup:
Things of note:
  CM13 does a handle_volatiles things for explosions.
  Explosions don't throw other explosives they explode said explosives.
  Fire should also ignite other explosives, ie. Cooking grenades and the like.
  
  Napalm and Chlorine Trifluoride need an entirely new entity effect that spawns a tile fire of some radius. This entity effect needs to check the solution container for water and the reagent being produced. If the amount of either of these two goes over the amount of water it would create a tile fire with the properties of the total amount of volatile substance in the container. A suggestion was to make it a separate reaction that spawns the effect, perhaps having actual napalm production need water as a catalyst, but that would make you only need 1u of water for any amount of napalm.
  
  ANFO needs to explode if you make more than 60.01u of it in a single reaction, it shouldn't check for the total amount of reagent in the container, therefore ReagentThreshold effect condition would not be suitable to check on before an explosion occurs.
  
  Can water mitigate ANFO explosion? Likely not.

PR #5969, opened 2025-03-20
Fixes PassThrough on map inserts

CyberTropic:
## About the PR

Space tiles are treated as pass through tiles

## Why / Balance

Bug fix, allows using non-square shapes on map inserts

## Technical details

The bounding box was being drawn for the whole transform and not each specific fixture

## Media

image

## Requirements

**Changelog**

:cl:
- fix: Fixed map insert pass through.


DrSmugleaf (approved):
Thanks for the PR!

PR #5973, opened 2025-03-21
increase tailstab range by .5

DummeHytteOst:
## About the PR
increases tailstab range by .5 (2>2.5)

apparently in parity you can hit someone from 2 tiles diagonally, which is about 2.83 tiles, but i think its better to not increase it that far, considering we use circles and not tiles, and i do not know if that is factual, but from how other abilities in cm13 work, i would imagine it is true.

## Why / Balance
identical reason as #5114 

## Requirements

**Changelog**
:cl:
- fix: Fixed tail stab range being 0.5 tiles too short (2 > 2.5).


DrSmugleaf (approved):
Thanks for the PR!

PR #5982, opened 2025-03-23
CLEAR FOR SADAR!!!

0tito:
## About the PR

AP Sadar rockets now gib marines

## Why / Balance

Boom, also parity i think

## Media

https://github.com/user-attachments/assets/cd5cc68d-3137-45f8-a726-dc7d96caef6b

## Requirements

**Changelog**

- tweak: AP Sadar rockets now gib marines on hit


0tito:
Also i coded this in 5 mins, i didnt check if there was something like this, i just created a new component, if theres stuff to fix lemme know, but it seems to work fine

DrSmugleaf (changes requested):
Looking into this it is only supposed to gib if the severity of the explosion is 200 after armor is factored in, so some armors like heavy, demolitionist and grenadier should resist the gib
Still have to decide if we want this type of gibbing in or not, but at a base level we would need that if we want it

0tito:
Alright form what I've heard this should be closer to parity

0tito:
https://github.com/user-attachments/assets/ade18e78-63f4-4b49-9a6c-3dda81ccaae1

Dutch-VanDerLinde:
image

This is the code for CM13 rocket gibbing, it basically makes the human explode twice but localized

Dutch-VanDerLinde:
350 is the severity, null direction, cause data is the thing that caused the explosion,

AverageNotDoingAnythingEnjoyer:
Don't forget to merge master for text fixes.

AverageNotDoingAnythingEnjoyer:
This should also work with regular rockets, since they deal 350 explosion damage but their armor pen is lower so it's less likely to happen

https://github.com/cmss13-devs/cmss13/blob/master/code/datums/ammo/rocket.dm#L34

PursuitInAshes:
Thanks for the PR!
I'm going to close this due to us generally wanting to be more restrictive on what can round remove (RR) someone compared to elsewhere. In the time between this PR being put up and now, it was decided that AP SADAR rockets are included in the list of things we don't want to port RRing from

PR #5983, opened 2025-03-23
Water now extinguishes fires

0tito:
## About the PR

RMC Water now extinguishes fire, same as vanilla ss14 water. I did not make it so that fire cannot be in water tho, so explosives and molotovs will still spawn fire in water, and if you are on top of it you will still get damaged.

## Why / Balance

Its kinda dumb it doesnt

## Media

https://github.com/user-attachments/assets/b010f02f-0808-4903-a7c8-3d5e6076d706

The video had to be smaller than 10mb so sorry about the potato quality

## Requirements

**Changelog**

- fix: Fixed water not extinguishing fire when you walk on it


DrSmugleaf (changes requested):
This is only supposed to instantly extinguish you if you resist fire while on water

PR #5992, opened 2025-03-24
Doafter bars now have the same opacity as the entity performing the action.

Arendian:
## About the PR

The opacity of a doafter bar is now the same as an entity's opacity.
Burrowers still see their own doafter bar while unburrowing, but nobody else sees it.
Lurkers, scouts and snipers have their doafter bar the same opacity as their invisibility.

## Why / Balance

Parity, the bars are still a little more visible than the CM13 circles even when transparent because of the used colours.

## Technical details

Before the doafter bar is drawn, it's alpha is set to the same value as the alpha of the entity's sprite.
If an entity has the  EntityActiveInvisibleComponent, the opacity is set to the same level as defined in the component.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Doafter bars now have the same opacity as the entity performing the action, except for some types of hostile doafters.

### DIFF
--- a/Content.Client/DoAfter/DoAfterOverlay.cs
+++ b/Content.Client/DoAfter/DoAfterOverlay.cs
@@ -118,20 +118,23 @@ public sealed class DoAfterOverlay : Overlay
                 }
 
                 //RMC14
-                // Don't show the doafter bar to other clients if the entity's sprite isn't visible.
-                if(!sprite.Visible && uid != localEnt)
-                    continue;
+                if (!doAfter.Args.ForceVisible)
+                {
+                    // Don't show the doafter bar to other clients if the entity's sprite isn't visible.
+                    if(!sprite.Visible && uid != localEnt)
+                        continue;
 
-                // Set the doafter bar alpha to the alpha of the sprite.
-                alpha = sprite.Color.A;
+                    // Set the doafter bar alpha to the alpha of the sprite.
+                    alpha = sprite.Color.A;
 
-                // The system using this component does not edit the sprite alpha, so we use separate logic for it.
-                var invisibleQuery = _entManager.GetEntityQuery<EntityActiveInvisibleComponent>();
-                invisibleQuery.TryGetComponent(uid, out var invisible);
+                    // The system using this component does not edit the sprite alpha, so we use separate logic for it.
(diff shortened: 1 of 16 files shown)


Vermidia:
There is a small difference - For scout when doing a "Hostile" Action, the circle is fully visible
Which actions are hostile are the ones that uses the specific red circle doafter bar. I believe this effects everything like this.

Arendian:
> There is a small difference - For scout when doing a "Hostile" Action, the circle is fully visible Which actions are hostile are the ones that uses the specific red circle doafter bar. I believe this effects everything like this.

The only relevant ones I could find were planting explosives and starting a fireman carry so I made those two exempt from the doafter bar being less visible.

Vermidia:
Stripping someone also counts as a "hostile" doafter, as well as suiciding, loading a mortar, and resiting handcuffs/legcuffs

Xeno wise prying open doors is visible, + a few other minor things

Everything else we don't have yet.

Arendian:
> Stripping someone also counts as a "hostile" doafter, as well as suiciding, loading a mortar, and resiting handcuffs/legcuffs
> 
> Xeno wise prying open doors is visible, so is devouring/hauling, disabling sensor tower + a few other minor things
> 
> Everything else we don't have yet.

I added most of the things mentioned here, invisible people loading the mortar isn't an issue I think as scout doesn't have the skillset for it(I still added it as an exception though just in case).
I tested in CM13 and resisting cuffs actually still has the doafter opaque, only if cuffing or uncuffing someone else is it visible. I  also didn't make devouring an exception as lurkers already get out of stealth when they try to devour someone.

Vermidia:
No, visible doafters are ones marked with BUSY_ICON_HOSTILE in CM, neither of those use that type of doafter in CM so they match opacity.

Arendian:
> is looking through binocs or wielding your gun invisible? (bonus points for grabbing an invisible scout bag also being invisible) as a rouny main those are what I find scouts doing all the time and lets me cap them when they should have had an easy drop on me.

Thanks for mentioning the dragging effect, that one now also has the same opacity as as the entity being dragged, whistles now also don't put an exclamation mark above invisible entities.
 And as Vermidia said, the effects you mentioned are the same opacity as the scout's sprite.

DrSmugleaf (approved):
Thanks for the PR!

PR #6010, opened 2025-03-29
Parasite fix

Dutch-VanDerLinde:
## About the PR
 fixes parasite blindness being applied incorrectly, since the parasite removes the blindness effect after it falls off the host's mask slot it doesnt correctly keep them blinded

makes blindness effect apply after the mask falls off

## Media

blindness works
image

## Requirements

:cl:
- fix: Fixed parasite infect blindness being applied incorrectly.


DrSmugleaf (approved):
Thanks for the PR!

Issue #3656, opened 2024-08-14
can see ship announcements as xeno


lzk228:
![Screenshot_20240814_182917](https://github.com/user-attachments/assets/3268ebb9-2f5e-4e1b-a335-3161db681006)

PR #6014, opened 2025-03-29
Rework marine latejoin and earlyleave announcement

chavonadelal:
## About the PR

Bringing to parity, as far as I know, announcements about latejoin and earlyleave. 

If it is CO, XO or SEA, a big announcement with a sound signal is made, the sound signal of SEA is different.
If it is the head of the department, a message is sent to the department radio channel and to the Almayer channel.
If the department does not have its own radio channel, a message is sent to the Almayer channel.
If it is not the head of the department, a message is sent to the department radio channel.
If the department does not have a radio channel but it is a combat marine, we send a message to his squad channel
## Why / Balance

The big announcement about the arrival of senior officers and SEA will not go unnoticed, people will be informed. An additional message on the Almayer radio channel if the head of the department has arrived will keep the Almayer staff informed so that they can contact the head if necessary.
Resolves https://github.com/RMC-14/RMC-14/issues/3656

## Technical details

If the job prototype has JoinNotifyCrew enabled, an announcement with sound will be made. If not, the message will be sent to the common channel and, if applicable, the department's radio channel.

The department is determined by checking the DepartmentPrototype list, and the radio channel is fetched from the DepartmentRadio property. If the department does not have a radio channel, the system checks if the player is a combat marine and, in that case, attempts to get the squad’s radio channel. To avoid duplicate messages, already processed channels are tracked using a HashSet.

If no department channel is found or if the player is the head of the department (checked via the HeadOfDepartment property), the message is additionally sent to the common channel.

Additional fields for logical non-hardcoded communication have been added to prototypes
## Media

CO/XO/SEA Arrival Announcement
image

Announcement of the arrival of the head of the department
image

Announcement of the arrival of a regular employee of the department
image

## Requirements

**Changelog**

:cl:to4no_fix
- tweak: Reworked announcements about latejoin and earlyleave of Marines. Late joining of CO/XO/SEA is accompanied by a large announcement with a sound signal. SEA has its own sound signal. Late joining of the head of department (squad leader) is accompanied by a message on the radio channel of the department (squad radio channel) and the radio channel of Almaer. Late joining of an ordinary employee of the department (squad) is accompanied by a message on the radio channel of the department (squad). At early leave, an announcement occurs according to the same logic, only without a large announcement.


Content.Server/GameTicking/GameTicker.Spawning.cs
```diff
@@ -237,30 +250,6 @@ private void SpawnPlayer(ICommonSession player,
             var jobName = _jobs.MindTryGetJobName(newMind);
             _admin.UpdatePlayerList(player);
 
-            if (lateJoin && !silent)
```
DrSmugleaf, line 240: It's better to  keep this code but add a && false to minimize line changes, then comment // RMC14 next to it

Content.Shared/Roles/DepartmentPrototype.cs
```diff
@@ -47,6 +48,18 @@ public sealed partial class DepartmentPrototype : IPrototype
     /// </summary>
     [DataField]
     public bool EditorHidden;
+
+    /// <summary>
+    /// RMC14 for logical communication of the department and its radio channel.
+    /// </summary>
+    [DataField]
+    public ProtoId<RadioChannelPrototype>? DepartmentRadio { get; private set; }
```
DrSmugleaf, line 56: This also needs to be moved to our own partial of DepartmentPrototype, its in Content.Shared/_RMC14/Marines/Roles

Content.Shared/Roles/JobPrototype.cs
```diff
@@ -153,6 +154,12 @@ public sealed partial class JobPrototype : IPrototype
         /// </summary>
         [DataField]
         public List<ProtoId<GuideEntryPrototype>>? Guides;
+
+        /// <summary>
+        /// RMC14 for arrival notification sound if <see cref="JoinNotifyCrew"/> true.
+        /// </summary>
+        [DataField]
+        public SoundSpecifier LatejoinArrivalSound { get; private set; } = new SoundPathSpecifier("/Audio/_RMC14/Announcements/ARES/sound_misc_boatswain.ogg");
```
DrSmugleaf, line 162: Same here, we have our own JobPrototype file

DrSmugleaf (changes requested):
Thanks for the PR!

chavonadelal:
I did what you asked + remembered about the announcement at early leave and also redid this!

chavonadelal:
I made the requested changes and moved the logic to a separate file

Content.Server/Bed/Cryostorage/CryostorageSystem.cs
```diff
@@ -238,6 +240,8 @@ public void HandleEnterCryostorage(Entity<CryostorageContainedComponent> ent, Ne
             _stationRecords.RemoveRecord(key, stationRecords);
         }
 
+        _marinePresenceAnnounce.AnnounceEarlyLeave(ent, recordId, station, jobName); // RMC14
+        if (false) // RMC14
```
DrSmugleaf, line 244: Doing return here is better semantically and fool proofs it in case upstream changes this method in the future
```suggestion
        return; // RMC14
```

DrSmugleaf (changes requested):

DrSmugleaf (approved):

PR #6017, opened 2025-03-29
Change how speed modifiers are calculated.

Arendian:
## About the PR

Speed modifiers now use a different formula to convert RMC14 movement speeds to CM13 movement delays before adding the modifiers, the movement delays from CM13 can be ported 1:1.
Slows and speed buffs now have a more significant impact on faster entities and a reduced effect on slower ones. In general, speed buffs have a lower overall impact.

## Why / Balance

Currently slows and speed buffs are using placeholder multiplier values, the formula in this PR brings it closer to parity.
Slows and superslows on xenos are currently too low and will increase with this PR.
Overall the speed increase of speed buffs has been decreased, the effect of slows has been increased.

## Technical details

The following formula is used to convert our movement speeds to CM13 values before adding the speed modifier and converting it back to RMC values: 1 / (Math.Max(1 / CurrentSprintSpeed * 10 + modifier, MaxSpeedModifier) / 10) / CurrentSprintSpeed

## Requirements

**Changelog**

:cl: Dygon
- tweak: Slows and super slows are now more effective on fast entities and have stronger effect on xenonids.
- tweak: Positive speed modifiers have their values tweaked and are less effective in general.

### DIFF
--- a/Content.Shared/_RMC14/Marines/Orders/MoveOrderComponent.cs
+++ b/Content.Shared/_RMC14/Marines/Orders/MoveOrderComponent.cs
@@ -15,7 +15,7 @@ public sealed partial class MoveOrderComponent : Component, IOrderComponent
     public SpriteSpecifier Icon = new Rsi(new ResPath("/Textures/_RMC14/Interface/marine_orders.rsi"), "move");
 
     [DataField, AutoNetworkedField]
-    public FixedPoint2 MoveSpeedModifier = 0.1;
+    public FixedPoint2 MoveSpeedModifier = -0.1;
 
     [DataField, AutoNetworkedField]
     public FixedPoint2 EvasionModifier = 5;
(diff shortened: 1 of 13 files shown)


Rainbeon:
Can you include some kind of comprehensive demonstration that this brings speed modifiers closer to parity? We have had like 3 different variations of automatic speed calculation attempts that have not worked out at all, so I am naturally suspicious of speed conversion automation with no provided hard data or even a video demo

Arendian:
> Can you include some kind of comprehensive demonstration that this brings speed modifiers closer to parity? We have had like 3 different variations of automatic speed calculation attempts that have not worked out at all, so I am naturally suspicious of speed conversion automation with no provided hard data or even a video demo

I'll try to make a video this weekend

Arendian:
I removed the changes to the speed modifiers and downsized this PR to only changing the slow and superslow calculations. I also added two videos to show that the speeds are now the same as in CM13.

PR #6029, opened 2025-03-30
Fix spotting and aimed shot

Arendian:
## About the PR

Fix targeting an entity crashing the server.
Also fixes lasers rendering below resin and other entities.
## Technical details

When the TargetedComponent is removed, the target is removed from the targeting list of any entities that were targeting it.

## Requirements

**Changelog**

no cl


DrSmugleaf:
I found a better fix for this but I will copy your changes to the overlay, ty for the pr

PR #6034, opened 2025-03-30
Adds missing groundside rocket warning sound

DummeHytteOst:
## About the PR
added the ground side sound warning for rockets.

i do believe this isn't the best sound to use groundside, or the sound parity intended to use, the original cm13 PR added 2 sounds, RocketPod_fire and IncommingRocket, incommingRocket is sadly not used anywhere in parity, not even by the original PR, i would be glad to switch the sound to IncommingRocket if its desired. 

here is the [IncommingRocket](https://github.com/cmss13-devs/cmss13/blob/0ec7c47a46ad4a6f3c9a6e381f7d8f04325e3bb0/sound/effects/IncomingRocket.ogg) sound which i believe was originally intended to be used as the ground side warning sound.
## Why / Balance

parity

## Requirements

**Changelog**

:cl:
- add: Added groundside warning sound for inbound rockets, played when the green dot appears.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/PowerLoader/Dropship/rmc_dropship_ammo.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/PowerLoader/Dropship/rmc_dropship_ammo.yml
@@ -1,4 +1,4 @@
-﻿- type: entity
+- type: entity
   abstract: true
   parent: RMCDropshipAttachmentBase
   id: RMCDropshipAttachmentAmmo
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_RMC14/Entities/Objects/PowerLoader/Dropship/rmc_dropship_ammo.yml
```diff
@@ -127,6 +127,7 @@
     travelTime: 6
     weapon: RMCDropshipAttachmentGuidedMissileLauncher
     soundCockpit: /Audio/_RMC14/Effects/rocketpod_fire.ogg
+    soundGround: /Audio/_RMC14/Effects/rocketpod_fire.ogg #IncommingRocket.ogg sounds better, and more accurate, but this is what is used in parity, i believe it to be a mistake in the original PR, but we can change it in the future
```
DrSmugleaf, line 130: This is supposed to be soundMarker, soundGround is used for the GAU delayed sound
```suggestion
    soundMarker: /Audio/_RMC14/Effects/rocketpod_fire.ogg #IncommingRocket.ogg sounds better, and more accurate, but this is what is used in parity, i believe it to be a mistake in the original PR, but we can change it in the future
```

DrSmugleaf (dismissed):
Thanks for the PR!

PR #6040, opened 2025-03-30
FIX CRUSHER ARMOR!

Dutch-VanDerLinde:
## About the PR
fixes crushers having the old legacy armor (removed datafield) and replaces it with xeno armor

:cl:
- fix: Fixed crushers having zero armor


Dutch-VanDerLinde:
Closed: See #6039

PR #6044, opened 2025-03-31
Fix map inserts not spawning

CyberTropic:
## About the PR

Map is initialized before map inserts are processed: the upstream update made it so that components get skipped by  EntityQueryEnumerator.MoveNext() if the entity is paused, so we initialize and unpause the map before processing.

## Why / Balance

Bug fix

## Media

*Map inserts actually spawning and not just a 1 in 10 million roll of bad odds*
image

## Requirements

**Changelog**

:cl:
- fix: Fixed map inserts not spawning.


DrSmugleaf (approved):
Thanks for the PR!

PR #6053, opened 2025-04-01
Fix pull effect

Dutch-VanDerLinde:
Makes it spawn on the server, it's spawning multiple pull effects and mispredicting


Arendian:
I made it so it spawns on client too so the opacity can be changed if an entity is invisible, When I tested it the gazillion effects were prevented by the IsFirstTimePredicted, is this not the case on the live server?.

DrSmugleaf:
It will spawn a client and server one right now, you could try changing the check to is client

Dutch-VanDerLinde:
> It will spawn a client and server one right now, you could try changing the check to is client

Done

Dutch-VanDerLinde:
Updated the PR a bit, made the pull effect despawn faster like CM13 so it doesnt linger and removed the effect from xeno lunge because the pull effect already exists

DrSmugleaf (approved):

PR #6058, opened 2025-04-03
Area Based Ambient Lighting and Weather

CyberTropic:
## About the PR

Change the Lighting and Weather systems to use the Areas system.
Adds weatherEnabled flag to areas on used maps.
This is preliminary setup for getting weather cycle and sunset system, but it allows weather and ambient light to be set from the console directly.

## Why / Balance

Parity and soul

## Technical details

For the lighting system, the upstream code just needed a check with the areas system, right after it checks for roof objects.
For the weather system, whenever we detect that there is an AreaGrid on a map, we reroute CanWeatherAffect function to CanWeatherAffectArea in our own RMCWeather system. This allows us to set whatever conditions we want without messing too much with upstream code. 
Minimal upstream changes.

## Media

image
image

https://github.com/user-attachments/assets/8728abde-6786-4bc7-b36e-9dbae6cb377c

## Requirements

**Changelog**

:cl:
- add: Added area based weather and lighting.

### DIFF
--- a/Content.Client/Overlays/StencilOverlay.Weather.cs
+++ b/Content.Client/Overlays/StencilOverlay.Weather.cs
@@ -40,7 +40,7 @@ public sealed partial class StencilOverlay
                 foreach (var tile in _map.GetTilesIntersecting(grid.Owner, grid, worldAABB))
                 {
                     // Ignored tiles for stencil
-                    if (_weather.CanWeatherAffect(grid.Owner, grid, tile, roofComp))
+                    if (_rmcWeather.CanWeatherAffectArea(grid.Owner, grid, tile, roofComp))
                     {
                         continue;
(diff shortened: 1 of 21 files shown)


DrSmugleaf (changes requested):
Won't this make it always lit outside in planets?
That's not intended, its supposed to be when a marine major happens you get sunrise, and at least in the past it used to be for the first minute or two for survivors as well

CyberTropic:
Nah, the MapLight component is set to #000000 by default, so everything will still be dark. This lets admin set ambient light to an arbitrary value for admemes and for the sunset system, and it'll only affect outdoor areas.

DrSmugleaf (approved):
Thanks for the PR!

PR #6066, opened 2025-04-04
fix(yaml): yaml linter crash and other errors

exincore:
## About the PR

Turns out that a missing `abstract: true` in `RMCFlagCarriableBase` from #6011 sent the poor YAML linter to the grave. Since the test was crashing and couldn't point out where the error was, it took several hours for me to debug and in that time I have learned the code is made of paper and glue. Anyway it's fixed now haha

After nurturing the YAML linter back to life I also fixed the errors it was complaining about.

## Why / Balance

Test fails. This also fixes wrong steelcrest defender armor, wrong base xeno explosion armor, and the missing lobby song.

## Technical details

yaml

## Requirements

**Changelog**

:cl:
- fix: (Already uploaded) Fixed Steelcrest Defender having 80 total frontal armor when fortified instead of 60.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
@@ -342,7 +342,6 @@
     durationMultiplier: 0.666
   - type: RMCCanBeFultoned
   - type: CMArmor
-    explosionArmor: 10
   - type: CriticalGraceTime
   - type: RMCObstacleSlamming
   - type: Buckle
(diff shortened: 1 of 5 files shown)


Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
```diff
@@ -342,7 +342,7 @@
     durationMultiplier: 0.666
   - type: RMCCanBeFultoned
   - type: CMArmor
-    explosionArmorStep: 10
+    explosionArmor: 10
```
DrSmugleaf, line 345: This just isn't supposed to be here at all, so it doesn't have a gameplay change in the end

DrSmugleaf (approved):

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
DrSmugleaf, line 345:
```suggestion
```

DrSmugleaf (changes requested):

PR #6072, opened 2025-04-05
Add lance corporal rifleman limit

Dutch-VanDerLinde:
## About the PR
Adds CM13's lance corporal rifleman limit of 7
If there are more than 7 lance corporal riflemen you will get PFC instead

:cl:
- add: Added a max limit of 7 lance corporal riflemen.


DrSmugleaf:
Asked a couple people that know military stuff well in case this was done in 13 to fix a realism problem and they said it's not a big deal, checking 13's PR it was done as part of a rank restructure to reduce the amount of E3s with a bundle of other changes (see https://github.com/cmss13-devs/cmss13/pull/6295)
I don't feel comfortable randomly demoting people that weren't in the lucky 7 in some rounds just because other people grinded the same amount of playtime as them, it devalues the time that they spent in this game

PR #6083, opened 2025-04-05
Add security survivor helmet and uniform

Dutch-VanDerLinde:
## About the PR
Adds the security helmet and uniform for security survivors

Needed because the upstream uniform doesnt have armor
Uses upstream sprites since the sprites in CM13 are made by blundir and we cannot use them


DrSmugleaf (approved):
Thanks for the PR!

PR #6084, opened 2025-04-05
Adds wield slowdown to the flamer

Siigiil:
## About the PR
The flamer was missing its wield slowdown. Which is pretty heavy.

## Why / Balance
Parity. The massive slowdown is one of the flamer's biggest downsides.

## Technical details
The flamer has `SLOWDOWN_ADS_INCINERATOR`, which is the same value as the smartgun. Tested on CM and here, both make you take about 8,71 seconds to move across 23 tiles.

## Requirements

**Changelog**

:cl: Sigil
- fix: The M34 incinerator now has its correct wield slowdown.


DrSmugleaf (approved):
Thanks for the PR!

PR #6088, opened 2025-04-05
Add security survivor playtime rank

Dutch-VanDerLinde:
## About the PR
Fixes corporate survivor ranks using the corporate liaison playtime tracker
Adds a 70 hour playtime rank for security survivor

It's cool!

:cl:
- add: Added a 70 hour playtime reward for security survivors, senior officer rank.


Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/security_survivor.yml
```diff
@@ -5,6 +5,10 @@
   description: cm-job-description-survivor
   playTimeTracker: CMJobSurvivorSecurity
   ranks:
+    RMCRankCivilianSeniorOfficer:
```
DrSmugleaf, line 8: Looking at survivor/security it just says `paygrades = list(PAY_SHORT_CPO = JOB_PLAYTIME_TIER_0)`, is this somewhere else?

Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/corporate_survivor.yml
```diff
@@ -7,15 +7,15 @@
...
       time: 630000 # 175 hours
     RMCRankWeYaSeniorExecutive:
     - !type:RoleTimeRequirement
-      role: CMJobLiaison
+      role: CMJobSurvivorCorporate
       time: 252000 # 70 hours
     RMCRankWeYaExecutive:
     - !type:RoleTimeRequirement
-      role: CMJobLiaison
+      role: CMJobSurvivorCorporate
```
DrSmugleaf, line 18: From the code in equipment presets it seems to read liaison time, not corporate survivor time

DrSmugleaf (changes requested):
Thanks for the PR!

Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/corporate_survivor.yml
Dutch-VanDerLinde: Neither a parity change but added because it having the playtime tracker of a different job is weird

Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/security_survivor.yml
Dutch-VanDerLinde: Yeah this isn't a parity change, just added for extra playtime rewards

Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/corporate_survivor.yml
LordEclipse: > Neither a parity change but added because it having the playtime tracker of a different job is weird

There is no way anyone is ever getting 175 hours as corposurv. That is a 1:6 chance vs. every other player rolling surv at the same time, and that is IF you get surv at all, not to mention the short life time. It also lets CL players use their CL characters.

DrSmugleaf: Might be interesting to take the sum of both but I'd have to think on it

Dutch-VanDerLinde:
Now it only adds security rank because the corporate colonist rank was fixed in the other PR

PR #6093, opened 2025-04-05
quebec squad and spawners

Ian321:
## About the PR

This PR adds the Quebec squad.

## Why / Balance

- Why add a new squad?
  - I have seen the echo spawners I made get used for admemes. This extra squad is for the case that echo was already taken.
- Why Quebec?
  - "q" WAS the only free radio key. ~~Now it's "Q".~~ Nevermind the key is not case-sensitive.

## Technical details

Why is it so much YML work to add a squad...

## Media

Screenshot From 2025-04-06 00-41-18
Screenshot From 2025-04-06 00-41-25
Screenshot From 2025-04-06 00-43-28

## Requirements

**Changelog**

:cl:
ADMIN:
- add: Quebec squad and spawners


DrSmugleaf (changes requested):
Forgot to note this down but this will require us to work how radio channels work so that they can use different prefixes like 13
Q is used by royal marines (in both 13 and 14)

PursuitInAshes:
Thanks for the PR!
I'm going to opt to close this for now as I don't think its particularly needed with the Echo Spawners already existing (Thank you for making those)

PR #6097, opened 2025-04-06
Gun attachment prediction and masterkey PBs

### LINKED ISSUE #5163: can't PB with underbarrel shotgun
read title :( very sad

Siigiil:
## About the PR
Title. Makes attachable guns predicted and lets you PB with the masterkey.

Closes #5163

## Why / Balance
Because normal guns are predicted.

## Technical details
This changes the `TryGetGun` method in the upstream gun system. I tried to figure out if there's a good way of handling this without doing that, but I don't think there is.

This way, the gun system just treats the attachment as the active gun and directly works with it. This has the side benefit of making the client not spew shoot requests at an attachment that shot too recently.

The other way is to intercept `RequestStopShootEvent` to redirect it at the attachment and make `SharedGunPredictionSystem.ShootRequested` target the attachment. The problem is that this might break if we or upstream make something new that requires the use of `TryGetGun` and doesn't account for attachments.

Basically, I think the small upstream change is a better solution for us than the alternative, especially long-term, but if you think it's better to avoid that, I'll change it.

## Media
Masterkey PB:

https://github.com/user-attachments/assets/a74cba4d-cdcf-4a6f-822b-d507288c8262

## Requirements

**Changelog**
:cl: Sigil
- fix: Shootable attachments are now predicted.
- fix: The masterkey can be used for PBs.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
+++ b/Content.Shared/Weapons/Ranged/Systems/SharedGunSystem.cs
@@ -1,6 +1,7 @@
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Numerics;
+using Content.Shared._RMC14.Attachable.Systems;
 using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.Random;
 using Content.Shared._RMC14.Weapons.Ranged.Flamer;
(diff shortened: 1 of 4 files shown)


NzzyFk:
i love you

DrSmugleaf (approved):
Thanks for the PR!

PR #6105, opened 2025-04-07
fix/many integration test fails

exincore:
## About the PR

Fixing whatever pops up on the integration test and closely-related bugs that I find on the way. There are a a few fails left, but those may take a bit more time.

- RMCPouchFuelTankFilled can fit normal size items (the fuel tanks) (but it already works because of whitelist, i think)
- RMCPouchFuelTankFilledExtinguisher has a prefix to distinguish it
- RMCKitHonorGuard is ginormous (contains ginormous item)
- Intel guidebook page shows the bursted colonist corpse again
- The hunting rifle RMCWeaponBoltActionRifle is a child of CMBaseWeaponRifle instead of RMCBaseBreechloader (manually checked the full yaml, no practical changes I could see)
- RMCWeaponFlamer has its ContainerContainer explicitly defined
- RMCStarShellBullet and RMCBatonSlugProjectile have their ContainerContainer explicitly defined
- WeaponLauncherM83 has the correct number of mag steps defined
- RMCWeaponRevolverSpearheadEmpty has its chambers listed out
- RMCShutterHybrisaWindow Door component set not to occlude when closed
- RMCCurtainMedicalOpen initial component state corrected (it is open so gets set noncollidable)
    - NOTE TO ALMAYER MAP OWNER. If this prototype is used in the Almayer, it may have some component yaml saved that are no longer necessary after this PR. I have access to the other maps so I was able to remove the extra yaml, similar to [1e6aff8](https://github.com/RMC-14/RMC-14/pull/6105/commits/1e6aff890b0609ae5660a5d94457b4fb65b9800b). This shouldn't break anything however.
- RMCPhoneboxBase and RMCPhoneboxBrokenBase initial physics state corrected (it is anchored so it gets set static)
- CMPlatformBroken initial physics state corrected (it has no fixtures so it gets set noncollidable)
- RMCBoxMagazineRifleM16Empty construction graph is fixed
- PostMapInitTest only tests _RMC14 maps and ignores upstream maps
- Orphaned entity in Trijent now has a loving parent
- Admin marine rejuvenate command has a description

## Why / Balance

Bugfixes, test fails.

## Technical details

yaml, some C# of PostMapInitTest

## Requirements

**Changelog**

Fixes: #6107

:cl:
- fix: Hybrisa window shutters no longer obstruct vision while closed.
- fix: The Basira-Armstrong bolt-action hunting rifle is no longer a double-barrel shotgun and can no longer accept nor fire shotgun shells.

### DIFF
--- a/Resources/Maps/_RMC14/shiva.yml
+++ b/Resources/Maps/_RMC14/shiva.yml
@@ -126224,9 +126224,6 @@ entities:
     - type: Transform
       pos: 116.5,150.5
       parent: 1
-    - type: ContainerContainer
-      containers:
-        gun_magazine: !type:ContainerSlot {}
   - uid: 15917
     components:
     - type: Transform


DrSmugleaf (approved):
holy shit thank you

PR #6116, opened 2025-04-09
Add map-based (survivor) jobs system

Dutch-VanDerLinde:
## About the PR
This system is useful for SPP events (so admins can put SPP preferences in the menu in the future)
Aswell as survivors, like FORECON and specific map-based survivor jobs

**This is just the SYSTEM for the map-based jobs, this does not add any content to the game**

## Technical details
A networked event is passed from the server to the client so the client can hide jobs that aren't available
Now an evil client can unhide these jobs but since survivors are managed on the server, it won't matter

Adds the base FORECON rifleman (WIP) prototype, the rest of their roles still need porting though

Adds a system to have specific survivor jobs per map, aswell as hiding jobs in the character menu per map too

## Media
image
image

**Changelog**
Not player facing YET


Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
```diff
@@ -998,11 +1002,17 @@ public void RefreshJobs()
...
+                var selectedPlanet = _clientGameTicker.SelectedPlanet;
+
+                var jobs = department.Roles // RMC14 map-based jobs
+                    .Select(jobId => _prototypeManager.Index(jobId))
                     .Where(job => job.SetPreference)
                     .Where(job => !job.Hidden)
                     .ToArray();
 
+                if (selectedPlanet != null) // RMMC14
+                    jobs = jobs.Where(job => job.AvailableOnPlanets == null || job.AvailableOnPlanets.Contains(selectedPlanet)).ToArray();
```
DrSmugleaf, line 1014: Wouldn't this hide all roles other than forecon if this is ever not null?
Either way it may be betterr to not dynamically update the list at all, in case you want to set it for the future and in case the UI doesn't refresh properly from experience

DrSmugleaf (changes requested):
Thanks for the PR!
For a later PR we want FORECON to only have 3 options, FORECON survivor, FORECON CO and FORECON synth (when those are in), we don't want each forecon role to be individually selectable

Content.Client/Lobby/UI/HumanoidProfileEditor.xaml.cs
Dutch-VanDerLinde: It will only hide jobs with the available planet field

DrSmugleaf: Ah yeah I see that now reading it again my bad
I'd prefer it removed so you can always change your FORECON preference though, so that it doesn't catch you by surprise
We can revisit that if we get a lot of planet specific roles in the future

Dutch-VanDerLinde: In CM13 I believe every map has special surv variants, with a menu you're able to select the variant with a dropdown next to the survivor pref (except for survivor CO and synth)

I can change the PR to make it similar to that unless you want all surv jobs to be split

DrSmugleaf: Probably easier to have the surv jobs be split yeah as long as they are distinct enough, if its just a variant of an existing one it can use the existing id

Dutch-VanDerLinde:
Reworking the entire PR

PR #6123, opened 2025-04-09
Fax-maxing

AverageNotDoingAnythingEnjoyer:
## About the PR

Replaced old fax area with a Fax Responder's outpost from CM13, the old one was moved to admin_fax_old

## Why / Balance

Looks cooler, might be used for fax responders in the future, has more factions

## Technical details

Adds the map itself, areas and one hull window variant. 
Currently there's a problem with some vendors and objects snapping to grid, changing their position in transform comp doesn't yield anything so if anyone finds a solution, I'd be glad to include it.
Also the map itself lacks a few props, those will have to be ported separately with upcoming maps as they are used there.

## Media

Screenshot 2025-04-09 at 15 41 20

## Requirements

**Changelog**

:cl: J C Denton
- tweak: Remapped the admin fax machines area.


mmails:
I noticed we lack a lot of consoles from the actual 13 fax rooms. Which ones are they?

AverageNotDoingAnythingEnjoyer:
> I noticed we lack a lot of consoles from the actual 13 fax rooms. Which ones are they?

Currently there's no personal computers, overwatch cameras computer (or maybe it's just renamed almayer cams computer, not sure) and health monitoring computers for UNMC and other factions

mmails:
You should wait for #6065 to be added as it adds SPP/PMC specific OW consoles

I think those are the ones you need anyway

AverageNotDoingAnythingEnjoyer:
> You should wait for #6065 to be added as it adds SPP/PMC specific OW consoles
> 
> I think those are the ones you need anyway

Honestly I'm not even sure cm13 has "overwatch cameras" console, since it uses the same prefab as almayer cameras console, so might be just an RP thing. 
Need to check though, since I'm not an expert in SDMM, I just map cool things

DrSmugleaf (approved):
Thanks for the PR!

PR #6129, opened 2025-04-09
Final obstacle slam fixes

Dutch-VanDerLinde:
Added a component to make it so you cant slam into mobs

https://github.com/user-attachments/assets/54451565-e094-4613-a5ff-402a5cb86ee9

As you can see, you won't slam into mobs


DrSmugleaf (approved):
Thanks for the PR!

PR #6130, opened 2025-04-09
Fix devouring not killing while crit

Dutch-VanDerLinde:
The changing mobstate was just changing the mobstate for a split second, the the mob did not die but it made the deathgasp
This makes it take enough asphyxiation damage to kill so the entity will actually die

### DIFF
--- a/Content.Shared/_RMC14/Damage/SharedRMCDamageableSystem.cs
+++ b/Content.Shared/_RMC14/Damage/SharedRMCDamageableSystem.cs
@@ -40,6 +40,7 @@ public abstract class SharedRMCDamageableSystem : EntitySystem
     [Dependency] private readonly DamageableSystem _damageable = default!;
     [Dependency] private readonly EntityWhitelistSystem _entityWhitelist = default!;
     [Dependency] private readonly MobStateSystem _mobState = default!;
+    [Dependency] private readonly MobThresholdSystem _mobThresholds = default!;
     [Dependency] private readonly INetManager _net = default!;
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Damage/SharedRMCDamageableSystem.cs
```diff
@@ -303,8 +305,16 @@ private void OnActiveDamageOnPulledMobState(Entity<ActiveDamageOnPulledWhileCrit
 
     private void OnActiveDamageOnPulledDevoured(Entity<ActiveDamageOnPulledWhileCritComponent> ent, ref XenoTargetDevouredAttemptEvent args)
     {
+        if (TryComp<MobThresholdsComponent>(ent, out var mobThresholds) && TryComp<DamageableComponent>(ent, out var damageable))
+        {
+            // Kill the mob
+            var lethalAmountOfDamage = mobThresholds.Thresholds.Keys.Last() - damageable.TotalDamage;
```
DrSmugleaf, line 311: You can do MobThresholdSystem.TryGetDeadThreshold here instead to get the damage threshold for death

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Damage/SharedRMCDamageableSystem.cs
Dutch-VanDerLinde: Done

DrSmugleaf (approved):
Thanks for the PR!

PR #6134, opened 2025-04-10
Parity survivor count (someday)

AverageNotDoingAnythingEnjoyer:
## About the PR

Changed CVARs pertaining survivor count to parity values

## Why / Balance

Parity.
But to expand further, this PR is not intended to be merged right now, only when hosts feel comfortable with these values for survs. Also since there are forks of RMC and there may be new ones in the future, I made this PR to make 1 survivor per 10 marines with a maximum of 8 the base value going forward, instead of current 1 survivor per 20 with a maximum of 6.
In other words, since CVARs can be changed by server hosts this one is just a baseline update.

## Technical details

Just changed 2 CVARs

## Requirements

**Changelog**

No cl

### DIFF
--- a/Content.Shared/_RMC14/CCVar/RMCCVars.cs
+++ b/Content.Shared/_RMC14/CCVar/RMCCVars.cs
@@ -244,13 +244,13 @@ public sealed class RMCCVars : CVars
         CVarDef.Create("rmc.power_load_multiplier", 0.01f, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCMarinesPerSurvivor =
-        CVarDef.Create("rmc.marines_per_survivor", 20, CVar.REPLICATED | CVar.SERVER);
+        CVarDef.Create("rmc.marines_per_survivor", 10, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCSurvivorsMinimum =
         CVarDef.Create("rmc.survivors_minimum", 2, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCSurvivorsMaximum =
-        CVarDef.Create("rmc.survivors_maximum", 6, CVar.REPLICATED | CVar.SERVER);
+        CVarDef.Create("rmc.survivors_maximum", 8, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCSpawnerMaxCorpses =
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/CCVar/RMCCVars.cs
```diff
@@ -244,13 +244,13 @@ public sealed class RMCCVars : CVars
         CVarDef.Create("rmc.power_load_multiplier", 0.01f, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCMarinesPerSurvivor =
-        CVarDef.Create("rmc.marines_per_survivor", 20, CVar.REPLICATED | CVar.SERVER);
+        CVarDef.Create("rmc.marines_per_survivor", 10, CVar.REPLICATED | CVar.SERVER);
```
DrSmugleaf, line 247: We changed this in the live to 12 and 7 which works fine without wounds and synthvivor
```suggestion
        CVarDef.Create("rmc.marines_per_survivor", 12, CVar.REPLICATED | CVar.SERVER);
```

Content.Shared/_RMC14/CCVar/RMCCVars.cs
```diff
@@ -244,13 +244,13 @@ public sealed class RMCCVars : CVars
...
 
     public static readonly CVarDef<int> RMCSurvivorsMinimum =
         CVarDef.Create("rmc.survivors_minimum", 2, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCSurvivorsMaximum =
-        CVarDef.Create("rmc.survivors_maximum", 6, CVar.REPLICATED | CVar.SERVER);
+        CVarDef.Create("rmc.survivors_maximum", 8, CVar.REPLICATED | CVar.SERVER);
```
DrSmugleaf, line 253:
```suggestion
        CVarDef.Create("rmc.survivors_maximum", 7, CVar.REPLICATED | CVar.SERVER);
```

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/CCVar/RMCCVars.cs
AverageNotDoingAnythingEnjoyer: Alright, then I'll change it to 15 and 7 too here

DrSmugleaf (approved):

PR #6136, opened 2025-04-10
Add different rank names for some xenos

Vermidia:
## About the PR

Title.

## Why / Balance

parity for Queen + Parasite (though para needs to be based on hugs not time). I also changed larva to go from Little -> Mature -> Big (its cute, CM just has them all be mature or bloody). And have 1 for King which is what started me working on this (Just gave them 1 because reaching 10 hours as King will probably not happen much. CM just doesn't give them any rank names eh).

## Technical details

New component that holds a dictionary of rank numbers to rank Loc.Ids. Instead of a switch a rank number is checked, if it exists in the dictionary that's what's used for the rank name. If it's not found it's null and it's just the base name as expected.

## Media

image

## Requirements

**Changelog**

:cl:
- fix: Fixed Parasites not having different rank titles. Larva also now have different titles.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/XenoRoleSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/XenoRoleSystem.cs
@@ -117,7 +117,10 @@ public sealed class XenoRoleSystem : EntitySystem
         if (HasComp<XenoMaturingComponent>(ent) || !TryComp<XenoRankNamesComponent>(ent, out var rankNamesComp))
             return;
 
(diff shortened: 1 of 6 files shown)


DrSmugleaf (changes requested):
For Empress we decided to save that name for a possible future whitelist role name distinction
For Parasite I would prefer to bundle that in with the change that makes it work with infects instead but we would need to start tracking infects before we actually make the switch so that people don't all get reset to 0, so it's probably fine for now, though the royal rank we also wanted to leave out for now in case we want to use it for something else soon after parity

DrSmugleaf (approved):
Thanks for the PR!

PR #6139, opened 2025-04-11
Make crates undestroyable by xeno claws, crate drag speed change

Vermidia:
## About the PR

Title.

## Why / Balance

Mostly parity. Crates can't be destroyed by xenos through attack in parity but I don't remember if we have a comp for just that so I made it require very sharp claws. They also have the 100 health they were supposed to.

Crate dragging speed is supposed to be the same as human dragging speed, so I tested it a bit to get them to mostly match up. The value for bodybags seemed to work the best here, so it's only off by a little probably.

## Technical details

YML.

## Requirements

**Changelog**

:cl:
- fix: Fixed xenos from being able to destroy crates with melee.
- fix: Fixed dragging crates not slowing you down.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Storage/crates.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Storage/crates.yml
@@ -51,7 +51,7 @@
     thresholds:
     - trigger:
         !type:DamageTrigger
-        damage: 50
+        damage: 100
       behaviors:
       - !type:DoActsBehavior
         acts: ["Destruction"]
@@ -100,6 +100,8 @@
     virtualRight: RMCVirtualCrateRight
     virtualLeft: RMCVirtualCrateLeft
   - type: RMCCanBeFultoned
+  - type: ReceiverXenoClaws
+    minimumClawStrength: VerySharp
 
 - type: entity
   parent: RMCCrateBase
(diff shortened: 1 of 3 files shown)


Resources/Prototypes/_RMC14/Entities/Structures/Storage/crates.yml
```diff
@@ -100,6 +100,8 @@
     virtualRight: RMCVirtualCrateRight
     virtualLeft: RMCVirtualCrateLeft
   - type: RMCCanBeFultoned
+  - type: ReceiverXenoClaws
+    minimumClawStrength: VerySharp
```
DrSmugleaf, line 104: We have ImmuneToMelee and ImmuneToUnarmed with applyToXenos: true, if either work for you

DrSmugleaf (changes requested):
Thanks for the PR!

Resources/Prototypes/_RMC14/Entities/Structures/Storage/crates.yml
Vermidia: Problem is that those cancel the attack. The attack still needs to happen (to open the crate) but not do damage. Since that opens the crate.

DrSmugleaf: Would be better to add a new tier of claws then and set it to that

Vermidia: Yeah that works.

DrSmugleaf (approved):

PR #6144, opened 2025-04-12
Flare packs can directly transfer their content to a flare pouch

Arendian:
## About the PR

Entities with the CMItemSlotComponent can now directly transfer their content to another entity with the ItemSlotComponent.
This means that for example flare packs can instantly transfer all their flares to a flare pouch by using the flare pack on the pouch.

## Why / Balance

The description of the flare pouch describes this interaction, but it currently doesn't exist.

## Technical details

When an entity with the CMItemSlotsComponent is interacted with, it checks whether the interacting entity has any item slots. If it does, it attempts to transfer all contents from the interacting entity into itself.

## Media

https://github.com/user-attachments/assets/adc2e022-e1fb-4b70-a00e-94f664b365b8

## Requirements

**Changelog**

:cl: Dygon
- add: Flare packs can now directly transfer their content to a flare pouch.

### DIFF
--- a/Content.Shared/_RMC14/Inventory/SharedCMInventorySystem.cs
+++ b/Content.Shared/_RMC14/Inventory/SharedCMInventorySystem.cs
@@ -229,7 +229,9 @@ public abstract class SharedCMInventorySystem : EntitySystem
     /// </summary>
     private void OnInteractUsing(Entity<CMItemSlotsComponent> ent, ref InteractUsingEvent args)
     {
-        if(!TryComp(args.Used, out ItemSlotsComponent? usedStorage) || !TryComp(ent, out ItemSlotsComponent? storage))
+        if(!TryComp(args.Used, out ItemSlotsComponent? usedStorage) ||
+           !TryComp(ent, out ItemSlotsComponent? storage) ||
+           args.Handled)
             return;
 
         SoundSpecifier? insertSound = null;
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Inventory/SharedCMInventorySystem.cs
```diff
@@ -222,6 +224,37 @@ private void OnSlotsEjectAttempt(Entity<CMItemSlotsComponent> ent, ref ItemSlotE
         }
     }
 
+    /// <summary>
+    ///     Try to transfer contents from ItemSlots of one entity to the ItemSlots of another.
+    /// </summary>
+    private void OnInteractUsing(Entity<CMItemSlotsComponent> ent, ref InteractUsingEvent args)
```
DrSmugleaf, line 230: This should be setting args.Handled to true if the interaction does anything (transfers any items or could have transferred any if not full), otherwise it may do multiple interactions at once

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Inventory/SharedCMInventorySystem.cs
Arendian: args.handled is now set to True if at least one item is transferred over.

DrSmugleaf (approved):

PR #6151, opened 2025-04-13
Add the ability to apply patreon custom item names via right-click verb.

BramvanZijp:
## About the PR
This PR makes 2 QOL changes/additions to the Patreon Custom Item Naming System:
1. You can now disable custom item names being automatically applied to the first item you dispense from a vendor.
2. You can now apply your custom patreon item name to an item via a right-click verb.

## Why / Balance
Explained in https://discord.com/channels/1168210010233376858/1359153311361798394
But TLDR: If you are working reqline you no longer need to worry about giving away your named item, and if you get an item from req, roundstart, etc, you can now name it.

## Technical details
Considering the fact that I used the tried and true programming tactic of Fucking around and Finding out for several hours, your guess is as good as mine. All I know is that I copied another true/false preference option and had to add a DB entry for it.

## Media
https://github.com/user-attachments/assets/c4a7cfb1-d8af-4a9e-8c31-edec57d31447

## Requirements

- [X] **I have added media to this PR** or it does not require an ingame showcase.

**Changelog**
:cl: BramvanZijp
- add: Patreon Supporters can now apply their custom item name to an item via a right-click verb if their name for that category has not already been applied to another item.
- tweak: Patreon Supporters can now disable their custom item name being automatically applied upon dispensing any item of that category, choosing to do so manually instead.


TurboTrackerss14:
Might be advisable to have this only useable once. 
Its bad enough to lose your firearm on the front line, but I would much rather not "lose" it and have a patreon supporter be able to claim they had it the entire shift because "it has my name on it" after the applied it to the stolen firearm.

BramvanZijp:
> Might be advisable to have this only useable once. Its bad enough to lose your firearm on the front line, but I would much rather not "lose" it and have a patreon supporter be able to claim they had it the entire shift because "it has my name on it" after they applied it to the stolen firearm.

You can only name one weapon per type as showed in the video, and if you label your gun, that stays. Also, if your gun has a patreon name already, somebody else cannot overwrite it.

BramvanZijp:
Also, before you tell me to move the yes/no button below the name list, I tried that, but due to the formatting, itll shift the button down to fully at the bottom of the entire page, so thats not ideal.

BramvanZijp:
Since I have time due to missing the update, going to try to re-format it in a way I can have the toggle below the item names without having it at the bottom of the page.

BramvanZijp:
> Might be advisable to have this only useable once. Its bad enough to lose your firearm on the front line, but I would much rather not "lose" it and have a patreon supporter be able to claim they had it the entire shift because "it has my name on it" after they applied it to the stolen firearm.

Ive elaborated the CL to reflect this, will also add an ingame label to specify you can only label 1 item per type.

BramvanZijp:
I think that should be it fully good to go now, UI updated, info label updated.

BramvanZijp:
Empty commit made, hoping that tests re-run and actually pass this time.

DrSmugleaf (changes requested):
This needs to just be able to reapply the name to an item through a verb, removing it from the old item, that way you don't need this setting
At that point editing what your named items setting is set to achieves the same thing with less settings
I had some work in progress code to do this, will link to this PR when I open a branch with it

DrSmugleaf:
See https://github.com/RMC-14/RMC-14/pull/6417

PR #6152, opened 2025-04-13
Various MP Armory Gun Locker Tweaks/Fixes

BramvanZijp:
## About the PR
Several Alert Gun Locker Changes:
1. Added a new basic gun locker, which the other 3 lockers target off of.
2. Made it so that the red alert gun safe can now once again be opened if you have armory access.
3. Made it so that the riot control safe now opens on blue alert
4. Fixed alert level triggers (unlock & open) only happening AT the target alert level, and not above it. (Meaning the armory would re-lock on code delta)
5. Prevents people from going into gun safes, which can get them trapped if they are in it when the alert level drops, and if they spend 2 minutes kicking their way out, the locker is then forced open until the alert level is updated again.
6. Alert gun safes will now close before locking if the alert level is lowered.

## Why / Balance
1. Because the parenting caused issues, such as parenting destruction components, making the ship gun safes need colony armory access, etc.
2. Parity + allows a warden/cmp to return gear to the armory that was not put back in before an alert level drop.
3. Parity, made possible due to number 4.
4. I think this is parity, but regardless, the riot locker shouldnt close again on red, and neither should close on delta.
5. Semi-parity, if they auto-lock in 13, all people inside get forced out, but making it so that people cannot go in at all seems more straight forward, and trapping people inside for 2 minutes is not ideal.

## Technical details
Turned an == into an <=, also added a base gun locker which the other 3 parent off of.
Also had to fix ContainerGettingInsertedAttemptEvent because smug did not apply it correctly. (Can't blame for that, who would expect the entity in that event to be the entity being put in, not the container itself which would have the blacklist comp.)

## Requirements

- [X] I have added media to this PR or **it does not require an ingame showcase.**

**Changelog**
:cl: BramvanZijp
- add: The MP riot control locker will now unlock automatically if the alert level is set to blue or above.
- remove: Marines can no longer go into gun safes, nor be locked in them by an unluckily timed alert level change.
- tweak: MP armory gun safes can now once again be locked/unlocked manually, requiring armory access to do so.
- fix: Fixed alert level locked safes/doors only opening on their exact target alert level, but not above it, such as red alert safes closing on delta.
- fix: Fixed alert level locked safes locking while still being open if the alert level drops, they now close first.


BramvanZijp:
Going to add a fix for the issue that lockers lock but dont close, meaning theyll be stuck locked but opened.

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
```diff
@@ -41,9 +41,9 @@ private void OnBlacklistInteractionAttempt(Entity<InteractedBlacklistComponent>
             args.Cancelled = true;
     }
 
-    private void OnInsertBlacklistContainerInsertedAttempt(Entity<InsertBlacklistComponent> ent, ref ContainerGettingInsertedAttemptEvent args)
+    private void OnInsertBlacklistContainerInsertedAttempt(ContainerGettingInsertedAttemptEvent args)
     {
-        if (args.Cancelled || ent.Comp.Blacklist is not { } blacklist)
+        if (args.Cancelled || !TryComp<InsertBlacklistComponent>(args.Container.Owner, out var insertBlacklist) || insertBlacklist.Blacklist is not { } blacklist)
```
DrSmugleaf, line 46: If you are returning when this component is not found, you should be subscribing the event to that component, not changing it to be a broadcast subscription by removing the component above

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
BramvanZijp: Willdo shortly

BramvanZijp: Oh righy, I know why I did this, and this is sadly an intended change:
ent doesnt HAVE the component, args.Container.Owner does, but it does not get called for the event, and I cant change the event without that being a sweeping breaking change. Ent is the entity getting PUT INSIDE the container, and it will never have insertblacklistcomponent.

BramvanZijp:
Changes cannot be applied.

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
BramvanZijp: What I said above is also the REASON why the system currently does not work.

DrSmugleaf: Change it to ContainerIsInsertingAttemptEvent then no?
That's directed at the container owner

BramvanZijp: THAT I can try.

BramvanZijp: Applied locally without errors, going to verify it still works in-game now.

BramvanZijp:
Changes may be appliable, testing in-game now.

BramvanZijp:
@DrSmugleaf applied, it works.

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
BramvanZijp: Works ingame, problem solved.

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
```diff
@@ -41,7 +41,7 @@ private void OnBlacklistInteractionAttempt(Entity<InteractedBlacklistComponent>
             args.Cancelled = true;
     }
 
-    private void OnInsertBlacklistContainerInsertedAttempt(Entity<InsertBlacklistComponent> ent, ref ContainerGettingInsertedAttemptEvent args)
+    private void OnContainerIsInsertingAttemptEvent(Entity<InsertBlacklistComponent> ent, ref ContainerIsInsertingAttemptEvent args)
```
DrSmugleaf, line 44:
```suggestion
    private void OnInsertBlacklistContainerIsInsertingAttempt(Entity<InsertBlacklistComponent> ent, ref ContainerIsInsertingAttemptEvent args)
```

Content.Shared/_RMC14/Interaction/RMCInteractionSystem.cs
```diff
@@ -16,7 +16,7 @@ public override void Initialize()
     {
         SubscribeLocalEvent<InteractedBlacklistComponent, GettingInteractedWithAttemptEvent>(OnBlacklistInteractionAttempt);
         SubscribeLocalEvent<NoHandsInteractionBlockedComponent, GettingInteractedWithAttemptEvent>(OnNoHandsInteractionAttempt);
-        SubscribeLocalEvent<InsertBlacklistComponent, ContainerGettingInsertedAttemptEvent>(OnInsertBlacklistContainerInsertedAttempt);
+        SubscribeLocalEvent<InsertBlacklistComponent, ContainerIsInsertingAttemptEvent>(OnContainerIsInsertingAttemptEvent);
```
DrSmugleaf, line 19:
```suggestion
        SubscribeLocalEvent<InsertBlacklistComponent, ContainerIsInsertingAttemptEvent>(OnInsertBlacklistContainerIsInsertingAttempt);
```

DrSmugleaf (approved):
Thanks for the PR!

PR #6156, opened 2025-04-14
Fix gun melee attack damage being too low

Arendian:
## About the PR

Damage from attachments like the bayonet is now also increased by the user's melee weapon skill.
The bonus melee damage when melee attacking a xenonid is now also applied to attachment bonus damage.

## Why / Balance

Parity.

A melee attack using a gun, with attachments that improve melee damage, now deals more damage against xenonids because the 50% bonus melee damage is now also applied to the attachment damage.
A melee attack using a gun, with attachments that improve melee damage, now deals more damage if the user has RMCSkillMeleeWeapons on level 1 or higher

Example values:
Base melee damage of the M42A2 shotgun: 5
Bayonet bonus damage when attached: 20
Damage multiplier when melee attacking a xeno: 1.5

Example calculation before this PR without melee skill: 5 * 1.5 + 20 = 27.5
Example calculation after this PR without melee skill: 5 * 1.5 + 20 * 1.5 = 37.5

Example calculation before this PR with melee skill 1: 5 * 1.5 * 1.25 + 20 = 29.375
Example calculation after this PR with melee skill 1: 5 * 1.5 * 1.25 + 20 * 1.5 * 1.25 = 46.875

## Media

Damage on a runner in CM13 after being hit by a weapon specialist's(level 1 melee skill) melee attack, using the M37A2 equipped with a bayonet.
image

## Technical details

Previously only the base melee damage was increased by the melee skill multiplier, a new method is now called on melee hit to also increase the bonus damage from attachments based on the user's melee skill.

## Requirements

**Changelog**

:cl: Dygon
- fix: Weapon attachment damage now accounts for the user's melee skill level and any other melee damage modifiers.

### DIFF
--- a/Content.Shared/_RMC14/Marines/Skills/SkillsSystem.cs
+++ b/Content.Shared/_RMC14/Marines/Skills/SkillsSystem.cs
@@ -4,6 +4,7 @@ using System.Runtime.InteropServices;
 using Content.Shared.Chemistry.Components.SolutionManager;
 using Content.Shared.Chemistry.EntitySystems;
 using Content.Shared.Chemistry.Reagent;
+using Content.Shared.Damage;
 using Content.Shared.Examine;
 using Content.Shared.Flash;
 using Content.Shared.Interaction;
@@ -82,7 +83,7 @@ public sealed class SkillsSystem : EntitySystem
         if (skill <= 0)
             return;
 
-        args.Damage *= 1 + 0.25 * skill;
+        args.Damage = ApplyMeleeSkillModifier(args.User, args.Damage);
     }
 
     private void OnAttemptHyposprayUse(Entity<MedicallyUnskilledDoAfterComponent> ent, ref AttemptHyposprayUseEvent args)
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Marines/Skills/SkillsSystem.cs
```diff
@@ -539,4 +540,11 @@ public float GetSkillDelayMultiplier(Entity<SkillsComponent?> user, EntProtoId<S
 
         return multiplier;
     }
+
+    public DamageSpecifier ApplyMeleeSkillModifier(EntityUid user, DamageSpecifier damage)
+    {
+        var skill = GetSkill(user, _meleeSkill);
+
+        return damage* (1 + 0.25 * skill);
```
DrSmugleaf, line 548:
```suggestion
        return damage * (1 + 0.25 * skill);
```

DrSmugleaf (approved):
Thanks for the PR!

Issue #6158, opened 2025-04-15
xenos can dodge aimed shot


Carolyn3114:
## Description
when using aimed shots on xenos, it can activate, shooting the bullet, but xenos can walk out of the way to dodge it. 

**Reproduction**

**Screenshots**

**Additional context**

PR #6159, opened 2025-04-15
Warrior lunge is homing

Tyzemol:
## About the PR
Warrior lunge is homing on the target, making it impossible to dodge after the lunge has started, with a time limit of course, no warriors flying across the map chasing a guy.

## Why / Balance
Ima be honest here i am a young warrior and every time i try to do a lunge, it misses, i have no idea how warrior wizards land their lunges but it has never worked for me, its pretty bad.

## Technical details
I simply give warrior a `HomingProjectile` component that aimed shot uses, and remove it once the warrior lands, which is called either directly by the `XenoLungeSystem` once warrior hits something, or by the `ThrownItemSystem` once the throw duration exprires.

## Media
To make change even visible at all the speed numbers have been divided by 7.

https://github.com/user-attachments/assets/82d26451-5152-49d6-b7a4-596e58ebd3d2

## Requirements

**Changelog**
:cl:
- fix: Warrior lunge should reliably hit targeted marine now.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Lunge/XenoLungeSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Lunge/XenoLungeSystem.cs
@@ -1,6 +1,7 @@
 using Content.Shared._RMC14.Damage.ObstacleSlamming;
 using Content.Shared._RMC14.Marines;
 using Content.Shared._RMC14.Pulling;
+using Content.Shared._RMC14.Weapons.Ranged.Homing;
 using Content.Shared._RMC14.Xenonids.Hive;
 using Content.Shared.Coordinates;
 using Content.Shared.Mobs.Systems;
@@ -44,6 +45,7 @@ public sealed class XenoLungeSystem : EntitySystem
 
         SubscribeLocalEvent<XenoLungeComponent, XenoLungeActionEvent>(OnXenoLungeAction);
         SubscribeLocalEvent<XenoLungeComponent, ThrowDoHitEvent>(OnXenoLungeHit);
+        SubscribeLocalEvent<XenoLungeComponent, LandEvent>(OnXenoLungeEnd);
 
         SubscribeLocalEvent<XenoLungeStunnedComponent, PullStoppedMessage>(OnXenoLungeStunnedPullStopped);
     }
@@ -75,7 +77,13 @@ public sealed class XenoLungeSystem : EntitySystem
         Dirty(xeno);
 
         EnsureComp<RMCObstacleSlamImmuneComponent>(xeno);
-        _throwing.TryThrow(xeno, diff, 30, animated: false);
+        //throw speed has to be slightly less than homing speed, otherwise if the target is running away the throw won't connect
+        _throwing.TryThrow(xeno, diff, 27, animated: false);
+
+        var homingProjectile = EnsureComp<HomingProjectileComponent>(xeno);
+        homingProjectile.Target = args.Target;
+        homingProjectile.ProjectileSpeed = 30;
+        Dirty(xeno, homingProjectile);
 
         if (!_physicsQuery.TryGetComponent(xeno, out var physics))
             return;
@@ -101,6 +109,12 @@ public sealed class XenoLungeSystem : EntitySystem
 
         ApplyLungeHitEffects(xeno, args.Target);
     }
+
+    private void OnXenoLungeEnd(Entity<XenoLungeComponent> xeno, ref LandEvent args)
+    {
+        RemComp<HomingProjectileComponent>(xeno);
+    }
+
 
     private bool ApplyLungeHitEffects(Entity<XenoLungeComponent> xeno, EntityUid targetId)
     {
--- a/Content.Shared/_RMC14/Xenonids/Lunge/XenoLungeSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Lunge/XenoLungeSystem.cs
@@ -115,7 +115,6 @@ public sealed class XenoLungeSystem : EntitySystem
         RemComp<HomingProjectileComponent>(xeno);
     }
 
-
     private bool ApplyLungeHitEffects(Entity<XenoLungeComponent> xeno, EntityUid targetId)
     {
         // TODO RMC14 lag compensation


BramvanZijp:
I suspect this wont be merged due to not being parity, but even ignoring that I am not sure this is a good idea.

RMC14, right now, is trying to be an identical game to CM13, just on the 14 engine. After thats done, personal changes for 14 are planned, but right now anything that strays away from parity is VERY unlikely to be accepted because it will progressively make it harder and harder to keep track of what changes still need to be updated to match 13, and what was intentionally done as a change for 14 only.

Furthermore, a LOT of xenos have these type of dash forward abilities, including runners. I find it odd that these are ignored, with the change only applying to warrior. While I don't like the term "mald pr", it does feel comparable to what is happening here, where a player gets frustrated with a specific part of the game they believe unfair or too hard, and then proceeds to make a change that "fixes" that very specific thing, without giving careful concern to greater game balance, and also if that change should be applied on a more global scale, which I think it should not. Aiming these abilities is a large part of the skill expression for both the melee xeno players using them, who may have spent time to learn and predict the way marines move, and recognize the timings needed to increase the likelyhood of a hit, and for the marines trying to make themselves harder to hit. I think making abilities for xenos that GUARANTEE a hit at long range is setting a very dangerous precedent that could very easily cause major issues if left unchecked.

VictorJob:
If the problem is hitting marines with the Warrior's charge, just... Git gud.

Jokes aside, it probably won't be merged due to parity as the comment above explained. It's always good to check by asking in the discord chats what would be good or not to do for a PR in the current state of the game in which we seek parity, normally purely aesthetic changes are welcome but even they don't completely escape parity.

Dutch-VanDerLinde:
Pretty sure the warrior lunge is homing or near instant in cm13

Tyzemol:
> I suspect this wont be merged due to not being parity, but even ignoring that I am not sure this is a good idea.
> 
> RMC14, right now, is trying to be an identical game to CM13, just on the 14 engine. After thats done, personal changes for 14 are planned, but right now anything that strays away from parity is VERY unlikely to be accepted because it will progressively make it harder and harder to keep track of what changes still need to be updated to match 13, and what was intentionally done as a change for 14 only.
> 
> Furthermore, a LOT of xenos have these type of dash forward abilities, including runners. I find it odd that these are ignored, with the change only applying to warrior. While I don't like the term "mald pr", it does feel comparable to what is happening here, where a player gets frustrated with a specific part of the game they believe unfair or too hard, and then proceeds to make a change that "fixes" that very specific thing, without giving careful concern to greater game balance, and also if that change should be applied on a more global scale, which I think it should not be applied at all; Aiming these abilities is a large part of the skill expression for both the melee xeno players using them, who may have spent time to learn and predict the way marines move, and recognize the timings needed to increase the likelyhood of a hit, and for the marines trying to make themselves harder to hit, such as scouts using their cloak to try to help evade xenos in their backlines. I think making abilities for xenos that GUARANTEE a hit at long range is setting a very dangerous precedent that could very easily cause major issues if left unchecked.

i am sorry i am not reading all that but the fact is simple: in 13 you click the ability and it hits the target, in 14 it half the time doesnt, this is not about parity, this is about fixing a broken ability, as for runner and other dashes, lunge is not a dash, its a targeted ability, you can not aim it ahead of a moving target like you can with a runner leap.

Dutch-VanDerLinde:
> I suspect this wont be merged due to not being parity, but even ignoring that I am not sure this is a good idea.
> 
> 
> 
> RMC14, right now, is trying to be an identical game to CM13, just on the 14 engine. After thats done, personal changes for 14 are planned, but right now anything that strays away from parity is VERY unlikely to be accepted because it will progressively make it harder and harder to keep track of what changes still need to be updated to match 13, and what was intentionally done as a change for 14 only.
> 
> 
> 
> Furthermore, a LOT of xenos have these type of dash forward abilities, including runners. I find it odd that these are ignored, with the change only applying to warrior. While I don't like the term "mald pr", it does feel comparable to what is happening here, where a player gets frustrated with a specific part of the game they believe unfair or too hard, and then proceeds to make a change that "fixes" that very specific thing, without giving careful concern to greater game balance, and also if that change should be applied on a more global scale, which I think it should not be applied at all; Aiming these abilities is a large part of the skill expression for both the melee xeno players using them, who may have spent time to learn and predict the way marines move, and recognize the timings needed to increase the likelyhood of a hit, and for the marines trying to make themselves harder to hit, such as scouts using their cloak to try to help evade xenos in their backlines. I think making abilities for xenos that GUARANTEE a hit at long range is setting a very dangerous precedent that could very easily cause major issues if left unchecked.

You don't "aim" the warrior lunge

Unlike the lurker or runner pounce, you can only use to to target mobs while runner pounce can be used to leap from one location to another

Once you use the ability on the mob, you'll lock on and lunge towards them

The problem with warrior lunge is that it's inconsistent, sometimes the lunge won't connect even though it 100% should, in CM13 you will almost never miss it

The warrior lunge is not long rang. If you're in range for the lunge (which is around 5 tiles max) and the target quickly moves out of range the ability will mispredict, "leaping" you forward for a split second and then you rubberband back to your original location while also putting the ability on cooldown

BramvanZijp:
Unless 13 ACTUALLY makes it homing

Vermidia:
so
`lunge_user.throw_atom(get_step_towards(affected_atom, lunge_user), grab_range, SPEED_FAST, lunge_user, tracking=TRUE)`
Both lunge and defender headbutt have this.

`// If we missed, but we are tracking and the target is still next to us and the turf we launched from, then we still count it as a hit
				hit_atom = LM.target`
				
So it's not homing but it does basically count it as a hit if you are within a tile.

DummeHytteOst:
> So it's not homing but it does basically count it as a hit if you are within a tile when missing, sort of. (Note the starting turf also has to be within a tile range of the target).

isn't that nearly identical to what projectile prediction does? 

i might have understood projectile prediction, but i think it basically tracks where the server side target is and if its within 1.5 tiles of the targets client then the projectile hits?

Vermidia:
kinda yeah.

Kadeo64:
good pr in theory but if #6158 is actually real idk how well this PR will work in actual server conditions.

DrSmugleaf (changes requested):
I added code for lunges to connect if they end adjacent to the target like 13, need to see if that fixes it on the live server next update first

PR #5997, opened 2025-03-26
Grass landing radius for small drone is limited

N0lim:
## About the PR
See title

## Why / Balance
Parity
https://github.com/cmss13-devs/cmss13/blob/4dfaca73cc5d9d08f79c487d4af079b5afbb1999/code/modules/mob/living/carbon/xenomorph/abilities/lesser_drone/lesser_drone_powers.dm

## Media

https://github.com/user-attachments/assets/87fafc29-6edb-4c21-b095-e9ae0467f01e

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

- fix: Fixed lesser drones being able to plant new weed nodes further than 5 tiles away from another weed node.

PR #6170, opened 2025-04-17
Fix of node placment range for lesser drones

N0lim:
## About the PR
Due to a https://github.com/RMC-14/RMC-14/pull/5997 error the placing area was 6x6 instead of 5x5

## Requirements

**Changelog**

:cl:
- fix: Fixed lesser drones being able to plant new weed nodes in 6x6 area instead of 5x5 from other nodes.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
@@ -292,7 +292,7 @@ public abstract class SharedXenoWeedsSystem : EntitySystem
     {
         var range = 5;
         var position = _mapSystem.LocalToTile(grid, grid, coordinates);
-        var checkArea = new Box2(position.X - range, position.Y - range, position.X + range, position.Y + range);
+        var checkArea = new Box2(position.X - range + 1, position.Y - range + 1, position.X + range, position.Y + range);
         var enumerable = _mapSystem.GetLocalAnchoredEntities(grid, grid, checkArea);
 
         foreach (var anchored in enumerable)


Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
```diff
@@ -292,7 +292,7 @@ public bool HasWeedsNearby(Entity<MapGridComponent> grid, EntityCoordinates coor
     {
         var range = 5;
         var position = _mapSystem.LocalToTile(grid, grid, coordinates);
-        var checkArea = new Box2(position.X - range, position.Y - range, position.X + range, position.Y + range);
+        var checkArea = new Box2(position.X - range + 1, position.Y - range + 1, position.X + range, position.Y + range);
```
DrSmugleaf, line 295: Does Box2.CenteredAround not work for this?

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
N0lim: No, changing this code to 

> var range = 10;
var position = _mapSystem.LocalToTile(grid, grid, coordinates);
var checkArea = Box2.CenteredAround(position, new Vector2(range, range));
var enumerable = _mapSystem.GetLocalAnchoredEntities(grid, grid, checkArea);

works the same as bugged code before (range = 9 is the same)

DrSmugleaf (approved):

PR #6172, opened 2025-04-17
Fix engine cooler and fuel enhancer sprites being flipped.

BramvanZijp:
## About the PR
Title, the rotations are now correct.

## Why / Balance
Bug

## Requirements

- [X] I have added media to this PR or **it does not require an ingame showcase.**

**Changelog**
:cl: BramvanZijp
- fix: Fixed Engine Coolers & Fuel Enhancers sprite being inverted when attached to a dropship.


DrSmugleaf (changes requested):
This shouldn't be a problem with the sprites themselves, it just needs to change the rotation in the code

BramvanZijp:
> This shouldn't be a problem with the sprites themselves, it just needs to change the rotation in the code

Even though the sprites ONLY for attachments are flipped? Like hold this side by side wirh the empty one.

BramvanZijp:
This is the currently already in-game sprite for the base texture:
fuel_base

But for some reason the overlay is flipped:
cooling_system_installed
fuel_enhancer_installed

You sure the correct action is to rotate those specific layers?

BramvanZijp:
May not even be possible to only rotate those specific layers with the way they are attached.

BramvanZijp:
> This shouldn't be a problem with the sprites themselves, it just needs to change the rotation in the code

The problem with the sprites themselves is that they are supposed to overlay the base engine attachment sprites, but the attached sprites are flipped compared to the base engine attachment sprite, so I will likely have to edit sprites either way.

BramvanZijp:
Requested changes were able to be applied after all.

DrSmugleaf (approved):
Thanks for the PR!

PR #6173, opened 2025-04-17
Don't show popups to nearby players while invisible

Arendian:
## About the PR

Invisible entities don't show popups to nearby players anymore.

## Why / Balance

Popups from actions like wielding and unwielding your gun while invisible would betray your location to any nearby players.
After this PR they only show for yourself.

## Technical details

If an entity with the XenoActiveInvisibleComponent or the EntityActiveInvisibleComponent does something that shows a popup, it only shows for the client doing the thing and not for other players.

## Requirements

**Changelog**

:cl: Dygon
- fix: Invisible entities no longer show popups to nearby players.

### DIFF
--- a/Content.Server/Popups/PopupSystem.cs
+++ b/Content.Server/Popups/PopupSystem.cs
@@ -1,5 +1,4 @@
-using Content.Shared._RMC14.Stealth;
-using Content.Shared._RMC14.Xenonids.Invisibility;
+using Content.Server._RMC14.Popups;
 using Content.Shared.Popups;
 using Robust.Server.GameObjects;
 using Robust.Server.Player;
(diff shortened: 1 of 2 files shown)


Content.Server/Popups/PopupSystem.cs
```diff
@@ -138,6 +140,10 @@ public override void PopupPredicted(string? message, EntityUid uid, EntityUid? r
 
             if (recipient != null)
             {
+                // RMC14 Don't show popups to other's while invisible.
+                if(HasComp<EntityActiveInvisibleComponent>(recipient) || HasComp<XenoActiveInvisibleComponent>(recipient))
```
DrSmugleaf, line 144: This only changes one of the popup methods, PopupEntity with a single EntityUid also needs to be changed
Would be better to create an RMCPopupSystem, create a ShouldShowPopup method and call that from here

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Server/Popups/PopupSystem.cs
Arendian: I created a separate RMCPopupSystem that  now does the same check.

PopupEntity only shows a popup if the entity is the direct target of it, I think those should still be shown regardless of if the entity causing the popup is invisible(for example trying to take equipment from someone).

DrSmugleaf (approved):

PR #6177, opened 2025-04-18
Add 25 minute end round timer once hijack begins

Vermidia:
## About the PR

Title.

## Why / Balance

parity.

## Technical details

New forceend field set once hijack begins (when the music starts). Checked everytime roundend is checked, and once time goes over it force ends the round in a xeno minor. Changed the xeno minor text to be more general because of this.

## Media

end

## Requirements

**Changelog**

:cl:
- fix: Fixed rounds not ending 25 minutes after hijack begins. (It's a xeno minor)
- tweak: Changed some round end texts.


mmails:
holy fucking soul you souled all over the place

Ian321:
Maybe instead of modifying the round end message add a new one. As it is possible to wipe out the xenos on board and still have time to get out.

Vermidia:
The idea is no matter what minor the xenos will fail because they didn't wipe out the marines before the ship crashes, that way the minor text works out for both (and a xeno minor is still a xeno win technically, it's a little more clear like this).

I can adjust it a little but not sure to what so it keeps both flavors of minor (It's more annoying to make this technically a separate victory type when in CM its the same.)

Ian321:
Maybe

```
[color=purple]The xenos hijacked a dropship[/color] [color=yellow]but where unable to claim the Almayer as their own![/color] 
```

also I feel like an alert that the ship will crash in X minutes every Y minutes is needed.

Vermidia:
I think in CM the time since hijack is listed in the status menu, so maybe it could be shoved on like the time panels? Like instead of showing the time or alert status they have time until the ship crashes on it? (Out of scope of this PR but something like that).

DrSmugleaf:
> The idea is no matter what minor the xenos will fail because they didn't wipe out the marines before the ship crashes, that way the minor text works out for both (and a xeno minor is still a xeno win technically, it's a little more clear like this. It's also the clear the ship is doomed no matter if you wipe them out or not on it).
> 
> I can adjust it a little but not sure to what so it keeps both flavors of minor (It's more annoying to make this technically a separate victory type when in CM its the same.)

It might be better to be able to pass in a custom round end message independent of the outcome
It can be nullable so that it picks the standard outcome message if null

Vermidia:
Did so. Since I did one for hijack timeout I also made a diff message for marine major timeout.

DrSmugleaf (approved):

PR #6189, opened 2025-04-19
Adding end-round statistics

N0lim:
## About the PR

TODO list
- [ ] Basic statistics system
- [ ] Statistics of damage
- [ ] Statistics of heal
- [ ] Statistics of cargo
- [ ] Statistics of building
- [ ] Statistics of equipment usage

## Why / Balance
For some cases it's parity, for some cases it's just fun to see what you did in round, for some cases it is information for game balance

## Technical details
Well, the general idea is as follows
We have systems in the game that do something, for example, they are responsible for taking damage, they send out events that notify other systems of what has happened when certain actions are taken.
Then we create systems that listen to these events and do something with them to turn it into some statistical elements, whether it's just text or animated graphics and other stuff that came to the author's mind.
The statistics elements should implement an appropriate interface so that they can all be processed in the same way in a logical (non-UI) system, what exactly it will do I have no idea at the moment, maybe it will just be called at the end of the round, request and collect all the accumulated statistics for the UI system.
The UI system itself is not remarkable, its task is to ask the logic system and draw it as it should. 
And lastly prototypes, I'm not sure if they will be needed, but in general they should represent a tree of statistical elements that should be collected and rendered in the right order.

изображение

## Requirements

**Changelog**

:cl:
- add: Added round-end statistics

### DIFF
--- a/Content.Server/_RMC14/Requisitions/RequisitionsSystem.cs
+++ b/Content.Server/_RMC14/Requisitions/RequisitionsSystem.cs
@@ -119,6 +119,8 @@ public sealed partial class RequisitionsSystem : SharedRequisitionsSystem
         elevator.Comp.Orders.Add(order);
         SendUIStateAll();
         _adminLogs.Add(LogType.RMCRequisitionsBuy, $"{ToPrettyString(args.Actor):actor} bought requisitions crate {order.Name} with crate {order.Crate} for {order.Cost}");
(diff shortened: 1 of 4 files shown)


PursuitInAshes:
Thanks for the PR!
Going to close this as I don't know if there is intent to return and someone else has started on working on reviving this, I wouldn't want you to come back and end up wasting time on it by doubling up on work.

PR #6199, opened 2025-04-20
Marines now get stunned and thrown by explosions, new cluster OB shockwave effects

Dutch-VanDerLinde:
## About the PR
From CM13
I am not sure if the stun times are exactly 100% right with cm13 will need to double check

Adds missing Cluster OB SHockwave effects

## Media

https://github.com/user-attachments/assets/5faad147-c811-4cd1-a5ad-c3c4fb8f8d06

Cluster OB effects: https://streamable.com/37skzk

:cl:
- fix: Fixed marines not being thrown, blinded and stunned from explosions.
- fix: Fixed Cluster OB not having a shockwave effect.

### DIFF
--- a/Content.Shared/_RMC14/Armor/CMArmorSystem.cs
+++ b/Content.Shared/_RMC14/Armor/CMArmorSystem.cs
@@ -109,6 +109,8 @@ public sealed class CMArmorSystem : EntitySystem
 
     private void OnGetArmor(Entity<CMArmorComponent> armored, ref CMGetArmorEvent args)
     {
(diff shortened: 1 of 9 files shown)


DrSmugleaf (changes requested):
Is the code path in 13 the same for this? This was done with xenos in mind

Dutch-VanDerLinde:
Updated the code to make it use CM13's human calcuations and added a blind and blur effect

DrSmugleaf (approved):
Thanks for the PR!

PR #6231, opened 2025-04-22
Fix tunnels filling with dead paras, + other fixes

### LINKED ISSUE #5279: No Way To Clear Tunnels Of Corpses
## Description

When two parasites/lesser drones die in a xeno tunnel, you are unable to fast travel from that spot unless an admin's around to delete the dead.

Vermidia:
## About the PR

Title. + Fixed tunnel blocking working wrong.

## Why / Balance

Parity. tunnels aren't supposed to block you if they're too full, they only block you if the destination is too full. Also they aren't supposed to get full of dead stuff. Fixes #5279

Also added examine text

## Technical details

XenoTunnelComp is always added to anything entering the tunnel that's a mob to ensure all the other stuff happens to it (because paras enter the tunnel then die). Some replacing of obsolete methods. Added stuff for examine. One thing was just to make it clearer which tunnel was being checked.

## Requirements

**Changelog**

:cl:
- fix: Fixed tunnels blocking travel if they were too full. They now block travel if the destination tunnel is too full.
- fix: Fixed dead parasites getting stuck in tunnels.
- fix: Non-hivemembers can now examine tunnels to see if they contain something.


DrSmugleaf (approved):
Thanks for the PR!

PR #6238, opened 2025-04-22
Use the right CanInsert for ItemSlot transfers

Arendian:
## About the PR

I changed the CanInsert check used for transferring entities between ItemSlots

## Why / Balance

#6213 Fixed the whitelist issue, but prevents transferring between ItemSlots(Flare pack to flare pouch for example) as it uses the wrong CanInsert function(I was tired and stupidly only tested if the whitelist was respected)

## Requirements

**Changelog**

No CL, I can't believe I missed this a second time.


DrSmugleaf (approved):
Thanks for the PR!

PR #6242, opened 2025-04-22
Increase rav empower range to match parity area

Vermidia:
## About the PR

Title.

## Why / Balance

Like tremor/stop/other area based moves, ravager's empower area was much small diagonally because circles vs squares. Range has been increases so the area is the same. (Area is 81 in CM, area of a circle with radius 5.078 is about 81)

## Requirements

**Changelog**

:cl:
- fix: Fixed Ravager's empower range being too small (4 > 5.078)


DrSmugleaf (approved):
Thanks for the PR!

PR #6253, opened 2025-04-24
Remove the time limit on the death area outside the docking bay

Arendian:
## About the PR

The entities causing everyone outside the docking bay to die now don't despawn after 30 seconds.

Another solution could be making it so the lifeboat doors don't reopen if something is on their tile when they close, I can change it to that if that is preferred.

## Why / Balance

The doors keep reopening if a corpse/body is on them, allowing players to still go into space after the kill area despawned..

## Requirements

**Changelog**

:cl: Dygon
- tweak: The kill area outside the docking bay doesn't despawn anymore.

### DIFF
--- a/Resources/Prototypes/Entities/Tiles/death.yml
+++ b/Resources/Prototypes/Entities/Tiles/death.yml
@@ -34,5 +34,3 @@
   - type: Tag
     tags:
     - HideContextMenu
-  - type: TimedDespawn
-    lifetime: 30


Vermidia:
For bodies getting stuck in the doors, I think CM pushes them outwards into space or something, so that's also a solution.

Dutch-VanDerLinde:
You can make it bolt shut instead of being 'locked'
Like how the dropship doors function, you basically cant open them
Making them actually bolt would also fix the bolt visuals for the lifeboat doors not working

DrSmugleaf (approved):
Can do the door part in a separate PR for now this is fine

PR #6254, opened 2025-04-24
the big iff refactor

Ian321:
## About the PR

Refactor of the IFF system to allow for dynamic configuration of IFF factions.

## Why / Balance

why not

## Technical details

- Removed FactionFrequencies, IFFFaction and UserIFF component and made the systems based on npc factions.
  - Moved the frequencies into the faction.
- Adjusted ItemIFF to take a npc faction.
- Fixed a ghosts being in the wrong faction..
  - The problem was, that the NpcFactionMemberComponent did not sync.
- ~~Fixed a bug caused by the random markings.~~

## Media

https://github.com/user-attachments/assets/7b76d27f-c0cc-4a1d-95f7-a1f118969ad4

## Requirements

**Changelog**
:cl:
ADMIN:
- add: IFF can now also be configured via the factions admin menu.


Ian321:
Have to also make it sync the faction data when the client starts.

Content.Shared/_RMC14/NPC/NpcFactionPrototype.cs
```diff
@@ -0,0 +1,10 @@
+using Content.Shared.Radio;
+using Robust.Shared.Prototypes;
+
+namespace Content.Shared.NPC.Prototypes;
+
+public sealed partial class NpcFactionPrototype : IPrototype
+{
+    [DataField]
+    public HashSet<ProtoId<RadioChannelPrototype>> Channels = new();
```
Dutch-VanDerLinde, line 9: Making it already use an existing system is great, but the lack of NPC factions being prototypes instead of entity prototypes makes it less modular, since you have to add a new datafield in the prototype rather than just added it to faction entity proto

Ian321: Without the extra IFFFaction Component it would be hard and a lot of duplicate code to find their radio channels. This way it has all relevant information in one place that is easy to query.

Furthermore, I don't think there are any other properties that are related to factions.

AverageNotDoingAnythingEnjoyer:
Merge master, Vro. Integration tests were fixed.

Ian321:
Is the tesla turret supposed to attack neutral targets?

DrSmugleaf (changes requested):
Merging npc factions and iff prototypes together makes it so you don't get any of the niceties of being able to split data in components and being able to raise events on the faction entity itself (see surgeries for a developed example), but more importantly makes it so that if you want to change the mob's npc faction separately from their intrinsic user IFF, you can't do that anymore
In an ideal world npc factions themselves would be entity prototypes but that would require upstream changes

Dutch-VanDerLinde:
u coming back to this?

if you are you could make it so theres no npc factions and it explicity uses IFF factions

Ian321:
> u coming back to this?
> 
> if you are you could make it so theres no npc factions and it explicity uses IFF factions

AI behavior uses the NpcFaction, so no.

One possibility would be to add another system for IFF friendlies that would require another UI to configure (IFF Console).

PR #6257, opened 2025-04-24
Fixes the economy of weapon specialist ammo

Arcticular1:
## About the PR
Changes the amount of weapon specialist ammo you can buy.

## Why / Balance
The main reason i made this PR is because the current meta for weapon specialists is buying grenades and exchanging them to req for specialist ammo due to the extreme value difference of grenades compared to the other ammo in the vendors. (basically putting yourself at a disadvantage for choosing ammo for your class).

This also becomes an IC issue in my opinion, as i do not think it makes sense that the UNMC is not providing their specialists enough ammo to a point where they specifically need to order ammunition from requisitions to be ready for an operation. 

This lines up the ammo prices in the vendor to the requestion prices. (each crate of ammunition being worth 3000) 

## Technical details
YML.

## Media
image

## Requirements

**Changelog**

:cl:
- tweak: Changed the amount of ammo weapon specialists get in their vendor.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/specialist.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/specialist.yml
@@ -36,27 +36,45 @@
     - name: Extra Scout Ammunition
       entries:
       - id: RMCMagazineRifleM4SPRA19Impact
+        name: A19 HV high impact magazine (10x24mm) (3)
         points: 40
+        spawn: 3
       - id: RMCMagazineRifleM4SPRA19Incendiary
+        name: A19 HV incendiary magazine (10x24mm) (3)
         points: 40
+        spawn: 3
       - id: RMCMagazineRifleM4SPRA19
+        name: A19 HV magazine (10x24mm) (5)
         points: 40
+        spawn: 5
     - name: Extra Sniper Ammunition
       entries:
       - id: CMMagazineSniperM96SFlak
+        name: M96S flak magazine (10x28mm) (5)
         points: 40
+        spawn: 5
       - id: CMMagazineSniperM96SIncendiary
+        name: M96S incendiary magazine (10x28mm) (5)
         points: 40
+        spawn: 5
       - id: CMMagazineSniperM96S
+        name: M96S marksman magazine (10x28mm) (5)
         points: 40
+        spawn: 5
       - id: RMCMagazineSniperXM43E1AntiMateriel
+        name: XM43E1 marksman magazine (10x99mm) (3)
         points: 40
+        spawn: 3
     - name: Extra Demolitionist Ammunition
       entries:
       - id: RMCRocket84mmAntiArmor
+        name: 84mm anti armor rocket (3)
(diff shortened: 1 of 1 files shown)


DummeHytteOst:
its intentional that they don't get that much extra special ammo, so they have to rely on requsition if they run out, like everyone else, we are also missing the inflation mechanic.

i do not believe this non parity change to be good for the game.

Arcticular1:
> its intentional that they don't get that much extra special ammo, so they have to rely on requsition if they run out, like everyone else, we are currently missing the inflation mechanic too.
> 
> i do not believe this non parity change to be good for the game.

Honestly i do agree, i was thinking about maybe making it in TOTAL what you can get in 2 crates - 6 rockets, 6 AMR mags, etc. The main issue is that i dont want to feel as if you're throwing the game by not doing nade tax (an example would be today a QM denied service to specialists who didnt give them nades)

QuietlyWhisper:
Would having the scaling requisitions order costs we don't have yet make a difference

PolarTundraa:
the scaling requisition order costs would help, but it still wouldn't stop grenades being the most outright money efficient option to purchase in the vendor

DummeHytteOst:
> Would having the scaling requisitions order costs we don't have yet make a difference

inflation would probably help, i imagine once grenades also start giving daze, stun etc. to marines, that it would also help.
these two changes would probably make marines start using grenades less haphazardly, and only when they actually need them, which in turn would decrees the demand for grenades.

QuietlyWhisper (commented):
I don't think giving specs 3-5x more ammo than they're supposed to get is a good solution without considering any balance implications of that

What I would suggest is removing grenades from this vendor all together and having the grenadier have their own specialist grenades instead

I'll ask smug for some opinions as well but I don't think this is the right way to go about it

DrSmugleaf (changes requested):
> The main reason i made this PR is because the current meta for weapon specialists is buying grenades and exchanging them to req for specialist ammo due to the extreme value difference of grenades compared to the other ammo in the vendors. (basically putting yourself at a disadvantage for choosing ammo for your class).

Even if you are playing optimally as QM, the only situations where it makes sense for specs to hand you grenades instead of buying their own ammo is where you have 1 or more grenadier specialists, or when you have a few marines that are very good with grenades, as you do not have enough money to buy more than 1 crate of ammo for each spec (if that) depending on the pop if you are playing properly and buying AP ammo, while HEDP grenades are one of the things you get the most of at req for free
The only thing this would fix is people misplaying assuming that it is worth it when it is not, which to be fair I am extremely tired of playing QM and telling specs to stop buying me grenades, because it is not a good play most of the time, and is just a carryover from the old req budget, but the solution is not handing specs a massively increased of ammo, that is a massive balance change
There isn't really a good fix for this until we rework the grenadier specialist post parity to one, not be bad, and two, have its own grenades that only its launcher can fire

Arcticular1:
I'm going to close this, as this fix is not without it's major setbacks in terms of balancing. 

I'm happy to look for compromises but in the meantime this PR just fundamentally not healthy for the game in it's current state.

PR #6263, opened 2025-04-24
Powered Floodlights

CyberTropic:
## About the PR

Colony floodlights now behave the same way as streetlights on Hybrisa: they only turn on when there is power in the colony.
~~Added a small floodlight variation of landing light which is always on, small floodlights that are currently at LZs will need to be replaced.~~ I got it wrong, the small floodlights currently ARE the landing lights.
In true parity, tall floodlights should be toggleable by a floodlight switch, and small floodlights should be toggleable individually, but that goes beyond a yml change (I'm pretty sure). 

## Why / Balance

Parity

## Media

*Unpowered at map start*
Screenshot from 2025-04-24 15-03-07

*Once power is turned on*
Screenshot from 2025-04-24 15-07-06

## Requirements

**Changelog**

:cl:
- add: Colony floodlights now use colony power.


Dutch-VanDerLinde:
theres supposed to be a floodlight switch that you can toggle in the lv-624 power room

CyberTropic:
> theres supposed to be a floodlight switch that you can toggle in the lv-624 power room

Yeah there's at least one on every map, but the switch doesn't do anything if the lights are unpowered.

DrSmugleaf (approved):
Thanks for the PR!

PR #6267, opened 2025-04-24
Fix baton slugs losing their acid after being shot again

Arendian:
## About the PR

The new baton slug spawned when a baton slug projectile hits anything now inherits the corrosion timer of the projectile that spawned it.

## Why / Balance

You were able to remove acid from baton slugs by just loading one and firing it again.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed HIRR baton slugs having their acid removed from them when fired.


DrSmugleaf (approved):
Thanks for the PR!

PR #6278, opened 2025-04-25
Katana parity damage

LightGreyStripe:
## About the PR

Katana's damage value
## Why / Balance

Parity
## Media

image
image

## Requirements

**Changelog**

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Melee/sword.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Melee/sword.yml
@@ -63,7 +63,7 @@
   - type: MeleeWeapon
     damage:
       types:
-        Slash: 15
+        Slash: 45
     soundHit:
         path: /Audio/Weapons/bladeslice.ogg
   - type: Item


LightGreyStripe:
media is from cm13 github for additional proof it's parity

mmails:
since when do we have a katana wtf isnt it just the upstream one

LightGreyStripe:
> since when do we have a katana wtf isnt it just the upstream one

this katana is in hyrbisa in jacks, on the table

mmails:
> > since when do we have a katana wtf isnt it just the upstream one
> 
> this katana is in hyrbisa in jacks, on the table

this is new tech to me thats insane

CyberTropic:
Instead of modifying the upstream one, you should make an RMCKatana version and put it in the /_RMC14 folder. It could then inherit from the base katana, like how the fireaxe does it for example. 👍

DrSmugleaf (changes requested):
exincore's review needs to be addressed, our own prototypes should be made where possible instead of modifying and relying on the upstream one

LightGreyStripe:
> > that sounds like a lot of work
> 
> and here's how to do it! ✨
> 
> In `Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Melee/sword.yml` add the prototype. You only need to list the components you change from the parents.
> 
> ```yaml
> - type: entity
>   parent: [ Katana, RMCBaseMeleeWeapon ]
>   id: RMCKatana
>   suffix: RMC
>   components:
>   - type: MeleeWeapon
>     damage:
>       types:
>         Slash: 45
>     soundHit:
>         path: /Audio/Weapons/bladeslice.ogg
> ```
> 
> Note that RMCBaseMeleeWeapon has a MeleeDamageMultiplierComponent which multiplies damage done to Xenonids by 1.5 by default.
> 
> You may also add a PryingComponent if the Katana is able to pry in CM13.
> 
> ```yaml
>     - type: Prying
>     speedModifier: 100000
> ```
> 
> And if the Katana does more damage while wielded in both hands there's a component for that too! Just make sure all the numbers add up right in the end.
> 
> ```yaml
>   - type: IncreaseDamageOnWield
>     damage:
>       types:
>         Slash: 10
> ```
> 
> Then replace all instances of `Katana` with `RMCKatana` in the RMC maps. No database migration needed, I think.
> 
> I think I covered most of it?

should i do this then? also the katana can't be wielded and it can't pry doors

DrSmugleaf:
> should i do this then? also the katana can't be wielded and it can't pry doors

Yeah

DrSmugleaf:
Done in https://github.com/RMC-14/RMC-14/pull/6551

PR #6280, opened 2025-04-25
Fix xenonid actions with a DoAfter being useable while resting

Arendian:
## About the PR

Resting now prevents using all actions that have a doafter.
If a xenonid rests during a  doafter, the doafter will be interrupted.
While checking CM13 to make sure the above changes were parity I also noticed that the RMC14 cooldown is 20 seconds instead of the 30 seconds in CM13, so I added that to this PR too.

## Why / Balance

Parity

## Requirements

**Changelog**

:cl: Dygon
- fix: Resting will now interrupt and prevent the use of all actions with a DoAfter.

### DIFF
--- a/Content.Shared/DoAfter/DoAfterArgs.cs
+++ b/Content.Shared/DoAfter/DoAfterArgs.cs
@@ -155,6 +155,13 @@ public sealed partial class DoAfterArgs
     /// </summary>
     [DataField]
     public bool RequireCanInteract = true;
+
+    /// <summary>
+    ///     RMC14
+    ///     If the doafter should break when the user rests.
+    /// </summary>
+    [DataField]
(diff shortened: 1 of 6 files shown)


Vermidia (changes requested):
We need this to be a general thing you can attach to doafters ideally. Many xeno doafters need something like this. (Bombard, Charge, etc)

Arendian:
> We need this to be a general thing you can attach to doafters ideally. Many xeno doafters need something like this. (Bombard, Charge, etc)

Resting now interrupts all DoAfters, I also moved the Bombard ability cooldown to another PR as it doesn't really fit the scope of this one anymore.

DrSmugleaf (approved):

PR #6288, opened 2025-04-27
Scale the cost of resin constructs with how many there are in an area

Arendian:
## About the PR

If more than 40% of the buildable tiles in an area has a resin construct on it, new resin constructs in that area are more expensive to build.

## Why / Balance

Parity.

## Technical details

If the amount of resin constructs divided by the buildable tiles in an area is higher than 0.4(density), the resin cost of new constructs in that area is increased using the following formula: base cost * (density + 0.35) * 2.
As an example: a resin wall with a base cost of 120 plasma that is built right when the threshold is exceeded will cost 120 * (0.4 + 0.35) * 2 = 180 plasma.

## Requirements

**Changelog**

:cl: Dygon
- tweak: The plasma cost of new xeno constructs now scales with how many existing constructs there are in an area if they exceed a threshold(40% of buildable tiles in an area).

### DIFF
--- a/Content.Shared/_RMC14/Areas/AreaSystem.cs
+++ b/Content.Shared/_RMC14/Areas/AreaSystem.cs
@@ -3,7 +3,9 @@ using System.Linq;
 using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.GameStates;
 using Content.Shared._RMC14.Warps;
+using Content.Shared._RMC14.Xenonids.Construction;
 using Content.Shared.Coordinates;
+using Content.Shared.Damage;
 using Content.Shared.GameTicking;
 using Content.Shared.Maps;
 using Content.Shared.Popups;
@@ -294,6 +296,21 @@ public sealed class AreaSystem : EntitySystem
         return area.Value.Comp.SupplyDrop;
     }
 
+    private void UpdateBuildableTiles(Entity<AreaGridComponent> grid, EntProtoId<AreaComponent> areaProto, EntityUid? anchored = null)
+    {
+        foreach (var (area, uid) in grid.Comp.AreaEntities)
+        {
+            if (areaProto != area || !TryComp(uid, out AreaComponent? areaComp))
+                continue;
+
+            if (HasComp<XenoConstructComponent>(anchored))
+                areaComp.ResinConstructCount++;
+
+            areaComp.BuildableTiles++;
+            Dirty(uid, areaComp);
+        }
+    }
+
     public override void Update(float frameTime)
     {
         try
@@ -322,6 +339,12 @@ public sealed class AreaSystem : EntitySystem
                         {
                             areaGrid.Colors[pos] = minimapColor.Color;
                             found = true;
+
+                            if (areaGrid.Areas.TryGetValue(pos, out var areaProto))
+                            {
+                                if (HasComp<DamageableComponent>(anchored))
+                                    UpdateBuildableTiles((ent,areaGrid), areaProto, anchored);
+                            }
                         }
 
                         if (_areaLabelQuery.HasComp(anchored))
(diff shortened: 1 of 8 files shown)


Content.Shared/_RMC14/Areas/AreaSystem.cs
```diff
@@ -322,6 +339,12 @@ public override void Update(float frameTime)
                         {
                             areaGrid.Colors[pos] = minimapColor.Color;
                             found = true;
+
+                            if (areaGrid.Areas.TryGetValue(pos, out var areaProto))
+                            {
+                                if (HasComp<DamageableComponent>(anchored))
+                                    AddBuildableTile((ent,areaGrid), areaProto, anchored);
+                            }
```
DrSmugleaf, line 347: You shouldn't be doing this in the middle of the minimap trying to color itself. You should pull this out to its own section of code. Anchored here is relying on it having a minimap color component, which isn't correct.

Content.Shared/_RMC14/Areas/AreaSystem.cs
```diff
@@ -339,13 +362,16 @@ public override void Update(float frameTime)
...
-                        continue;
-                    }
+                        AddBuildableTile((ent, areaGrid), area);
 
+                        if (areaComp.MinimapColor != default)
+                        {
+                            areaGrid.Colors[pos] = areaComp.MinimapColor.WithAlpha(0.5f);
+                            continue;
+                        }
+                    }
```
DrSmugleaf, line 374: Same here

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
```diff
@@ -373,15 +387,25 @@ private void OnXenoSecreteStructureDoAfter(Entity<XenoConstructionComponent> xen
         var coordinates = GetCoordinates(args.Coordinates);
         if (!coordinates.IsValid(EntityManager) ||
             !xeno.Comp.CanBuild.Contains(args.StructureId) ||
-            !CanSecreteOnTilePopup(xeno, args.StructureId, GetCoordinates(args.Coordinates), true, true))
+            !CanSecreteOnTilePopup(xeno, args.StructureId, GetCoordinates(args.Coordinates), true, true) ||
+            !_area.TryGetArea(GetCoordinates(args.Coordinates), out var area, out _))
```
DrSmugleaf, line 391: This area check should be done as part of the density check, not stop the whole method from running, in case someone loads a map without areas

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
Arendian: I will probably be able to address the reviews soon

DrSmugleaf (dismissed):

Content.Shared/_RMC14/Xenonids/Construction/SharedXenoConstructionSystem.cs
DrSmugleaf: Done in [`b467dc2` (#6288)](https://github.com/RMC-14/RMC-14/pull/6288/commits/b467dc25bba4b47429a448cfd299a99ee52238a4)

Content.Shared/_RMC14/Areas/AreaSystem.cs
DrSmugleaf: Done in [`b467dc2` (#6288)](https://github.com/RMC-14/RMC-14/pull/6288/commits/b467dc25bba4b47429a448cfd299a99ee52238a4)

Content.Shared/_RMC14/Areas/AreaSystem.cs
DrSmugleaf: Done in [`b467dc2` (#6288)](https://github.com/RMC-14/RMC-14/pull/6288/commits/b467dc25bba4b47429a448cfd299a99ee52238a4)

PR #6311, opened 2025-04-30
Fix bleedout happening on fire/on weeds/as larva

Vermidia:
## About the PR

Title.

## Why / Balance

Bug.

## Technical details

Conditions for returning were wrong, so if you weren't on fire it returned, but it also returned if you didn't heal on weeds and weren't on them, but it needs to only return if you aren't on fire, don't heal off weeds, and are off weeds. Otherwise bleedout is 0.

## Media

https://github.com/user-attachments/assets/be2a4c00-59c7-4eed-b0fa-014222272982

## Requirements

**Changelog**

:cl:
- fix: Fixed bleedout still occurring on fire, on weeds, and to larva in most cases.


DrSmugleaf (approved):
Thanks for the PR!

PR #6321, opened 2025-05-01
Movement based devour breakouts

Vermidia:
## About the PR

Title.

## Why / Balance

Parity. 

## Technical details

DevouredComponent now stores time between struggle attempts. You can no longer attack while devoured, instead attempting to move with use the weapon in your hand to do damage - this gets the weapons melee damage with the only modifier being x0.625 (for reference in parity it's random between 1/4th of the weapons damage and the full weapons damage, this is the average of those). The xeno melee modifier does not apply here. Gib chance was not added in this PR. Note that you must reattempt to move and can't hold down a key - this is a limitation of moveinput, but is technically how it works parity wise so shrug.

Note in general I think this means the DPS while devoured is a bit higher than it was because of this, slightly? Can be adjusted if needed

## Media

https://github.com/user-attachments/assets/88c631da-4e67-42ea-9328-d9991771ecbb

## Requirements

**Changelog**

:cl:
- fix: Fixed attack rate and damage while devoured to be more accurate.
- fix: You can no longer click in harm mode to attack while devoured. Instead moving or clicking (out of harm mode) will attempt to attack the devourer.


Vermidia:
I did try melee but problem is I think that would require modifying the melee system since I couldn't figure out how to make it eat a melee attack.

Vermidia:
This mostly works with overriding melee attack now. It mispredicts a bit with move inputs and attacks but works well otherwise.

DrSmugleaf (dismissed):

PR #6326, opened 2025-05-02
Mortar kit storage fix 2

Vermidia:
## About the PR

Title.

## Why / Balance

~So the mortar kit was originally a removeonlystorage but that broke things, but removing that made the box basically the best way to store it. It's been a bit and the original issue is still present but we got the better belts and backpacks now, so I'm temporarily fixing it in a different way that preserves how the box is supposed to work.~

~Now instead of the mortar kit spawning on the map, a package containing the kit spawns which spawns the kit, which works. (Not including any recent breakage of remove only storage for some pill packets and MREs).~

Just changes the proto back.
Also included was a recent parity change to change how many rangefinders are in the kit.

Needs the storage fix.

## Technical details

Just YML.

## Requirements

**Changelog**

:cl:
- fix: Fixed mortar kit being able to have items reinserted into it.
- fix: Mortar kit now has 1 rangefinder and 1 laser designator, instead of 2 rangefinders


DrSmugleaf (approved):
YAML looks good not 100% sure on the unpacking since it might entice QMs to not fill the belt and pack with shells but probably fine
Need to fix RemoveOnlyStorage which I think would be better fixed by making storage fill force insert and excluding the items from whitelist like 13 does
Mortar is going to be even more dogshit to carry around and use but that's parity so

DrSmugleaf (changes requested):
Thinking about this more if the new kit package unpacking is just to work around the bug I think I would rather fix it properly first so we can do it properly like 13

Vermidia:
Sure. I'll just change it to fix the YML then.

DrSmugleaf (approved):
Changes look good, going to wait until laser designator lasing for mortar is done before merging this

PR #6328, opened 2025-05-02
Fix airburst grenades hitting marines or xenos before reaching their destination, when shot from an underbarrel grenade launcher

Arendian:
## About the PR

The underbarrel launcher now arcs.
The underbarrel launcher now has IFF, this is only used so the projectiles of any grenades it shoots can inherit the IFF faction of the shooter.

## Technical details

I made the GiveAmmoIFF method check for the outer container instead of the ContainingContainer so that it also checks correctly for attachments(before this it would check the gun for it's IFF faction if an attachment did the shooting). This is needed to make the starshell projectiles  inherit the IFF of the attachment.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed airburst grenades ricocheting off of marines or xenos when shot from an underbarrel grenade launcher.
- fix: Fixed starshell projectiles not having working IFF when shot from an underbarrel grenade launcher.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/IFF/GunIFFSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/IFF/GunIFFSystem.cs
@@ -138,7 +138,7 @@ public sealed class GunIFFSystem : EntitySystem
         {
             owner = gun;
         }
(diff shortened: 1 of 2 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #6336, opened 2025-05-02
Adjust fire damage of projectiles and the CAS laser.

Arendian:
## About the PR

- The CAS laser cannon now deals twice as much initial damage on a direct hit.
- Resisting fire is now half as effective, making it so you have to drop and roll twice for most fire sources.(green and CAS laser fires still require only one).
- The fire damage over time applied by projectiles now deals 50% more damage per second.
- You now take double damage if a fire tile spawns below you and you don't move, previously you had to move before the game would notice you were on a fire tile.

## Why / Balance

Parity. The fire damage from projectiles and CAS laser is currently lower than in CM13.

- Getting hit directly by the CAS laser cannon now deals 200 damage up from 100,
- Stepping into a fire tile from the CAS laser now instantly inflicts 75 damage.

## Technical details

The OnIgniteOnProjectileHit method passed the wrong values to the Ignite method, it passed Stacks(default value of 20) as intensity and intensity(default value of 30) as duration. Now it passes the proper values, making it so the intensity of all projectiles using the default IgniteOnProjectileHitComponent(almost all of them) now have an intensity of 30 instead of 20. Their duration is now also 20 instead of 30.

I halved the amount of seconds the firestacks would be reduced while resisting from 2 to 1, effectively making to so you remove 10 firestacks per resist instead of 20. You lose 10 stacks per second for 1 second, instead of the previous 10 stacks per second for 2 seconds.

You now receive the SteppingOnFireComponent even if the tile failed to ignite you.

## Media

https://github.com/user-attachments/assets/f6f12c44-b707-4a85-84f1-3d9c7aafaf1b

## Requirements

**Changelog**

:cl: Dygon
- tweak: Resisting fire(Drop and Roll) is now half as effective as before, it now removes 10 fire stacks instead of 20.
- tweak: A direct hit from the CAS laser now deals 200 damage instead of 100, walking into a laser fire tile deals 75 damage.
- fix: Fire from incendiary projectiles now deals 50% more damage.
- fix: You no longer need to move before you start taking double fire damage when a fire tile spawns below you.
- fix: Fortifying as a defender while on a fire tile no longer prevents taking double fire damage.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
@@ -425,7 +425,7 @@ namespace Content.Server.Atmos.EntitySystems
             _stunSystem.TryParalyze(uid, flammable.ResistDuration, true, force: true);
 
             // TODO FLAMMABLE: Make this not use TimerComponent...
-            uid.SpawnTimer(2000, () =>
+            uid.SpawnTimer(1000, () =>
             {
                 flammable.Resisting = false;
                 Dirty(uid, flammable);
--- a/Content.Shared/_RMC14/Atmos/IgniteOnProjectileHitComponent.cs
+++ b/Content.Shared/_RMC14/Atmos/IgniteOnProjectileHitComponent.cs
@@ -8,9 +8,6 @@ namespace Content.Shared._RMC14.Atmos;
 public sealed partial class IgniteOnProjectileHitComponent : Component
 {
     [DataField, AutoNetworkedField]
-    public int Stacks = 20;
-
-    [DataField, AutoNetworkedField]
     public int Intensity = 30;
 
     [DataField, AutoNetworkedField]
--- a/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
+++ b/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
@@ -119,7 +119,7 @@ public abstract class SharedRMCFlammableSystem : EntitySystem
 
     private void OnIgniteOnProjectileHit(Entity<IgniteOnProjectileHitComponent> ent, ref ProjectileHitEvent args)
     {
-        Ignite(args.Target, ent.Comp.Stacks, ent.Comp.Intensity, ent.Comp.Duration, false);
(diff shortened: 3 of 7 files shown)


0tito:
Surely this doesnt thanos snap boilers

BramvanZijp:
> Surely this doesnt thanos snap boilers

Ive killed boilers before with just laser fire, so it probably does.

Arendian:
I undid the impact damage change, since it's possible on RMC14 to get hit by multiple shots at once if you stand in between tiles. When that is addressed it can be set to 200 again.

DrSmugleaf:
> I undid the impact damage change, since it's possible on RMC14 to get hit by multiple shots at once if you stand in between tiles. When that is addressed it can be set to 200.

Using chains should fix that, see how spray acid does it

Arendian:
> > I undid the impact damage change, since it's possible on RMC14 to get hit by multiple shots at once if you stand in between tiles. When that is addressed it can be set to 200.
> 
> Using chains should fix that, see how spray acid does it

I just realized it already spawns a chain, so it wasn't an issue. I initially tested it by spawning individual fires through the spawn menu instead of flying down and calling CAS on myself. This means it should be fine to merge with the damage at 200.

Arendian:
I added a video showing how much damage the CAS laser does on a direct and indirect hit using 3 different xenonid castes.

DrSmugleaf (changes requested):
Make sure you are testing this in 13 with the laser cannon 

https://github.com/user-attachments/assets/9727d8bb-6f13-4fd0-95d6-78464536de30

DrSmugleaf:
I copied the resist change to https://github.com/RMC-14/RMC-14/pull/6486 for now until you can check that the dps is the same as 13, if you can record it as well

Arendian:
> Make sure you are testing this in 13 with the laser cannon
> 
>  lasercannon.mp4

It looks like boilers don't take the impact damage but smaller xeno's do as runner's do lose all their health when directly hit, drones also lose the correct amount. I'll go check if all T3 xeno's are immune to the impact damage from the cannon or if it's just the boiler.

Arendian:
I forgot I still had this PR. Most of the stuff it was going to do was done in different PR's, so it's not needed anymore.

PR #6337, opened 2025-05-02
Lightning Effect

CyberTropic:
## About the PR

Adds lightning effects using ambient light system, and plays a thunder sound.
Fixes rmclight command autocomplete.

## Why / Balance

Parity

## Technical details

Lightning is handled the same way as in parity: every 5 seconds, when there's not a special light animation going on, we do a check with some probability that's given per weather event, and if we pass we do a lightning color sequence and play a thunder sound. This uses the ambient light system which only shows the lightning outdoors.

## Media

https://github.com/user-attachments/assets/4c3e5156-c4f9-42bb-8d7e-8b7504b70a12

## Requirements

**Changelog**

:cl:
- add: Added lightning effects when it rains.
- fix: Fixed rmclight command for ambient lighting changes.


mmails:
THIS IS SO FUCKING SOUL!! SPEEDMERGE!!!

DrSmugleaf (approved):
Thanks for the PR!

PR #6339, opened 2025-05-03
Lesser Carrier

Vermidia:
## About the PR

It's just lil. Admin spawn only.

## Why / Balance

Cute...Sprites by Sigma Draconis.

It has the following traits from lesser drones:
tier 0
dies off weeds
no crit state
all abilities are not as good as base caste.
size and evasion.
Needs 2 hands to hold egg.

From Carrier:
It can store 2 parasites and 1 egg at a time. Parasites don't die while held.
It has more health than a lesser drone (160->200)
It has less plasma than a lesser drone (200->150) and regens 0.5 less then carrier (3.5 -> 3)
It can't plant resin traps, but it can have 1 sustained egg like eggsac carrier (who gets 4). The range for sustaining an egg has also been lowered from 14 to 7.
It does 5 more damage than a lesser.
It has the same speed as a carrier.
It has 1/2 as much explosion armor.
Egg plant time is slower (1 -> 2)
Tackle chance is not as good (min 3 max 5 instead of min 2 max 4, 40% chance instead of 50%. Drone Lesser is 4-5 0.35 chance. Stun time is default instead of increased.)

Also comes with it's own map blip.

## Media

xaby

## Requirements

**Changelog**

no cl players will only know she exists when they spot her.


SigmaTheDragon:
Xaby

Kadeo64:
Xaby

DrSmugleaf (dismissed):
Thanks for the PR!

PR #6348, opened 2025-05-04
Nerf M79 & Buff M83 Firerates

Keer-Sar:
## About the PR

Swaps the baseFireRate yml entry in the M79 and M83 with each other. 

M79 was 0.833 => Now 0.3125
M83 was 0.3125 => Now 0.833
## Why / Balance

M79 is accessable and usable by ALL MARINES
- Six are usually stocked in requisitions
- Any Marine can buy one for 30 points from their barracks vender
- No additional skill checks are needed to use it optimally
- Limit of use is by requisition's budget. If req does not hand out enough grenades then the marine is SOL.

M79 is better than the Weapon specialist's M83
- Firerate is faster than M83
- Reload speed is mechanically capped at how fast you can cycle thru keybindings
- Wield delay is the same for both M83 and M79
- It is common to see Grenadier Weapon Specialists take a M79 from Requisitions into battle instead of their Special M83 gun that only they can use and is un-meltable.
## Technical details

Swaps the numbered used in: 

m79_grenade_lancher.yml
type: RMCSelectiveFire
         baseFireRate: 0.833 [OLD => 0.3125]

m83_grenade_launcher.yml
type: RMCSelectiveFire
         baseFireRate: 0.3125 [OLD => 0.833]

## Media

https://github.com/user-attachments/assets/4ae97f78-f45a-4b08-996d-fbf10e8200bb

## Requirements

**Changelog**

:cl:
- tweak: The M79 and M83 grenade launchers' fire rates are swapped with each other.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m79_grenade_launcher.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m79_grenade_launcher.yml
@@ -18,7 +18,7 @@
     soundGunshot:
       path: /Audio/_RMC14/Weapons/Guns/Gunshots/m79_shoot.ogg
   - type: RMCSelectiveFire
-    baseFireRate: 0.833
+    baseFireRate: 0.3125
   - type: BallisticAmmoProvider
     whitelist:
       tags:
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m83_grenade_launcher.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m83_grenade_launcher.yml
@@ -1,4 +1,4 @@
-﻿- type: entity
+- type: entity
   name: M83 grenade launcher
   parent: [ RMCBaseWeaponGrenadeLauncher ]
   id: WeaponLauncherM83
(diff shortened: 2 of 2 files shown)


DrSmugleaf:
We aren't doing micro balance adjustments until a later date for the same reason that neither the Healer Drone, MOU53, Lurker, Vanguard Praetorian, incendiary OB, or mortar have been changed.
If we were to do a balancing pass it would need to be a coordinated one on the whole game.
We have neither the people, time, resources, or policy yet to focus on these kind of changes for things that are not different because of the fundamental differences between 13 and 14 (see barricade and resin structure invincibility frames for something that is because of a fundamental difference).
Once we do we can revisit this and any other balance issues.

As for the PR itself:
I'm aware that the non spec grenade launcher is better than the spec grenade launcher in the vast majority of cases, the way 13 deals with this is admin enforcement forcing you to use your spec gear (the grenadier spec isn't allowed to use the M79) which is a shit solution and not one we are going to be doing.
Also, you chose these numbers extremely arbitrarily. This makes the M79 single shot standalone gun worse than having a U1 grenade launcher attachment with a capacity of 3 grenades, which is fundamentally unworkable as a change.
This also doesn't handle the change that we eventually want to make (post-parity) that changes what kind of grenades the grenadier spec uses, to fix other issues as well, such as their grenades being stolen and being the only spec that can get value after their death, as their ammo can be reused by other marines where the ammo of every single other spec cannot (barring foxtrot, or a sniper spec using amr ammo, which isn't a problem in parity either with a cap of 1 spec per archetype instead of 2), leading to it being risky to supply a spec ammo if they end up dying.

As for a change that is parity, and you could do if you want to look into it and bring the two launchers closer together, the M79 has a very brief delay after loading that doesn't allow it to fire
I'm not sure where that is in the code of 13 exactly but here's a video example

https://github.com/user-attachments/assets/84eb85f7-b753-4760-bda7-ecafb533e8c0

PR #6368, opened 2025-05-06
Fix airburst grenades not bouncing at the proper angle, and their projectiles always having the over penetration penalty

Arendian:
## About the PR

Rebound:
- Airburst grenades now rebound at the angle of incidence when they collide instead of always rebounding at 180 degrees.
- Airburst grenades now collide with the same things a normal grenade shot from a launcher would.

Projectile spawning:
- If an airburst projectile detonates on top of an entity, it will now deal the damage of 5 projectiles and remove 5 projectiles from the total projectiles in the grenade. Any leftover projectiles will ignore the directly hit target.
- There is no hit limit of 5 projectiles per target anymore if the grenade didn't directly hit the target.

## Why / Balance

Parity.

When I ported over the airburst grenades, I accidentally always gave them the over penetration behavior instead of only when directly hitting an entity.

## Technical details

The direction of an airburst grenade is now changed every tick using the projectile's velocity.
Instead of adding 180 degrees to the angle when the grenade collides, it now uses the physics system to bounce when colliding in the same way as an entity thrown at a wall would.

The mask is changed to ItemMask instead of Impassable | BulletImpassable and it's fixture is now hard.

When a projectile grenade spawns it's first projectile, it will now search for any entity within 1 tile and remove 5 projectiles from the total projectiles it contains per found entity.  Entities found this way are added to a list that prevents them from being hit by any leftover projectiles from the same projectile grenade.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Airburst grenades now rebound at the angle of incidence instead of always rebounding at 180 degrees.
- fix: Airburst grenade projectiles now receive the overpenetration penalty only when a target is directly impacted by the grenade. For example, with the Hornet grenade, you can potentially hit all 15 projectiles if you aim directly in front of a target, instead of the previous limit of 5.

### DIFF
--- a/Content.Server/Explosion/Components/ProjectileGrenadeComponent.cs
+++ b/Content.Server/Explosion/Components/ProjectileGrenadeComponent.cs
@@ -4,7 +4,7 @@ using Robust.Shared.Prototypes;
 
 namespace Content.Server.Explosion.Components;
 /// <summary>
-/// Grenades that, when triggered, explode into projectiles
+///     Grenades that, when triggered, explode into projectiles
 /// </summary>
 [RegisterComponent, Access(typeof(ProjectileGrenadeSystem), typeof(RMCProjectileGrenadeSystem))]
 public sealed partial class ProjectileGrenadeComponent : Component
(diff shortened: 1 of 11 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #6377, opened 2025-05-08
Fix the Vanguard's pierce ability doing twice as much damage as intended

Arendian:
## About the PR

The Vanguard's pierce ability now doesn't hit the same target twice.
Before this PR the same target could be hit twice, making an unarmored target take 90 damage in one hit instead of 45, and regenerating the shield even if you only hit only one target.

## Why / Balance

Bugfix.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fix the Vanguard's pierce ability dealing twice as much damage as intended

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Pierce/XenoPierceSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Pierce/XenoPierceSystem.cs
@@ -13,6 +13,7 @@ using Robust.Shared.Audio.Systems;
 using Robust.Shared.Map.Components;
 using Robust.Shared.Network;
 using Robust.Shared.Player;
+using Robust.Shared.Timing;
 
 namespace Content.Shared._RMC14.Xenonids.Pierce;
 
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Xenonids/Pierce/XenoPierceSystem.cs
```diff
@@ -41,7 +44,7 @@ public override void Initialize()
 
     private void OnXenoPierceAction(Entity<XenoPierceComponent> xeno, ref XenoPierceActionEvent args)
     {
-        if (args.Handled)
+        if (args.Handled || !_timing.IsFirstTimePredicted)
```
DrSmugleaf, line 47: Is this necessary? I have it fixed on my branch as well and it didn't seem like it

Content.Shared/_RMC14/Xenonids/Pierce/XenoPierceSystem.cs
```diff
@@ -83,6 +87,9 @@ private void OnXenoPierceAction(Entity<XenoPierceComponent> xeno, ref XenoPierce
 
             foreach (var ent in _pierceEnts)
             {
+                if (_hitEnts.Contains(ent))
+                    continue;
+
```
DrSmugleaf, line 92: This can be just `if (!_hitEnts.Add(ent))` just above hits++

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf:
Done in https://github.com/RMC-14/RMC-14/pull/6419

PR #6406, opened 2025-05-10
Fix deafness popups

Dutch-VanDerLinde:
Small fix, makes the popup client only so it doesnt appear to other marines


DrSmugleaf (approved):
Thanks for the PR!

PR #6407, opened 2025-05-10
Fix crate misprediction

Dutch-VanDerLinde:
## About the PR
The ``ResistLocker`` component was serverside only, which was causing the misprediction (thank you @Vermidia for discovering this)
Makes it use a shared component, RequsitionsCrate

:cl:
- fix: Fixed crate drag slowdown mispredicting. (Rubberbanding when you pull them)


DrSmugleaf (approved):
this is epic

PR #6418, opened 2025-05-12
Hijack hibernation + burrowed surge

Vermidia:
## About the PR

Title.

## Why / Balance

parity!

## Technical details

On Hijack Distress signal will now also go through xenos and delete them, ghosting players with a message. All xenos that count for slots get added as 1 burrowed for the hive.

For surge we get all the players alive on the ship that are infectable, add their role weights, and times it by a CVAR (whos value starts at 0.5) minus the amount of xenos. And we get the bigger of this vs the min value of 5 (also in a CVAR).

Once the pooled larva value is gotten a component is added to the hive to generate burrowed while a hivecore is up, reseting surge delay (which decreases as it makes burrowed) if hivecore disappears (how CM handles it because the timers are on the core). If it hits 0 or less it removes itself. There are hive announces for the start and end of the burrowed surge.

Only thing untested is if the role weight stuff works as that code has to get the job proto of each marine and spit out the roleweight.

## Media

https://github.com/user-attachments/assets/f6cd69c2-d64a-4c80-b8be-f1d58f6ef2c5

https://github.com/user-attachments/assets/9eda55d7-896f-4310-b69a-d655a10b747f

## Requirements

**Changelog**

:cl:
- fix: Fixed groundside xenos not becoming burrowed larva once hijack begins.
- add: Ported hijack burrowed larva surge. When hivecore is up, burrowed larva will be generated overtime depending on the initial disparity in numbers between marines and xenos. It generates slightly faster each time a new burrowed is made but speedup resets if hivecore is destroyed.

### DIFF
--- a/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
+++ b/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
@@ -12,9 +12,11 @@ using Content.Server.Chat.Managers;
 using Content.Server.Fax;
 using Content.Server.GameTicking;
 using Content.Server.GameTicking.Rules;
+using Content.Server.Ghost;
 using Content.Server.Ghost.Roles.Components;
 using Content.Server.Mind;
 using Content.Server.Players.PlayTimeTracking;
+using Content.Server.Popups;
 using Content.Server.Power.Components;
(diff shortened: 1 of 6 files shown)


DrSmugleaf (approved):
thank god

PR #6422, opened 2025-05-12
berserker apprehend speed buff fix

Tyzemol:
## About the PR
Apprehend speed buff was supposed to last for the entire duration of the ability and not removed after the attack.

Dont think i need to list it in changelog since its a fix for an update that is not out yet.

## Technical details
Since XenoActiveCripplingStrikeComponent is what keeps track of the speed buff duration i had to add a bool that is turned off after one attack, so the component can stay and apply speed buff for the rest of the ability duration.

## Requirements

**Changelog**
:cl:
- fix: Fixed Berserker Ravager Apprehend bonus movement speed being removed after an attack instead of lasting for the full 5 seconds.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Crippling/XenoActiveCripplingStrikeComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Crippling/XenoActiveCripplingStrikeComponent.cs
@@ -11,6 +11,9 @@ public sealed partial class XenoActiveCripplingStrikeComponent : Component
     public TimeSpan ExpireAt;
 
     [DataField, AutoNetworkedField]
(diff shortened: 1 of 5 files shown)


Dutch-VanDerLinde:
Thanks

Dutch-VanDerLinde (commented):
Add a next slash check on the damage multiply code

If I'm correct, lurkers will do 1.2x damage during the crippling strike duration instead of just the first hit

Dutch-VanDerLinde (approved):

Dutch-VanDerLinde:
It's good I misread the code

DrSmugleaf (approved):

PR #6431, opened 2025-05-12
Fix throwing knives killing the user

Dutch-VanDerLinde:
## About the PR
removed this so disarming wouldnt have the throw effects
but didnt know the Ctrl+Q throws use the same system

:cl:
- fix: Fixed throwing knives damaging the user


DrSmugleaf (approved):
bless

PR #6497, opened 2025-05-20
Fix praetorian's acid ball ignoring the hit limit

Arendian:
## About the PR

Being hit by a projectile spawned by the acid ball now counts as being hit by the acid ball instead of each projectile individually.

## Requirements

**Changelog**

:cl: Dygon
- fix: The Praetorian's acid ball can once again only hit a maximum of four projectiles on the same target.


DummeHytteOst:
this bug is also being fixed by #6486, looks like #6486 is also adding a bit of extra stuff in CMClusterGrenadeSystem.cs

DrSmugleaf (approved):
Looks good my other changes were to make sure that what the acid ball hits doesn't get counted as a max hit towards acid ball

PR #6510, opened 2025-05-21
Survivor fixes

Dutch-VanDerLinde:
## About the PR
FORECON SL spawns with slug box instead of a stack in CM13
image

Removes duplicate meta.json entry
Allows primary weapons to spawn inhand, like CM13
Currently if you have no suit storage or its full, the gun will not spawn

:cl:
- fix: Fixed civilian survivors not spawning with primary weapons.


DrSmugleaf (approved):
bless

PR #6529, opened 2025-05-22
reword forecon commander

QuietlyWhisper:
Changes Forecon CO to be a Forecon Commander/Major

Reduces a lot of bloat in marine SOP / CO guidelines by not considering them a "Commanding Officer" by title or permissions.

"High ranking officers" (Major+) have been given their own permissions and SOP in place of this, which only needs to dictate the additional rights they receive rather than also what they do not.

This was done so I don't need to write what they're not allowed to do while not being the commanding officer of the ship but being a commanding officer. If that sounds confusing, it will make sense why I added this.

*This is still a whitelisted role only accessible by CO whitelisted players.

image

### DIFF
--- a/Resources/Locale/en-US/_RMC14/job/command.ftl
+++ b/Resources/Locale/en-US/_RMC14/job/command.ftl
@@ -21,8 +21,7 @@ CMJobExecutiveOfficer = Executive Officer
 
 # Special forces
 rmc-job-prefix-forecon = FORECON
-rmc-job-prefix-forecon-co = FORECON CO
-
+rmc-job-prefix-forecon-co = FORECON Maj.
 # Admin only right now.
 cm-job-name-high-command = High Command
 
--- a/Resources/Locale/en-US/_RMC14/job/command.ftl
+++ b/Resources/Locale/en-US/_RMC14/job/command.ftl
@@ -21,7 +21,8 @@ CMJobExecutiveOfficer = Executive Officer
 
 # Special forces
 rmc-job-prefix-forecon = FORECON
-rmc-job-prefix-forecon-co = FORECON Maj.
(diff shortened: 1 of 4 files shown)


Resources/Locale/en-US/_RMC14/job/survivor.ftl
```diff
@@ -30,8 +30,8 @@ rmc-job-greeting-forecon = You are UNMC Force Reconnaissance (FORECON) marine fr
 
   You are [bold][color=#51A16C][font size=16]NON-HOSTILE to the UNMC![/font][/color][/bold]
 
-rmc-job-name-survivor-co = Survivor CO
-rmc-job-description-survivor-co = You are a commanding officer of the UNMC.
+rmc-job-name-survivor-co = FORECON Commander
```
Dutch-VanDerLinde, line 33: This role and description is the role in the job selection. This job never appears ingame, and is just a preference for the CO selection. In the future, we'll have the New Varadero CO and CLF commander

Resources/Locale/en-US/_RMC14/job/command.ftl
```diff
@@ -21,8 +21,7 @@ CMJobExecutiveOfficer = Executive Officer
 
 # Special forces
 rmc-job-prefix-forecon = FORECON
-rmc-job-prefix-forecon-co = FORECON CO
-
+rmc-job-prefix-forecon-co = FORECON Maj.
```
Dutch-VanDerLinde, line 24: So FORECON Maj. Maj. ?
Seems a little weird

Dutch-VanDerLinde: Never mind their name is still commander

But how about changing it to FORECON COM. OR FORECON CDR?

PR #6537, opened 2025-05-22
Add a spawn as job menu name for job prototypes, fix FORECON names

Dutch-VanDerLinde:
## About the PR
Some job names share the same name with others, it'll add a duplicate entry in the spawn as job menu
Adds a datafield to job prototype which allows it to have a custom spawn menu name

image

:cl:
ADMIN:
- add: Added different job spawn menu names for FORECON and WeYA PMC operators.


DrSmugleaf (approved):
thank fuck
need to fix playtime tracking in the future too but that's for later

PR #6540, opened 2025-05-23
Add unconscious status

### LINKED ISSUE #3941: Demo Spec AP Rocket needs to blind + mute
## Description

It's a paralyze effect like para infects and symptoms, so it needs to do both of these with the stun. They do effect even fortified defenders who don't get the stun.

(I did check in cm to make sure)

**Reproduction**

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Title.

## Why / Balance

parity! A lot of things already attempted to add it by adding blindness/deafness/etc seperately, but now I've seperated it so it can be tracked by itself.

Fixes #3941

## Technical details

Added TryKnockOut to the sizeStun system, and also changed some stuff in it to support knockouts (the one for flashbang stuff). had to modify a line in the status effects system otherwise stun would never have a linked component if it was added after knockout (leading to perma stun from batons).

The stun resistance mods now check daze & knockout (parity) and to make sure stuff isn't applied twice, KnockOut adds the relevant components directly, and readds them when statuses go away just incase, and doesn't remove them if a status would keep ticking on when removed, to also prevent that breaking. Shouldn't be issues since it's the only one that does this, hopefully.

I tested a few cases (by itself, with a longer stun duration vs knockout duration, with a shorter stun duration vs knockout duration) to make sure it works generally.

## Media

https://github.com/user-attachments/assets/149837eb-1511-4582-b2c5-156eb1cb2f65

https://github.com/user-attachments/assets/b3826587-3e41-42b4-8850-b55f972d48ff

## Requirements

**Changelog**

:cl:
- fix: Fixed unconscious not being it's own status effect. Various causes of it now correctly treat it as such, and it can be shaken out of.
- fix: Fix AP M5-ATL Rockets not causing unconsciousness


RangerXVII:
Will this let vampire lurkers bite your head off? In the ability description it says that you can bite the head off of unconscious people.

Issue #4547, opened 2024-10-20
Canister Parity


Vermidia:
## Description

Canisters can only be refilled by:
Chemical Dispensers
Reagent Tanks

Yeah...that's it

**Reproduction**

**Screenshots**

**Additional context**

Dutch-VanDerLinde:
Fixed?

AverageNotDoingAnythingEnjoyer:
@Vermidia Verm, close the issue, Verm

Dutch-VanDerLinde:
I'm verm

PR #6544, opened 2025-05-23
Injector + Reagent Canister Tweaks

### LINKED ISSUE #5681: Canister pouch visuals
## Description

The color of the reagent container is supposed to show through while in the pouch. Also shows whether an injector is in it or not.

**Reproduction**

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Title.
Custom normal injectors from CM are now in.
Reagent Canisters properly can only be used in dispensers and with reagent tanks, and can't be filled/emptied in other ways.
Custom injectors no longer have the refillable component, it's been replaced with a SoltuionStorageFillable comp.
Ported the gas tank inhands.
Removed the extra tricord injector sprite, not sure if it was correct or not before but it is not if it wasn't.
Visuals for having an injector + tank (colored).

Non-parity:
Custom Injector's names have been made more consistent (same across diff max volumes for all injectors)
Custom Injectors now have a stripe to indicate size (along with the name still)
Fill and inhand sprites for pressurized canisters.

## Why / Balance

Parity, bugs, and QoL. Custom Injector names for EZs were all over the place while Medic & normal Custom were consistent with eachother.

Fixes #5681, #4547

## Technical details

2 new comps: 1 for rmcpressurized solution (used for handling reagent tanks) and 1 for storagefillable (so reagent pouches + fuel tank backpack work ok). Dispenser already checked for mixable so it's fine, the tank was to make sure that worked.

YML for injectors was moved around a bit, so custom ones won't have the refill at a vend text (they don't start with any to refill) etc.

## Media

containers
injectors
lathe

## Requirements

**Changelog**

:cl:
- fix: Fixed pressurized canisters allowing transfer from non-reagent dispensers and non-reagent tanks
- fix: Fixed custom autoinjectors being refillable from bottles
- add: Added custom autoinjectors that work like the default autoinjectors (slow without skill but don't need skill), printable from the medilathe
- tweak: Changed custom autoinjector names to be consistent
- add: Added colored stripes to autoinjectors based on volume, pressurized canisters now have different fill sprites.
- add: Canister Pouches now show when they have a tank (and what color reagent is in it) and when they have an injector inserted.
- add: You can now flush canisters in reagent container pouches.


Content.Client/_RMC14/Medical/StorageRefillContainerVisualizerSystem.cs
```diff
@@ -0,0 +1,38 @@
...
+
+        if (!AppearanceSystem.TryGetData<Color>(uid, SolutionContainerStoreVisuals.Color, out var color))
+            return;
+
+        if (!sprite.LayerMapTryGet(SolutionContainerStoreVisuals.Base, out var colorLayer))
+            return;
+
+        if (!_container.TryGetContainer(uid, component.ContainerId, out var container) ||
+    !container.ContainedEntities.TryFirstOrNull(out var contained) || (!_solution.TryGetDrainableSolution(contained.Value, out var drainable, out var sol) &&
+    !_refillable.TryGetPressurizedSolution(contained.Value, out drainable, out sol)) || sol.Volume == 0)
```
DrSmugleaf: This makes it slightly clearer for which condition is grouped, on its own line
```suggestion
        if (!_container.TryGetContainer(uid, component.ContainerId, out var container) ||
            !container.ContainedEntities.TryFirstOrNull(out var contained) ||
            (!_solution.TryGetDrainableSolution(contained.Value, out var drainable, out var sol) && !_refillable.TryGetPressurizedSolution(contained.Value, out drainable, out sol)) ||
            sol.Volume == 0)
```

DrSmugleaf (approved):
Thanks for the PR!
The autoinjector stripes are very nice as well

PR #6546, opened 2025-05-23
Spec bug fix

### LINKED ISSUE #6531: Specialists roles show improperly on comms.
## Description

title
**Reproduction**

pick WS, select spec, talk in chat
**Screenshots**

Image

Image

Image

Image

Collected the holy rombus of pictures
**Additional context**

RangerXVII:
## About the PR
The original WS icon is now the scout icon, and a WS that hasn't picked a kit has no icon. I also got rid of the _super annoying_ space that was in front of the spec prefixes. (fixes #6531)

## Why / Balance
It was kind of weird that the scout didn't have its own icon, which, it doesn't really, but now it doesn't belong to the base WS.

## Technical details
Changed some pictures and deleted a space that didn't look like it should be there.

## Media
No kit:
Content Client_Zadv05aFlV
Scout kit selected:
Content Client_MfWkXLCsXs
Yay it's fixed!
Content Client_IVARMYPt4g

## Requirements

**Changelog**

:cl:
- tweak: Scout weapon specialist now has it's own icon.
- tweak: Weapon specialists that have not chosen a kit now have no icon.
- fix: Fixed a random space being in front of the weapon specialist prefixes.


chavonadelal:
if you want the role to be without icons the best solution would be to not give an icon at all instead of giving it but giving it empty. looks a bit dirty

chavonadelal (changes requested):
if you want the role to be without icons the best solution would be to not give an icon at all instead of giving it but giving it empty. looks a bit dirty

RangerXVII:
If i just deleted the icon files, would that work?

chavonadelal:
> If i just deleted the icon files, would that work?

No. You need to search for a place where the sprite that you no longer want to issue as a roundstart is mentioned and remove its mention from there. Most likely, this is a prototype of the work

RangerXVII:
Yeah I deleted it and...
Content Client_2cQ8k6dBIE

RangerXVII:
What should i do about the tacmap icon?

Dutch-VanDerLinde:
> What should i do about the tacmap icon?

it doesnt matter, why would a WS be deploying without their kit

RangerXVII:
Um.... very severe stupidity? I'll just do a rifleman one, because they are just a rifleman without a kit.

chavonadelal:
> What should i do about the tacmap icon?

you also need to remove her mention from the prototype work

chavonadelal:
although... perhaps it's worth not touching, since you have nothing to offer in return if, by some stupidity, a specialist without a whale ends up on the planet. you simply won't be able to find him on the map

RangerXVII:
I think that's good, with an empty picture you get just the background square, which is perfect.

Content.Server/Radio/EntitySystems/RadioSystem.cs
```diff
@@ -104,7 +104,7 @@ public void SendRadioMessage(EntityUid messageSource, string message, RadioChann
 
         if (TryComp(messageSource, out JobPrefixComponent? prefix))
         {
-            var prefixText = (prefix.AdditionalPrefix != null ? $" {Loc.GetString(prefix.AdditionalPrefix.Value)}" : "") + Loc.GetString(prefix.Prefix);
+            var prefixText = (prefix.AdditionalPrefix != null ? $"{Loc.GetString(prefix.AdditionalPrefix.Value)}" : "") + Loc.GetString(prefix.Prefix);
```
DrSmugleaf: The idea for this was for the space to be in the middle but it was bugged, as in (SCT Spc)

Resources/Prototypes/_RMC14/Roles/Jobs/Marines/weapons_specialist.yml
```diff
@@ -20,7 +20,7 @@
     RMCRankLanceCorporal: []
   startingGear: CMGearWeaponsSpecialist
   dummyStartingGear: CMGearWeaponsSpecialistEquipped
-  icon: "CMJobIconWeaponsSpecialist"
+  icon: "CMJobIconEmpty"
```
DrSmugleaf: This will make it also have no icon on the role selection menu, which is not intended

DrSmugleaf (changes requested):
Thanks for the PR!

Resources/Prototypes/_RMC14/Roles/Jobs/Marines/weapons_specialist.yml
RangerXVII: Well what should it be then? I can't think of an icon for all the specs, should I just keep it as the scout icon?

Content.Server/Radio/EntitySystems/RadioSystem.cs
RangerXVII: fixed, it works now

Resources/Prototypes/_RMC14/Roles/Jobs/Marines/weapons_specialist.yml
DrSmugleaf: Yeah that's fine for now, though it would probably be easier to leave out the icon change and have the prefix change until someone sprites an icon for the scout

RangerXVII:
Okay it's just a bugfix now lmao

RangerXVII:
Well sorry it took so long.... such a simple fix took a couple weeks

DrSmugleaf (approved):

PR #6550, opened 2025-05-24
Make corporate survivor rank up from the sum of both corpo surv and corporate liaison.

Dutch-VanDerLinde:
## About the PR
Getting the times for corpo surv is almost impossible due to how surv functions, you might not live, so having the CL give the ranks was reasonable. The problem with this is that CL is a different role than corpo surv, some players might not want to play it if they want a higher rank in their role.

This PR makes it use both CL and corpo surv time, the sum of it.

## Requirements

:cl:
- tweak: Corporate survivor now ranks up from sum of time in liaison and time in corporate survivor.


DrSmugleaf (approved):

QuietlyWhisper:
I wouldnt have merged this personally, RIP anyone who actually played the role for the medal ranks

We have these for account progression, they don't mean anything- the whole point is to play the role to obtain them

PR #6554, opened 2025-05-24
Make shutters and bolted doors block larvae and parasites

Arendian:
## About the PR

I changed the shutter's collision layer from 'airlock' to 'wall', preventing larvae from escaping during lockdowns in the hangar and surgery rooms, or containment areas in research and the CL's office.

When a door is bolted while closed it now changes it's fixture to prevent small entities from passing through it.

## Why / Balance

Parity, in CM13 larva and parasites can't go under shutters or bolted doors.

## Media

CM13 behavior:

https://github.com/user-attachments/assets/0612b1e1-55b7-4c4a-a86f-09487668c0a9

## Technical details

I don't think a new component that stores the exact same values will be needed, I also didn't want to cause issues by changing the name of an upstream component. This is why I opted to use the LayerChangedOnWeldComponent for the bolting state of doors instead of making a new one or renaming the existing one. Renaming it to something like LayerChangeOnWeldOrBoltComponent is also an option, since everything that should change layer on weld should also do it on bolt if it has one.

## Requirements

**Changelog**

:cl: Dygon
- fix: Closed shutters and bolted doors now block larvae and parasites.


DrSmugleaf (commented):
Looks fine though probably need vent crawling before we merge this

DrSmugleaf (dismissed):

PR #6559, opened 2025-05-25
Mappable Tents

SirChopwood:
## About the PR
Added static tents that can be mapped for now. They do not have any special functionality or the ability to be moved right now.

## Why / Balance
For mapping. Will add function later.

## Technical details
Needed to patch the camo system to allow multiple layer mappings, rn set it to allow 5. (Camo code is already bad so didnt make it much worse, is on the TODO list to fix.)

## Media
image
image

## Requirements

**Changelog**
:cl: Ramiris
- add: Added tents!
- tweak: Allowed camo to be applied to 5 layers at once.


Carolyn3114:
is there not a way to make the roof fully transparent when you're under it? I'm sure it'd be less of an eyestrain to use

SirChopwood:
> is there not a way to make the roof fully transparent when you're under it? I'm sure it'd be less of an eyestrain to use 

Its using the existing upstream sprite fade, same as things like Hybrisa gantries and various trees. So simple answer, at least for now, is no. 

Personally i found it rather soulful anyway to have it just fade the roof when you're inside/hovering on it. (The walls still obscure you a bit but made sure you cant full hide behind them.)

DrSmugleaf (approved):
Thanks for the PR!

PR #6564, opened 2025-05-25
Made the M4 free for SL

UnicornOnLSD:
## About the PR

NOT PARITY !!! But hear me out.
Transferred the M4 in the SL vend from point buyable (16 points) to free, and CHOOSE between that and M4 and B12

## Why / Balance

It hardly does.
Stat wise B12 and M4 compensate eachother.
Defense wise, M4 is worse than B12 except for acid. However, M4 got one more storage slot.

Thus, both of these armors being on par, both should be of the same price, in this case free.

## Technical details

yml, nothing too hard, check code change, very small.

## Media

fergedghe
efeefef

## Requirements

**Changelog**

:cl:
- add: SLs can now choose between B12 and M4 for free (instead of having to buy the M4)

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/leader.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/leader.yml
@@ -31,11 +31,6 @@
       - id: RMCMOU53Case
       - id: RMCCaseXM88
       - id: RMCKitEngineering
-    - name: Armors
-      entries:
-      - id: RMCArmorM4Zipties
-        name: leader M4 pattern marine armor (zipties)
-        points: 16
     - name: Clothing Items
       entries:
       - id: RMCBackpackRTO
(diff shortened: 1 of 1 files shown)


UnicornOnLSD:
After checking with community's general consensus, I could mark the B12 as recommended for player's sake.

UnicornOnLSD:
After discussing further with different members of the community. It was agreed that M4 is generally speaking a downgrade compared to the B12. As such I made the B12 recommended.

UnicornOnLSD:
image
(it works)

UnicornOnLSD:
fair enough.

UnicornOnLSD:
Thought about it a little and figured.

>a new player specially is way more likely to default to the B12 if it is the only option at the top, which is what we want

New players often see the purchasable M4 (after picking B12) in the vend and think it's an upgrade (because you have to PAY for it), buying it, making M4 a possibility while B12 is recommended is better. 
Other experienced players usually buy it for the drip and that's their choice, better for both sides.

UnicornOnLSD:
Here is a recent proof of my point

image


Brand new SL who geared up and likely saw the M4 being point buyable and fell into the trap of "it costs points so it must be an upgrade"

DrSmugleaf (approved):

PR #6568, opened 2025-05-26
XvX improvements

Vermidia:
## About the PR

Converts as many abiltities as possible to work better Xeno vs Xenos (going over the cm code to make sure the cases are correct per ability)
Also fixes weeds of not your hive healing you and a few other things.

## Why / Balance

Parity.

## Technical details

So the main bulky is now theres a HiveVsHive part of XenoSystem, that stores the const modifiers for the xeno stuff, and applies the effects if the target matches the conditions of it (like CM). A few gas functions also had to be modified to set hive as they weren't before. Also xenos now do 1.5x damage to other xenos, as is parity.

Also CanAbilityAttackTarget now returns true if a xeno or marine comp was detected, abilities only working on marines has their own is xeno comp in there afterward.

A few other changes were done:
Charge was reworked to not rely on the hit non-marines part of canabilityattacktarget (which that part has been removed entirely and substituted with better checks) and now properly instant destroys stuff like doors, tables, etc. There's a comp for that.

Theres now a onFriendlyWeeds func so hive is also checked.

Stomp had to be messed with a bit since some of the things used the xeno mults and some didn't (CM inconsistency)

Brutalize's damage was moved out of being based on melee mult - in CM it's flatly based on melee damage upper.

Screech now adds duration, instead of refreshing (CM parity)

Anything supposed to daze should now daze, for the most part.

Spray acid no longer refreshes stun if you're downed. There is currently a bug where if a xeno would predict going crit/dead it get stucks in the downed state, not sure how to fix that but it seems to only be on xenos.

The Gas DOT isn't 100% parity yet, xenos instead of taking the full armored damage only take 5 damage instead, / 2 if the caste doesn't take fire damage for some reason. Thats not in this PR, only the inhalation (AP) part doing extra damage.

I think thats everything. Note I didn't fix pheros working or hivemind chat being global for all xenos, as that requires more indepth reworking.

## Media

https://github.com/user-attachments/assets/d7ffbe27-af4c-4618-839a-2d9ef8198331

## Requirements

**Changelog**

:cl:
- fix: Fixed base crusher's charge not deleting or doing enough damage against some objects.
- fix: Fixed screech not stacking with existing stuns.
- fix: Fixed ground acid stunning if you were downed.


Kadeo64:
👀👀👀👀👀👀👀👀

DrSmugleaf (dismissed):

PR #6569, opened 2025-05-26
Armylathe and prototypes for casings.

AlexisMoonlight:
## About the PR
Add armylathe, casings made in armylathe.

## Why / Balance
Parity, needed for OT department

## Technical details
All sprites and values taken from CM13. At the moment, casings serve no purpose.

## Media

Armylathe

## Requirements

**Changelog**
:cl:
- add: Added armylathe and casings for a future Ordnance Technician.


Resources/Prototypes/_RMC14/Entities/Structures/Machines/Lathe/armylathe.yml
```diff
@@ -0,0 +1,110 @@
+- type: entity
+  parent: [ BaseMachinePowered, RMCConstructibleMachine ]
+  id: CMArmylathe
```
DrSmugleaf: You should be parenting this from CMAutolathe, a lot of the components are the same or similar, which lets you remove any common lines of YAML

DrSmugleaf (changes requested):
Thanks for the PR!

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Lathe/armylathe.yml
AlexisMoonlight: Sure, that is an easy change. I do want to point out that medical lathe does not follow this convention, which is why I did it that way in the first place. I guess medilathe has more differences in the yml, though not by much.

DrSmugleaf (approved):

PR #6575, opened 2025-05-27
Adding a max players limit to planetary maps

varttist:
## About the PR
Added a limit on the maximum number of players at which a planetmap can be selected as a candidate for voting

## Why / Balance
Not having a maximum online limit on maps is bad

## Requirements


DrSmugleaf (approved):
Thanks for the PR!
Don't need a changelog since it does not actually add a max player count to any of our maps, otherwise players may get confused

PR #6596, opened 2025-05-30
Dropship & fire stuffs

Vixting:
## About the PR

fire collision polling replaced with events 
dropship uses events instead of timer
exclusion zone of fire under dropship 

## Why / Balance

better fire perf
fix  dropship with low timer causing explosion to happen to dropship
fixes set on fire when moving at point of crash

## Technical details

fire changes
dropship changes 
i think GetShipExclusionZones can be done better 
should work fine on actual ship but cant test
some code formatting

## Requirements

**Changelog**
fix: fixes set on fire when moving at point of crash

### DIFF
--- a/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
+++ b/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
@@ -484,7 +484,7 @@ public abstract class SharedRMCFlammableSystem : EntitySystem
             if (!spawned.Add(target)) continue;
             var nextRange = SpawnFire(target, spawn, chain, range, intensity, duration, out var cont, exclusionZones);
             if (nextRange == 0 || cont) continue;
-            Timer.Spawn(TimeSpan.FromMilliseconds(10), () =>
+            Timer.Spawn(TimeSpan.FromMilliseconds(50), () =>
             {
                 try { SpawnFires(spawn, target, nextRange, chain, intensity, duration, spawned, exclusionZones); }
                 catch (Exception e) { Log.Error($"Error occurred spawning fires:\n{e}"); }
@@ -552,10 +552,13 @@ public abstract class SharedRMCFlammableSystem : EntitySystem
         return false;
     }
 
+    /// <summary>
+    ///     Spawns fire in a cone shape in the direction the entity is facing.
+    /// </summary>
     private void SpawnFireCone(Entity<DirectionalTileFireOnTriggerComponent> ent, EntityCoordinates center, int? intensity = null, int? duration = null)
     {
-        var chain = _onCollide.SpawnChain();
         if (_net.IsClient) return;
+        var chain = _onCollide.SpawnChain();
         ent.Comp.DiagonalRange = (int)Math.Floor((double)ent.Comp.Range / 2);
         Dirty(ent);
         var initialShot = !ent.Comp.InitialSpread;
--- a/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
+++ b/Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
@@ -30,7 +30,9 @@ using Content.Shared.Tag;
 using Content.Shared.Whitelist;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
+using Robust.Shared.GameObjects;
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
```diff
@@ -129,10 +147,9 @@ private void OnTileFireMapInit(Entity<TileFireComponent> ent, ref MapInitEvent a
 
     private void OnTileFireVaporHit(Entity<TileFireComponent> ent, ref VaporHitEvent args)
     {
-        if (_net.IsClient)
-            return;
-
+        if (_net.IsClient) return;
```
DrSmugleaf: You need to get your editor to parse the .editorconfig at the project's root, these syntax changes are unnecessary and also against the style guide that's in the repo

DrSmugleaf (changes requested):
For fire igniting you on hijack landing, it should be enough to check if the fire is terminating when igniting, but if not this is fine as a fix too

Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
Vixting: i will update it & test termination instead

Vixting: I couldn't get the termination check to work (didnt prevent fire igniting you) but idk if i did something wrong.

fixed the styling issues though (i think)

Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
```diff
@@ -339,6 +342,171 @@ private void OnFlammableIgnite(Entity<FlammableComponent> ent, ref RMCIgniteEven
...
+    }
+
+    private void OnIgniteDamageCollide(Entity<RMCIgniteOnCollideComponent> ent, ref DamageCollideEvent args)
+    {
+        Ignite(args.Target, ent.Comp.Intensity, ent.Comp.Duration, ent.Comp.MaxStacks);
+    }
+
+    private void OnIgniteComponentInit(Entity<RMCIgniteOnCollideComponent> ent, ref ComponentInit args)
+    {
+        Timer.Spawn(TimeSpan.FromMilliseconds(50), () => CheckForExistingOverlaps(ent));
```
DrSmugleaf, line 384: Is the timer here necessary? They are typically bad form, since they don't serialize properly and depending on test execution might make them break as well
If this is to wait until component data is valid, you might want to try using MapInitEvent instead

Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
```diff
@@ -339,6 +342,171 @@ private void OnFlammableIgnite(Entity<FlammableComponent> ent, ref RMCIgniteEven
...
+            }
+        }
+    }
+
+    private void TrackFireContact(EntityUid fireSource, EntityUid target)
+    {
+        if (!_fireContacts.TryGetValue(fireSource, out var contacts))
+        {
+            contacts = new HashSet<EntityUid>();
+            _fireContacts[fireSource] = contacts;
```
DrSmugleaf, line 446: This set should be stored on a component on the fireSource entity, static data for entities in entity systems is usually a bad idea, specially since it's not getting cleared on round restart cleanup event in this
Same with _extinguishContacts

Content.Shared/_RMC14/Atmos/SharedRMCFlammableSystem.cs
```diff
@@ -339,6 +342,171 @@ private void OnFlammableIgnite(Entity<FlammableComponent> ent, ref RMCIgniteEven
...
+        {
+            contacts.Remove(target);
+            if (contacts.Count == 0)
+                _extinguishContacts.Remove(extinguishSource);
+        }
+    }
+
+    private bool IsEntityTouchingFire(EntityUid entity)
+    {
+        foreach (var contacts in _fireContacts.Values)
```
DrSmugleaf, line 486: This should also just be getting tracked on the entity, this would be horrible for performance checking every single contact

DrSmugleaf (changes requested):

Vixting:
I will rework this in the future maybe, many things can be done better

PR #6614, opened 2025-06-01
Screen Announces

NotInButIn:
## About the PR
This PR adds messages(example - overwatch transmit) at the top of the screen like in CMSS13 also it adds a mission start message when the dropship first flies out and plays music if it has it

## Why / Balance
Parity? i think

## Technical details
N/A

## Media
Later

## Requirements

**Changelog**
:cl: NotInButIn


NotInButIn:
So for now I don't have much time because I'll continue a little later on 
Most of it's already done.

PursuitInAshes:
Thanks for the PR!
We're going to go with #9426, which I believe might have used this as a base. Thank you for your work

PR #6629, opened 2025-06-05
 Fixed not being able to give gibbed marines medals

chavonadelal:
## Why / Balance

Heroes should receive the medals they deserve!
## Technical details

Added logic for recording the necessary data at the moment of a marine's gibbed, with the help of which we use the modified logic that supports the issuance of a medal only by LastPlayerId and the recorded name.

I have no way to test this in a guaranteed way but it should work.
## Requirements

**Changelog**

:cl: to4no_fix
- fix: Fixed not being able to give gibbed marines medals.


Content.Server/_RMC14/Commendations/CommendationSystem.cs
```diff
@@ -68,4 +68,54 @@ public override async void GiveCommendation(
...
+            catch (Exception e)
+            {
+                Log.Error($"Error saving commendation to database, giver: {giverName}, receiver: {receiverName}, round: {round}:\n{e}");
+            }
+        }
+        catch (Exception e)
+        {
+            Log.Error($"Error giving commendation by last player id, giver: {giver.Owner}, lastPlayerId: {lastPlayerId}");
+        }
+    }
```
DrSmugleaf: Don't copy and paste the entire method, make it take in the receiver's name and user id

Content.Server/_RMC14/Marines/ControlComputer/MarineControlComputerSystem.cs
```diff
@@ -1,5 +1,43 @@
...
-public sealed class MarineControlComputerSystem : SharedMarineControlComputerSystem;
+public sealed class MarineControlComputerSystem : SharedMarineControlComputerSystem
+{
+    public override void Initialize()
+    {
+        base.Initialize();
+        SubscribeLocalEvent<MarineComponent, BeingGibbedEvent>(OnMarineGibbed);
+    }
+
+    private void OnMarineGibbed(EntityUid uid, MarineComponent component, ref BeingGibbedEvent ev)
```
DrSmugleaf: This should be using a single Entity<MarineComponent> argument instead of the EntityUid and a component separately

DrSmugleaf (changes requested):

chavonadelal:
Added localization support for xeno jelly

PR #6640, opened 2025-06-07
Resin traps are easier to see and hit, and don't reapply stuns on knocked down entities.

Arendian:
## About the PR

Acid resin traps now don't reapply a stun when you are dragged through one. This means that you are stunned for the intended 4 seconds by one trap, instead of 7 seconds (you get a 4 second stun reapplied right when the 3 second acid trap duration expires).
The resin trap sprites now have weeds as part of their sprite, this makes their sprite the size of an entire tile. The bigger sprite means that you can shoot the trap by just aiming at the tile it is on, even when the brightly colored part is hidden beneath another object.. Traps are now also easier to spot when hidden beneath objects because even when a nearby weed node is destroyed, the tile with the trap will still look weeded.

Acid resin traps no longer reapply a stun when dragged through, ensuring you're only stunned for the intended 4 seconds instead of 7 (previously, a 4-second stun would be reapplied right as the 3-second acid trap effect expired).

Additionally, resin trap sprites now include weeds as part of their design, making them cover the full tile size. The larger sprite allows you to shoot the trap by simply aiming at the tile, even if the filled part is hidden under another object. Traps are also easier to spot when hidden beneath objects, as the tile will still appear weeded, even if a nearby weed node is destroyed.

All acid, sprayed or from traps, has their hitbox reduced to be the same as a fire tile to prevent only a tiny part of your sprite clipping into the tile making you get hit. Without this change traps hit people outside their trigger range.

## Why / Balance

Parity.

## Media

https://github.com/user-attachments/assets/02f043eb-b050-44de-a390-4511ee7b5e05

## Requirements

**Changelog**

:cl: Dygon
- tweak: Resin trap sprites now have weeds as part of their sprite, making them more visible and easier to target.
- tweak: Xeno acid effects now have their hitbox reduced to match the size of fire tiles. meaning you'll only get hit if a larger portion of your sprite clips into the acid.
- fix: Acid traps no longer apply or reapply their paralyze effect when you are already downed.


Vermidia:
I already fixed the stun, they no longer stun you if you're downed, like in CM13. (It might be in the XvX PR)

Also instead of changing the sprites you can just add weeds as a layer like done with hive core/other structures.

Arendian:
> I already fixed the stun, they no longer stun you if you're downed, like in CM13. (It might be in the XvX PR)

We'll only have a single conflicting line, if you don't mind that I'll just keep it in here in case your PR takes a bit longer to review as it is a lot bigger.

> Also instead of changing the sprites you can just add weeds as a layer like done with hive core/other structures.

Thanks a lot for this, that is a way better way to do it.


Test fail is unrelated.

DrSmugleaf (approved):

DrSmugleaf:
In 13 I believe it takes the weeds texture from the weeds underneath when planted instead, but this works for now

PR #6647, opened 2025-06-07
Replaces magazines in m1984 weapon cases with AP rounds (MP sidearm change)

TeaTotalTemplar:
Updates the magazines provided in the M1984 Weapon Case to AP, to match ammo provided in the other weapon cases

## About the PR

Added RMCWeaponMagazineM1984AP to the M1984 Weapon Case whitelist, and replaces the 6 standard magazines with AP instead
## Why / Balance

Of the three sidearm choices available to MPs, the M1984 is the only one that does not come with AP magazines in reserve. This leaves it lacking against armored targets (such as bad-actor marines and post-hijack xenos) whereas the M77 and M44 will deal their full damage. 

This change replaces the reserve ammunition in the 1984 case with their AP variant, which deals 25 damage, on par with the M77 (before accounting for weapon modifiers). The 1984 otherwise has a higher rate of fire than the M77, but a lower total capacity.
I believe this change will make the 1984 a more attractive prospect for some MPs whilst also preventing the common occurance of the QM's supply of AP magazines being gobbled up by those MPs.

According to testing conducted by the MP enjoyers of MarineCord, this brings the TTC (time to crit) on an EOD-clad marine as follows:
M1984 AP - 7 shots
M77 - 7 shots
M44 Marksman - 4 shots 

Concerns of balance are that the higher firerate of the 1984 (and its affinity with the burstfire assembly) might make the 1984 TOO strong, but i feel the lower magazine size of the firearm (and limited supply of AP 9mm) negates this somewhat.

## Technical details

Changes whitelist of M1984 guncase to allow AP magazines and replaces the 6 standard mags with AP
## Requirements

**Changelog**

:cl:
- tweak: Replaces the standard magazines in the MPs 1984 Weapon Case with their AP variant. No more raiding the QM's stash.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Cases/gun_cases.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Cases/gun_cases.yml
@@ -130,13 +130,14 @@
     whitelist:
       tags:
       - CMWeaponPistolM1984
+      - RMCMagazinePistolM1984AP
       - CMMagazinePistolM1984
       - RMCAttachmentRailFlashlight
   - type: StorageFill
     contents:
     - id: RMCAttachmentRailFlashlight
     - id: CMWeaponPistolM1984
-    - id: CMMagazinePistolM1984
+    - id: RMCMagazinePistolM1984AP
       amount: 6
 
 # RSh9 Assault Revolver


DrSmugleaf (changes requested):
It being 7 shots like another gun does not matter when the M1984 is by far the fastest firing one of the group, where you should be measuring actual time, not shots.

https://github.com/user-attachments/assets/8155d96e-0d5b-4a85-ab7d-59d765316d57

This feels like a very hastily made PR for multiple reasons:
One is we don't even have HP magazines added, which would fit MP a lot better.
Two is we don't have limb targeting yet, where you could shoot someone somewhere other than the torso, and the armor values vary per limb.
The actual damage values for a marine with 30 armor (uniform + medium armor to the chest), firerate aside, is:
Regular M1984: 40 damage, 10 AP, 22 final damage.
AP M1984: 25 damage, 40 AP, 25 final damage.
HP M1984: 55 damage, 0 AP, 23 final damage.
Use this graph to test values https://www.desmos.com/calculator/fcmhz8gyd6, X is armor, Y is final damage, D on the left is initial damage.

And this is best case scenario, where you are shooting someone on the chest where they are most armored (which right now is anywhere in the body) with a medium armor value. The case where it becomes better as you said is at higher armor values (B12 is 25 bullet and EOD is 35 bullet, plus 10), where the non-AP variants both do around 15 damage to 40 armor.

As for it matching other weapon cases, the reason the M1984 doesn't get special ammo in the case is because it has an insane fire rate compared to the weapons in those other cases. You are making M1984 AP insanely more common with this change, and there would be nothing stopping an MP from handing them out as its gear found in squad vendors.

If you want to do this, add HP magazines (hollowpoint) and then we can consider what ratio of each (Reg/AP/HP) to give them on the case.

Issue #5730, opened 2025-02-25
CPR Doafter Times


Vermidia:
## Description

In CM it's 4 seconds no skill, x0.75 if you have medical 2+, x0.35 instead if you have master medical skill (4)

**Reproduction**

**Screenshots**

**Additional context**

Utmanarn:
It is first at skill level 3 that you get the x0.75 speed boost so it does not effect the combat medics.

Vermidia:
Fixed by #6656

PR #6656, opened 2025-06-08
Made medical skills impact CPR

Utmanarn:
Medical skills now impact the time it takes to perform CPR.

## About the PR

Made the Medical skill impact CPR timer.

## Why / Balance

Parity.
Solves: #5730

## Technical details

Added a if/else check to test against the performers medical skill to decide the time it takes to perform CPR.

## Media

None.

## Requirements

**Changelog**

:cl:
- tweak: CPR time now depends medical skill.

### DIFF
--- a/Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
+++ b/Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
@@ -1,4 +1,5 @@
 using Content.Shared._RMC14.Marines;
+using Content.Shared._RMC14.Marines.Skills;
 using Content.Shared.Atmos.Rotting;
 using Content.Shared.Damage;
 using Content.Shared.Damage.Prototypes;
@@ -11,6 +12,7 @@ using Content.Shared.Mobs.Systems;
 using Content.Shared.Popups;
 using Robust.Shared.Network;
 using Robust.Shared.Player;
+using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
 
 namespace Content.Shared._RMC14.Medical.CPR;
@@ -25,6 +27,9 @@ public sealed class CPRSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popups = default!;
     [Dependency] private readonly SharedRottingSystem _rotting = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly SkillsSystem _skills = default!;
+
+    public static readonly EntProtoId<SkillDefinitionComponent> SkillType = "RMCSkillMedical";
 
     // TODO RMC14 move this to a component
     [ValidatePrototypeId<DamageTypePrototype>]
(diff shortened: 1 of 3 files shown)


Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
```diff
@@ -177,8 +182,15 @@ private bool StartCPR(EntityUid performer, EntityUid target)
 
         EnsureComp<ReceivingCPRComponent>(target);
 
-        // TODO RMC14 less time for skilled doctors
-        var doAfter = new DoAfterArgs(EntityManager, performer, TimeSpan.FromSeconds(4), new CPRDoAfterEvent(), performer, target)
+        // If the performer has skills in medical their CPR time will be reduced.
+        var delay = TimeSpan.FromSeconds(4);
```
Dutch-VanDerLinde: image
You can add delay modifiers to the medical skill prototype and use SkillsSystem.GetSkillDelayMultiplier()

Also, make sure you don't hardcode the 4 seconds and put a new datafield on the CPR popup

Dutch-VanDerLinde: If the skill modifiers are directly hardcoded to the CM13 CPR system then just keep it tho

Utmanarn: From what I can see in the CM13 code, it is set at the HUMAN_STRIP_DELAY which is set to 4 seconds. Then they multiply it by the medical skill duration.

For medical the skill durations are 0.35 for master level and 0.75 for level doctor level.

I could not find the GetSkillDelayMultiplier() before so that's why I implemented it directly.

Utmanarn: Check now if you think the new system works better.

I took all the values directly from the CM13 codebase.

Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
```diff
@@ -177,8 +182,10 @@ private bool StartCPR(EntityUid performer, EntityUid target)
 
         EnsureComp<ReceivingCPRComponent>(target);
 
-        // TODO RMC14 less time for skilled doctors
-        var doAfter = new DoAfterArgs(EntityManager, performer, TimeSpan.FromSeconds(4), new CPRDoAfterEvent(), performer, target)
+        // If the performer has skills in medical their CPR time will be reduced.
+        var delay = TimeSpan.FromSeconds(4 * _skills.GetSkillDelayMultiplier(performer, SkillType));
```
Dutch-VanDerLinde: Make sure the delay isn't hardcoded to 4 seconds

Add a datafield to one of the cpr components

Dutch-VanDerLinde (changes requested):

Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
Utmanarn: Yep! Will do!

Utmanarn: Well, I added it to the "ReceivingCPR" component as there was no other one that made sense which already exists.

DrSmugleaf (approved):

PR #6657, opened 2025-06-08
Revert "Make corporate survivor rank up from the sum of both corpo surv and corporate liaison."

QuietlyWhisper:
Reverts RMC-14/RMC-14#6550

I believe anyone who actually played the role for the medal ranks should be rewarded for doing so.

We have these for account progression, they don't mean anything - the whole point is to play the roles they belong to obtain them, otherwise we may as well do this for all similar roles. 

Doing this removes long term account progression otherwise, and is not something I want to see pushed to the other roles because it was done in this one.

### DIFF
--- a/Resources/Locale/en-US/_RMC14/job/cm-role-timers.ftl
+++ b/Resources/Locale/en-US/_RMC14/job/cm-role-timers.ftl
@@ -3,7 +3,6 @@ role-timer-medical-roles = any medical roles
 role-timer-engineering-roles = any engineering roles
 role-timer-dropship-roles = any dropship roles
 role-timer-dropship-pilot-roles = any dropship pilot roles
-role-timer-corporate-roles = any corporate roles
 role-timer-survivor-roles = any survivor roles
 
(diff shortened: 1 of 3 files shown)


Dutch-VanDerLinde:
So the corporate survivor uses the shipside liaison playtime trackers for their playtime ranks 

Could it be considered to make it use the corporate survivor tracker instead of the liaison one?

Dutch-VanDerLinde:
Also, what do you mean by "I believe anyone who actually played the role for the medal ranks should be rewarded for doing so."?
You'll still get the medal, it's only the rank that changes

Considering that security surv gets a playtime rank I think we should change the tracker from liaison to corporate survivor

Dutch-VanDerLinde:
So I added the corporate job groups originally so a plat CL could get exec spec corporate colonist, while a plat corporate colonist can also get exec spec.

Because for some reason they used the Liaison playtime trackers. I understand if you don't want CL time to contribute to corpo colonist ranks because they are different roles .

DrSmugleaf:
@Dutch-VanDerLinde I asked Whisper and he can talk with you about it after the update tomorrow to make sure there's no issues with it

Dutch-VanDerLinde:
Ok

PR #6664, opened 2025-06-09
Mechanical enforcement of correspondent

0tito:
## About the PR

Correspondent now has CorrespondentComponent, it makes it so that if he tries to shake marines he falls into the floor, and it also makes it so that he cannot shoot guns, similar to how the clowns clumsy component works.
But this could damage actual self defense!! Thats why i added this new gun that the correspondent can shoot, ammo is unobtainable and its damage sucks, its only useful for self defense rp!

## Why / Balance

I keep running into combat correspondents, some ahelps go through, others dont, admins arent on 24/7 and that is okay, so this will make it easier for everyone.

## Media

Still WIP i gotta get a sprite for the gun and magazine
This is a video i took earlier, im too lazy to record a new one 

https://github.com/user-attachments/assets/25c365fc-b28f-4dd2-af68-70b4df582510

## Requirements

**Changelog**

### DIFF
--- a/Content.Shared/Weapons/Ranged/Components/GunComponent.cs
+++ b/Content.Shared/Weapons/Ranged/Components/GunComponent.cs
@@ -260,6 +260,13 @@ public sealed partial class GunComponent : Component
     public bool ClumsyProof = false;
 
     /// <summary>
+    /// Whether or not someone with the
+    /// correspondent trait can shoot this
+    /// </summary>
+    [DataField]
+    public bool CorrespondentProof = false;
+
+    /// <summary>
     /// Firing direction for an item not being held (e.g. shuttle cannons, thrown guns still firing).
     /// </summary>
     [DataField]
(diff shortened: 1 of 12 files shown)


Arendian:
Just making them unable to shake is already enough in my opinion, there is absolutely no reason to stun someone for a whopping 10 seconds for trying. There are multiple ways where you can accidentally click someone and you shouldn't be stunned for 10 seconds for that, one example is trying to pickup your camera after being knocked down but a marine is laying near it.

I also think not having any firearms training is already enough of a deterrent, as you're unable to hit anything that's not right in your face.

0tito:
The humble combat correspondant with a shotgun running into xenos does not agree, also the stun for shaking is so that they only do it once, after that they will learn to not do it again and just stay on the backline rping, no reason to accidentally shake somoene if you have your camera in your hand and you arent interacting with marines

BarteG44:
Needs more sovl. Make the punishment make actual sense IC, even if it's just flavortext like "you try to shake [marine] but you trip and fall!"

PursuitInAshes:
Thanks for the PR!
We discussed this on discord a while ago and the thoughts were this likely wasn't needed, I brought it up again and the RP expectations around correspondents have shifted slightly so this likely isn't needed anymore atleast.

As a general note, almost all ahelps are handled at somepoint but admins are not always in round to address them in that precise moment

PR #6665, opened 2025-06-09
Bug fix: Space drug conservation of mass

x5tink:
## About the PR
Something something conservation of mass, something something insert witty comment here.

## Why / Balance
While looking at the new stuff i noticed a couple of the chems in the Space drugs merge, created more chems than input. This seems wrong.

## Technical details
Corrected the two chems to only produce the correct amount based on input chems.

## Media
https://github.com/user-attachments/assets/6b8ff1bf-ccf7-4d6b-9860-ba296bbec7de

## Requirements

**Changelog**
:cl:
- fix: Fixed space drug Conservation of Mass

### DIFF
--- a/Resources/Prototypes/_RMC14/Recipes/Reactions/medicine.yml
+++ b/Resources/Prototypes/_RMC14/Recipes/Reactions/medicine.yml
@@ -191,7 +191,7 @@
     RMCHydrogen:
       amount: 1
   products:
-    RMCMethylphenidate: 3
+    RMCMethylphenidate: 2
 
 - type: reaction
   id: RMCCitalopram
@@ -201,7 +201,7 @@
     RMCCarbon:
       amount: 1
   products:
-    RMCCitalopram: 3
+    RMCCitalopram: 2
 
 - type: reaction
   id: RMCParoxetine


Arykh:
I want to agree with you, but the almighty Parity will smite thee. 😭 

IMG_0205

I did alittle digging and the 2:3 ratio for these two meds have been like that since the initial commit of cm13. Even looking at baystation12, the base that cm13 is built from, it's still 2:3 and I can't find any reason why it's this ratio.

I haven't looked at every ss13 servers, but those that have these two meds have the same ratio.

It's up to smug and whisper if they wana change it.

Arykh:
You could potentially do a pr in cm13 and then parity it here. *hint* *hint* *wink* *wink*

x5tink:
Or i could not, if RMC wants the fix they can have, if they dont, just reject this. I'm not going to go work on cm13.

x5tink:
As a note, Ive also noticed Epinephrine and Imidzoline also does this, where 3u chems go in, 2u chems come out. Ive not done anything with that pending the outcome of this PR

Arykh:
As you might imagine, epinephrine and imidazoline's input output ratio of 3:2 is parity. This will also be true for Alkysine(heals brain damage).
At least it's not magically creating more than you put in lol.

x5tink:
Yeah i did note that one, but figured i'd point it out at the same time,

DrSmugleaf (changes requested):
As long as this can't be broken down into more of its ingredients, units are not necessarily an indication of mass, you can put in large molecules that break into more smaller molecules

x5tink:
> As long as this can't be broken down into more of its ingredients, units are not necessarily an indication of mass, you can put in large molecules that break into more smaller molecules

I'm not aware of anyway to break down these drugs, but im not that well versed in chemisty outside of medical in rmc.

It was simply created as its supprising when you come across one of these recipes, (see Epi and Imi, and these drugs), they dont follow what you might naturally expect from doing other chemisty work, IE chems input = sum of input chems, as output chem.

If your fine with chemicals producing more than the input sum of chemicals then by all means please go ahead and close this ticket.

x5tink:
Closing to not maintain this further. Feel free to open in the future

PR #6699, opened 2025-06-11
Barricades block leaps, lunges, melee attacks and tackles from the front

Arendian:
## About the PR

Adding barbed wire to a barricade now increases it's max health by 50.
Any xeno that attempts to leap or lunge over a barbed barricade will be stopped and stunned.
Trying to melee attack or tackle over a barricade from the front now makes you attack the barricade instead of the targeted entity. The chance for this to happen is the current health percentage of the barricade, with a minimum chance of 30%.

Tail Stab is now a targeted attack instead of a line.
Leap actions now ignore barricades without barbed wire and go over them.

## Why / Balance

Parity.

## Technical details

When an attack or tackle is attempted a raycast will check if there are any entities with the DirectionalAttackBlockerComponent in the way, if there are, a check is done to see if the attack is coming from the front of any found entities. The chance for the entity to block the attack is based on it's current health, every time it takes damage, or when the entity fails to block an attack, a new random number between 0 and 1 is generated. If the random number is higher than the current health percentage of the barricade, the next attack it receives will not be blocked.

## Media

https://github.com/user-attachments/assets/b689e49c-104d-4015-aa88-eeff9973ea66

## Requirements

**Changelog**

:cl: Dygon
- tweak: Applying barbed wire to a barricade now increases it's max health by 50.
- tweak: Attempting to attack or tackle over a barricade from the front now has a chance to make you attack the barricade instead. The chance of this occurring ranges from 100% to 30%, depending on the barricade's current health.
- tweak: Trying to leap or pounce over a barbed barricade will now stun you.
- tweak: All leaps now ignore barricades without barbed wire.
- tweak: Tail Stab is now a targeted action and requires you to click on your target.
- fix: The grenadier armor's leap protection no longer blocks ravagers and praetorians.


texas1234567889:
small correction. its supposed to stun only if the barricade is barbed, otherwise it just lets you leap right over (unsure how many xenos this applies to, but i tested ravager. runner. lurker. vanguard praetorian. base praetorian. and even warrior. and they all were able to leap over an unbarbed cade, t3s arent stunned by leaping into a barbed cade t1s and t2s are) 

shit quality because i had to get it under 10mb but you also seem to take damage when you slam into a barbed cade and get stunned, t3s dont take the damage so the damage is linked to the cadeslam stun i'd guess

https://github.com/user-attachments/assets/d8989625-b7ee-4e08-a808-181d3131a531

Arendian:
> small correction. its supposed to stun only if the barricade is barbed, otherwise it just lets you leap right over (unsure how many xenos this applies to, but i tested ravager. runner. lurker. vanguard praetorian. base praetorian. and even warrior. and they all were able to leap over an unbarbed cade, t3s arent stunned by leaping into a barbed cade t1s and t2s are)

Thanks for that clarification, barricades now only stun while barbed.

Dutch-VanDerLinde:
Does this allow runners and lurkers to leap over Cades if they're not barbed?

Dutch-VanDerLinde:
Also, nice work. This system could be used for directional slashing

Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
```diff
@@ -168,6 +175,7 @@ private void OnXenoLeapDoAfter(Entity<XenoLeapComponent> xeno, ref XenoLeapDoAft
         leaping.LeapSound = xeno.Comp.LeapSound;
         leaping.LeapEndTime = _timing.CurTime + TimeSpan.FromSeconds(direction.Length() / xeno.Comp.Strength);
 
+        _obstacleSlamming.MakeImmune(xeno, 0.5f);
```
Dutch-VanDerLinde, line 184: You shouldn't need this? Leap code doesn't use throwing, which obstacle slamming listens to throwing events. Unless it was changed

Arendian: To prevent clipping through the barricade the RMCLeapProtection knocks back any entities that slam into it a bit, that knockback causes obstacle slamming.

Arendian:
> Does this allow runners and lurkers to leap over Cades if they're not barbed?

It currently does not, but I think if I change the leap logic, from adding an impulse to "throwing" the xeno, it will make them able to go over. I think this is why the warrior does go over with it's lunge.

DrSmugleaf (changes requested):
You probably want to use deterministic rng like gun prediction/lag compensation does by using Xoroshiro64S instead, otherwise fast attacks will make the client and server roll different numbers and mispredict
If you do that you can put all the code in shared and not have to store a roll

Arendian:
I replaced the stored roll with a XoroShiro64 seed, logging shows the seed staying the same for the client and server. I also changed Tail Stab into a targeted ability, it's now blocked by barricades too.

Arendian:
> Does this allow runners and lurkers to leap over Cades if they're not barbed?

Tables and barricades without barbed wire are now ignored during a leap.

DrSmugleaf (approved):
Thanks for the PR!

PR #6707, opened 2025-06-11
Fix close queen scream stacking with far scream effects

Vermidia:
## About the PR

Title.

## Why / Balance

Bug. CM checks range and doesn't stack those.

## Technical details

Moved the close effects to check first, dupe the set to a temp hashset and don't do far effects if it was in the set.

## Requirements

**Changelog**

:cl:
- fix: Fixed close Queen scream stacking with far queen scream effects.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Screech/SharedXenoScreechSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Screech/SharedXenoScreechSystem.cs
@@ -26,6 +26,7 @@ public sealed class XenoScreechSystem : EntitySystem
     [Dependency] private readonly XenoSystem _xeno = default!;
 
     private readonly HashSet<Entity<MobStateComponent>> _mobs = new();
+    private readonly HashSet<Entity<MobStateComponent>> _closeMobs = new();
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Xenonids/Screech/SharedXenoScreechSystem.cs
```diff
@@ -57,27 +58,32 @@ private void OnXenoScreechAction(Entity<XenoScreechComponent> xeno, ref XenoScre
             _audio.PlayPvs(xeno.Comp.Sound, xeno);
 
         _mobs.Clear();
-        _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.StunRange, _mobs);
+        _entityLookup.GetEntitiesInRange(xform.Coordinates, xeno.Comp.ParalyzeRange, _mobs);
+
+        HashSet<Entity<MobStateComponent>> closeTargets = _mobs.ToHashSet(); //Duplicate
```
DrSmugleaf: This should be a field on the system that gets cleared instead

DrSmugleaf (changes requested):

PR #6712, opened 2025-06-12
Removes SO job scaling nearly entirely.

UnicornOnLSD:
## About the PR

Removed the job scaling for staff officers. Now by default you start with 4 available, and a 5th one will be available if there are 160+ marines.

## Why / Balance

No matter the pop. There will be multiple squads.

As of now, moves members are limited (SLs, WSes, SGOs) and encryption key do not automatically transfer. Squad merging is not reliable for low pop.

Staff officer is best enjoyed when you can focus on one squad to overwatch. Not having to manage three squad comms because it's 70 pop and the other SO is dedicated on alpha.

The cap makes CIC lacking in staff that it needs. It makes the role less enjoyable due to forcing it to do extra work load.

Consequently, it makes XO worse to play because you do not have proper relay (the SO enacting the plan to respective squads) between you and squads. And as such makes CIC less reliable for marines as a whole.

See feature post for community discussion/opinion: "Remove SO cap"
https://discord.com/channels/1168210010233376858/1382468386214379530

## Technical details

yml soup
## Media

N/A
## Requirements

**Changelog**

:cl:

- remove: Removed SO job scaling. CIC will now have 4 SO slots at any pop, up to 5 at 160+ pop.

### DIFF
--- a/Resources/Prototypes/_RMC14/rmc_job_slot_scaling.yml
+++ b/Resources/Prototypes/_RMC14/rmc_job_slot_scaling.yml
@@ -22,9 +22,9 @@
         min: 4
         max: 8
       CMStaffOfficer:
-        factor: 40
+        factor: 160
         c: 1
-        min: 2
+        min: 4
         max: 5
       CMDoctor:
         factor: 25


QuietlyWhisper:
I think 4 SOs should always be available yes

PR #6715, opened 2025-06-12
Make warriors unable to tackle their lunge victim

Arendian:
## About the PR

If a Warrior attempts to tackle their lunge target while it is still stunned by the lunge, the tackle will be replaced with a light attack

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Warriors can no longer tackle targets that are stunned by their own lunge effect.

### DIFF
--- a/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
+++ b/Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
@@ -2,6 +2,7 @@ using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Numerics;
 using Content.Shared._RMC14.CCVar;
+using Content.Shared._RMC14.Weapons.Melee;
 using Content.Shared.ActionBlocker;
 using Content.Shared.Administration.Logs;
 using Content.Shared.CombatMode;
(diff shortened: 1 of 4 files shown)


Content.Shared/_RMC14/Weapons/Melee/MeleeAttackAttemptEvent.cs
```diff
@@ -0,0 +1,9 @@
+using Content.Shared.Weapons.Melee.Events;
+using Robust.Shared.Map;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared._RMC14.Weapons.Melee;
+
+[ByRefEvent]
+[Serializable, NetSerializable]
+public record struct MeleeAttackAttemptEvent(NetEntity Target, AttackEvent Attack, NetCoordinates Coordinates , List<NetEntity> PotentialTargets,NetEntity? Weapon = null);
```
Dutch-VanDerLinde:
```suggestion
public record struct MeleeAttackAttemptEvent(NetEntity Target, AttackEvent Attack, NetCoordinates Coordinates, List<NetEntity> PotentialTargets, NetEntity? Weapon = null);
```

Dutch-VanDerLinde: indentatioin

Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
```diff
@@ -381,6 +382,29 @@ private bool AttemptAttack(EntityUid user, EntityUid weaponUid, MeleeWeaponCompo
 
                 if (!Blocker.CanAttack(user, target, (weaponUid, weapon), true))
                     return false;
+
+                // RMC14
+                if (target != null)
```
DrSmugleaf: You need to recheck if it can attack non-disarm and if the weapon id is the same as the target id, same as above
You might be better off moving this up before this switch, and changing what attack refers to there though, for code duplication

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/Weapons/Melee/SharedMeleeWeaponSystem.cs
Arendian: I made it so the code runs after the switch, because I need the target of the attack or disarm.  The target is retrieved inside the switch. I also moved the code to an RMC class instead of the upstream one.

DrSmugleaf (approved):

PR #6720, opened 2025-06-12
Solaris Ridge Soul Update

Dutch-VanDerLinde:
## About the PR
Adds the Double barrel from CMSS13 to solaris ridge 

Adds props, bloodstains, etc to the areas around the hospital, filtration and bar
Not all the props are added but most of the ones around the above areas should be in.

(FROM CMSS13, double barrel)
image

:cl:
- tweak: Added more soul to Big Red (Solaris Ridge)


Vermidia:
Note that our double barrel is the old one so the new stats need to be ported with it.

Dutch-VanDerLinde:
> Note that our double barrel is the old one so the new stats need to be ported with it.

Stats fixed

DrSmugleaf (approved):

PR #6724, opened 2025-06-13
Add Neuro hallucinations + fix process time + other fixes

Vermidia:
## About the PR

Title. Neuro only processes every 1 second now (parity). Also returned blindness to it, since that should prevent any weirdness that was happening before, parasite blindness + others getting stuck aren't reported too often so it should be ok.

## Why / Balance

Parity.

## Technical details

Fixed rejuvenate not regening stamina correctly (it just full heals now always).

NeuroComp stores some hallucination sounds, and some names were changed for stuff since they happen per tick and not necessarily per second.

So when a hallucination is added, it's type is rolled from a weighted list (probabilities based on the CM code), and then run. If it's a lingering hallucination a comp is ensured and added to a list for that to be run. That query goes and checks the timing of each thing on the list and runs it if it can, adding it to be removed and adding the returned tuple to be added later (to not cause foreach issues).

The actual processes are simple, it's mostly audio. Audio is randomly offset from the initial player's location to seem more believable.  OB Alerts, Firemission warnings, and mortar warnings will also fill the chat to match their actual counterparts somewhat. The xeno hallucinations have some stun attached (CM knocks you down) as well. The stage counter of the tuple is changed, and timespan is updated as it's returned. Like CM, the CAS hallucination is pretty long.

## Media

https://github.com/user-attachments/assets/c119e3ba-adca-4c62-80f6-e57e28e6c8c9

https://github.com/user-attachments/assets/3bd88ba1-2728-4733-bca1-ba5e92f3da3f

(Stamina damage and neurotoxin tick down was turned off for testing, CAS was bugged a bit here and I forget to set a random position so it always used the player as the center. Frequency was also tuned up a bit here.)

## Requirements

**Changelog**

:cl:
- fix: Fixed Neurotoxin not blinding
- fix: Fixed Neurotoxin not causing hallucinations


Content.Shared/_RMC14/Xenonids/Neurotoxin/NeurotoxinComponent.cs
```diff
@@ -79,4 +90,65 @@ public sealed partial class NeurotoxinComponent : Component
...
+    [DataField]
+    public Dictionary<string, float> Hallucinations = new()
+    {
+        {"AlienAttack", 0.05f},
+        {"OB", 0.05f},
+        {"Screech", 0.06f},
+        {"CAS", 0.08f},
+        {"Mortar", 0.18f},
+        {"Giggle", 0.27f},
+        {"Sounds", 0.31f}
```
DrSmugleaf: It would be better for these strings to be a new enum, both for YAML validation and having data types used represent their use

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf (approved):

PR #6747, opened 2025-06-14
Fix the bipod's scatter and slow values.

### LINKED ISSUE #6761: HAR with Bipod and scope accuracy bug.
## Description
The accuracy of using an Har with bipod and scope bugs when you wield it before deploying causing an negative accuracy buff it seems.

**Reproduction**
Have an HAR with bipod and reflex sight/red dot
Wield -> deploy bipod 
The accuracy does propperly work when you do the following.
Deploy bipod -> wield

**Additional context**
I have tested this with other guns like the M54MK2 and it doesnt expiernce the same problem with the same attachment.
I have tested this only with the red dot scope and reflex scope so their might be more such cases with other scopes i havent tested.

https://github.com/user-attachments/assets/b0a681ad-1a4d-4118-8c16-7d4950c35737

Arendian:
## About the PR

The bipod now doesn't applies multiple stacking slows anymore, only a slow and a  superslow.(which don't stack)

The bipod now also refreshes the burst scatter multiplier when deployed or undeployed, this means that the scatter reduction when using a deployed bipod is now properly applied. Previously it would only apply if the weapon had at least one other attachment that would refresh the burst scatter multiplier, and if it was deployed before wielding.

## Why / Balance

Parity and bug fix.
Resolves #6761

## Technical details

Previously a wielded M54CE2 with only a bipod would have a max scatter of 90 degrees in automatic mode with the bipod deployed and no other attachments, this is now 50 degrees because the burst scatter value gets refreshed upon deploying/undeploying the bipod.

If the M54CE2 was used with a bipod AND reflex sight, it would only apply the bipod scatter reduction if it was first deployed and then wielded, this is because the reflex sight would call a refresh for burst scatter multiplier upon wielding the gun. If the gun was first wielded(refresh burst scatter multiplier because of reflex) and then deployed the bipod, the scatter wouldn't change at all because the action of deploying the bipod would not refresh the burst scatter value.

## Requirements

**Changelog**

:cl: Dygon
- fix: Improperly folding the bipod attachment now applies a weaker slowdown effect.
- fix: The bipod attachment now correctly applies it's scatter reduction effect when toggled.

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Components/AttachableTemporarySpeedModsComponent.cs
+++ b/Content.Shared/_RMC14/Attachable/Components/AttachableTemporarySpeedModsComponent.cs
@@ -16,10 +16,10 @@ public sealed partial class AttachableTemporarySpeedModsComponent : Component
     public List<TemporarySpeedModifierSet> Modifiers = new();
 
     [DataField, AutoNetworkedField]
-    public TimeSpan SlowDuration = TimeSpan.FromSeconds(2);
+    public TimeSpan SlowDuration = TimeSpan.FromSeconds(4);
 
     [DataField, AutoNetworkedField]
-    public TimeSpan SuperSlowDuration = TimeSpan.FromSeconds(1);
+    public TimeSpan SuperSlowDuration = TimeSpan.FromSeconds(2);
 }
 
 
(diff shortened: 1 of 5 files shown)


Vermidia:
I was unware temporary speed mods were a thing. I already added the speed mods on undeploy through the method BreakSlow (Which are the same values but doubled because remember CM handling status effect results in the numbers x2 90% of the time).

Arendian:
> I was unware temporary speed mods were a thing. I already added the speed mods on undeploy through the method BreakSlow (Which are the same values but doubled because remember CM handling status effect results in the numbers x2 90% of the time).

I forgot about the doubling also counting for slows. I removed the method you made since the already existing component allows for more flexible slow durations.

DrSmugleaf (approved):

PR #6757, opened 2025-06-15
Resprite Mateba

VictorJob:
## About the PR

This PR alters mateba sprites with the intention of removing the noise from the original sprites.

## Media

image

## Requirements


Arcticular1:
Personally i do not like this resprite, it feels more akin to the "classic" mateba variant, opposed to the standard, as well as it feel's like it doesn't fit the look of a real Mateba.
image

QuietlyWhisper (changes requested):
I don't like the major loss of detail, these don't look good

PR #6758, opened 2025-06-15
Synth fixes

Dutch-VanDerLinde:
## About the PR
Removes synth's crit threshold (parity)
Removes their damage slowdown
Synths aren't affected by neurotoxin slash or spit/slowing spit

:cl:
ADMIN:
- fix: Fixed synths being affected by xeno neurotoxin spit and slashes
- fix: Fixed synths having a crit threshold


Content.Shared/_RMC14/Xenonids/Paralyzing/XenoParalyzingSlashSystem.cs
```diff
@@ -80,6 +81,13 @@ private void OnXenoParalyzingSlashHit(Entity<XenoActiveParalyzingSlashComponent>
                 continue;
             }
 
+            if (HasComp<SynthComponent>(entity))
```
DrSmugleaf, line 84: This should probably be its own component in the future but this is fine for now

DrSmugleaf (approved):

PR #6767, opened 2025-06-16
Drop pouches for medical + name refactor

x5tink:
## About the PR
While looking at something else i noticed that in CM13 medical staff have access to the drop pouch in their vendors, so this adds that, Also i renamed the Drop pouch from CMWebbingPouch to CMWebbingDropPouch for clarity, as i got awful confused when i looked at it, and this is a bit clearer.

## Why / Balance
Parity and Clarity... Hey that ryhmes....

## Technical details
Renamed all the CMWebbingPouch to CMWebbingDropPouch for clarity
Added drop pouch to medical vendors

## Media
Before:
DropPouchBefore

After:
DropPouchAfter

PR Rename:
image

## Requirements

**Changelog**
:cl:
- add: Added drop pouch to medical vendors


DrSmugleaf (changes requested):
Thanks for the PR!
You don't need to edit existing map files, you can just add CMWebbingPouch: CMWebbingDropPouch in migration.yml to automatically change it in any existing maps next time they are saved, and when they are loaded

x5tink:
I was not aware the migrations thing was an uh... thing xD I've removed the map renames, added to the migration file as you suggested.
I Then renamed them all to RMCWebbingDropPouch, to be more inline with the other renames, but noticed some inconsistency in the names, in the migrations file. Is there a specific reason to name something CM vs RMC?

x5tink:
Should have fixed the conflict....

DrSmugleaf:
> I was not aware the migrations thing was an uh... thing xD I've removed the map renames, added to the migration file as you suggested. I Then renamed them all to RMCWebbingDropPouch, to be more inline with the other renames, but noticed some inconsistency in the names, in the migrations file. Is there a specific reason to name something CM vs RMC?

CM is old IDs, any new IDs that aren't for job prototypes specifically should be prefixed RMC

x5tink:
Rebased and fixed the migrations file again.

DrSmugleaf (approved):

PR #6768, opened 2025-06-16
Add Xeno evolution groundside alert for people on the planet

Dutch-VanDerLinde:
## About the PR
Soul + Parity
Synths have their own message

https://github.com/user-attachments/assets/ddca8f6d-1776-4d03-8c6e-f256faaf5253

:cl:
- add: Colonists will now be an alerted when a new xeno caste unlocks


Dutch-VanDerLinde:
> please move these warnings to popups, they were invented in ss14 for this very purpose - not to clutter up the chat. I would really not want degradation and breaking of all the game design achievements of ss14 when porting content in this way. so that one fine day we do not discover that the chat is as terrible and built-in as in the game that we are supposedly improving to modern game design standards

The message is just far too long to fit as a popup, not everyone would be able to read it in time

chavonadelal:
> > please move these warnings to popups, they were invented in ss14 for this very purpose - not to clutter up the chat. I would really not want degradation and breaking of all the game design achievements of ss14 when porting content in this way. so that one fine day we do not discover that the chat is as terrible and built-in as in the game that we are supposedly improving to modern game design standards
> 
> The message is just far too long to fit as a popup, not everyone would be able to read it in time

it is not intended for chat and this cannot be the reason why it is clogging it, you can always reduce the text size or remove the bold font, please do not disappoint the players who came to play the improved remake and not to see how the build degrades, this should be taken very seriously

QuietlyWhisper:
> The message is just far too long to fit as a popup, not everyone would be able to read it in time

I agree with this but I also agree that we're putting too much text in the chat box lately, I'd like to see what smug thinks of this

You could pace it into two separate popup messages? I think anything that you'd want to read over more than once (like announcements) should be in chat log, but anything single use like this makes more sense to pop up.

Dutch-VanDerLinde:
> > The message is just far too long to fit as a popup, not everyone would be able to read it in time
> 
> I agree with this but I also agree that we're putting too much text in the chat box lately, I'd like to see what smug thinks of this
> 
> You could pace it into two separate popup messages? I think anything that you'd want to read over more than once (like announcements) should be in chat log, but anything single use like this makes more sense to pop up.

this seems like a lot of work for something so small

Dutch-VanDerLinde:
it's a pretty important alert anyhow, incase you missed it while being afk

JavaVoidWasTaken:
It would be important to make this toggleable!

PVE GM's might want to not let their players suddenly hear an ear piercing screech during their sessions.

DrSmugleaf (approved):
soul
can worry about the pve toggle later most pve rounds don't have players groundside early enough for this to matter

PR #6783, opened 2025-06-17
Provost MP requirement fixes

noctyrnal:
## About the PR

Provost enforcers and leaders now require MP time.

**Changelog**

:cl: noctyrnal
- fix: Provost Enforcers and Provost Team Leaders now require 5 hours of MP playtime.


AverageNotDoingAnythingEnjoyer:
Bless

t0wnshark:
imo ought to be ten rather than five since five isn’t even enough to hit the first playtime
milestone on MP

PR #6788, opened 2025-06-18
I eep, therefore I am. [xenonid rest keybind]

Arykh:
## About the PR

Made the rest keybind activate xeno's rest action.
## Why / Balance

Parity keybind.
## Technical details

Press shift + r
checks if xeno
eep
## Media

https://youtu.be/xWPIsmTsqA0
## Requirements

**Changelog**

:cl:
- tweak: Xenonids can now bind a key to their rest action. (Shift+R by default)


Content.Shared/_RMC14/Xenonids/Rest/XenoRestBindInputSystem.cs
```diff
@@ -0,0 +1,33 @@
...
+        base.Initialize();
+
+        CommandBinds.Builder
+            .Bind(CMKeyFunctions.RMCRest,
+                InputCmdHandler.FromDelegate(session =>
+                    {
+                        var ent = session?.AttachedEntity;
+                        if (ent != null && HasComp<XenoComponent>(ent.Value))
+                        {
+                            var ev = new XenoRestActionEvent();
```
DrSmugleaf: You can't raise an action event like this, it will have none of the data that it usually does, like the performer, action id, etc
It also will not check for cooldowns
You need to find the action on the xeno and then call ActionsSystem.TriggerAction

Content.Shared/_RMC14/Xenonids/Rest/XenoRestBindInputSystem.cs
```diff
@@ -0,0 +1,73 @@
...
+            .Register<XenoRestBindInputSystem>();
+    }
+
+    public override void Shutdown()
+    {
+        base.Shutdown();
+        CommandBinds.Unregister<XenoRestBindInputSystem>();
+    }
+
+    private bool TryFindActionByPrototype(EntityUid owner,
```
Dutch-VanDerLinde:
```suggestion
    public bool TryFindActionByPrototype(EntityUid owner,
```
make this public, it could be used by other systems

Arykh:
OH NO, IT CRASHED THE CLIENT. :disintegrate:

Arykh:
Screw it!
I’m gonna create a new CMKeyFunction.RMCXenoRest and have it use shift+r by default too. I can then use handle: true and not have to rely on janky timers that you can bypass easily by spamming the input. Doing it this way will stop TriggerAction from sending multiple RaisePredictiveEvent and crashing the client.

Arykh:
It is done!
Having handle: true stops TriggerAction from sending multiple RaisePredictiveEvent and crashing the client!
I've decided to keep the simple cooldown check to stop input spam and as an extra precaution against giga edge case scenarios. No one likes it when their game crashes. 😄

Content.Client/Input/ContentContexts.cs
```diff
@@ -86,7 +86,7 @@ public static void SetupContexts(IInputContextContainer contexts)
             human.AddFunction(ContentKeyFunctions.Arcade2);
             human.AddFunction(ContentKeyFunctions.Arcade3);
 
-            // actions should be common (for ghosts, mobs, etc)
+            // actions should be common (for ghosts, mobs, etc.)
```
DrSmugleaf: Should be etc. but this is an upstream line of code so best left unchanged for merge conflicts
```suggestion
            // actions should be common (for ghosts, mobs, etc)
```

Content.Client/_RMC14/Xenonids/Rest/XenoRestKeybindSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+                InputCmdHandler.FromDelegate(_ =>
+                {
+                    var timing = IoCManager.Resolve<IGameTiming>();
+                    var curTime = timing.CurTime;
+                    if (_pendingRestTrigger && curTime < _pendingRestTriggerUntil)
+                        return;
+                    var ent = _playerManager.LocalEntity;
+                    if (ent == null || !ent.Value.IsValid() || !HasComp<XenoComponent>(ent.Value))
+                        return;
+                    if (!_findActionByPrototype.TryFindActionByPrototype(ent.Value, "ActionXenoRest", out var actionId, out var actionComp))
```
DrSmugleaf: This should be an action BaseEvent check, some code in RMC14 does this, but it goes action.BaseEvent is type, with type in this case being XenoRestActionEvent
If you wanted it to stay it would need ActionXenoRest to be a static readonly EntProtoId<InstantActionComponent> field so that it gets automatically validated by the YAML linter

Content.Client/_RMC14/Xenonids/Rest/XenoRestKeybindSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+    private TimeSpan _pendingRestTriggerUntil = TimeSpan.Zero;
+
+    public override void Initialize()
+    {
+        base.Initialize();
+        CommandBinds.Builder
+            .Bind(CMKeyFunctions.RMCXenoRest,
+                InputCmdHandler.FromDelegate(_ =>
+                {
+                    var timing = IoCManager.Resolve<IGameTiming>();
```
DrSmugleaf: You can put a dependency to IGameTiming at the top of this file and use that instead

Content.Client/_RMC14/Xenonids/Rest/XenoRestKeybindSystem.cs
```diff
@@ -0,0 +1,64 @@
...
+            .Register<XenoRestKeybindSystem>();
+    }
+
+    public override void FrameUpdate(float frameTime)
+    {
+        base.FrameUpdate(frameTime);
+        var timing = IoCManager.Resolve<IGameTiming>();
+        if (_pendingRestTrigger && timing.CurTime >= _pendingRestTriggerUntil)
+            _pendingRestTrigger = false;
+    }
```
DrSmugleaf: You should be checking the time above instead of changing the value of a boolean depending on the time here
Are you sure you need the time check anyways? This is only done on the client, so it's not very useful
Either way these values should have been stored in the component instead, but you shouldn't need it or these

Content.Shared/_RMC14/Actions/FindActionByPrototype.cs
```diff
@@ -0,0 +1,27 @@
+using Content.Shared.Actions;
+
+namespace Content.Shared._RMC14.Actions;
+
+public sealed class FindActionByPrototype : EntitySystem
```
DrSmugleaf: The file and this class should end with System, since it's an EntitySystem
You don't need it though, since you can check the action by event type instead of ID and delete this file

Content.Shared/_RMC14/Xenonids/Rest/RMCXenoRestComponent.cs
```diff
@@ -0,0 +1,4 @@
+namespace Content.Shared._RMC14.Xenonids.Rest;
+
+[RegisterComponent]
```
DrSmugleaf: This should also be a NetworkedComponent since it's in Shared
You can delete it though, it's not used

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
```diff
@@ -90,6 +90,7 @@
   - type: MobMover
   - type: Emoting
   - type: Actions
+  - type: RMCXenoRest
```
DrSmugleaf: If you wanted to keep this you should add this to the bottom since its an RMC specific component, to have them together, could have called it XenoRest component though, and update the file and class names
You don't need it though, so just remove this when you remove its file

DrSmugleaf (changes requested):

Arykh:
Tested. No crashes and it works.
### XIPPEE 🎉🎉🎉🎉🎉

DrSmugleaf (approved):

PR #6793, opened 2025-06-18
decrease minimum usable range by mortar, by taking from maximum range

QuietlyWhisper:
## Why / Balance

- Makes the mortar more usable at closer ranges, at the expense of longer ranges (by the same amount of tiles)
- This will no longer require you to be as far back from the front, but will require you to move up with the front.

**Changelog**

:cl: Whisper
- tweak: Decreased the minimum range required for mortars by 10 tiles in exchange for the maximum range reduction of 10 tiles. You will be able to shoot it closer but not as far by the same amount.


DrSmugleaf (approved):

AverageNotDoingAnythingEnjoyer:
Everything was non-parity, until parity struck
https://github.com/cmss13-devs/cmss13/pull/9549

PR #6801, opened 2025-06-20
Recycling dropship parts and ammo

Arendian:
## About the PR

You can now recycle printed dropship ammo and parts at the dropship part fabricator, this refunds 80% of the point cost.
Ammo that is partially used refunds less points based on how much ammo is left, 70% of ammo left means you only get: 0.8 * 0.7 = 56% of the original point cost refunded instead of 80%.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- add: Dropship parts and ammo can now be recycled at the fabricator, this returns 80% of their point value. Ammo that is partially used refunds less points based on how much ammo is left, 70% of ammo left means you only get: 0.8 * 0.7 = 56% of the original point cost refunded instead of 80%.


Resources/Prototypes/_RMC14/Entities/Objects/PowerLoader/Dropship/rmc_dropship_fabricator.yml
```diff
@@ -2,7 +2,7 @@
   parent: BaseMachinePowered
   id: RMCDropshipFabricator
   name: dropship part fabricator
-  description: A large automated 3D printer for producing dropship parts. # You can recycle parts or ammo in it, and get 80% of your points back, by clicking it while holding them in a powerloader claw. # TODO RMC14
```
DrSmugleaf, line 5: Is this removal intended? It's intended to be part of the description when implemented

Arendian: Ah I thought the entire comment was part of the TODO when I removed that, I added the previously commented out part as part of the description now.

PR #6802, opened 2025-06-20
feat: rank localization

TornadoTechnology:
## About the PR
Using Loc.GetString, to be able to translate ranks.

## Why / Balance
Ru forks.

## Requirements

### DIFF
--- a/Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
+++ b/Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
@@ -76,17 +76,18 @@ public abstract class SharedRankSystem : EntitySystem
     }
 
     /// <summary>
-    ///     Gets the rank name of a given mob.
+    /// Gets the rank name of a given mob.
     /// </summary>
     public string? GetRankString(EntityUid uid, bool isShort = false, bool hasPaygrade = false)
     {
         var rank = GetRank(uid);
-        if (rank == null)
+
+        if (rank is null)
             return null;
 
         if (isShort)
         {
-            if (rank.FemalePrefix == null || rank.MalePrefix == null)
+            if (rank.FemalePrefix is null || rank.MalePrefix is null)
                 return rank.Prefix;
 
             if (!TryComp<HumanoidAppearanceComponent>(uid, out var humanoidAppearance))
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
```diff
@@ -76,17 +76,18 @@ public void SetRank(EntityUid uid, ProtoId<RankPrototype> from)
     }
 
     /// <summary>
-    ///     Gets the rank name of a given mob.
+    /// Gets the rank name of a given mob.
```
DrSmugleaf: The rest of the file (and default by .editorconfig) is to have 4 spaces before the comment, you don't need to make this change
```suggestion
    ///     Gets the rank name of a given mob.
```

Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
```diff
@@ -76,17 +76,18 @@ public void SetRank(EntityUid uid, ProtoId<RankPrototype> from)
...
     /// </summary>
     public string? GetRankString(EntityUid uid, bool isShort = false, bool hasPaygrade = false)
     {
         var rank = GetRank(uid);
-        if (rank == null)
+
+        if (rank is null)
```
DrSmugleaf: Same for this, we style most null checks with == and !=, not is and is not
```suggestion
        if (rank == null)
```

Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
```diff
@@ -76,17 +76,18 @@ public void SetRank(EntityUid uid, ProtoId<RankPrototype> from)
...
             return null;
 
         if (isShort)
         {
-            if (rank.FemalePrefix == null || rank.MalePrefix == null)
+            if (rank.FemalePrefix is null || rank.MalePrefix is null)
```
DrSmugleaf: Same here
```suggestion
            if (rank.FemalePrefix == null || rank.MalePrefix == null)
```

Content.Shared/_RMC14/Marines/Roles/Ranks/SharedRankSystem.cs
```diff
@@ -96,19 +97,16 @@ public void SetRank(EntityUid uid, ProtoId<RankPrototype> from)
...
-        else if (hasPaygrade && rank.Paygrade != null)
-        {
-            return "(" + rank.Paygrade + ")" + " " + rank.Name;
-        }
-        else
-        {
-            return rank.Name;
-        }
+
+        if (hasPaygrade && rank.Paygrade is not null)
```
DrSmugleaf: Same here
```suggestion
        if (hasPaygrade && rank.Paygrade != null)
```

DrSmugleaf (changes requested):
Thanks for the PR!

PR #6828, opened 2025-06-24
SetOperationName command

noctyrnal:
## About the PR

cro gets dragged into the CS mines because his hunger is never satiated
adds a command called "SetOperationName" that updates the upcoming round's operation name, used ideally for PVE operations or canon round akin to valliant fall. 

## Why / Balance

QoL, now the round alerts for PVE rounds can match up to actual operation names given to the sessions.

## Technical details

Adds a new Command called `SetOperationNameCommand.cs`, on run we verify if we're in the lobby - if not we prevent the command - if we are, we use the new function `SetCustomOperationName` in `CMDistresssignalRuleSystem.cs` to update OperationName.
Updates `GetRandomOperationName` in `CMDistresssignalRuleSystem.cs` to return OperationName and set `_usingCustomOperationName` to false when called if the value of OperationName isn't null and if `_usingCustomOperationName` is true

## Media

https://github.com/user-attachments/assets/f41600a8-024f-4598-9606-89415ede0828

## Requirements

**Changelog**

no cl since not player facing


Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1779,6 +1784,9 @@ public void CancelPlanetVote()
 
     private string GetRandomOperationName()
     {
+        if (OperationName != null)
+            return OperationName;
+
```
DrSmugleaf: Wouldn't this make the operation name be reused for future rounds too?
You probably want to make this only last one round

noctyrnal: fixed, gave it a bool condition

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1779,6 +1786,15 @@ public void CancelPlanetVote()
...
     private string GetRandomOperationName()
     {
+
+        if (_usingCustomOperationName && OperationName != null)
+        {
+            _usingCustomOperationName = false;
+
+            return OperationName;
+        }
+        ReloadPrototypes();
```
DrSmugleaf: Is this intended to be here? It will just re-parse valid operation names which doesn't seem necessary

DrSmugleaf (changes requested):

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
noctyrnal: i might have a disease and added this for 0 reason, removed it

Dutch-VanDerLinde:
Merge master to fix tests

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1781,6 +1788,14 @@ public void CancelPlanetVote()
 
     private string GetRandomOperationName()
     {
+
+        if (_usingCustomOperationName && OperationName != null)
```
DrSmugleaf:
```suggestion
        if (_usingCustomOperationName && OperationName != null)
```

Content.Server/_RMC14/Rules/SetOperationNameCommand.cs
```diff
@@ -0,0 +1,30 @@
...
+using Robust.Shared.Toolshed;
+using Robust.Shared.Utility;
+
+namespace Content.Server._RMC14.Rules;
+
+[ToolshedCommand, AdminCommand(AdminFlags.Round)]
+public sealed class SetOperationNameCommand : ToolshedCommand
+{
+
+    [CommandImplementation]
```
DrSmugleaf:
```suggestion
    [CommandImplementation]
```

DrSmugleaf (dismissed):
Thanks for the PR!

PR #6841, opened 2025-06-25
Fix green fire not reducing the armor of fortified defenders and the queen

Arendian:
## About the PR

Fortified defenders and the queen are now affected by the armor shred applied by green fire.
Entities that don't move when a green fire tile spawns below them now have their armor reduced instantly, instead of having to move first.

## Why / Balance

Parity.

## Technical details

The SteppingOnFIreComponent, which is responsible for the armor modifier, is now applied before checking if the entity has the FlammableComponent.

A nearby entity check is now done after the GetContactingEntities check to see if an entity is still standing on fire, this is because an entity with a static body type, for example an anchored fortified defender, doesn't have any contacts with fire tiles.

## Requirements

**Changelog**

:cl: Dygon
- fix: Green fire now reduces the armor of fortified defenders and the queen.


deepdarkdepths:
Does the armor reduction show up on the HUD?

Dutch-VanDerLinde:
> Does the armor reduction show up on the HUD?

Yes

Dutch-VanDerLinde:
Merge master to fix tests

DrSmugleaf (approved):

PR #6845, opened 2025-06-26
Savannah: Episode IV - Return of the Ship

Tunguso4ka:
## About the PR
- Adds Savannah back into default map pool.
- Ups maximum player count from 75 to 90.

## Why / Balance
In RMC-14 **`S`** stands not only for a Soul, but for Savannah too.

## Media
https://tunguso4ka.github.io/map-viewer/?map=Savannah

## Requirements

**Changelog**
:cl:
- add: LCS-14 Savannah returns as the lowpop ship map.


Tunguso4ka:
YAML Linter fail is not related to this PR.

Dutch-VanDerLinde:
Merge master to fix tests

PursuitInAshes:
Thanks for the PR!
Going to close this, Savannah now will be PVE only/for other events for us just due to pop being significantly higher than when this PR was made. We'll just never see it and even if we get that low; Its an extra "thing" to juggle/learn for new players which was a concern voiced to me during discussion.

Issue #4888, opened 2024-11-20
you shouldn't be able to build or plant weed nodes under window frames


Dutch-VanDerLinde:
if you jump on a window frame you can place weed nodes under them

PR #6854, opened 2025-06-27
Prevent resin structures being built on occupied tiles.

Arendian:
## About the PR

Resin structures can no longer be placed on tiles that are already occupied by another structure.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix: Resin structures can no longer be placed on tiles that are already occupied by another structure.


Dutch-VanDerLinde:
does this fix #4888?

Arendian:
> does this fix #4888?

~~It doesn't, but I can look into making it so it works for placing weeds too. It would need a different check because weeding is allowed in some places where constructing isn't.~~

It does now.

DrSmugleaf (approved):

PR #6858, opened 2025-06-28
Fix barbed wire build time

### LINKED ISSUE #6819: Barbed Wire crafting speed
## Description

In construction menu openable with Z while holding some metal in hand Wire takes around 1 second to make.
While in construction menu openable with G Wire takes 0 time to make, it's instant.
Seems one of these two has incorrect crafting time.

Dutch-VanDerLinde:
Fixes #6819 

Makes the construction graph use the RMC Prototype
:cl:
- fix: Fixed building barbed wire from the menu not having a do-after.


Arendian:
The Parity menu does have a x5, x10 and x20 option for barbed wires already that takes the same time as crafting 1, making it way easier to bulk craft with a DoAfter. Would this be easy to add to the SS14 build menu too?

Dutch-VanDerLinde:
> The Parity menu does have a x5, x10 and x20 option for barbed wires already that takes the same time as crafting 1, making it way easier to bulk craft with a DoAfter. Would this be easy to add to the SS14 build menu too?

I've already added this, press Z with metal in hand. Though this doesn't exist in the upstream construction menu because of UI complicated stuff

Arendian:
> I've already added this, press Z with metal in hand. Though this doesn't exist in the upstream construction menu because of UI complicated stuff

Yeah I saw it in the Z menu, if it's not easy(UI code my beloved) to add to the upstream menu it can probably be left out

DrSmugleaf (commented):
Might be worth leaving the do after out since it forces you to make it 1 by 1 compared to the other UI but I would have to think on it

PR #6860, opened 2025-06-29
Add MP vests to the MW and CMP vendors, remove CMP access to MP vendor.

BramvanZijp:
## About the PR
Added the regular MP vest to the CMP and MW vendors. (Includes a comment for a potential custom variant of the vest for these roles)
Removed the ability for the CMP to access the regular MP Vendors, given they have their own vendor now via the Senior Officer Vendor.

## Why / Balance
These roles, especially CMP (but also MW, esp. on lowpop) will, just like regular MPs, get into scenarios where they have to run around a lot, and having a vest they can wear that does not slow them down for use during code green on the ship makes sense. 
Also, CMP has had their own vendor for a while, them having access to the MP vendor no longer makes sense.

For the vests, I am willing to remove it from the warden vendor if you believe this should stay as their standard armor only to discourage the warden leaving the brig too frequently, though I think that they should have the option, especially for lowpop.

## Requirements

**Changelog**
:cl: BramvanZijp
- add: Added an MP Vest to the Military Warden and Chief MP's vendors.
- remove: The Chief MP can no longer access the MP equipment vendor.


Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/police.yml
```diff
@@ -109,6 +108,7 @@
       takeAll: CMArmor
       entries:
       - id: CMArmorM3Warden
+      - id: RMCArmorVestMP # TODO: Possible custom variant
```
QuietlyWhisper:
```suggestion
      - id: RMCArmorVestMP
```

Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
```diff
@@ -841,6 +841,7 @@
       takeAll: CMStandard
       entries:
       - id: CMArmorM3WO
+      - id: RMCArmorVestMP # TODO: Possible custom variant
```
QuietlyWhisper:
```suggestion
      - id: RMCArmorVestMP
```

QuietlyWhisper (changes requested):
This should be choose one and not take all

BramvanZijp:
> This should be choose one and not take all

Changes applied.

PR #6864, opened 2025-06-29
XM51 Stock gives Burst

### LINKED ISSUE #6861: XM51 burst mode
## Description
The XM51 stock should allow you to activate a burst mode, similar to the M4SPR custom burst. I believe it's a two shot burst

noctyrnal:
## About the PR

The XM51 stock now allows for a 2 shot burst. Fixes #6861 

## Why / Balance

Parity

## Technical details

`OnAttemptShoot` in `SharedPumpActionSystem` returns if `BurstActivated` is true in the gun component of the parent entity, allowing for multiple shots from a pump action shotgun in a burst without having to pump for each shot in said burst.
`Pump` in `SharedPumpActionSystem` returns true if `BurstActivated` is true too, same logic as `OnAttemptShoot` - This is done to prevent pumping mid burst.
Added the same logic as the other two mentioned functions to `OnAttemptShoot` in `PumpActionSystem` so the client wont get a popup telling them they need to pump to shoot when the gun is doing a burst shot.

## Media

https://github.com/user-attachments/assets/f6c39a11-c7f7-4e92-a498-e673817a19be

## Requirements

**Changelog**

:cl: noctyrnal
- add: The XM51 can now do a two shot burst when the stock is attached.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
@@ -74,7 +74,6 @@ public abstract class SharedPumpActionSystem : EntitySystem
     {
         if (TryComp<GunComponent>(ent.Owner, out var gun) && gun.BurstActivated)
         {
-            _popup.PopupClient($"BurstActivated: {gun.BurstActivated}", user, user);
             return true;
         }
 
--- a/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
@@ -3,6 +3,7 @@ using Content.Shared.Examine;
 using Content.Shared.Popups;
 using Content.Shared.Weapons.Ranged.Events;
 using Content.Shared.Weapons.Ranged.Systems;
+using Content.Shared.Weapons.Ranged.Components;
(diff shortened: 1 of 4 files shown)


BarteG44:
huge

Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
```diff
@@ -65,6 +72,12 @@ private void OnEntRemovedFromContainer(Entity<PumpActionComponent> ent, ref EntR
 
     public bool Pump(Entity<PumpActionComponent> ent, EntityUid user)
     {
+        if (TryComp<GunComponent>(ent.Owner, out var gun) && gun.BurstActivated)
+        {
+            _popup.PopupClient($"BurstActivated: {gun.BurstActivated}", user, user);
```
DrSmugleaf: I assume the popup was testing code?
```suggestion
        if (TryComp<GunComponent>(ent.Owner, out var gun) && gun.BurstActivated)
        {
```

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Weapons/Ranged/SharedPumpActionSystem.cs
noctyrnal: oops yes i forgot to remove this

DrSmugleaf (approved):

PR #6871, opened 2025-06-30
Lurkers can't tackle anymore during crippling strike

Arendian:
## About the PR

If a lurker tries to tackle while having crippling strike active it will now do a light attack instead.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix:  Lurker tackles now convert into a light attack when Crippling Strike is active.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Crippling/XenoCripplingStrikeSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Crippling/XenoCripplingStrikeSystem.cs
@@ -31,6 +31,7 @@ public sealed class XenoCripplingStrikeSystem : EntitySystem
         SubscribeLocalEvent<XenoCripplingStrikeComponent, XenoCripplingStrikeActionEvent>(OnXenoCripplingStrikeAction);
 
         SubscribeLocalEvent<XenoActiveCripplingStrikeComponent, MeleeHitEvent>(OnXenoCripplingStrikeHit);
+        SubscribeLocalEvent<XenoActiveCripplingStrikeComponent, MeleeAttackAttemptEvent>(OnActiveCripplingStrikeMeleeAttempt);
         SubscribeLocalEvent<XenoActiveCripplingStrikeComponent, RefreshMovementSpeedModifiersEvent>(OnActiveCripplingRefreshSpeed);
         SubscribeLocalEvent<XenoActiveCripplingStrikeComponent, ComponentRemove>(OnActiveCripplingRemove);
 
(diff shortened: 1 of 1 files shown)


pumpkin-pwr:
this should also apply to sentis paralyze, parity
also as a note to leave here this was made as QOL for 13 because of intents we dont have.

Arendian:
> this should also apply to sentis paralyze, parity also as a note to leave here this was made as QoL for 13 because of intents we dont have.

Since the restriction on Neuro Slash is a QoL change and it doesn’t reset the attack delay, along with RMC having a dedicated tackle button, I think it would be better to allow Sentinels to still use tackle while Neuro is active.
This is so they keep the option to not use it if they for any reason don't want to, since in RMC you can't really accidentally tackle instead of slash. I don't play a lot of sentinel though, so if forcing an attack on tackle for neuro slash would feel better it can be added too.

The limitation was initially added to crippling strike in CM because of it's attack delay reset being used to tackle, so for crippling strike it was not a QoL change.

pumpkin-pwr:
> > this should also apply to sentis paralyze, parity also as a note to leave here this was made as QoL for 13 because of intents we dont have.
> 
> Since the restriction on Neuro Slash is a QoL change and it doesn’t reset the attack delay, along with RMC having a dedicated tackle button, I think it would be better to allow Sentinels to still use tackle while Neuro is active. This is so they keep the option to not use it if they for any reason don't want to, since in RMC you can't really accidentally tackle instead of slash. I don't play a lot of sentinel though, so if forcing an attack on tackle for neuro slash would feel better it can be added too.
> 
> The limitation was initially added to crippling strike in CM because of it's attack delay reset being used to tackle, so for crippling strike it was not a QoL change.

no theyre both QoL changes made to Both abilties, your reasoning is the exact reason it shouldnt really apply to lurker but its parity so if they want it than it is what it is

Arendian:
> no theyre both QoL changes made to Both abilties, your reasoning is the exact reason it shouldnt really apply to lurker but its parity so if they want it than it is what it is

If you look at the PR that added the forced harm on crippling strike to CM you'll see that is has the "Balance" tag, it resetting the tackle delay is specifically mentioned as the reason for it being changed. The sentinel's neuro slash PR has the "Quality of Life" tag and specifically mentions it being a QoL change.

I am not against adding forced harm to neuro slash at all, I just want the neuro slash to behave in a way that feels best, since it doesn't affect balance at all if it's forced or not, if that is adding forced harm to it then I'll gladly add it.

pumpkin-pwr:
ooh you right, thought this was a super old change before they used git, https://github.com/cmss13-devs/cmss13/pull/3608 pr link for if anyone reads this

PR #6873, opened 2025-06-30
Fix some xeno abilities having too much range

Arendian:
## About the PR

I adjusted the range of some xeno abilities.

## Why / Balance

Parity.
Abilities that have their range reduced, the number is the range in tiles:

- Defender's Headbutt: 3 > 2
- Burrower's Tremor: 4 > 3
- Spitter's Charged Spit: 6 > 4
- Queen's Acid(not neuro) Spit: 6 > 5
- Sentinel's Slowing Spit: 6 > 5
- Praetorian's Acid Spit: 7 > 6

## Requirements

**Changelog**

:cl: Dygon
- fix: Reduced the range of the Defender's Headbutt, Burrower's Tremor, Queen's Acid Spit, Sentinel's Slowing Spit and Praetorian's Acid spit by 1 tile. Reduced the range of the Spitter's Charged Spit by 2 tiles.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Headbutt/XenoHeadbuttComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Headbutt/XenoHeadbuttComponent.cs
@@ -18,7 +18,7 @@ public sealed partial class XenoHeadbuttComponent : Component
     public DamageSpecifier Damage = new();
 
     [DataField, AutoNetworkedField]
-    public float Range = 3;
+    public float Range = 2;
 
     [DataField, AutoNetworkedField]
     public float ThrowForce = 1;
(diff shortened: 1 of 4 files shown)


Vermidia:
burrower's tremor range was changed forever ago to be based on AOE area already, so it's previous number should be correct. (4 = about 3.9 if you do the math)

Vermidia:
The range of defender is 3
```
var/max_distance = 3 - (fendy.crest_defense * 2)
```
crest def is already handled in there

I think the spits were increased by 1 for a reason but I forget

What I do know is spitter charged spit and Queen's charged should be the same

Arendian:
> burrower's tremor range was changed forever ago to be based on AOE area already, so it's previous number should be correct. (4 = about 3.9 if you do the math)

image
image

Tremor does not stun in the second image, but does in the first. Or do you mean that we made the range non-parity on purpose?

Vermidia:
yes
we did because RMC uses circles for range and not tiles, so 3.9... radius = the area of a 7x7 square

Vermidia:
So I can confirm on CM:
Headbutt is 3 range
sentinel spit is 6
charged spit is 5 (both)
prae is 7

Arendian:
> So I can confirm on CM: Headbutt is 3 range sentinel spit is 6 charged spit is 5 (both) prae is 7

I already see what happened with the projectile ranges, while looking through the CM files I compared the range of x projectile in CM13 with the range of y ability in RMC14 for some reason(I probably just need to go to sleep). Thanks for the quick reply!

DrSmugleaf (approved):

PR #6891, opened 2025-07-02
Add a short cooldown to Dancer abilities and Warrior's Lunge when they are used on an invalid target

### LINKED ISSUE #3056: Warrior Lunge doesn't cooldown on miss
## Description

Warrior's Lunge ability is intended to be balanced around being able to miss, and getting punished with a cooldown for clicking wildly. This should last for 1.5s but does not currently work

The problem codewise seems to be that abilities only use ActionComponents that target either Entities or WorldCoordinates. Fixing this will need a third type that can target either, and have its Event include nullable args for both a target EntityUid and Coordinates, as well as the component itself including an alternate UseDelay that can be set.

The way projectiles work might be helpful for this, since they can aim anywhere but have altered behavior when targeting an entity.

**Reproduction**

**Screenshots**

**Additional context**

Arendian:
## About the PR

Warrior's lunge, Dancer's Impale and Dancer's Tail Trip now have a 1.5-second cooldown if used on an invalid target. Nothing also counts as an invalid target.

## Why / Balance

Resolves #3056
Now that we have collision, clicking the targets sprite is not an issue.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Warrior's Lunge, Dancer's Impale and Dancer's Tail Trip now have a 1.5-second cooldown if used on an invalid target.


Dutch-VanDerLinde:
I believe vampire lurker tail jab is also supposed to have this cooldown like tail stabs

Though tail jab works more like a CM13 tail stab so I think tail stab should use the same system as tail jab

0tito:
This is ✨horrible✨ in my opinion, first of all 1.5 is too much, also mispredicts will probably make this painful, also, sprite clicking is still an issue, even with collision,because xenos are too fat and their sprites take way more space than what they should. But is it parity tho, because if it is i guess we well just have to deal with it

0tito:
Ok i just read the issue, yeah if it is parity i guess its fine, I still agree with this tho

>While this is a thing in CM and also effects dancer etc, dunno if we should have it just due to sprite clicking being harder than tile clicking + sprite blocking etc.

Arendian:
> Ok i just read the issue, yeah if it is parity i guess its fine, I still agree with this tho

These abilities don't work on tile click in CM13 from my testing, you need to sprite click there too, the sprite overlap was mainly an issue when we didn't have collision yet because of people rendering under xenos. Altough our sprites are a bit bigger than in CM13, it's impossible for your target to be below it since humans render over xenos.

PolarTundraa:
i'm mainly concerned as to how this will interact with lag as it's a decently common occurence to (particularly with fast-moving targets) click on them with no effect

DrSmugleaf:
> first of all 1.5 is too much,

1.5 is how long it takes in 13

>also mispredicts will probably make this painful

This makes no sense, mispredicts cannot "make this painful", because in a mispredict you would not miss on the server, so there would be no cooldown, hence the MIS in MISpredict

@0tito I have spent an ungodly amount of hours correcting people on their plainly incorrect takes about prediction and lag compensation on Discord and in-game, please make sure you are not contributing to the mind virus that completely misunderstands what prediction is or does on GitHub

DrSmugleaf (dismissed):

PR #6892, opened 2025-07-03
make eviscerate doAfter not cancel

Binbagg:
## About the PR
BreakOnMove = true

## Why / Balance
fix it being cancelled by moving rav

## Requirements

**Changelog**

:cl:
- fix: Eviscerate do-after being cancelled if you get moved.


Vermidia:
Break on move breaks if you move too, it does both.

Binbagg:
Gg

Dutch-VanDerLinde:
> Break on move breaks if you move too, it does both.

Well no eviscerate sets your move speed to 0 so you can't  move

Vermidia:
technically yes but it's prob just to have the rooted thing, as this can mean a xeno can pull a rav who it's eviserating around

PR #6901, opened 2025-07-04
Add paradropping from the dropship using parachutes

Arendian:
## About the PR

If you jump out of a dropship with a parachute equipped you will safely land on the ground, if the pilot has a target selected you will drop near it.
The dropship part fabricator can now print the HPU-1 paradrop deployment system for 50 points.
The requisitions vendor now vends parachutes.

Depending on the situation, one of these 4 things will now happen if you jump out of a dropship:

1. A target is selected and you are wearing a parachute: You paradrop near the targeted location.
2. A target is selected and you are not wearing a parachute: You crash land near the targeted location.
3. No target is selected and you are wearing a parachute: You will paradrop at a random location.
4. No target is selected and you are not wearing a parachute: The same behavior as before this PR, you crash land at a random location.

~~There are still some small issues, most of them being visual ones, that need to be fixed before this should be merged:~~
- [x] Fix parachutes not being visible to anyone that is outside of PVS range of the paradropping entity's  initial sprite offset location.
- [x] Fix the job icon not having the same offset as the sprite of a paradropping entity for some reason.
- [x] ~~Render a paradropping entities sprite outside of FOV.~~
If the upper part of your screen is outside your FOV because your vision is blocked, you will only start seeing the sprite once it reaches an area that is inside of your FOV. I assumed it wasn't supposed to be like that, but I checked and this is parity. 
- [x] Only open the back doors instead of all doors.(I made the doors, but they need to be mapped on the dropships which I can't do)
- [x] Add more info to the UI of the deployment system

## Why / Balance

Parity.

## Technical details
 
~~I assumed that the dropships use RMCCrashLandBarrier entities that are part of it's grid to crash land people, if this is not the case some changes are required.~~ It now also works when an entity gets reparented to an FTLMap. The Dropships will need the doors with specified location(Aft, Port, Starboard) mapped, in the future these can also be used for locking only a specific set of doors instead of all of them at once.

If a paradrop target is selected using the dropship weapon controls the grid will get the ActiveParaDropComponent and all doors with the Aft location will be bolted open. If this component is removed, either by the target disappearing , the ship starting it's landing procedures, or the pilot cancelling the targeting, the doors will be bolted close again.

Before an entity crashes to the planet, an event is raised that can be cancelled, if the event is cancelled the crash land won't happen. The ParaDropSystem listens to this raised event, and will cancel it if the grid the entity is leaving either has the earlier mentioned ActiveParaDropComponent, or if the falling entity has the ParaDroppableComponent. The ParaDroppableComponent is granted when an entity wears a parachute on their back, crates have this component inherently.

The ActiveParaDropComponent decides if a fall will be at a specific location, the ParaDroppableComponent decides if the entity will paradrop instead of crash land.

When an entity is paradropped, it's teleported to it's new location with an offset sprite. The offset will return to 0 during the paradrop animation, while the animation is happening the entity can't move, has no collision, can't be attacked, can't be ignited, can't be knocked back and can't take any damage. The actual location of the entity during the animation is the tile it "lands" on when the animation is finished, so any throw items or shot bullets will come from that location even if the sprite appears to be somewhere else.

## Media

https://github.com/user-attachments/assets/85943235-243a-4fc2-b239-e34f7ba56457

https://github.com/user-attachments/assets/79014ea9-fc96-4379-aa70-c288c5d3931d

## Requirements

**Changelog**

:cl: Dygon
- add:  Dropship fabricators can now print the "HPU-1 paradrop deployment system", installing this module in a dropship allows it to open it's doors mid-flight after selecting a viable paradrop target. Players that walk out of a dropship that is mid-flight while a target is selected will land near it.
- add:  Throwing a crate out of the dropship will now make it land near the location targeted by the pilot.
- add: Requisitions vendors now stock parachutes, having a parachute equipped will prevent you from falling to your death if you walk off a dropship mid-flight. If no target is selected by the pilot when you fall out you will paradrop to a random location.
- fix: When falling out of the dropship you can't land at underground areas anymore.


XxSWAG-MASTER-OVERLORDxX:
holy soul

CatoChristopher:
Please speedmerge this I need to watch privates who forgot their parachutes land dead infront of me

DrSmugleaf:
> I assumed that the dropships use RMCCrashLandBarrier entities that are part of it's grid to crash land people, if this is not the case some changes are required. The Dropships will need the doors with specified location(Aft, Port, Starboard) mapped, in the future these can also be used for locking only a specific set of doors instead of all of them at once.

@Arendian the dropships don't have any barriers beyond the door blockers for cades and such
The code just checks if you are directly on a map with ftl component, in CrashLandableSystem.cs line 52

Arendian:
> @Arendian the dropships don't have any barriers beyond the door blockers for cades and such The code just checks if you are directly on a map with ftl component, in CrashLandableSystem.cs line 52

Thanks for the clarification, it should now also work using the way you mentioned. The PR should be ready for review now.

Arendian:
I replaced the old video and wrote a summary of the technical details

noctyrnal:
for the paradrop fail animation, you should consider speeding it up rapidly.

Arendian:
> for the paradrop fail animation, you should consider speeding it up rapidly.

It currently takes 1.5 seconds, how short do you think it should be?

noctyrnal:
> > for the paradrop fail animation, you should consider speeding it up rapidly.
> 
> It currently takes 1.5 seconds, how short do you think it should be?

https://github.com/user-attachments/assets/97ec4064-af79-4285-98f9-4e0479949b33
you can in this clip here how fast it is

Arendian:
> https://github.com/user-attachments/assets/97ec4064-af79-4285-98f9-4e0479949b33 you can in this clip here how fast it is



https://github.com/user-attachments/assets/347b46de-1f82-4054-bce8-f2dc81f14792

This is the new speed, it looks about the same speed as that clip I think.

noctyrnal:
> > https://github.com/user-attachments/assets/97ec4064-af79-4285-98f9-4e0479949b33 you can in this clip here how fast it is
> 
>  FasterCrash.mp4 
> This looks about the same speed as that clip

beautiful shit

Dutch-VanDerLinde:
crashlanding should make you unrevivable, pretty sure its like this in cm13

Arendian:
> crashlanding should make you unrevivable, pretty sure its like this in cm13

I think it currently does 10k damage since I didn't change anything about it, making someone completely unrevivable is easy to add though.

Kadeo64:
> I think it currently does 10k damage since I didn't change anything about it, making someone completely unrevivable is easy to add though.

don't take this as 100% fact but iirc smugleaf specifically made it so that supply drops falling on someone don't gib them because having them be revivable is.. funny? or something. may or may not apply here but you'd have to ask smugleaf.

Dutch-VanDerLinde:
> > I think it currently does 10k damage since I didn't change anything about it, making someone completely unrevivable is easy to add though.
> 
> don't take this as 100% fact but iirc smugleaf specifically made it so that supply drops falling on someone don't gib them because having them be revivable is.. funny? or something. may or may not apply here but you'd have to ask smugleaf.

Supply drops don't gib you yet because we don't have the crate falling animation
Smug said she'd consider adding the gibbing after we get this animation

Arendian:
> Supply drops don't gib you yet because we don't have the crate falling animation Smug said she'd consider adding the gibbing after we get this animation

We now have one, so crate gib might be closer than we think.....
I do think that a huge amount of damage is more fun as it gives the doctors more stuff to do. If we get newmed for example we can just fracture every single limb and add as many injuries as possible , making it a challenge for a doctor to heal them in time. This does sound more engaging than just having another corpse added to the pile.

Parity is almighty though, so adding the permadeath is probably the thing to go for until Smug says otherwise, or we reach post-parity.

FallingDownofMadness:
After looking, I can tell that the landing animation, when a Marine jumps out of a shuttle into space and before the animation of landing on the planet begins, is faster than in the video PR

https://github.com/user-attachments/assets/27ac89bf-e806-4816-a80b-add4b448e016

Content.Client/_RMC14/ParaDrop/ParaDropSystem.cs
```diff
@@ -0,0 +1,157 @@
...
+
+    private void OnComponentRemove(Entity<SkyFallingComponent> ent, ref ComponentRemove args)
+    {
+        if (!TryComp<SpriteComponent>(ent, out var sprite) ||
+            TerminatingOrDeleted(ent))
+        {
+            return;
+        }
+
+        sprite.Scale = ent.Comp.OriginalScale;
```
DrSmugleaf, line 127: Need to stop the animation if the component is removed as well

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Client/_RMC14/ParaDrop/ParaDropSystem.cs
Arendian: The animation now stops before the sprite scale is reset

DrSmugleaf (approved):
Thanks for the PR!

PR #6935, opened 2025-07-06
Allows the smartscope to fit on the M54C Mk2. (parity)

Spaghetti-bit:
## About the PR

Allows the smart scope to be attached to the Mk2, as it is in CM13

## Why / Balance

Parity

## Technical details

Added the `RMCAttachmentB8SmartScope` to the `WeaponRifleM54C` rail attachment whitelist

## Media

Examples of parity: 
image
image

## Requirements

**Changelog**

:cl:
- tweak: Allows the B8 smart scope to attach to the M54C Mk2 rail.


noctyrnal:
this is because iff got changed on cm13, so it was weak enough to fit onto the m4a1. If we allowed the smartscope to go on the m54c without the iff changes then it would be an insane unintended buff.

Spaghetti-bit:
Would you like to see the alternative IFF added to prevent it from firing? I'll close this and add that system here instead.

noctyrnal:
> Would you like to see the alternative IFF added to prevent it from firing? I'll close this and add that system here instead.

yeah ideally you should port them together, would be nice to see o7

DrSmugleaf:
> Would you like to see the alternative IFF added to prevent it from firing? I'll close this and add that system here instead.

We would need alt IFF for this yes, problem is that alt IFF doesn't work very well with latency since you might shoot on the screen but not the server, or the other way around, so we haven't changed it yet because of that and because IFF here hasn't been as much of a problem as in 13

Spaghetti-bit:
> > Would you like to see the alternative IFF added to prevent it from firing? I'll close this and add that system here instead.
> 
> We would need alt IFF for this yes, problem is that alt IFF doesn't work very well with latency since you might shoot on the screen but not the server, or the other way around, so we haven't changed it yet because of that and because IFF here hasn't been as much of a problem as in 13

I've implemented a rudimentary version of the alt IFF. It isn't fool proof but it uses a ray vector before the gun shoots to prevent it just like it does in CM13. I'll refine it tomorrow.
image

PursuitInAshes:
Thanks for the PR!
I'm going to opt to close this for now due to this requiring Alternate IFF, a system which we don't have desire for due to both technical concerns (Latency, as an example) and gameplay concerns (Caused by our higher population than CM13).

PR #6944, opened 2025-07-06
Fix xenos not being able to add plasma to hive structures

Dutch-VanDerLinde:
## About the PR
In CM13 any xeno with a plasma bar can repair structures, not just builders

## Media

:cl:
- fix: Fix some xenos not being able to add plasma to hive structures


FeelFlows:
While they can add plasma to build structures, I'm pretty sure they can't repair ones that are already built, like damaged clusters.

PR #6967, opened 2025-07-08
Escape pod fixes

Dutch-VanDerLinde:
## About the PR
Makes the escape pod computer actually explode over the 3 person limit
Fixes the escape pod doors being locked shut when it crashes, in CM13 it just forces it unlocked

## Technical details
image
Trigger explosive is called, but the computer doesn't have an explosive component.

## CM13 Code
image
image
image

:cl:
- fix: Fixed escape pods not exploding when it's launched over the maximum occupancy, and not reopening the door.


DrSmugleaf (approved):

DrSmugleaf:
Need to test if this correctly spreads the explosion but this is fine for now

PR #6976, opened 2025-07-09
Actually fix escape pods

Dutch-VanDerLinde:
## About the PR
Makes it queue an explosion on a valid unblocked tile
currently it'll explode inside a wall

changes the door incorrectly referencing the escape pod computer instead of the grid's child


Dutch-VanDerLinde:
ready for review

PR #6980, opened 2025-07-09
Aegis event improvements & new version

DoguhanOzgurAkca:
## About the PR
Improves various parts of the AEGIS event and adds a new 'mode' to it. Experimenting continues.

To make review easier, I am just gonna list the changes properly. Smug already has too much work.

1) There is a new way to run the AEGIS event. Straight from the lobby. How it works is that you un aegis:lobby command, an it makes it so that  aegis box and corpse spawns (randomly chosen from ones that you have) spawn, and then after a minute, you also get the announcement and ASRS items. This version also makes it so that the aegis box is closer to FOB, but you have to find the aegis ID card with a pinpointer.
2) There are 3 more aegis Command with better descriptions
Aegis:normal which is the normal version of aegis
Aegis:status which shows if AEGIS event is scheduled to happen in the next round
Aegis:reset which cancels AEGIS event from happening
2) Pinpointer added with new modified sprites by sigma. It is currently only used for the aegis key card and nothing else.
3) Firing the AEGIS OB now gives you 5 intel points. Overall, this is to incentivize people using it more, and also make it have more of a positive to marines since it was hard so far.

## Why / Balance

Experimenting with how it all works, especially menu run commands. AEGIS has been a rathe bad event because of how hard it is to use for marines. This also experiments with how to have a round start objective when we don't know which LZ marines will land at.

Hopefully a step to putting aegis into the game rotation without admin intervention as well.

## Media

These are the placements of aegis crate and scientists corpses that spawn in the map. Blue circles are crate spawns, and the pink ones are scientist corpse spawns. One is selected every round in random. It would be amazing if it could be based on the LZ but I couldn't think of a way to make that possible without somehow forcing the CiC to pick a LZ in the first minute of the game or something.

Balance wise, since marines have to find the keycard as well, I wanted to keep the AEGIS crate one big step away from FOB. This will cause issues on a few maps, I am sure. 

download (3) (2)
We got new pinpointer sprites by the glorious @SigmaTheDragon  idk what I would do without em.

https://github.com/user-attachments/assets/51586a3c-1790-4ef8-80f9-82aaa3e26bb9
This isn't the final sprites, pointer arrows are different, but I didn't record em.

## Requirements

**Changelog**


DoguhanOzgurAkca:
i swear to fucking god this is the fourth solaris merge conflict I AM GOING INSANE

PR #6987, opened 2025-07-10
Adds variety mortar crate for requisitions

Stop-Signs:
## About the PR
Adds the option to purchase a crate with 2 of each kind of mortar shell.

## Why / Balance
Convenience really, most mortar people appreciate having more of all without forcing req to spend 4k. Technically you lose 167.7777777777777 dollars each time you purchase one due to the flares costing less per unit than the other two, which is a reasonable price for convenienve

## Technical details
This is my first time contributing to the RMC codebase, if there are any conventions I failed to follow or anything I left out please let me know so I can improve in future PR's

## Requirements

**Changelog**

:cl:
- add: Added a mixed mortar shells crate that requisitions can order.

### DIFF
--- a/Resources/Prototypes/_RMC14/Catalog/Cargo/requisitions_catalog.yml
+++ b/Resources/Prototypes/_RMC14/Catalog/Cargo/requisitions_catalog.yml
@@ -251,6 +251,8 @@
         crate: RMCCrateMortarShellIncendiary
       - cost: 1000
         crate: RMCCrateMortarShellFlare
+      - cost: 1500
+        crate: RMCCrateMortarShellMixed
 #    - name: Operations
 #      entries: []
 #    - name: Reagent tanks
--- a/Resources/Prototypes/_RMC14/Catalog/Fills/Crates/mortar.yml
+++ b/Resources/Prototypes/_RMC14/Catalog/Fills/Crates/mortar.yml
@@ -37,3 +37,17 @@
     contents:
     - id: RMCMortarShellFlare
       amount: 6
+
+- type: entity
+  parent: RMCCrateMortarAmmo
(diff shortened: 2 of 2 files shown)


DoguhanOzgurAkca:
afaik, not parity
and also you never really want a mixed crate shell as a mortar operator. It just makes little sense. I don't want any incendiary shells basically, and I am only going to use like a crate or two of cam shells for most operations but I can use up to 4-6 crates of HE.
Also I don't know if its going to happen or not soon but 13 has another shell type, so you won't be able to fit 6 in a box when there is 4 types.

FeelFlows:
> afaik, not parity and also you never really want a mixed crate shell as a mortar operator. It just makes little sense. I don't want any incendiary shells basically, and I am only going to use like a crate or two of cam shells for most operations but I can use up to 4-6 crates of HE. Also I don't know if its going to happen or not soon but 13 has another shell type, so you won't be able to fit 6 in a box when there is 4 types.

Have you considered that you're not the only mortar operator, and that some do want *some* amount of incend shells?

DoguhanOzgurAkca:
> > afaik, not parity and also you never really want a mixed crate shell as a mortar operator. It just makes little sense. I don't want any incendiary shells basically, and I am only going to use like a crate or two of cam shells for most operations but I can use up to 4-6 crates of HE. Also I don't know if its going to happen or not soon but 13 has another shell type, so you won't be able to fit 6 in a box when there is 4 types.
> 
> Have you considered that you're not the only mortar operator, and that some do want _some_ amount of incend shells?

I mean its not just about incendiary shells. If you do want them, you can just order as well.
We had a talk about this in the discord, and again, I have never met any mortar operator who would use this outside of edge cases.

PR #6988, opened 2025-07-10
root doafter

Vixting:
## About the PR

Adds ability to root entities performing doafters
Please tell me which other doafters should be rooted as this is not all of them

## Why / Balance

qol

## Media

https://github.com/user-attachments/assets/12828eeb-4493-4747-b6bd-52a2cefe93bf

## Requirements

**Changelog**

:cl:
add: Certain doafters will now root entities in place

### DIFF
--- a/Content.Server/Medical/DefibrillatorSystem.cs
+++ b/Content.Server/Medical/DefibrillatorSystem.cs
@@ -184,6 +184,7 @@ public sealed class DefibrillatorSystem : EntitySystem
             DuplicateCondition = DuplicateConditions.SameEvent,
             TargetEffect = "RMCEffectHealBusy",
             MovementThreshold = 0.5f,
+            RootEntity = true
         });
     }
 
--- a/Content.Shared/DoAfter/DoAfterArgs.cs
+++ b/Content.Shared/DoAfter/DoAfterArgs.cs
@@ -8,6 +8,10 @@ namespace Content.Shared.DoAfter;
 [DataDefinition]
 public sealed partial class DoAfterArgs
 {
+    //RMC
+    [DataField]
+    public bool RootEntity;
+
     /// <summary>
     ///     The entity invoking do_after
     /// </summary>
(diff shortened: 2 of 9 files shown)


Arendian:
It should also make it so people can't walk through you at all, in the current iteration it looks like you can still force yourself through the rooted entity.

Vixting:
> It should also make it so people can't walk through you at all, in the current iteration it looks like you can still force yourself through the rooted entity.

I don't know how it works in CM, but it does block.


https://github.com/user-attachments/assets/f8924ff3-f332-4bcc-9dad-fd168619b5e0

Arendian:
> I don't know how it works in CM, but it does block.

In CM it makes it so people can't pass through, in your PR you can still walk through entities with only a tiny slowdown.


https://github.com/user-attachments/assets/8793d194-c101-4634-bdde-020719928895

Vixting:
Isn't this just how collision works in SS14 with circular bounding boxes? Like I could make it like CM but I feel like it would just be worse, this feels better to play with. I guess I could just make it repel you more / look at how it handles walls collisions. (I'm not that familar with how its handled tbh, this just the simplest way to achieve it without modifying upstream code very much)

Current collision stuff is designed so you can slip past other mobs if you have a low velocity & min penetration depth. 

I'm not interested in doing so, so up to Smug if she wants to close this / someone else wants to work on it. It probably wouldn't be to hard to do, but upstream code would need to be modified more or it would probably be jank. But maybe there is a simple way to fix this I'm not seeing.



> > I don't know how it works in CM, but it does block.
> 
> In CM it makes it so people can't pass you, in your PR you can still walk through entities with only a tiny slowdown if your collision area isn't too big.
> DoAfterBlock2.mp4

DrSmugleaf (approved):
Yeah this is fine, to address the collision issue we need to make the hitbox for mob collision bigger but that is gonna take some work to make it work properly

PR #7008, opened 2025-07-12
Fix most known master issues

noctyrnal:
## About the PR

Fixes most known issues caused by upstream merge.
- Gave plating friction 1 since they got a new friction field that was set to 0.2, causing you to slip around on plating.
- Fixed toggle visuals on lights and toggle lighting, since ToggleableLightVisuals got split into ItemTogglePointLight and ToggleableVisuals
- Fixed missing flame state inhands for candles (not caused by upstream but just a bug i found)
- Fixed tool pouches when StorageFilled with items that had a Storedrotation other than 0 causing massive duplication of slots, entities in the slots to not be visible and not being able to insert into the tool pouch.
- Fixes disposal units being invisible.
- Unrelated to upstream merge but fixes wrong loadout for liaison / weya goon zombie spawner, and fixes wrong stairs on sorokyne.
There might be a few more issues so expect an update or follow up to this PR.

## Why / Balance

bugs bad

## Technical details

yaml sprite, no c# thankfully.

## Requirements

**Changelog**

no cl since wont be player facing (players arent going to experience these bugs before this fix is on the server)


AverageNotDoingAnythingEnjoyer:
Now let's see Cro's business card (test results)

Resources/Prototypes/Tiles/plating.yml
```diff
@@ -6,7 +6,7 @@
   isSubfloor: true
   footstepSounds:
     collection: FootstepPlating
-  friction: 0.2
+  friction: 1
```
AverageNotDoingAnythingEnjoyer, line 9: The sin of editing upstream files....

AverageNotDoingAnythingEnjoyer: Actually, don't we have all CMPlating inherit from CMPlatingBase or something?
If it's affecting all tiles, just edit this one or make it

Tunguso4ka: i allow it

noctyrnal: > Actually, don't we have all CMPlating inherit from CMPlatingBase or something? If it's affecting all tiles, just edit this one or make it

this has already been edited before, alot of our code uses upstream plating instead so even if i make a new plating i'd need to update like 500 instances of plating being referenced too.

noctyrnal:
i have no idea what demonic shit is going on with test fails but even upstream stuff is borked

AverageNotDoingAnythingEnjoyer:
> i have no idea what demonic shit is going on with test fails but even upstream stuff is borked

Heisentests, mayhaps

noctyrnal:
draft to fix disposal units

Resources/Prototypes/Tiles/plating.yml
DrSmugleaf, line 9: This was last edited by @Rainbeon to be 0.2 in her pulling speed fix PRs, are you sure this is correct? It might be something else at play unless friction now works differently

DrSmugleaf (changes requested):

Resources/Prototypes/Tiles/plating.yml
DrSmugleaf: Nvm it got changed in https://github.com/space-wizards/space-station-14/pull/36847

DrSmugleaf (approved):
ty boss

PR #7026, opened 2025-07-14
Allow multiple jobs to use the same playtime tracker

Dutch-VanDerLinde:
## About the PR
Makes it so playtime is shared between survivor variants, basically
So you dont need 500 playtime trackers for variants, removing a SHITTON of copypaste

FORECON (Eventually, SPP SOF will use this tracker also) now get a Military Survivor playtime tracker.

Removes the map survivor job groups since they're redundant now

Ideally, this should be merged before the update, so players don't get playtimes on the maps with survivor variants

## Technical details
Getting a job from a playtime tracker now gets the FIRST available job, since most trackers will only be used once.
As a result, playtime trackers now have an optional 'name' and 'department' field.

If a playtime tracker doesnt have a name it'll use the job's name

UPSTREAM PR: https://github.com/space-wizards/space-station-14/pull/38977

## Possible issue
Since the original FORECON playtime trackers are removed, people who have FORECON playtime will have that reset
I believe this is ok, as FORECON is a decently rare role so people shouldnt have that much playtime on it

## Media
Playtime medal as an NV survivor variant
image

Doctor survivor and scientist survivor trackers, shared through all variants
image

## Requirements

**Changelog**
not player facing


yagwog:
Has the option ever been explored to also have a separate tracker;

Let's say, one totaling xeno, then marines, then survivors.

I'm mainly considering xenonids as it's very hard to tell when one unlocks the next pre/postfix for xenos.

Dutch-VanDerLinde:
> Has the option ever been explored to also have a separate tracker;
> 
> 
> 
> Let's say, one totaling xeno, then marines, then survivors.
> 
> 
> 
> I'm mainly considering xenonids as it's very hard to tell when one unlocks the next pre/postfix for xenos.

This wouldn't be too hard to make

Dutch-VanDerLinde:
conflicts fixed

Dutch-VanDerLinde:
ahhh 2000 conflicts

Dutch-VanDerLinde:
Pls add, people are getting survivor playtime on the variants :(

Resources/Prototypes/_RMC14/Roles/Jobs/Survivor/Sorokyne_Strata/liaison.yml
```diff
@@ -3,7 +3,9 @@
   id: CMJobSurvivorHalcyonDynamicsLiaison
   name: rmc-job-name-survivor-halcyon-dynamics-liaison
   description: cm-job-description-survivor
-  playTimeTracker: CMJobSurvivorHalcyonDynamicsLiaison
+  playTimeTracker: CMJobSurvivorCorporate
+  ranks:
+    RMCRankWeYaExecutive: []
```
Dutch-VanDerLinde:
```suggestion
```

Dutch-VanDerLinde:
added trijent survs

NotInButIn:
Oh it would be better if you could make trackers transfer their time to some existing trackers, because it is a shame to lose time

DrSmugleaf (dismissed):

PR #7038, opened 2025-07-15
Disables being able to place glass tiles

noctyrnal:
## About the PR

Makes CMSheetGlassBase no longer parent from SheetGlassBase and instead BaseItem with the needed components added. 

## Why / Balance

Prevents players from being able to place glass floors which were immersion breaking (why can i see space through the floor of the 2nd deck of the Almayer). Upstream tiles are also bad.

**Changelog**

:cl: noctyrnal
- fix: Glass sheets no longer let you place upstream glass tiles.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Materials/Sheets/glass.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Materials/Sheets/glass.yml
@@ -16,10 +16,41 @@
   maxCount: 50
 
 - type: entity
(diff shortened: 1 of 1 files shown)


BarteG44:
alas, no more "home design is my passion" aah alamo renovations

BadaBoomie:
> alas, no more "home design is my passion" aah alamo renovations

This feels targeted

DrSmugleaf (approved):

PR #7053, opened 2025-07-16
Xeno announcements for queen death and hive core destruction cooldowns

path-to-aaru:
## About the PR

Adds support for scheduled xeno announcements, including a message when the queen death cooldown has ended as well as when a new hive core can be built. To support this cleanly, the existing xeno caste unlock announcements were folded into the same timed scheduling system, avoiding redundant polling logic in the update method. Future time-based announcements can now be easily added.

## Why / Balance
The 5-minute cooldown after a queen dies is only visible/checkable by attempting to evolve into a queen. Same goes for the 5 minute cooldown on making a new hive core after the previous one is destroyed, you can only check it by trying to place a hive core. This adds new announcements for both via a simple announcement scheduler. Xeno caste unlock announcements were moved to the same scheduler.

## Technical details
New local events for queen death, hive core destruction, and caste unlock. These events are raised from SharedXenoHiveSystem and XenoPylonSystem. XenoHiveSystem subscribes to these events and handles them server-side by setting the appropriate announcement to be scheduled.

## Media
msgs

## Requirements

**Changelog**
- add: Queen death cooldown announcement
- add: Hive core destruction cooldown announcement
- tweak: Moved existing caste unlock announcements to use the same time scheduling system

### DIFF
--- a/Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
@@ -51,12 +51,14 @@ public sealed class XenoHiveSystem : SharedXenoHiveSystem
         public TimeSpan AnnounceAt;
         public AnnouncementType Type;
         public string MessageKey;
+        public List<EntProtoId>? Unlocks;
     }
 
     private enum AnnouncementType
     {
         QueenCanEvolve,
-        CoreBuildable
+        CoreBuildable,
+        EvolutionUnlock
     }
 
     public override void Initialize()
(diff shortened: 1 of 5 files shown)


Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
```diff
@@ -38,16 +38,39 @@ public sealed class XenoHiveSystem : SharedXenoHiveSystem
 
     private readonly List<string> _announce = [];
     private readonly EntProtoId _defaultHive = "CMXenoHive";
+    private readonly Dictionary<EntityUid, List<TimedAnnouncement>> _pendingTimedAnnouncements = new();
```
DrSmugleaf, line 41: This should be tracked by adding a component to these entities or spawning new entities, not by keeping track of them in a system's dictionary like this
Otherwise you have to manually keep it in sync if the entity is deleted, the round restarts, etc

DrSmugleaf (approved):
Thanks for the PR!

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
DrSmugleaf, line 41: This should be tracked by adding a component to these entities or spawning new entities, not by keeping track of them in a system's dictionary like this
Otherwise you have to manually keep it in sync if the entity is deleted, the round restarts, etc

DrSmugleaf (changes requested):

Content.Server/_RMC14/Xenonids/Hive/XenoHiveSystem.cs
path-to-aaru: I'll re-write it, thank you. Would it make sense to track them via the hive component?

DrSmugleaf: Could do that yeah

PR #7062, opened 2025-07-17
Make 175 hour DCC rank Warrant Officer

Dutch-VanDerLinde:
## About the PR
This seems like a mistake, especially considering it shared the pilot playtime tracker before (but then was fixed)
Currently, a 175 hour DCC can outrank a <25 hour pilot, which is strange considering the DCC are the subordinate to the pilots

So now it's
Corporal -> Sergeant -> Staff Sergeant -> Warrant Officer

**Changelog**
update not live yet


BramvanZijp:
I was confused about this too, good change IMO.

QuietlyWhisper:
It was a typo yeah, thanks

PR #7125, opened 2025-07-21
Returns Plushies to the people

x5tink:
THEY STOLE OUR PLUSHIES!

## About the PR
A previous commit : https://github.com/RMC-14/RMC-14/pull/6938 Removed plushies from everyone but civilian roles.
I've reverted this so people can have plushies back. It keeps the civilian categories as is from that commit, but returns the plushie category for everyone else.

## Why / Balance
Don't steal my shork bro

## Technical details
Few reverts, thats it.

## Media
CMO loadout:
image

CMSurv Loadout:
image

## Requirements

**Changelog**
:cl:
- fix: Returned plushies to the people!


noctyrnal:
parity that you're not meant to have them, also kinda makes sense (little too goofy for serious marines to have plushies)

path-to-aaru:
> parity that you're not meant to have them, also kinda makes sense (little too goofy for serious marines to have plushies)

if you've ever done a barracks inspection in the military you've seen more plushies than you could count

x5tink:
> parity that you're not meant to have them, also kinda makes sense (little too goofy for serious marines to have plushies)

So what, its a cosmetic thats nice to have and people enjoy it, why take that away from people. How does a survivor running round with one make it better? Not everything has to be super serious all the time.

x5tink:
> parity that you're not meant to have them, also kinda makes sense (little too goofy for serious marines to have plushies)

I'm just going to add further to my previous comment before bed. 
As regular things in the RMC/CM universe, we have A heavily armed pizza company, HEFA knights, Crayons, Eating said crayons (and the trip it gives), Fantastical alien marines species, miniatures of players scattered all around,  Crazy foods like Copy-Pasta, Golden apples (and doctors delight), Clowns tears, sashimi ( and that it practially murders you ). To list only a few things that are a 'little too goofy', in comparison to a few plushies.

snebl:
> parity that you're not meant to have them

Isn't parity just for balance and mechanical changes? iirc cosmetic stuff doesn't have to be parity, non-human marines for example; since they don't have species quirks.

mmails:
I do agree that it's goofy for marines to run around with plushies, they're better left as admin spawned only, especially because they're not even CM plushies and don't match CM's style

xfloodie:
Plushies are just cosmetics and a lot of people really enjoy them. RMC is not exactly a milsim and even if it were, you'd be surprised to see how many silly items you would find on IRL soldiers as trinkets. We do have a number of non-parity features already implemented both visual and mechanical, and plushies- which are just another way for players to express what their characters like- should not have been taken away in the first place. To add onto that, the psych treatment room on the second floor already has a therapy plushie, why would people not be carrying similar objects for RP reasons?

QuietlyWhisper:
Personally I dont mind people having these since I had already adjusted them to be less annoying

I don't really care much for the "is this parity" or "is this serious enough" conversations, one because we're allowed to make our own decisions with a guideline, and the last because I think the players are capable of policing themselves if theyre being disruptive with these (which shouldnt really be possible since they were already adjusted not to be, unless something was missed)

As far as I know it wasnt intentional from either me or smug to remove the loadout options. We did discuss not having them as loadout options when they were first added all those months ago, but concluded to adjusting them to not have sound spam.

DrSmugleaf (approved):

PR #7134, opened 2025-07-21
Fixed scouts not actually needing their armor to use their cloak

### LINKED ISSUE #5501: Scout can still use their cloak without their armor equipped
## Description

Scout's cloak can only be equipped when they are wearing their armor, but you can just take your armor off after equipping it; the cloak will stay and it's ability still works.

**Reproduction**

- Be a scout WS.
- Put on your armor.
- Put on your cloak.
- Take off your armor.
- You have become death, destroyer of T2s.

**Screenshots**

Image
(wearing cloak without armor.)

snebl:
## About the PR

With this PR scouts will actually need to at least have their armor on them in order to use their cloak. There was already a check for when equipping the cloak, but players could easily just unequip and ditch the armor after equipping it solely to wear the cloak. 

Now when the armor is dropped; the cloak will fall off along with it.

## Why / Balance

Resolves #5501

## Technical details

`ClothingRequireEquippedComponent` now has an `autoUnequip` field, if this is set to `true` then the clothing will drop when something within it's whitelist is dropped.

This field is false by default so that it doesn't affect anything else that's also using the component.

## Requirements

**Changelog**

:cl:
- tweak: Scouts will now drop their cloak when dropping their required armor.

### DIFF
--- a/Content.Shared/_RMC14/Clothing/RMCClothingSystem.cs
+++ b/Content.Shared/_RMC14/Clothing/RMCClothingSystem.cs
@@ -1,4 +1,5 @@
 using Content.Shared._RMC14.UniformAccessories;
+using Content.Shared.Clothing.Components;
 using Content.Shared.Clothing.EntitySystems;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Inventory;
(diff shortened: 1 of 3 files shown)


snebl:
fun fact: you can cheese the armor check by just having the armor in your hands, putting on the cloak, and then dropping the armor. You don't even have to equip the armor in the first place.

That's why i decided to do it this way (dropping when required clothing is dropped) instead of dropping when the armor is unequipped.

Content.Shared/_RMC14/Clothing/RMCClothingSystem.cs
```diff
@@ -98,6 +104,19 @@ private void OnRequireEquippedBeingEquippedAttempt(Entity<ClothingRequireEquippe
...
 
+    private void OnDropped(Entity<ClothingComponent> ent, ref DroppedEvent args)
+    {
+        var slots = _inventory.GetSlotEnumerator(args.User);
+        while (slots.MoveNext(out var slot))
+        {
+            if (slot.ContainedEntity is not { } contained)
+                continue;
+
+            if (TryComp<ClothingRequireEquippedComponent>(contained, out var requiresEquipped) && requiresEquipped.AutoUnequip && _whitelist.IsValid(requiresEquipped.Whitelist, ent.Owner))
```
DrSmugleaf, line 102: I don't think this matters for current usages of this but you would need to check if no other item fulfills the whitelist before force unequipping

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Clothing/RMCClothingSystem.cs
snebl: Oh i thought it was an AND list but it looks an OR one so yeah let me fix that

snebl:
Made a new private method so that its not as messy and made `OnRequireEquippedBeingEquippedAttempt` to use it since its the exact same logic.

I tested it with lizard plushies and it all works as expected. should be good now.

PR #7138, opened 2025-07-21
Make survivor job rolling more random

Dutch-VanDerLinde:
## About the PR
Currently, survivors are always rolled in a specific order:
- Engi
- Doctor
- Security
- Corporate
- Scientist
- Civilian

This PR changes it so the order of the jobs are shuffled every round. I believe this to be a good change as engi/doc rollers will have a way higher chance of rolling survivor than a security, corporate, scientist, or civilian player. This PR gives a fair chance for everybody, and is also parity, as jobs are rolled all together in CM13 rather than a specific order.

## Media
1st attempt: Player 1 rolled Corporate Colonist, and Player 2 rolled Engineer colonist. Player 1 got corpo
image

2nd attempt: Player 1 rolled Engineer Colonist, and Player 2 rolled Security colonist. Player 1 got engi.
image

:cl:
- tweak: Survivor jobs now roll more randomly, instead of rolling in a specific order every round.

### DIFF
--- a/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
+++ b/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
@@ -686,6 +686,13 @@ public sealed class CMDistressSignalRuleSystem : GameRuleSystem<CMDistressSignal
 
             if (comp.SpawnSurvivors)
             {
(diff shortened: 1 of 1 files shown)


AverageNotDoingAnythingEnjoyer:
Oh, so this is why we have only engi and doctor survs all the time

DrSmugleaf (approved):

PR #7167, opened 2025-07-21
Multiple bug fixes 

### LINKED ISSUE #7160: Unable to dig up sandbags and wrench cades.
## Description
Your currently unable to dig up placed sandbags and wrench cades.

**Reproduction**
Place down either an sandbag or cade. 
Use an entrenching tool to dig up the sandbag.
Or use an an wrench to unachor an cade

https://github.com/user-attachments/assets/9b41502f-d715-4111-aa3f-539eff11d66c

Arendian:
## About the PR

- Tools can be used on barricades again.
- You're now unable to fall onto the Almayer.(I think)
- Parasite leaps won't be blocked by resin doors they are hiding under.

## Why / Balance

Bug fixes
Resolves #7160
## Technical details

- The InteractUsingEvent was always being handled by the BarbedComponent, even if it did nothing. This prevented any other components from handling this event, for example the AnchoringComponent.
- I'm unable to reproduce the falling onto the Almayer issue on the Dev map, which has the same roof level as the Almayer, so I am not 100% sure if it's actually fixed.  I added another check that shooting a CAS weapon uses to see if the target is valid. On the Almayer you're able to select a flare as a target for paradropping, while on the Dev map the check works as it should and prevents targeting.
- Same issue as in #7124 but this time with constructs. Slipped through testing because it only happens when the construct is from the same hive as the leaping entity.

## Requirements

**Changelog**

:cl: Dygon
- fix: Barricades can be anchored and unanchored, and sandbags can be disassembled again.
- fix: You're now unable to drop or fall onto the Almayer.
- fix: Parasites already under a resin door won't be blocked by it during their leap.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
@@ -7,6 +7,7 @@ using Content.Shared._RMC14.Damage.ObstacleSlamming;
 using Content.Shared._RMC14.Entrenching;
 using Content.Shared._RMC14.Pulling;
 using Content.Shared._RMC14.Stun;
+using Content.Shared._RMC14.Xenonids.Construction;
 using Content.Shared._RMC14.Xenonids.Hive;
 using Content.Shared._RMC14.Xenonids.Invisibility;
 using Content.Shared._RMC14.Xenonids.Plasma;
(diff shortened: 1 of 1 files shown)


noctyrnal:
WW shooting in air already fixed in #7153

DrSmugleaf (approved):

PR #7174, opened 2025-07-21
Smoke nades no longer detonate when deleted

### LINKED ISSUE #6622: deleting smoke grenades detonates them
## Description
~~using the "erase mode" function of the admin entity spawner on~~ deleting an m40 hsdp grenade AT ALL, such as through admin spawned chasms or xeno acidification, causes it to immediately produce a smoke cloud. this does not seem to happen with the other m40 grenades

**Reproduction**
>spawn or otherwise acquire an m40 hsdp grenade
>open the admin entity spawner menu
>turn on erase mode
>erase the grenade

**Screenshots**

**Additional context**
its possible this occurs whenever the entity is deleted *at all*, ~~meaning tossing it in the ASRS might also cause this to happen.~~ EDIT: the ASRS does not delete items. however, other ways of deleting entities causes this.

snebl:
## About the PR

The m40 smoke grenade now uses `SpawnOnTrigger` instead of `SpawnOnTerminate` to spawn it's effect. This doesn't affect the grenades behavior aside from it no longer detonating when deleted by other means. 

## Why / Balance

Fixes #6622

## Technical details

None.

## Requirements

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Throwable/grenades.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Throwable/grenades.yml
@@ -109,9 +109,8 @@
     delay: 2
   - type: Sprite
     sprite: _RMC14/Objects/Weapons/Grenades/m40smoke.rsi
-  - type: SpawnOnTerminate
-    spawn: RMCSmoke
-    projectileAdjust: false
+  - type: SpawnOnTrigger
+    proto: RMCSmoke
   - type: SoundOnTrigger
     sound: /Audio/Effects/smoke.ogg
   - type: DeleteOnTrigger


snebl:
I tested in the dev world and it seems to still work fine.

Dutch-VanDerLinde:
does it spawn the same amount of smoke like before

snebl:
> does it spawn the same amount of smoke like before

It's my understanding that its just spawning a different prototype that does the smoke, both before and after the change. So i don't see why it would be different, but i checked again just to be sure and yeah its the exact same.

DrSmugleaf (approved):

snebl:
image

RosySaturniidae:
seal clapping. thank you

PR #7203, opened 2025-07-22
Head top marking layer is now above hats.

Adrian16199:
## About the PR

Makes head top markings go above the helmets/hats etc etc.
I will *note* that some species even by basic markings might look eh with the helmet so i could just make it that the helmets will be the same still while normal hats will have this. 
Instances of when the helmet would look really weird (this only really happens with skrella) I want to be fully honest and not just pretend that i checked everythin and its gonna look *chefs kiss*
image

## Why / Balance

So to be honest. It genuily icked me with the fact how hats at the front? looked fine, you can see ears (some of em) but then you look at the side and you see that there is *nothing there*, this incosistency is just killing me inside and ss13 has this. (ss13 isnt a good argument, i know but mostly the idea is comin from there)
Plus I believe it just makes species look better with horns/ears not being overlapped by hats, etc etc
## Technical details

Pretty much simple YML
## Media

Zrzut ekranu 2025-07-22 202550

image

image

image

image

image

image

image

image

## Requirements

**Changelog**

:cl:
- tweak: Changed layering of head top marking to be above hats. Ears/horns will be now visible.

### DIFF
--- a/Resources/Prototypes/Entities/Mobs/Species/base.yml
+++ b/Resources/Prototypes/Entities/Mobs/Species/base.yml
@@ -43,12 +43,12 @@
     - map: [ "enum.HumanoidVisualLayers.FacialHair" ]
     - map: [ "enum.HumanoidVisualLayers.Hair" ]
     - map: [ "enum.HumanoidVisualLayers.HeadSide" ]
-    - map: [ "enum.HumanoidVisualLayers.HeadTop" ]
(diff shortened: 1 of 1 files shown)


Adrian16199:
I will note regardin this PR that I could also make it that helmets could have option to just tug in ears incase felinid enjoyers would hate this (or the other 2 species), thumbs up (to this comment, not the PR) if you would want that or somethin.

PR #7209, opened 2025-07-22
Fix Tumble, Lunge, Charge, and Headbutt not always hitting

Arendian:
## About the PR

Lunge, Charge, Headbutt and Tumble will now consistently hit targets in range. Lunge was the most noticeable, the other actions only at the edge of their range.
Lunge, Tumble and Headbutt won't overshoot their target anymore, the xeno's velocity is set to 0 when they hit their target.
Charge and Tumble(only without hitting anything) still overshoot a single tile, but they only hit entities within their ability range.

## Why / Balance

Bug fix

## Technical details

These actions throw the xeno at coordinates and apply their effect on throw hit. Friction made it so they would land before reaching their target.

## Requirements

**Changelog**

:cl: Dygon
- fix: Lunge, Tumble, Headbutt and Charge consistently hit their target again.
- fix: Lunge, Tumble and Headbutt won't move past their target anymore after hitting them.


DrSmugleaf (approved):
Thanks for the PR!

PR #7211, opened 2025-07-22
map template

noctyrnal:
no more wrong map stuff

map template file that does everything you need, all set up. you can just load it and map to it.


DrSmugleaf (approved):
Thanks for the PR!

PR #7234, opened 2025-07-23
Fix certain entities blocking resin structure construction while they shouldn't

Arendian:
## About the PR

Platforms, barricades and open doors no longer block resin structures.
Beds, vendors and foldable chairs are now anchored.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix: Platforms, barricades and open doors/shutters/podlocks no longer block resin structures.

### DIFF
--- a/Content.Shared/Foldable/FoldableComponent.cs
+++ b/Content.Shared/Foldable/FoldableComponent.cs
@@ -1,3 +1,4 @@
+using Content.Shared._RMC14.Folded;
 using Robust.Shared.GameStates;
 
 namespace Content.Shared.Foldable;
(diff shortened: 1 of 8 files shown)


path-to-aaru:
Would it be possible to also add this for window shutters? If they're opened before melting a window, you then have to wait til the window melts and then melt the shutters as well before you can build on that tile.

Arendian:
> Would it be possible to also add this for window shutters? If they're opened before melting a window, you then have to wait til the window melts and then melt the shutters as well before you can build on that tile.

Shutters, podlocks and blast doors are all considered doors, so it's also fixed for those. I'll edit the changelog to make this more clear.

PR #7255, opened 2025-07-25
Fix flaregun flare ignites lasting too long

Vermidia:
## About the PR

Title.

## Why / Balance

Parity.

## Technical details

In CM, starshell is firestacks 2, and flares are firestacks 2.5.
Starshell is duration 2, so flares are 3. (Duration can only be applied as an int...so I rounded up)

## Requirements

**Changelog**

:cl:
- fix: Fixed flaregun flares and airburst starshells ignites applying for too long. Starshell ignites also no longer do less damage.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Tools/flares.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Tools/flares.yml
@@ -344,7 +344,7 @@
     radius: 2.0
     energy: 7.0
   - type: IgniteOnProjectileHit
-    intensity: 3
+    duration: 3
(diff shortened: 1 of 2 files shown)


Arendian:
Stacks in CM should be ported as duration to RMCs.(Woops, I did that wrong for starshell) Starshell projectiles should also have a duration of 2, and intensity as the default value instead of what it currently is on master.

Vermidia:
Is the duration 1:1 with time or is it like status effects and is 1:2

Arendian:
> Is the duration 1:1 with time or is it like status effects and is 1:2

it's 1:1

PR #7268, opened 2025-07-25
Training grenade fixes

Arendian:
## About the PR

Training grenades now become small sized again after they "explode".
A training grenade will no longer be primed automatically when thrown, if it has ever been shot from a grenade launcher.

## Why / Balance

Bug fix.

## Technical details

An entity's original size is stored on the ChangeItemSizeOnTimerTriggerComponent before it gets changed, when a TriggerEvent happens the item size is set back to this stored original size.
The TriggerOnThrowEndComponent is now removed when a TriggerEvent happens.

## Requirements

**Changelog**

:cl: Dygon
- fix: Training grenades no longer explode without priming when shot from a grenade launcher, and now regain their original size after 'exploding.


DrSmugleaf (approved):
Thanks for the PR!

PR #6946, opened 2025-07-06
Adds Intel spawns to CORSAT

FeelFlows:
## About the PR

Adds intel spawns to CORSAT

## Why / Balance

CORSAT was taken out of rotation on CM13 before intel became a proper thing. Having made intel maps for each map, i've taken a look at all of them and took various elements when adding CORSAT's spawns. The number of intel markers is nearly identical to Hybrisa, and takes a similar distribution to Hybrisa.

This will let CORSAT be used for Distress Signal in case one wishes to.

## Technical details

This is my first PR ever, was told I could do this all through one yml file, as far as I am aware I haven't disturbed anything other than adding intel markers. Please let me know if I broke something or otherwise messed up :)
## Media

image
image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**
:cl:
- add: Added intel spawners to CORSAT.

PR #7279, opened 2025-07-26
Adds Sensor Tower to CORSAT

FeelFlows:
## About the PR

A second PR which updates CORSAT into the modern world, as sensor towers did not exist in CM13 when it was removed from rotation. This is a follow-up PR to #6946 which added intel markers to CORSAT.

## Why / Balance

Non-parity, as CORSAT was taken out of rotation before sensors were added to the game. While CORSAT is a PVE map only, this will make it playable in Distress Signal, if that was ever wanted. It's also just good for PVE if a gamemaster wants to use it as a prop for a story.

I chose to place the sensor tower in the ice biodome after analyzing scant information on how CORSAT has been historically played, it's to my understanding that the east LZ is the most popular and optimal one, as it has the most relaxed and open path towards the hive, letting them easily enter the two eastern biodomes before proceeding to hive. I was planning to pick one of the biodomes as it would both make the most sense to add them there, and require no modifying of the actual map layout, but picking either of the two ones on the east would make the already optimal east LZ the no-brainer pick, as it would mean easy access to Sensors.
On the other hand, the north LZ faces immediate urban environments, with long hallways & corners before they can even reach any of the biodomes. For this reason, I picked to place the sensor tower in the ice biodome, as it gives the north LZ some positives over the east LZ, with it being closer. It's still viable for the east LZ, though now they have to make a purposeful choice to push direct west from the solaris biodome, and push through a similar amount of hallways as north LZ, rather than head to hive directly.

The Gamma symbol represents the ice biodome, and so its inclusion next to the Gamma science lab has soul factor.
## Technical details

modifies corsat.yml, only changed stuff at the sensor tower as far as i'm aware (new mapper)

## Media

Screenshot 2025-07-26 230015
Photos_oeDog2HFxE

## Requirements

**Changelog**

:cl: Max_Power
- add: Added a sensor tower to CORSAT


XxSWAG-MASTER-OVERLORDxX:
I was personally thinking it could have randomized placement like Solaris, and it could be placed where you did, but also a location within the grassy dome (but very good RP)

PR #7288, opened 2025-07-27
Added [check] tag and also fixed some paperwork stuff to do with my previous addition

yagwog:
## About the PR
Fixed stamps not properly de-populating [form] and [signature] tags
Fixed also the bottom margin getting eaten up by excess line height not accounted for by in line button elements.

added a new tag for check boxes, tick, cross and leave empty to your hearts content

Also made the buttons appear a bit nicer and uniform

## Why / Balance
paperwork soulium and also less headache for people who just want to get paperwork out the fuckin way and RP

## Technical details
Added a dynamic bottom margin that accounts for button tags.
- Buttons are extremely hard, if not impossible, to modify at this point in time, unless there's changes to toolbox
- The larger buttons were messing up in line characters on the same line, adding too much line height, I reduced the line height to account for neatness when appearing alongside other things.
- I can't change where the text appears within the buttons, again, will have to wait for robust toolbox updates.

## Media

https://github.com/user-attachments/assets/c40b1128-078f-45d5-95e6-0fdd9c3cbad7

## Requirements

**Changelog**
:cl:
- fix: Stamps not working correctly with tags
- fix: Bottom margin not correctly extending for tags
- added: [check] tags
- fix: made buttons nicer to look at.

### DIFF
--- a/Content.Client/Paper/UI/PaperWindow.xaml.cs
+++ b/Content.Client/Paper/UI/PaperWindow.xaml.cs
@@ -40,6 +40,9 @@ namespace Content.Client.Paper.UI
         // If paper limits the size in one or both axes, it'll affect whether
         // we're able to resize this UI or not. Default to everything enabled:
         private DragMode _allowedResizeModes = ~DragMode.None;
+        
+        // Store original margin to restore when switching modes
+        private Thickness _originalContentMargin;
 
         private readonly Type[] _allowedTags = new Type[] {
             typeof(BoldItalicTag),
@@ -50,7 +53,8 @@ namespace Content.Client.Paper.UI
             typeof(ItalicTag),
             typeof(MonoTag),
             typeof(FormTagHandler),
-            typeof(SignatureTagHandler)
+            typeof(SignatureTagHandler),
+            typeof(CheckTagHandler)
         };
 
         public event Action<string>? OnSaved;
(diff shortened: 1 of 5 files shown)


yagwog:
tested immensely for max working and not non working

on sod

AverageNotDoingAnythingEnjoyer:
Would it be possible to make the choice menu be smaller and appear slightly above the cursor on press?
I imagine filling out documents with a lot of checks might become tiring if the menu will constantly appear on top

yagwog:
To add some context on the PR:
- Buttons are extremely hard, if not impossible, to modify at this point in time, unless there's changes to toolbox
- The larger buttons were messing up in line characters on the same line, adding too much line height, I reduced the line height to account for buttons, I also reduced the size of the fill buttons as they occur within areas so the surrounding text won't be messed up.
- I can't change where the text appears within the buttons, again, will have to wait for robust toolbox updates.

@AverageNotDoingAnythingEnjoyer 
> Would it be possible to make the choice menu be smaller and appear slightly above the cursor on press? I imagine filling out documents with a lot of checks might become tiring if the menu will constantly appear on top

I think covering the document itself with a UI element goes against the idea of it being an accessibility addition if it isn't clear what the question is or what they're filling for. The buttons are attached to the paperwork so in my opinion it isn't too far.

yagwog:
image

DrSmugleaf (approved):
Thanks for the PR!

PR #7293, opened 2025-07-27
Make X17 riot charge no longer require police skill.

Dutch-VanDerLinde:
## About the PR
This breaching charge spawns on TSEPA constables survivors aswell in the Hybrisa armory. Since it's impossible for survivors (Hybrisa security variants no longer have police skill due to a recent change) to use this item, the police skill requirement has been removed, now is engineering 1 like the other breaching charges.

## Requirements

:cl:
- tweak: The X17 riot charge no longer requires police skill to use.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Explosives/breaching_charge.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Explosives/breaching_charge.yml
@@ -58,9 +58,6 @@
     - state: icon
       map: ["base"]
   # TODO Shrapnel volume 20
-  - type: RequiresSkill
-    skills:
-      RMCSkillPolice: 1
 
 - type: entity
   parent: RMCExplosiveBreachingCharge


QuietlyWhisper:
Is this used anywhere else? It seems to be crowd control equipment, which makes sense for the skill-

If MP are intended to have this or something, it might be better to replace the charges with the parent charges instead?

If not, then this change is fine.

Dutch-VanDerLinde:
> Is this used anywhere else? It seems to be crowd control equipment, which makes sense for the skill-
> 
> If MP are intended to have this or something, it might be better to replace the charges with the parent charges instead?
> 
> If not, then this change is fine.

No this is only used on hybrisa currently

texas1234567889:
do believe theyre used in specific CMB/MP ERTs as well, saw them use it to blast open the liaisons office on 13 once

QuietlyWhisper:
ah if it's only part of ert loadouts as well that's fine enough for me, change is fine

DrSmugleaf (approved):
Thanks for the PR!

PR #7294, opened 2025-07-27
Don't clobber cursors every frame

PJB3005:
This fixes various UI cursors like I-bars and window resizing.


DrSmugleaf (approved):
holy shite

PR #7301, opened 2025-07-28
Queen Gut improvements

noctyrnal:
## About the PR

Gutting as the queen now has a pop up for starting and finishing, shakes the target as they're being gibbed, and splatters their items rather than dropping them all in one neat pile.

## Why / Balance

Looks alot cooler and makes more sense (turning someone into organs shouldn't make all their items nicely pile up)

## Technical details

Added `RMCGibSystem` which stores `ScatterInventoryItems` which will fling items off an entity. Designed to be used with gibbing.
`SharedXenoGutSystem` now has Locs for starting and finishing, uses `DoJitter` when starting the do-after on a target, will remove the Jitter effect when the do-after is cancelled and now calls `ScatterInventoryItems` when gibbing a target.

## Media

https://github.com/user-attachments/assets/d0e2f3d2-0567-4256-8487-3e635f04568f

## Requirements

**Changelog**

:cl: noctyrnal
- add: Gibbing a person as a queen will now pop up text, cause the target to shake and make their dropped items splatter out.

### DIFF
--- a/Content.Shared/_RMC14/Gibbing/RMCGibSystem.cs
+++ b/Content.Shared/_RMC14/Gibbing/RMCGibSystem.cs
@@ -1,0 +1,46 @@
+using Content.Shared.Inventory;
+using Robust.Shared.Physics.Systems;
+using Robust.Shared.Random;
+
+namespace Content.Shared._RMC14.Gibbing;
+public sealed class RMCGibSystem : EntitySystem
+{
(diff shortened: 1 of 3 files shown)


Spaghetti-bit:
Thoughts of making it spawn blood decals?

DrSmugleaf (approved):
Thanks for the PR!

PR #7313, opened 2025-07-29
Survivor job rolling 2

Dutch-VanDerLinde:
## About the PR
Reworked survivor rolling again. I figured out a solution for this, so here we are
This brings survivor rolling to parity (EXCEPT for rolling survivor being one job and putting you in a random job if you don't get your preferred role. Many people would rather just not play if they dont get their preferred survivor job so i decided against it)

## Technical details
OLD survivor rolling:

The loop would go through each job priority, (high to low). Then, it'd loop through all the jobs and lists of players in the survivor candidates dictionary (meaning, if engi colonist got put at the top, and 4 people rolled engi, those 4 people had a guaranteed chance of getting engi colonist)

NEW survivor rolling:
It loops the amount of times the total survivor count is (so, in a live round, 7)
Then, each loop, it gets a random survivor job and attempts to get a random player who had that job on the specific priority.
Jobs that ignore the maximum survivor count are rolled separately like CM13.

Basically each player has an equal chance of rolling survivor

Both survivor rolling method goes through each job priority

## Requirements

**Changelog**
:cl:
- tweak: Survivor rolling has been reworked, now every survivor job has an equal chance of rolling. Civilian survivor no longer rolls last.

### DIFF
--- a/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
+++ b/Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
@@ -740,7 +740,7 @@ public sealed class CMDistressSignalRuleSystem : GameRuleSystem<CMDistressSignal
                 var selectedSurvivors = 0;
                 for (var prio = priorities - 1; prio >= 0; prio--)
                 {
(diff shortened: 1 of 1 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #7338, opened 2025-07-31
Tackling nested + nest retargeting

Vermidia:
## About the PR

Title.

## Why / Balance

Told it's Parity + to make it work with clicks.

## Technical details

You can now tackle nested/nested & infected hosts. The attackattempt func in the xeno system no longer stops disarmed attacks.

To make this easier, if a xeno attacks a xeno nest, it will try to redirect the attack to it's occupant, both disarms and attacks. Ofc at this point the xeno attack code will cancel it if they're infected. To make this work however I had to move the nest offsets slightly, and make the south offset inline with the others, so they're farther down the wall now, because they were being counted as being obstructed. But as a bonus I fixed nests not rotating on creations so.

## Media

(nested sprites fix + where the front goes now)
image

## Requirements

**Changelog**

:cl:
- fix: Fixed not being able to tackle nested hosts. Trying to attack/tackle a nest will also now redirect the attack to the occupant, if it's from a xeno.
- fix: Fixed nest sprites always using the south sprite.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Construction/Nest/XenoNestSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Construction/Nest/XenoNestSystem.cs
@@ -4,6 +4,7 @@ using Content.Shared._RMC14.Chat;
 using Content.Shared._RMC14.Ghost;
 using Content.Shared._RMC14.Inventory;
 using Content.Shared._RMC14.Map;
+using Content.Shared._RMC14.Weapons.Melee;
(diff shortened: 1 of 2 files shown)


pumpkin-pwr:
Random side note, can you use the code that makes tackling nest tackle the host and make it so when you use a parasite on the nest it redirects onto the host?

Would be convenient as often you have to tell new players to shove themselves into the nest and use the right click drop down.

Vermidia:
It worked in this case cause we have a melee override event. I'll see what I can do though

PR #7362, opened 2025-08-03
increase default tacmap size

noctyrnal:
image
just makes its 1000 by 800 now instead of 600 by 600


DrSmugleaf (approved):
Thanks for the PR!

PR #7363, opened 2025-08-03
Disabler rework (rmc)

Elysium206:
## About the PR

I used components from the stun baton and existing rifles to add a right click melee function to the disabler. I also edited the stamina system slightly to make melee hits call an event to actually drain battery (they were not before)

## Why / Balance

This is a purely stylistic choice, as I find the large batons to be somewhat silly. in real life tasers can often be used to "drive stun" a suspect. This is done by the front of the taser being pushed into the individual rather than shot.
image
The Stun baton is already a one hit stun, so this is not any better. But due to the fact there is now less equipment to carry, I ended up with the disabler have 5 less charges than the baton to try and balance that fact out at least a little. MPs can still carry whichever they wish or both.

## Technical details

- In the RMCStaminaCystem.Cs I Added a callout for event when using the "RMCStaminaDamageComponent". Currently the stun component would not functionally take battery charge per melee hit until i added that.
- In the Taser.yml I added multiple components related to Alt click melee stun functionality.

## Media

https://github.com/user-attachments/assets/7effe9e9-c3ac-42da-b4f7-0dc55a2e35f2

https://github.com/user-attachments/assets/1df24299-20a7-4bca-add4-13645c2de17e

## Requirements

- [ X ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ X ] I have added media to this PR or it does not require an ingame showcase.

- [ X ] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Disabler's now have a melee stun function similar to Drive Stunning!

### DIFF
--- a/Content.Shared/_RMC14/Stamina/RMCStaminaSystem.cs
+++ b/Content.Shared/_RMC14/Stamina/RMCStaminaSystem.cs
@@ -13,6 +13,7 @@ using Content.Shared.StatusEffect;
 using Content.Shared.Throwing;
 using Content.Shared.Weapons.Melee.Events;
 using Content.Shared.Wieldable.Components;
+using Content.Shared.Damage.Events;
 using Robust.Shared.Network;
 using Robust.Shared.Player;
 using Robust.Shared.Timing;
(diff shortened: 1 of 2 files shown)


QuietlyWhisper:
Maybe I'm missing something but the amount of charges is still the same after this change

Elysium206:
Disablers currently have 16 charges. This should put them at 15, which is 5 less than the batons 20

DrSmugleaf (approved):

PR #7370, opened 2025-08-04
High visibility marking for rubber ammunition

BreakpointDebug1234:
## About the PR

Added high visibility markings to the rubber ammunition available in the MP RiotTech using pink tape (blue and yellow already used elsewhere)
## Why / Balance

Non-lethal ammunition should NOT have the same sprites as legal ammunition this could lead to a multiple of incidents such as using lethal ammunition on mutineers. This PR clearly highlights rubber mags as being as such.
## Technical details

- Dedicated sprites for the rubber mags based on the existing mags but with a pink stripe to mark that this is non lethal
- The bullets for the loaded mag layer now have black tips to further highlight this is rubber ammunition

## Media

M4SPR
M54CE
M63
M77
M1984

## Requirements

**Changelog**
:cl:
- tweak: Rubber mags are now marked by a pink stripe


QuietlyWhisper:
Usually non-lethal weapons/ammo are marked in orange, is orange used anywhere else?

If so, is pink used anywhere else, that we may not have yet?

QuietlyWhisper:
I also recommend adding to the licenses that the modifications are made by you (or whoever did them)

texas1234567889:
> Usually non-lethal weapons/ammo are marked in orange, is orange used anywhere else?
> 
> If so, is pink used anywhere else, that we may not have yet?

pink is used for toxin mags (dont think we've ported those yet) however it'd probably be fine to use pink for nonlethals instead as they're admeme only

DoguhanOzgurAkca:
> Usually non-lethal weapons/ammo are marked in orange, is orange used anywhere else?
> 
> If so, is pink used anywhere else, that we may not have yet?

Technically incendiary is orange-ish

BreakpointDebug1234:
> Usually non-lethal weapons/ammo are marked in orange, is orange used anywhere else?

I know this is the case however **yellow is already used to mark extended mags**. Blue would be the next best thing however M4SPRA19 uses blue to mark the mags so it would add further confusion. Since I created the PR I have found out that in CM13 they are marked with a dark blue but still not sure if people would be confused by the two SPR mags. 
**I don't have an issue with changing the sprites if this is the solution.**

> I also recommend adding to the licenses that the modifications are made by you (or whoever did them)

Will do so in a few hours.

> pink is used for toxin mags (dont think we've ported those yet) however it'd probably be fine to use pink for nonlethals instead as they're admeme only

This certainly is an issue. I'll consider changing the color.

**Build failure is unrelated to this PR.**

BreakpointDebug1234:
I've decided to change the color of the mags to the one used in CM13 (not the exact shade but close to it). I do still think pink would've been a bit better for visibility (even if it is unrealistic) however this would create problems down the line for whomever ports toxin mags so I've decided it's better to avoid this issue.

Licenses have been modified to include the person who modified them.

**As usual any build failure is unrelated to the PR.**

Let me know if any other changes are required.

DrSmugleaf (approved):
Thanks for the PR!

PR #7372, opened 2025-08-04
Disable Civilian Access to the CLF Patch in Loadouts

PursuitInAshes:
## About the PR
Remove `PatchCLFRMC` from the Civilian loadout group (the one survivors get, and a few other roles). I've disabled it, but with a comment, instead of removing it. So someone doesn't accidentally re-enable it for parity reasons.

## Why / Balance
This patch being accessible is setting *someone* up for a rule violation (Survivors or MPs or Civi roles, etc). It is an actively hostile faction's patch. Survivors may get the wrong idea that by wearing it they can be even remotely antagonistic (then causing a rule violation) or an MP will see someone with it and get the wrong idea they they are a proper CLF and kill them (then causing a rule violation). By outright removing it, we aren't encouraging players to break the rules anymore. It just reduces admin workload outright.

## Technical details
YAML

## Media
N/A - But I have verified this functions ingame.

## Requirements

**Changelog**
:cl:
- tweak: Civilians can no longer access the CLF Patch through loadouts.

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Loadouts/loadout_groups.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Loadouts/loadout_groups.yml
@@ -493,7 +493,7 @@
   - CorporateBrownJacketRMC
   - CorporateBlueJacketRMC
   - JacketWindbreakerRMC
-  
+
 - type: loadoutGroup
   id: SynthUniformAccessoriesRMC
   name: rmc-loadout-group-synthetic-accessories
@@ -559,7 +559,7 @@
   - PatchWeYaRMCFury161
   - PatchTSERMC
   - PatchCECRMC
-  - PatchCLFRMC
+  # - PatchCLFRMC # RMC14 - Disabled intentionally to reduce admin burden.
   # TODO Halcyon Dynamics patch
   # TODO cowboy hat
   # white cowboy hat TODO


XxSWAG-MASTER-OVERLORDxX:
The CL also gets access to this patch

PursuitInAshes:
Its removed from the civilian loadout group outright, so they won't get it either. I think the similar logic applies. The CL shouldn't be outwardly antagonistic to the UN. Reporters are guests on a UN vessel, same logic applies.

Edited the PR to be more accurate in the removal.

Resources/Prototypes/_RMC14/Roles/Loadouts/loadout_groups.yml
```diff
@@ -559,7 +559,7 @@
   - PatchWeYaRMCFury161
   - PatchTSERMC
   - PatchCECRMC
-  - PatchCLFRMC
+  # - PatchCLFRMC # RMC14 - Disabled intentionally to reduce admin burden.
```
Dutch-VanDerLinde, line 562: Just straight up remove it, there's no reason why this should be a loadout item ever

PursuitInAshes: I specifically commented it out because I'm worried otherwise someone would just put it back due to it being parity to exist, not knowing we removed it. Hence leaving it commented out with the explanation next to it.

DrSmugleaf (approved):
my survivor players are animals

PR #7413, opened 2025-08-07
Acid blood splash

N0lim:
## About the PR
Added acid blood to xeno. Damage higher than 12 would activate acid splash. Acid splash spawns decals. If acid splash deal damage to someone it can't activate in next 3 seconds. Standard xeno would deal 25 heat damage, lesser drone 15, parasite 5. Radius was recalulated with formula r = x/sqrt(pi) to save area. For example 3x3 tiles radius recalculated as r = 3/sqrt(pi). 

## Why / Balance
Parity
Splash code
https://github.com/cmss13-devs/cmss13/blob/26ce387c7130c12c6afb3eb1e5f9dda25cabbb85/code/modules/mob/living/carbon/xenomorph/damage_procs.dm#L250-L297
Damage trigger code
https://github.com/cmss13-devs/cmss13/blob/4012064e765da5f274f22a125d431d29b9c3ee7d/code/modules/mob/living/carbon/xenomorph/damage_procs.dm#L149-L166
Gib code
https://github.com/cmss13-devs/cmss13/blob/00692062e308c8ed892f83bbca9289c6249eb304/code/modules/mob/living/carbon/xenomorph/life.dm#L483-L490
Gib trigger code
https://github.com/cmss13-devs/cmss13/blob/4012064e765da5f274f22a125d431d29b9c3ee7d/code/modules/mob/living/carbon/xenomorph/death.dm#L146-L163

## Media
Radius of splash from taking damage
Radius of splash from taking damage
Radius of splash from gib (imitation now, not real gib)
изображение
Xeno don't spawn decals after splash until cooldown ends

https://github.com/user-attachments/assets/e5db4bd1-4074-49c0-9062-f141c2fd8c85

Xeno don't activate splash until damage is high enough

https://github.com/user-attachments/assets/1df6b36f-c9da-4f7e-9840-66413812a0cc

## Requirements

**Changelog**
:cl:
- add: Added acid blood to xeno. Damage and death causes it to spray. If you are close enough, damage will be dealt depending on the type of xeno.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashSystem.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 using Content.Shared.Damage;
 using Content.Shared.Damage.Prototypes;
 using Robust.Shared.Prototypes;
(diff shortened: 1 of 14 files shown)


Dicerson:
Seems cool! I wonder - how do xenos ever get gibbed? I've... never seen that. Is it an OB only thing?

N0lim:
> Seems cool! I wonder - how do xenos ever get gibbed? I've... never seen that. Is it an OB only thing?

Each xeno has a base chance of gib, and this chance increases with the amount of damage that was taken upon death (excess damage / 2)
Base chance for standard xeno 5%, larva 25%, parasite 75%, lesser drone 100%

Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashSystem.cs
```diff
@@ -0,0 +1,145 @@
...
+            foreach (var type in bruteProto.DamageTypes)
+            {
+                _bruteTypes.Add(type);
+            }
+        }
+    }
+
+    private void ActivateSplash(EntityUid uid, AcidBloodSplashComponent comp, float splashRadius)
+    {
+        Spawn(comp.BloodDecalSpawnerPrototype, uid.ToCoordinates()); // create decal, probability inside prototype
```
DrSmugleaf: These should probably either be capped to a max per tile or delete old ones, see RMCScorchSystem for an example of the former

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashSystem.cs
DrSmugleaf: RandomDecalSpawner seems to support this as well, with MaxDecalsPerTile

N0lim: MaxDecalsPerTile just doesn't work for some reason, SnapGrid doesn't either, although I'd prefer to put decal directly under xeno instead of randomizing the position (or just a little bit, radius 0.2-0.3)

DrSmugleaf: It might be because it's a spawner and not the actual id, you would need to do it like RMCScorchSystem then

DrSmugleaf: Done in [`e8cc294` (#7413)](https://github.com/RMC-14/RMC-14/pull/7413/commits/e8cc2942f91928dbc3805b57109fa9b73b79a9ca)

PR #7417, opened 2025-08-07
Fixed all Normandy walls being named "alamo wall"

snebl:
## About the PR

Every Normandy wall was named "alamo wall" because of inheritance. Each wall inherits from `CMBaseNormandyWall` but also from a corresponding alamo wall (to save on typing i'd imagine). The name gets taken from the first prototype and not the last one so i just swapped the parent order for every normandy wall so that they would all end up with the name "normandy wall" instead.

Exceptions:
- line 953: uses `CMNormandyWall84` and not `CMBaseNormandyWall` because that's how it was before.
- line 980: was already in the correct order.

This will not affect mapping because i did not change any prototype IDs, only the parent order.

## Technical details

No code changes.

## Media

{E076B370-42D1-40C5-821D-7A6A1E15E899}
image

## Requirements

**Changelog**

:cl:
- fix: Normandy walls are no longer all named "alamo wall".

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Dropships/normandy_walls.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Dropships/normandy_walls.yml
@@ -12,7 +12,7 @@
     sprite: _RMC14/Structures/Dropships/normandy.rsi
 
 - type: entity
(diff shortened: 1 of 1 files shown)


snebl:
image

PR #7440, opened 2025-08-09
Gives the DP and GP their own icons.

RangerXVII:
## About the PR

Gave the dropship and gunship pilots different icons, colored after the alamo and normandy.

## Why / Balance

They are two different job, I always though it was weird they had the same icons.

It is also just nice to know which pilot you saw at a glance, instead of needing to examine them.

## Technical details
YAML and RSI changes.

## Media

PilotIcons

## Requirements

**Changelog**

:cl:

- tweak: The Dropship Pilot and Gunship Pilot now have different icons.


BarteG44:
Parity has two different icons for GP and DP, which work much better than a recolor imo, not to mention that making one gray and the other yellow implies a rank structure similar to something like that of CE and an MT or Doctor and nurse

QuietlyWhisper:
> Parity has two different icons for GP and DP, which work much better than a recolor imo, not to mention that making one gray and the other yellow implies a rank structure similar to something like that of CE and an MT or Doctor and nurse

Can you post example pictures of them

BarteG44:
> > Parity has two different icons for GP and DP, which work much better than a recolor imo, not to mention that making one gray and the other yellow implies a rank structure similar to something like that of CE and an MT or Doctor and nurse
> 
> Can you post example pictures of them

image

my memory failed me, since these are just in the observe menu and the hud uses the same icon we have currently, still i believe a different shape is a superior option without the insinuation of one pilot outranking the other

QuietlyWhisper:
I do think a new set of shapes would be better, but I don't particularly like that one.

Maybe we can come up with something better, even just two entirely new icons that look similar but you can easily tell the difference between them

RangerXVII:
Content Client_2x0sFyUsw3

Here is a picture I should have put here a long time ago.

> Parity has two different icons for GP and DP, which work much better than a recolor imo, not to mention that making one gray and the other yellow implies a rank structure similar to something like that of CE and an MT or Doctor and nurse

Since they both have the same frame color I don't really think it's a big deal. Officers have a copper or gold frame while enlisted, non-combatants have silver.

I found that while looking through map blips and thought it looked weird, so I didn't copy that.

I just thought a recolor would look interesting so I did that.

If you want a new shape that would not be this PR because I have no idea what I would put for that.

QuietlyWhisper:
You could try keeping the same color and inverting the direction of the icon, that's the best I could think of, but I'm not certain if that would give it a distinct appearance or not

PR #7453, opened 2025-08-11
Fix for buckled objects colliding with walls

ExponentialWizard:
## About the PR
Buckling now sets physics to soft, preventing movement altering collisions

## Why / Balance
This fixes buckling into chairs getting pushed out when the chair is close to a wall, canceling the buckle. Most noticeable on the dropship edge chairs. Also prevents objects like water tanks from colliding with buckled objects as a side effect.

## Technical details
Buckling sets physics of the buckle to soft, and unbuckling sets physics to hard. I avoided modifying collision masks, since it seems easy to lose track of the original mask when multiple systems are modifying collision. It seems a little hacky but it was the simplest solution I could come up with

## Requirements

:cl:
fix: Fixed being kicked out of chairs by walls

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
@@ -26,6 +26,8 @@ using Robust.Shared.Physics.Components;
 using Robust.Shared.Physics.Events;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Utility;
+using Robust.Shared.Physics;
+using System.Linq;
 
 namespace Content.Shared.Buckle;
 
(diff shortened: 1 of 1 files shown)


Content.Shared/Buckle/SharedBuckleSystem.Buckle.cs
```diff
@@ -416,8 +418,14 @@ private void Buckle(Entity<BuckleComponent> buckle, Entity<StrapComponent> strap
         var gotEv = new BuckledEvent(strap, buckle);
         RaiseLocalEvent(buckle, ref gotEv, true);
 
-        if (TryComp<PhysicsComponent>(buckle, out var physics))
+        if (TryComp<PhysicsComponent>(buckle, out var physics) && TryComp<FixturesComponent>(buckle, out var fixtures))
```
DrSmugleaf: Review comment somehow got lost, but it would probably be better to do this by adding a new component, adding that on buckle, removing it on unbuckle, and subscribing it to PreventCollideEvent to cancel all collisions
Other code may set a physics body's hardness value which would break this

ExponentialWizard: Thanks, this is so much simpler and lets me stop collisions with walls only so stuff like charger works properly.

PR #7463, opened 2025-08-11
Add Hive Pylon map blip

DummeHytteOst:
## About the PR

Adds a map blip and tracking for hive pylons
includes an update to the xenonid icon guidebook page, with the addition of the hive pylon.
## Why / Balance

while not parity, i believe its a very good QoL addition, that just makes sense to have, currently the hive pylon is, from what i know, the only xeno structure that doesn't have a map blip, which might just be an oversight in parity, or no one bothered to make it.
## Technical details

the pylon.png is just supply.png xenoified, as i thought that would make most sense, i believe the meta.json is in order.
## Media

image
image
image

## Requirements

**Changelog**

:cl:
- add: The hive pylon has gotten a map blip, and is now visible on the tactical map for xenonids, you can view the new map blip in the xenonid map icons guide book page.


Aurallianz:
@QuietlyWhisper pour vous.

NeYz0s:
The new icon seems to be the cause of a bug breaking the map for xenos, whenever a pylon is built the map breaks (can't see anymore icons, labels or whatever the queen was drawing) and when it's destroyed the max fixes itself

PR #7465, opened 2025-08-11
Fix for dead things rendering behind rollerbeds

ExponentialWizard:
## About the PR
dead things get their draw depth set to be above whatever they are buckled to if the buckle doesn't face north. Things on rollerbeds are always drawn on top.

## Technical details
BuckleVisualSystem now sets the draw depth of dead things to draw above the strap. Rollerbed buckles now get set to always be drawn above. 

This might not work if GetDrawDepthEvent is given to CMMobStateSystem before BuckleVisualSystem. I don't know what controls the order these activate in, and I am afraid of refactoring the buckling draw depth system, so I just went with the quick fix. Also there's another draw depth system in BuckleSystem that I think is being overridden?

This still doesn't fix draw depth ignoring camera rotation

## Requirements

- 
:cl:
- fix: Fixed dead things rendering behind rollerbeds.


Aurallianz:
LGTM

PR #7469, opened 2025-08-12
Sets Mortar FireDelay to 0

FeelFlows:
## About the PR

Functionally removes the firedelay on the mortar, letting you fire it as fast as you can type in coordinates and load shells.

## Why / Balance

Parity! I've always noticed how they sometimes fire to shells one after the other, first it was explained to me that it was part of the laser designator changes, but further research and discussions got me to finally test it, and i've found out that it's pretty much never HAD a fire delay! The cooldown on it applies to actually opening up the UI, which can be circumvented by just keeping the UI open the whole time while using it, you see this in the CM13 demo chat log.

Balance wise, it lets you combo shells to have greater firepower, drop an HE and immediately drop an incend to seriously hurt xenos. You still need to deal with dial & offset do-afters if you want to do really powerful things, so functionally outside of double tapping the cooldown is about 50% of what it used to be.

## Technical details

Changes a 9 to a 0, zero experience in C# but it works when testing and nothing seemed wrong.

## Media

https://github.com/user-attachments/assets/cfbd87b7-8eed-47c6-a4d8-3403f8e83b96

https://github.com/user-attachments/assets/0b66385b-c395-409d-8e5f-a638e4209be0

## Requirements

**Changelog**

:cl: Max_Power
- fix: Fixed the mortar tube having a cooldown before allowing you to fire another shell.


Aurallianz:
LGTM

Dicerson:
Ohk but we NEEEED THAT SCREENSHAKE ON FIRING IT THE SOUUL THE SOUUUL

PR #7485, opened 2025-08-13
Add Vent crawling

Vermidia:
## About the PR

Title.

## Why / Balance

Parity!

## Technical details

New system for handling vent crawling + comps and events. How it works is every time you have an opportunity to move (it works off a move delay) you will try to move in the last inputed direction, stored on the vent crawling comp, or none if you stopped pressing buttons. This makes it mostly smooth and predicted.

TRay scanner had some changes as one of the new comps basically gives Tray vision without the item, as I didn't want to deal with the base TRay comp. This is needed to mimic (somewhat) CM's behavior and also cause the client will mispredict exiting pipes if it can't see under the floor.

Overlay system & comp for those with the TRay comp given on vent crawling so they can see themselves and other vent crawling ents if they see the pipes. (Not parity but cool)

Pipes got migrated to vent crawling pipes as they use new comps for it. Sprites for the overlays by me (they could be better but they work for now). These ones snapgrid center, idk if that'll break something or not, but it's how most CM pipes work (far as I can tell).

Like CM you can't transport non-small hosts through vents and primed aciders can't enter vents either.

Didn't change pipe stats, since not sure what CM's are with how destructible they are, can be changed later.

Two more vent crawling non-parity stuff: Xenos in containers no longer count towards xenos being alive(no pipe stall) and you can't exit a pipe into a wall (safety). One final thing is ghosts get the vent crawl overlay icons, but only for visible pipes (as they don't have the tray part, so aghosts would need to use a scanner to see them for instance).

## Media

https://github.com/user-attachments/assets/b3a3e241-ab28-4fd0-bee0-61254f8318f6

## Requirements

**Changelog**

:cl:
- add: Added vent crawling. T0s, most T1s, and some T2s can vent crawl to move quickly through pipes from one location to the next.
- fix: Non-Queen xenos in containers (tunnels, vents) no longer count as alive for round end checks.


XxSWAG-MASTER-OVERLORDxX:
I assume those icons in the pipes are visible to ghosts?

Never mind i cant read (for others that also cant read, yes, yes you can see it as a ghost)

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1272,7 +1272,7 @@ private void CheckRoundShouldEnd()
                 if (!xeno.ContributesToVictory)
                     continue;
 
-                if (HasComp<ThunderdomeMapComponent>(xform.MapUid))
+                if (HasComp<ThunderdomeMapComponent>(xform.MapUid) || _containers.IsEntityInContainer(xenoId))
```
DrSmugleaf: I think in 13 they made it so that if Queen is alive it won't forfeit the round even if all xenos are in tunnels, can't remember for sure
Might want to do that and/or add a check for there being more than 3 xenos (like marine minor victories)

DrSmugleaf (changes requested):

DrSmugleaf (dismissed):
Thanks for the PR!
I was not looking forward to coding this personally

Dicerson:
If I recall correctly, the doafter time for going into a pipe scales with xeno size - so the smaller bugs climb in faster. But in the video it appears that senti and larva have the same doafter time.

Was I misled/misinformed?

Vermidia:
Misinformed, the doafter time is always the same.

PR #7490, opened 2025-08-13
Roof Area Alert Changes

noctyrnal:
## About the PR

Removes the hivecore and pylon unique icons and protection naming from the Roof Alert. Moves the textures to a more fitting location.

## Why / Balance

The hivecore and pylon icons and names were being used to metagame positions of them, especially with the hivecore on the almayer (i've personally seen cases of all of this happening multiple times). Theres nothing a xeno can do to counter this, and the only difference it would make outside of changes in roof protection is reveal the existence of a core.
I removed the unique parts of it, while keeping the new info about protections.
The textures were kept in `Structures/Machines` instead of `Interface/Alerts` which made finding the textures hard, so I moved them.

## Technical details

just removed / modified a few lines in `AreaInfoSystem.cs`

## Media

hive core now displays with the level 4 icon, pylon with level 3.

## Requirements

**Changelog**

:cl: noctyrnal
- tweak: The hivecore and hive pylons no longer have unique identifiers on the Roof Alert. The only difference now is in the protection information.


DoguhanOzgurAkca:
This is parity. Hive core and Hive pylon create resin roofs so their influence can be seen if you inspect a roof. 
It also makes sense. They create literal resin roofs that marines can see. That is how they block CAS/OB etc.

Technically it can be made so that you don't have the area alert on Almayer. This was kinda requested because shipside roles has no reason to have it.

PR #7533, opened 2025-08-18
Adds CMB and ZSPA shuttles

SharkSnake98:
## About the PR
Adds a pair of CMB and ZSPA shuttle, useable for PVE or Admin events, and a pair of CMB and ZSPA shuttles for ERTs.

## Why / Balance
Would be nice to have more distinct shuttles with unique atmospheres and looks, and the CMB deserves a shuttle of its own.

## Technical details
YML

### CMB ERT
image

### ZSPA ERT
image

###ZSPA
image

###CMB
image

## Requirements

**Changelog**
:cl:
- add: Added 4 new shuttles, two for ERT and two for PvE/Admin events
- add: Added spawners for the new shuttles


QuietlyWhisper:
You'd need to expand the ERT spawn area if you ever want to see these be used (90% of staff will not have access to mapping permissions to load these)

Additionally, the TSE/TSE ERT already has a dropship, and you need to make sure these can actually fit on the Almayer's external hangers, which iirc are noticeably smaller (I can't really tell if they would just by looking at it, maybe count the tiles)

The CMB/CMB ERT might already have a dropship in parity, because they are called when cargo abuses black market, so you might want to look into that as well

If in comparison we like these dropships more though, they could just replace some of the current ones

ShootsTheMoon:
I like that they have attachment points for machine guns we don’t have yet :)

QuietlyWhisper (changes requested):
see above

SharkSnake98:
> Additionally, the TSE/TSE ERT already has a dropship, and you need to make sure these can actually fit on the Almayer's external hangers, which iirc are noticeably smaller (I can't really tell if they would just by looking at it, maybe count the tiles)

The ERT shuttle does fit in the external docking bays, I counted the tiles; the other versions are intended to be used for PvE.

SharkSnake98:
> Whisper's comments + You should be using the "Empty Space" tiles under the external walls where plating is visible.

I will try but the TSE ships do seem to have the tiles hardcoded into their sprites. I'll see what I can do.

PR #7593, opened 2025-08-23
Mateba / WW can fire in the air while moving

noctyrnal:
## About the PR

Title, Do-After no longer breaks while moving. Flare gun do-after unchanged for air-shots.

## Why / Balance

Feels silly that a CO is forced to stand still to fire in the air, also prevents people moving them to stop the do-after.

## Technical details

Added 1 variable to the `RMCAirShotComponent`, BreakOnMove is now tied to the variable which is true by default.

## Media

https://github.com/user-attachments/assets/39e598cc-718d-42c6-b472-f8be1a2c928c

## Requirements

**Changelog**

:cl: noctyrnal
- tweak: You can now move while firing the Mateba or Winter Wyvern in the air.

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/RMCAirShotComponent.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/RMCAirShotComponent.cs
@@ -26,6 +26,12 @@ public sealed partial class RMCAirShotComponent : Component
     public bool RequiresCombat;
 
     /// <summary>
+    ///     If the do-after breaks when moving.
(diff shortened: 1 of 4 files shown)


Aurallianz:
Technically not parity with 13.
Just marking this down.

Arcticular1:
> Technically not parity with 13. Just marking this down.

Most of the cosmetic options / vanity options are not parity for CO (they get more here).

DrSmugleaf (approved):

PR #7611, opened 2025-08-24
Implements laser targeting mode for mortars

chavonadelal:
## About the PR

Adds an targeting toggle to the mortar, which lets users link a laser designator, and fire directly on that laser designators lases as an alternative to firing on coordinates.

## Why / Balance

Parity! This addition gives mortar operators a a second choice on how they can fire on a target in order to give them more flexibility.

"Lase mode", toggleable with Alt+Clicking on the mortar, lets mortar operators link any laser designator to the mortar. The mortar will then automatically aim at any laser mark lased by laser designator, as is the case with CAS targets. The benefits of this mode is that the mortar operator can fire on targets much faster, and the mortar shells have 100% accuracy in this mode. The drawbacks compared to coordinate mode, is that a lase has to be active for the mortar operator to fire on the target. The mortar operator also cannot offset the mortar, so it is strictly direct fire. If the laser designator stops the lasing, the mortar will automatically reset, unable to fire until a laser target is acquired again. The mortar operator cannot lase and try to fire the mortar at the same time, so they will need to find a buddy that can lase targets for them.

The aim is for the "laser designator targetting" mode to not be a direct upgrade or downgrade to coordinate firing mode, but a sidegrade, where the mortar operator can switch modes if needed.
## Technical details

The mortar system has been modified to implement a new targeting mode. The logic of all checks is preserved, the new mode does not ignore or change them, only the exceptions of the shell deviation are changed.
## Media

https://github.com/user-attachments/assets/f3e32b7d-a8dd-4275-8c76-db32d65486d7

## Requirements

**Changelog**

:cl: to4no_fix
- add: A second targeting mode added to the mortar — laser targeting mode. Switch the mortar to laser targeting mode, link your laser designator (only one can be linked at a time), and then designate a target with it. In this mode, the mortar will automatically targeting on the laser mark, signaling you with a beep (its description will also change). After that, you only need to load a shell. This targeting method completely removes shell deviation — the shell will land exactly on target.
- tweak: In the mortar kit, rangefinders replaced by laser designators.

### DIFF
--- a/Content.Shared/_RMC14/Mortar/LinkMortarLaserDesignatorDoAfterEvent.cs
+++ b/Content.Shared/_RMC14/Mortar/LinkMortarLaserDesignatorDoAfterEvent.cs
@@ -1,0 +1,18 @@
+using Content.Shared.DoAfter;
+using Robust.Shared.Network;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared._RMC14.Mortar;
+
+[Serializable, NetSerializable]
+public sealed partial class LinkMortarLaserDesignatorDoAfterEvent : DoAfterEvent
+{
+    public NetEntity LaserDesignator;
+
+    public LinkMortarLaserDesignatorDoAfterEvent(NetEntity laserDesignator)
+    {
+        LaserDesignator = laserDesignator;
+    }
+
+    public override DoAfterEvent Clone() => this;
+}
--- a/Content.Server/_RMC14/Mortar/MortarSystem.cs
+++ b/Content.Server/_RMC14/Mortar/MortarSystem.cs
@@ -5,12 +5,14 @@ using Content.Shared._RMC14.Areas;
 using Content.Shared._RMC14.Map;
 using Content.Shared._RMC14.Mortar;
 using Content.Shared._RMC14.Rules;
+using Content.Shared._RMC14.Rangefinder;
 using Content.Shared.Maps;
 using Robust.Server.Containers;
 using Robust.Shared.Map;
 using Robust.Shared.Random;
 using Robust.Shared.Timing;
 using static Content.Shared.Popups.PopupType;
+using Content.Server.Chat.Systems;
 
 namespace Content.Server._RMC14.Mortar;
 
@@ -25,6 +27,7 @@ public sealed class MortarSystem : SharedMortarSystem
     [Dependency] private readonly RMCPlanetSystem _rmcPlanet = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly ChatSystem _chat = default!;
 
     protected override bool CanLoadPopup(
         Entity<MortarComponent> mortar,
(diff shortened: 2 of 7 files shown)


Aurallianz:
dreamseeker_3cL3vLPTBv

~~it should be one lasedes~~

~~If you move away and cancel the linking do-after it breaks the mortar permanently~~

Im not sure if this is intentional but when you switch the mortar mode it clicks SEVERAL times

There should be a slight delay before you can actually fire a mortar while the mortar "adjusts" to aim at the laser.

The mortar needs to ensure that it cannot target a Lasedes that's too close to the mortar (It should make a separate popup and noise if it is too close of a Lasedes).

chavonadelal:
> If you move away and cancel the linking do-after it breaks the mortar permanently

what do you mean?

Laser targeting mode does not change the checks, the mortar will not fire if it is too close or too far or the target is under a roof

Aurallianz:
> > If you move away and cancel the linking do-after it breaks the mortar permanently
> 
> what do you mean?
> 
> Laser targeting mode does not change the checks, the mortar will not fire if it is too close or too far or the target is under a roof

I've clarified above

Content.Shared/_RMC14/Mortar/LinkMortarLaserDesignatorDoAfterEvent.cs
```diff
@@ -0,0 +1,18 @@
+using Content.Shared.DoAfter;
+using Robust.Shared.Network;
+using Robust.Shared.Serialization;
+
+namespace Content.Shared._RMC14.Mortar;
+
+[Serializable, NetSerializable]
+public sealed partial class LinkMortarLaserDesignatorDoAfterEvent : DoAfterEvent
```
DrSmugleaf:
```suggestion
public sealed partial class LinkMortarLaserDesignatorDoAfterEvent : SimpleDoAfterEvent
```

Content.Shared/_RMC14/Mortar/LinkMortarLaserDesignatorDoAfterEvent.cs
```diff
@@ -0,0 +1,18 @@
...
+{
+    public NetEntity LaserDesignator;
+
+    public LinkMortarLaserDesignatorDoAfterEvent(NetEntity laserDesignator)
+    {
+        LaserDesignator = laserDesignator;
+    }
+
+    public override DoAfterEvent Clone() => this;
```
DrSmugleaf:
```suggestion
```

DrSmugleaf (changes requested):
Thanks for the PR!

DrSmugleaf:
> Im not sure if this is intentional but when you switch the mortar mode it clicks SEVERAL times
> 
> There should be a slight delay before you can actually fire a mortar while the mortar "adjusts" to aim at the laser.
> 
> The mortar needs to ensure that it cannot target a Lasedes that's too close to the mortar (It should make a separate popup and noise if it is too close of a Lasedes).

Also need this addressed

chavonadelal:
Done. There are also warnings about invalid targets (too close, etc.) and about the target being lost.

https://github.com/user-attachments/assets/3cc3b0a8-5a41-4ee3-9b43-88a454a20c18

DrSmugleaf (approved):
Very nice pr ty

PR #7660, opened 2025-08-26
Make `setlooc` permission require +ADMIN not +SERVER

PursuitInAshes:
## About the PR
`setlooc` is now useable inherently with the +Admin flag rather than being restricted to +SERVER.

## Why / Balance
This feels much more suitable and is consistent with the OOC and DChat toggles. Doesn't make sense for admins to have only 2/3 (Esp when the other two are much more serious and the outlier is less severe imo). Restricting +SERVER is important, but we can give this to +ADMIN without causing issues imo.

Whisper semi-approval via discord.

## Technical details
Changed a permission flag.

## Media
N/A

## Requirements

**Changelog**
- tweak: Setlooc is now useable with the +ADMIN permission instead of the +SERVER permission.

### DIFF
--- a/Content.Server/Chat/Commands/SetLOOCCommand.cs
+++ b/Content.Server/Chat/Commands/SetLOOCCommand.cs
@@ -6,7 +6,7 @@ using Robust.Shared.Console;
 
 namespace Content.Server.Chat.Commands;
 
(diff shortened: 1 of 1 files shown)


Aurallianz:
LGTM

PR #7664, opened 2025-08-26
Leaping parasites no longer collide with the resin door they start under

blueDev2:
## Technical details

On leap, parasite gains the "PreventCollideComponent" which prevents collision with the closest colliding resin door.
Once the leap ends, parasite loses this component.
## Media

https://github.com/user-attachments/assets/37717d71-e74a-453a-97b0-30e2136489ab

## Requirements

**Changelog**

:cl: blueDev2
- fix: Fixed leaping parasites colliding with the resin door they start under.
- fix: Fixed parasites not becoming visible over doors during leap doafters. This makes them able to be hit by bullets and melee attacks.


Aurallianz:
LGTM

Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -297,6 +318,8 @@ private void OnParasiteLeapDoAfter(Entity<XenoParasiteComponent> ent, ref XenoLe
 
     private void OnParasiteLeapStopped(Entity<XenoParasiteComponent> ent, ref XenoLeapStoppedEvent args)
     {
+        RemComp<PreventCollideComponent>(ent);
```
DrSmugleaf: This might get run during state application, and if it does it will make the client spam refresh PVS if its removed instantly instead of deferred
```suggestion
        RemCompDeferred<PreventCollideComponent>(ent);
```

DrSmugleaf (dismissed):
Thanks for the PR!
Code looks good, but before we merge this we need parasites to layer over doors while channeling and using leap, and for them to be able to be hit while doing so

image

blueDev2:
> Thanks for the PR! Code looks good, but before we merge this we need parasites to layer over doors while channeling and using leap, and for them to be able to be hit while doing so
> 
> image

When the parasite it channeling, do you want any shot hitting the resin door to hit the parasite or should it require a direct shot like resin traps require?

DrSmugleaf:
> When the parasite it channeling, do you want any shot hitting the resin door to hit the parasite or should it require a direct shot like resin traps require?

I don't think shots can, maybe if it's a pb? Would have to test it, but melee hits definitely can

blueDev2:
> > When the parasite it channeling, do you want any shot hitting the resin door to hit the parasite or should it require a direct shot like resin traps require?
> 
> I don't think shots can, maybe if it's a pb? Would have to test it, but melee hits definitely can


https://github.com/user-attachments/assets/238b54de-54ce-4c93-9d97-ee32cfc49b1a

PBs do work, which does cancel the leap

DrSmugleaf (approved):

PR #7684, opened 2025-08-27
Added mutineer leader

### LINKED ISSUE #7655: Add mutineer leader
## Description
In CM mutineers work a bit like revs, like with head revs and all that. I don't think we want that; however, the idea of a head mutineer could help with reducing the likely hood of a mutiny just going guns blazing and gives people a person to rally behind.

Now no matter how hard i tried to i could not get the "convert" ability to work for the leader mutineer on other people, this leads me to assume it uses consent of the person on the other end, and i could be reasonable on board with that. 
You should not be forced to be a mutineer, but it should be offered and if a mutiny were in effect, it would help a ton if there was an IC way for them to become a mutineer

TLDR: 
1. Make it so there is a Leader mutineer antag smite. (give it a mutineer icon with a gold outline or something)
2. Give that mutineer leader the ability to ask people if they want to be a mutineer.

rokudara-sen:
## About the PR
Closes #7655 
- Add mutineer leader component with consent-based recruitment ability
- Localize leader status messages and mutiny invitation dialog

## Why / Balance
- Gives mutinies a leader to coordinate with willing recruits without forcing players into the role

## Technical details
- Leader component grants a recruit action and uses shared mutineer icon
- Mutiny system validates recruit targets and opens an invitation EUI
- Adds localization strings and an action prototype for recruitment

Tested:
Not being able to recruit someone that has already been recruited.

## Media
image
image
image
image
image

## Requirements

**Changelog**
:cl:
- add: Added a mutineer leader role that can invite marines to mutiny through a consent window.

### DIFF
--- a/Content.Server/_RMC14/Marines/Mutiny/MutinySystem.cs
+++ b/Content.Server/_RMC14/Marines/Mutiny/MutinySystem.cs
@@ -164,10 +164,7 @@ public sealed class MutinySystem : SharedMutinySystem
         if (!TryComp<MindContainerComponent>(args.Target, out var mind) || !mind.HasMind)
             return;
 
-        if (_jobs.MindHasJobWithId(mind.Mind, "CMMilitaryPolice") ||
-            _jobs.MindHasJobWithId(mind.Mind, "CMMilitaryWarden") ||
-            _jobs.MindHasJobWithId(mind.Mind, "CMChiefMP") ||
-            HasComp<SynthComponent>(args.Target))
+        if (HasComp<MutinyImmuneComponent>(args.Target) || HasComp<SynthComponent>(args.Target))
             return;
 
         if (!TryComp<ActorComponent>(args.Target, out var actor))
--- a/Content.Shared/_RMC14/Marines/Mutiny/MutinyImmuneComponent.cs
+++ b/Content.Shared/_RMC14/Marines/Mutiny/MutinyImmuneComponent.cs
@@ -1,0 +1,12 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Marines.Mutiny;
+
+/// <summary>
+///     Marks a character as ineligible for mutiny recruitment.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class MutinyImmuneComponent : Component
+{
+}
+
--- a/Resources/Prototypes/_RMC14/Roles/Jobs/AuxiliarySupport/senior_enlisted_advisor.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Jobs/AuxiliarySupport/senior_enlisted_advisor.yml
@@ -85,6 +85,7 @@
         sprite: /Textures/_RMC14/Interface/map_blips.rsi
         state: background_medical
     - type: RMCTrackable
+    - type: MutinyImmune
 
 - type: startingGear
   id: CMGearSeniorEnlistedAdvisor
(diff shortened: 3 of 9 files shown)


rokudara-sen:
Please, if you have any test case ideas that I should try first, comment here.

BreakpointDebug1234:
Can you recruit MPs (MP, MW, CMP) to the mutiny?
Because you shouldn't be able to, server rules say that MPs may only side with commander or remain NEUTRAL.

What about synths?
I don't think there's a rule about them but it would be silly for synths to be able to mutiny.

crazy1112345:
> Can you recruit MPs (MP, MW, CMP) to the mutiny?
> Because you shouldn't be able to, server rules say that MPs may only side with commander or remain NEUTRAL.
> 
> What about synths?
> I don't think there's a rule about them but it would be silly for synths to be able to mutiny.

Synth mutiny stuff will be covered in their programming, we dont have anything about synths released yet because synths dont actually exist yet.

XxSWAG-MASTER-OVERLORDxX:
As per SOP synths and MP staff can't join a mutiny, so if that could be made a mechanic it would be nice (make it so it just does not work on a synth or a MP)

NEVERMIND, I SEE YOU ALREADY DID THAT

x5tink:
It might sound dumb, but maybe also stop you being able to recruit the XO/CO as well? and funny as that would be, its not exactly SOP but yeah.....

AverageNotDoingAnythingEnjoyer:
> It might sound dumb, but maybe also stop you being able to recruit the XO/CO as well? and funny as that would be, its not exactly SOP but yeah.....

It's not rule allowed too, so it's a good idea, yes. Include SOs too tho

QuietlyWhisper:
Approve of the concept and system, as a code review I recommend using a component on the job yml like "MutinyImmune" or something and checking for that component instead of doing a hardcoded check of specific jobs, this would mean making adding any future jobs to the mutiny immune list easier and not have to add to the code, as well as keeping the code overall cleaner

rokudara-sen:
Okay, the following roles have this new component: MP, Warden, CMP, SEA, SO, XO, and CO

HolySSSS:
What about corporate liaison? I am somewhat concerned to see one joining the mutiny for any reason, more so one attacking command or MP.

mmails:
Add this to synthetics too. They are unable to partecipate in a mutiny. They can either defend command (non lethally) or stay neutral

QuietlyWhisper:
Civilian Correspondants as well, good catch

Frogge6645:
Would be kind of awesome to add the upstream rev theme whenever you join up as a callback :))))

XxSWAG-MASTER-OVERLORDxX:
> Would be kind of awesome to add the upstream rev theme whenever you join up as a callback :))))

you and i both know, as cool as it would be. it would set the wrong precedent around what you are supposed to do
HOWEVER, i think a short custom theme could be ok to add at a later date

rokudara-sen:
Now that I think about it, the component should be added to all non-UNMC roles, no?

XxSWAG-MASTER-OVERLORDxX:
> Now that I think about it, the component should be added to all non-UNMC roles, no?

smug said "yeh"

Aurallianz:
Why are we making certain groups immune? Just make certain groups recruitable no?

AverageNotDoingAnythingEnjoyer:
> Why are we making certain groups immune? Just make certain groups recruitable no?

This might be a better idea, considering the amount of survs variants to be ported still and many possible event and PvE roles to be made/ported in the future.

ShootsTheMoon:
Should the mutineer leader be given leadership skill?

QuietlyWhisper:
> ShootsTheMoon

this is not within the scope of this pr

QuietlyWhisper:
> Why are we making certain groups immune? Just make certain groups recruitable no?

Whichever results in less effort needed, whitelist or blacklist the end result is the same
Though, you're probably right in that it makes more sense to do who is able to be recruited rather than not, since that's only almayer-based roles vs the entire codebase and future content

Sorry for the inconvenience, @Aurallianz is correct in that being the more effective implementation.

PR #7704, opened 2025-08-28
Make RMCPopOutWindow use OSWindow instead

PJB3005:
Removes manual usage of CreateWindowRoot. OSWindow is more convenient to use and is more future-proof.

### DIFF
--- a/Content.Client/_RMC14/UserInterface/RMCPopOutWindow.xaml.cs
+++ b/Content.Client/_RMC14/UserInterface/RMCPopOutWindow.xaml.cs
@@ -18,7 +18,7 @@ public partial class RMCPopOutWindow : DefaultWindow
     public event Action? OnFinalClose;
     public event Action? OnPopout;
 
-    private IClydeWindow? _popOutWindow;
+    private OSWindow? _popOutWindow;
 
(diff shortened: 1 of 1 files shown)


Content.Client/_RMC14/UserInterface/RMCPopOutWindow.xaml.cs
```diff
@@ -40,30 +40,25 @@ public void SetBui(BoundUserInterface bui)
         button.OnPressed += _ =>
         {
             OnClose -= FinalClose;
-            var clyde = IoCManager.Resolve<IClyde>();
-            var monitor = clyde.EnumerateMonitors().First();
```
PJB3005, line 44: I wasn't sure why you were intentionally selecting the first monitor. I'm not even sure that's guaranteed to be the *primary* monitor. Is this necessary? If so I can restore it (it'd need an engine update though).

DrSmugleaf: No real reason other than just passing a monitor into the args, wasn't sure if it was necessary to pass one in or not

PR #7760, opened 2025-09-01
Defenders can tailslam while fortified again.

Spaghetti-bit:
## About the PR

Fixes defenders being too chunky for their own good, allowing them to tail slam while fortified again.

## Why / Balance

bug fix.

## Technical details

Actions that check for "InRangeUnObstructed" are no longer blocked by the user's collision.
A basic predicate was added to the function in `OnInRangeUnobstructedUseAttempt` of `RMCActionsSystem.c`
> Is user entity and obstructed? No? Obstructed. Is user entity and obstructed? Yes? Unobstructed. 

## Requirements

**Changelog**

:cl:
- fix: Fixed defenders being unable to tail slam while fortified.

### DIFF
--- a/Content.Shared/_RMC14/Actions/RMCActionsSystem.cs
+++ b/Content.Shared/_RMC14/Actions/RMCActionsSystem.cs
@@ -164,8 +164,7 @@ public sealed class RMCActionsSystem : EntitySystem
         if (args.Target is not { } target)
             return;
 
-        EntityUid userUid = args.User;
-        if (!_interaction.InRangeUnobstructed(ent.Owner, target, ent.Comp.Range, predicate: entity => entity == userUid))
+        if (!_interaction.InRangeUnobstructed(args.User, target, ent.Comp.Range))
             args.Cancelled = true;
     }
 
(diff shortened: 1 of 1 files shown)


Aurallianz:
Yay Lambda, LGTM

Content.Shared/_RMC14/Actions/RMCActionsSystem.cs
```diff
@@ -164,7 +164,8 @@ private void OnInRangeUnobstructedUseAttempt(Entity<ActionInRangeUnobstructedCom
         if (args.Target is not { } target)
             return;
 
-        if (!_interaction.InRangeUnobstructed(ent.Owner, target, ent.Comp.Range))
+        EntityUid userUid = args.User;
+        if (!_interaction.InRangeUnobstructed(ent.Owner, target, ent.Comp.Range, predicate: entity => entity == userUid))
```
DrSmugleaf: You should be able to use args.User instead of ent.Owner for the first parameter to fix this, check if that fixes it instead

Spaghetti-bit: Will do, give me a moment to boot it up. :)

Spaghetti-bit:
```suggestion
        if (!_interaction.InRangeUnobstructed(args.User, target, ent.Comp.Range))
```

Issue #7791, opened 2025-09-04
Fix `RMCAIConsoleHERMES` Name


PursuitInAshes:
## Description
Name has a `.` at the end

Issue #7589, opened 2025-08-23
SMG Arm brace bugged allowing things to be wielded and used as if both hands empty.


Thereisabear:
## Description
The SMG arm brace when locked allows things to that require wielding to be wielded and fired.  e.g UGLs, Spec weps.
Bullet accuracy for both guns is has a wide deviation but things like grenades flares and batons...no issue goes exactly where aimed.
Pistol in the other hand with arm braced smg bullet accuracy isn't affected. Fires as if wielded.

**Reproduction**
Lock arm brace on SMG.  
Grab weapon that requires wielding e.g M85A1  in other hand
Press Wield button.   Fire away.

Unable to attach clip of my testing here, will ping with shortclip

PR #7793, opened 2025-09-05
RMCAIConsoleHERMES fix

Spaghetti-bit:
## About the PR

M63 Armbrace
https://github.com/RMC-14/RMC-14/issues/7589

RMCAIConsoleHERMES:
https://github.com/RMC-14/RMC-14/issues/7791

## Why / Balance

bug fix

## Technical details

New local event subscription in RMCWieldableSystem. Iterates over the hands to make sure (after attempting to wield) that there isn't two different items in the player's hands. If there is? It forces an unwield.

## Requirements

**Changelog**
:cl: Spaghetti
- fix: Fixed AI Hermes console name.

### DIFF
--- a/Content.Shared/_RMC14/Wieldable/RMCWieldableSystem.cs
+++ b/Content.Shared/_RMC14/Wieldable/RMCWieldableSystem.cs
@@ -133,19 +133,15 @@ public sealed class RMCWieldableSystem : EntitySystem
     private void OnItemWielded(Entity<WieldableSpeedModifiersComponent> wieldable, ref ContainerGettingRemovedAttemptEvent args)
     {
         if (!args.Cancelled || !TryComp(wieldable.Owner, out TransformComponent? xform) ||
-            !xform.ParentUid.Valid ||
-            !TryComp(xform.ParentUid, out HandsComponent? hands))
+                !xform.ParentUid.Valid ||
+                !TryComp(xform.ParentUid, out HandsComponent? hands))
         {
             return;
         }
-        foreach (var name in hands.Hands.Keys)
+
+        if (TryComp(args.EntityUid, out UnremoveableComponent? unremoveable))
         {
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Wieldable/RMCWieldableSystem.cs
```diff
@@ -126,6 +130,24 @@ private void OnItemWielded(Entity<WieldableSpeedModifiersComponent> wieldable, r
...
+        foreach (var name in hands.Hands.Keys)
+        {
+            if (hands.ActiveHandId == name)
+                continue;
+            if (name != null && wieldable.Owner.ToString() != name)
+            {
+                _wieldable.UnwieldAll(xform.ParentUid, force: true);
+            }
+        }
+    }
```
DrSmugleaf: Does subscribing to can drop item event on UnremoveableComponent not work for this? Either way you shouldn't be comparing a name to the string format of an entity's id

Spaghetti-bit: image

Another subscription is already set to exactly that, I cleaned up the code to not check for strings. Cleaned up the code :)

DrSmugleaf: That's for container removal, not drop attempts
DropAttemptEvent is the one you are looking for

DrSmugleaf: I see the problem, the drop attempt event isn't directed at the item
A new event is a cleaner way to do this, specially compared to subscribing to an unrelated gun component

DrSmugleaf:
Done in https://github.com/RMC-14/RMC-14/pull/7806

DrSmugleaf (dismissed):
Actually, going to merge this for the name fix

PR #7797, opened 2025-09-05
Fix mapped shotguns

noctyrnal:
M42A2 -> M42A1 for planet maps
cant add to migrations or it'll vanish all M42A2s (we want them on the almayer and in cases and whatnot)


Aurallianz (approved):
LGTM

DrSmugleaf (approved):

Issue #7769, opened 2025-09-02
xeno drag gets stopped by ANY tackle (marine and ally xeno)


Binbagg:
## Description
you can stop a xeno from dragging anything by right click tackling their sprite, as a marine and allied xeno

**Reproduction**
have xeno drag, tackle them as another xeno, both same hive, first xenos drag gets stopped

have xeno drag, tackle them as a marine, xenos drag gets stopped

also works with scout invis, and sometimes doesn't do it**Screenshots**

https://github.com/user-attachments/assets/3945f86c-014e-4058-bb73-f3821b00e0ca

Spaghetti-bit:
Peak? Just kidding, I'll make a fix, that doesn't seem intended.

DrSmugleaf:
Fixed in https://github.com/RMC-14/RMC-14/pull/7806

PR #7801, opened 2025-09-05
Xeno drags no longer get stopped by being disarmed or tackled.

Spaghetti-bit:
## About the PR

Fixes https://github.com/RMC-14/RMC-14/issues/7769

Leadered xenos still bully non-leadered xenos, still have the same affect as before; knock them down, stun for a short time... which makes the non-leadered xeno lose their pull.

## Why / Balance

bug fig

## Technical details

`RMCPullingSystem` watches for the `DisarmedEvent` and handles it without doing anything else if the target has the `XenoComponent`

## Requirements

**Changelog**

:cl: Spaghetti
- fix: Xenos no longer lose their drag/pull when disarmed/tackled by others.

### DIFF
--- a/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
+++ b/Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
@@ -9,6 +9,7 @@ using Content.Shared.Interaction;
 using Content.Shared.Interaction.Components;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.MouseRotator;
+using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Movement.Components;
 using Content.Shared.Movement.Pulling.Components;
 using Content.Shared.Movement.Pulling.Events;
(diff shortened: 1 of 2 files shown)


blueDev2:
I think you need to sort the XenoComponent disarm event handler so that it happens before the HandsComponent disarm event handler.

Aurallianz (changes requested):
This appears to work fine.
This breaks shove stunning for Marine Players.
Xeno Queen cannot shove stun...

Spaghetti-bit:
> This appears to work fine. This breaks shove stunning for Marine Players. Xeno Queen cannot shove stun...

What?? That doesn’t make any sense. I was able to tackle stun as defender and queen, I’ll triple check in a moment

Content.Shared/_RMC14/Pulling/RMCPullingSystem.cs
```diff
@@ -249,6 +253,13 @@ private void OnMeleePullAttempt(Entity<MeleeWeaponComponent> ent, ref PullAttemp
             args.Cancelled = true;
     }
 
+    private void OnDisarmed(Entity<XenoComponent> ent, ref DisarmedEvent args)
+    {
+        if (TryComp(args.Target, out XenoComponent? xeno))
+            args.PopupPrefix = "disarm-action-shove-";
+            args.Handled = true;
```
DrSmugleaf, line 260: Only the second line is wrapped in the if statement, either add curly braces or fix the indentation

DrSmugleaf:
Done in https://github.com/RMC-14/RMC-14/pull/7806

PR #6563, opened 2025-05-25
Splits marines in the crew manifest into squads

4310V343k:
## About the PR

Marines are now split into squads and are actually sorted based on their weight

## Why / Balance

I was playing with my friend and they said "I hate that I can't see which squad those marines belong to in the crew manifest". Well, now you can

## Technical details

There were multiple comments signifying the performance issues, so I tried to resolve them as I was modifying that code anyway. Client code is around two orders of magnitute faster due to a reverse lookup
Server code now build the crew manifest based on a queeu in Update, rather than on each change. Should significantly improve it's performance on round start (one build instead of n builds for n players)

## Media

https://github.com/user-attachments/assets/831799eb-7361-4b40-bedf-7f22824c2e58

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

- [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Marines in the crew manifest are now sorted and split into squads!

PR #7884, opened 2025-09-10
Squadded Crew Manifest fix

4310V343k:
## About the PR

Continuation fix for #6563
Mobs which are given a squad immediately after they spawn in get into the crew manifest too

## Why / Balance

It actually works instead of only after a squad change

## Technical details

`OnSquadMemberAdded` is called in the middle of the spawn, before the station record is created, so it wasn't updated as there was nothing to update yet. It is created in PlayerSpawnCompleteEvent, and we can update the squad afterwards

## Requirements

### DIFF
--- a/Content.Client/CrewManifest/UI/CrewManifestListing.cs
+++ b/Content.Client/CrewManifest/UI/CrewManifestListing.cs
@@ -101,7 +101,7 @@ public sealed class CrewManifestListing : BoxContainer
             AddChild(new CrewManifestSection(
                 _prototypeManager,
                 _spriteSystem,
-                Loc.GetString(department.Name),
+                department.Name,
                 depEntries));
         }
 
@@ -113,7 +113,7 @@ public sealed class CrewManifestListing : BoxContainer
             AddChild(new CrewManifestSection(
                 _prototypeManager,
                 _spriteSystem,
-                Loc.GetString(squad),
+                squad,
                 squadEntries));
         }
     }


Content.Client/CrewManifest/UI/CrewManifestListing.cs
```diff
@@ -101,7 +101,7 @@ public void RMCAddCrewManifestEntries(CrewManifestEntries entries)
             AddChild(new CrewManifestSection(
                 _prototypeManager,
                 _spriteSystem,
-                Loc.GetString(department.Name),
```
chavonadelal, line 104: for what? you shouldn't have touched it

Content.Client/CrewManifest/UI/CrewManifestListing.cs
```diff
@@ -113,7 +113,7 @@ public void RMCAddCrewManifestEntries(CrewManifestEntries entries)
             AddChild(new CrewManifestSection(
                 _prototypeManager,
                 _spriteSystem,
-                Loc.GetString(squad),
```
chavonadelal, line 116: Also there

chavonadelal (changes requested):

Content.Client/CrewManifest/UI/CrewManifestListing.cs
4310V343k: It's redundant, as stated in the commit message. CrewManifestSection also calls Loc.GetString before display

DrSmugleaf (approved):

PR #7900, opened 2025-09-13
Give the Queen, XO, and CO the bold leader chat that SLs and HLs get.

BramvanZijp:
## About the PR
Title, it also applies to the CO survivors.

## Why / Balance
Queen should make very logical sense. Your hive leaders get it to lead charges, but you should too if you are war queening.
For the CO, XO, and Surv COs, its both to allow them to be heard more easily when deploying, and to make briefings by them easier to hear. While this COULD be fixed with megaphone spam, I feel like this should be a thing if SLs get it too.

## Technical details
The queen has a special component that gives it the word of the queen ability, which now also grants the big chatbox.
For XO, CO, and the CO Survs, they now get a InnateCommandSpeechComponent, which grants it.

## Requirements

**Changelog**
:cl: BramvanZijp
- add: The Xeno Queen, XO, CO, and Survivor Commanders now get the bold chat font that SLs and HLs have.


Content.Client/Chat/UI/SpeechBubble.cs
```diff
@@ -251,6 +253,8 @@ protected override Control BuildBubble(ChatMessage message, string speechStyleCl
                     speechStyleClass = message.SpeechStyleClass;
                 else if (entityManager.HasComponent<SquadLeaderComponent>(senderUid) || entityManager.HasComponent<HiveLeaderComponent>(senderUid))
                     speechStyleClass = "commanderSpeech";
+                else if (entityManager.HasComponent<XenoWordQueenComponent>(senderUid) || entityManager.HasComponent<InnateCommandSpeechComponent>(senderUid))
+                    speechStyleClass = "commanderSpeech";
```
QuietlyWhisper, line 254: Remove the check for `XenoWordQueenComponent` and just add `InnateCommandSpeechComponent` to the queen prototype. I think it would be redundant, and this form of checking unrelated components is bad form.
```suggestion
                else if (entityManager.HasComponent<SquadLeaderComponent>(senderUid) || entityManager.HasComponent<HiveLeaderComponent>(senderUid) || entityManager.HasComponent<InnateCommandSpeechComponent>(senderUid))
                    speechStyleClass = "commanderSpeech";
```

QuietlyWhisper (changes requested):

Content.Client/Chat/UI/SpeechBubble.cs
BramvanZijp: Changes applied.

chavonadelal (changes requested):
You have invented the wheel, `RMCSpeechBubbleSpecificStyleComponent` was recently added, study it and add the same `commanderSpeech`style to its data field. Implemented specifically for such cases. No code changes required.

BramvanZijp:
> You have invented the wheel, `RMCSpeechBubbleSpecificStyleComponent` was recently added, study it and add the same `commanderSpeech`style to its data field. Implemented specifically for such cases. No code changes required.

Ill look into it, thank you.

BramvanZijp:
> You have invented the wheel, `RMCSpeechBubbleSpecificStyleComponent` was recently added, study it and add the same `commanderSpeech`style to its data field. Implemented specifically for such cases. No code changes required.

That solution breaks after using a megaphone, meaning until that is fixed, this is not re-inventing the wheel, simply inventing a properly functional wheel which does not yet exist.

chavonadelal:
> > You have invented the wheel, `RMCSpeechBubbleSpecificStyleComponent` was recently added, study it and add the same `commanderSpeech`style to its data field. Implemented specifically for such cases. No code changes required.
> 
> That solution breaks after using a megaphone, meaning until that is fixed, this is not re-inventing the wheel, simply inventing a properly functional wheel which does not yet exist.

I looked into it and there really is a bug, but it can be fixed in 10 lines of code. Instead of fixing the bug, creating a component that performs exactly the same function looks wrong.

DrSmugleaf (approved):
Should turn this into an event later but this is fine for now

PR #7914, opened 2025-09-15
fix girders being weak

Vixting:
## About the PR

Fix girders from being broken by low castes

## Why / Balance

parity

## Requirements

**Changelog**

:cl: Vixting
fix: fix girders being weak

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/ClawSharpness/XenoClawsSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/ClawSharpness/XenoClawsSystem.cs
@@ -1,4 +1,4 @@
-﻿using Content.Shared.Damage;
+using Content.Shared.Damage;
 using Content.Shared.Damage.Prototypes;
 using Content.Shared.Doors.Components;
 using Content.Shared.Weapons.Melee;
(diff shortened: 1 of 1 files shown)


Resources/Prototypes/_RMC14/Entities/Structures/Walls/girder.yml
```diff
@@ -55,6 +55,9 @@
       - !type:PlaySoundBehavior
         sound:
           collection: MetalBreak
+  - type: ReceiverXenoClaws
+    minimumClawStrength: Sharp
```
DrSmugleaf: This is supposed to be very sharp and tier < 2, so it doesn't apply to tier 3

Resources/Prototypes/_RMC14/Entities/Structures/Walls/girder.yml
```diff
@@ -104,6 +107,9 @@
...
```
DrSmugleaf: Same here

Vixting:
Changes made, should be fine now

Content.Shared/_RMC14/Xenonids/ClawSharpness/XenoClawsSystem.cs
```diff
@@ -55,8 +57,20 @@ private void OnReceiverDamageModify(Entity<ReceiverXenoClawsComponent> ent, ref
...
+                    damage = new DamageSpecifier(_protoManager.Index(_clawsDamageGroup), receiver.MaxHealth / receiver.HitsToDestroy);
+                }
+            }
+            else
+            {
+                if (claws.ClawType.CompareTo(receiver.MinimumClawStrength) >= 0)
+                {
+                    damage = new DamageSpecifier(_protoManager.Index(_clawsDamageGroup), receiver.MaxHealth / receiver.HitsToDestroy);
+                }
+            }
```
DrSmugleaf: These both need to be compared at the same time and zero the damage otherwise, for girders
From 13
```
f((M.caste && M.caste.tier < 2 && M.claw_type < CLAW_TYPE_VERY_SHARP) || unacidable)
		to_chat(M, SPAN_WARNING("Our claws aren't sharp enough to damage [src]."))
```

Vixting: Oh I see now

Vixting:
So as long as you meet either requirment you do dmg

PR #7924, opened 2025-09-15
Move automatic LZ selection to 10 minutes

DoguhanOzgurAkca:
## About the PR

Moves automatic  LZ selection from 25 to 10 minutes in
Sends a reminder to the command channel if they haven't picked the LZ yet at 8 mins

## Why / Balance

1) Its good for people to know which LZ it is before deployment time.
2) Events depend heavily on LZ and some preperation, this helps with that a lot. 

I  probably need a better text for the reminder, though. If anyone writes that, it would be good
## Requirements

**Changelog**

:cl:
- tweak: Moved automatic LZ selection from 25 minutes in to 10 minutes in. A reminder is sent in command channel if one is not picked in 8 minutes.


AverageNotDoingAnythingEnjoyer:
Are you sure it's a good idea?
Sometimes I prefer to announce the chosen LZ at brief or right after, from time to time people may have long discussions on what course of action is the best on the given map, etc.
In other words, 10 minutes is quite a small time frame.

BramvanZijp:
I feel like 10 minutes is too little time. I could see 15 minutes with a reminder at 12, which lines up with dropship refuel in 15, but 10 is too little IMO.

DoguhanOzgurAkca:
Ogey, gonna be a long post, but I wanted to explain the reasons.

So, lets get the facts in first.

1) You are allowed to deploy at 15 minutes.
2) Most deployments happen in between 15 and 20 minutes at the moment.
3) People like to deploy early, at least most of them, especially on some maps like Chances that warrants both IC and OOC extra attention and speed.

What are my issues with the LZ selection not happening early:

1)As a non-CiC marine player, especially one that plays bravo or plays objective based, I wanna know what gear I should get or what kind of a mental plan I should make. This changes a lot based on the LZ. Unlike others, I actually am very much affected by which LZ we are going with. An example is if we use batteries for comms, or the comms placement in general.
2) As a CiC player, I never had an moment where I couldn't have a LZ choice in mind by TEN minutes, even in new maps.
3) As the guy who made Aegis, and is working on the Behemoth
 - I need to have enough time to actually place all of the props, AND let both factions know BEFORE Marines start deploying or even the briefing. 
 - I wanna want the CiC and Xenos to have enough time to take in the information of the event, their position and plan something out.

The placement of basically any groundside objective depends on which LZ marines are going to be primarly landing at, so you HAVE to place groundside objective according to that.
 
Way to fix all of these problems, especially the third which is the most problematic, is to make it so that the LZ is selected a fair bit before the allowed deployment time (so 15 minutes). My first idea was to make it so that it was decided in the first 5-6 minutes actually, but some talks with the hosts just made me go with 10. 

Putting this timer on anything after 10 minutes has the following issues.
1) The person setting the event might or might not have enough time to properly set things up AND tell the factions in time.
2) Marines might already deploy or be deploying before they are informed of the extra objective on the ground, which goes for xenos who should try to protect it.
3) You are still not informed about which LZ you are going at.

I think if you think that 10 minutes is too short, you should instead try to change the 15 minute timer to deployment, and make it a different number closer to 20. 

@AverageNotDoingAnythingEnjoyer  @BramvanZijp

QuietlyWhisper (dismissed):
approve of concept

Content.Server/_RMC14/Dropship/DropshipSystem.cs
```diff
@@ -66,9 +70,12 @@ public sealed class DropshipSystem : SharedDropshipSystem
     private EntityQuery<DoorBoltComponent> _doorBoltQuery;
 
     private TimeSpan _lzPrimaryAutoDelay;
+    private TimeSpan _lzPrimaryReminderDelay;
     private TimeSpan _flyByTime;
     private TimeSpan _hijackTravelTime;
 
+    private bool _lzReminderSent;
```
DrSmugleaf, line 78: This will only be sent once until the entire server restarts
You need to reset this in RoundRestartCleanupEvent
This also needs to be able to be disabled for PvE

DrSmugleaf (changes requested):

PursuitInAshes:
I'm going to close this for the moment as the author has left and is unlikely to return to revive this.
If you wish to revive this, the above issues will need to be addressed. The easiest way would be a CVar which can be set for Normandy.

PR #7930, opened 2025-09-16
fix some storage oversights

QuietlyWhisper:
## About the PR

- Belts can hold anything that is a ballistic ammo provider, allowing unintended affects such as storing specifically underbarrel grenade launchers and underbarrel shotguns. This was fixed by blacklisting all underbarrel attachments, but is likely not the best solution to this. 
    - Alternatives could be increasing the item size of these problem attachments, at the cost of taking up more space in intended storages, like bags.

:cl:
- fix: Fixed being able to store underbarrel attachment guns in the ammo load rig.


DrSmugleaf (commented):
Mortar fix should be done now with https://github.com/RMC-14/RMC-14/pull/6326

Resources/Prototypes/_RMC14/Entities/Clothing/Belt/belts.yml
```diff
@@ -385,12 +388,18 @@
     tags:
     - RMCBeltMedical
 
+- type: Tag
+  id: CMBeltMortar
+
```
QuietlyWhisper:
```suggestion
```

Resources/Prototypes/_RMC14/Entities/Clothing/Belt/belts.yml
```diff
@@ -385,12 +388,18 @@
...
 - type: entity
   parent: CMBeltBaseStorage
   id: CMBeltMortar
   name: M276 pattern mortar operator belt
   description: An M276 load-bearing rig configured to carry ammunition for the M402 mortar, along with a sidearm.
   components:
+  - type: Tag
+    tags:
+    - CMBeltMortar
```
QuietlyWhisper:
```suggestion
```

PR #7944, opened 2025-09-18
Adds pre-25 hive core limited areas

DoguhanOzgurAkca:
## About the PR

PR adds the limitation from CM13 about hive core placement.
Before 25 minutes (I remember CM13 timers being funky, this might be a lower number like 20 because of the lobby counting or something) only certain locations can have a hive core

## Why / Balance

1) Frontline core is unsoul
2) Frontline core is kinda too good imo
3) Parity
## Technical details

Makes it so that every area that exists has this 25 minute limitation just like CM13
Adds an area component that makes areas exempt from this limitation 

Important problem:
In CM13, from what I could read, how the system works is that some areas DO have some parts of them be exempt, but it also makes it so that either all or the one area that queen spawns in gets this exemption as well. I couldn't properly understand it. 
I will be honest that I am a bit too busy to go over every area, and i don't trust myself on it too much. Some maps like Shiva or Trijent have too many small areas that could cause issue. I could ge tback to this, but if anyone else handles it, it would be better. 

Until then, if merged, this PR would fuck over a lot of maps like Trijent. Some maps like Solaris and Hybrisa SHOULD work fine however.

## Media

https://github.com/user-attachments/assets/f49a5b64-8827-4c12-82dc-9e0073f5a155

## Requirements

**Changelog**

### DIFF
--- a/Content.Shared/_RMC14/Areas/AreaComponent.cs
+++ b/Content.Shared/_RMC14/Areas/AreaComponent.cs
@@ -102,20 +102,18 @@ public sealed partial class AreaComponent : Component
     /// If null, no restriction applies. If set, hive cores cannot be built
     /// for this duration after round start.
     /// </summary>
-    [DataField]
+    [DataField, AutoNetworkedField]
     public TimeSpan? UnoviableTimer;
 
     /// <summary>
     /// Whether the unoviable timer is currently active and blocking hive core construction.
-    /// Server-side only, no need for networking.
     /// </summary>
-    [DataField]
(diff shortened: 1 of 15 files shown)


DoguhanOzgurAkca:
Seems like the video I sent is wrong as well, lol. The SDMM says both of those areas that I show should be blocked.

Again, someone has to go through maps for this to work.

PursuitInAshes:
Opting to close this as the Author is no longer apart of RMC and is unlikely to return. This has already been revived at #9821 .

PR #7980, opened 2025-09-22
Make welding tank explosions actually scary.

OmerF-afk:
## About the PR

It changes the Fuel tank's YML, making the explosion more scary. In parity, wise explosion is way scarier from my tests. I've been able to observe 170 damage burn and brute on drones 1 tile close to explosion and such, which goes to 90-90 and 30-30 on the second and third tile. I tried to get it closer to that by experimenting with ExplosionUI for a while. Should be way closer to what it feels like in 13. (Still missing the flame tiles. It should spawn while exploding.) Also changes the explosion type to RMC.
## Why / Balance

Parity. Explosions from the welding tanks actually have to hurt.
## Technical details

.YML change
## Media

image

## Requirements

**Changelog**

:cl:
- tweak: Adjusted the Fuel Tank Explosions to be scarier.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Storage/reagent_tank.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Storage/reagent_tank.yml
@@ -94,8 +94,8 @@
   - type: Explosive
     explosionType: RMC
     totalIntensity: 120
-    IntensitySlope: 6.0
-    MaxIntensity: 450.0
+    intensitySlope: 6.0
+    maxIntensity: 450.0
   - type: ReagentTank
     tankType: Fuel
   - type: Destructible
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Storage/reagent_tank.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Storage/reagent_tank.yml
@@ -92,8 +92,10 @@
         - ReagentId: WeldingFuel
           Quantity: 1000
   - type: Explosive
-    explosionType: Default
-    totalIntensity: 60
+    explosionType: RMC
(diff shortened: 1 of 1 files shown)


Aurallianz (changes requested):
failures related

Resources/Prototypes/_RMC14/Entities/Structures/Storage/reagent_tank.yml
```diff
@@ -92,8 +92,10 @@
         - ReagentId: WeldingFuel
           Quantity: 1000
   - type: Explosive
-    explosionType: Default
-    totalIntensity: 60
+    explosionType: RMC
+    totalIntensity: 120
+    IntensitySlope: 6.0
+    MaxIntensity: 450.0
```
Arykh:
```suggestion
    intensitySlope: 6.0
    maxIntensity: 450.0
```

Arykh: Apparently the first letters should be lowercase.

OmerF-afk: Ah! Thank you.

DrSmugleaf (changes requested):
Need fuel tanks to not explode when destroyed by xenos and non-bullets non-explosive-sources before this is merged

Dutch-VanDerLinde:
Probably just check for damage type trigger, check for piercing.
If it's not piercing then just destroy the tank as normal

OmerF-afk:
image

I have no idea also implement this functionality. Does this self implement itself with RMC explosion type?

PR #7986, opened 2025-09-23
Combat Correspondent read-only channels added

CatAndHats:
## About the PR
Creates a new correspondent encryption key that makes command, intel, and jtac muted, and removes the marine command and common encryption keys from the reporter headset.

## Why / Balance
As per a discord feature discussion around how Command comms are always filled with chatter making it hard to use and that currently 17 roles have command comms access, Whisper said that they would consider pruning access to Command comms (along with a cultural change of telling people to clear comms), and when anyone discussed removing command comms from roles the Correspondent was always in their lists. However now we have the ability to create read-only channels; so I copied the muted channels that the liason has to make sure the correspondents aren't filling vital comms with chatter.

## Technical details
simple yml. A copy of the marinecommand encryption key with a name change and command/intel/jtac set as muted, marine common is now the reporter default comms channel. A new key was created because there was no other key usable.

## Requirements

**Changelog**
:cl:
- tweak: Correspondents now have the Command, JTAC, and Intel channels as read-only.


Aurallianz:
Hi, this would be better served by keeping those comms but using this when it's merged. #8043

QuietlyWhisper (changes requested):
see above

CatAndHats:
> see above

Changes made

PR #7995, opened 2025-09-24
Make the command pouch no longer corrodible by xenos.

BramvanZijp:
## About the PR
Title, command pouch can no longer be melted by xeno acid.

## Why / Balance
Doublechecked this with warcat in deadchat, but this pouch is designed to contain a CO's sidearm and the Command Tablet, both of which are key items that are not intended to be corrodible by xenos. The fact that the pouch itself IS corrodible seems like a bug/oversight.

## Requirements

**Changelog**
:cl: BramvanZijp
- fix: The Commanding Officer's Command Pouch is no longer corrodible/meltable by Xenos.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Clothing/Pouches/storage.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Clothing/Pouches/storage.yml
@@ -149,6 +149,8 @@
     whitelist:
       tags:
       - Sidearm
+  - type: Corrodible
+    isCorrodible: false
 
 #Filled Command Pouch
 - type: entity


AverageNotDoingAnythingEnjoyer:
Actually, is it intended that Acid destroys the items within a container too? 
The confusion might be because of that.

Dutch-VanDerLinde:
> Actually, is it intended that Acid destroys the items within a container too? 
> 
> The confusion might be because of that.

Acidified container supposed to drop the items

AverageNotDoingAnythingEnjoyer:
Seems PR is redundant then, we need this thing ported instead.

BramvanZijp:
Then somebody will have to fix that, but I still feel like given this is a CO exclusive item that cannot be replaced if lost, it should still be unmeltable.

BramvanZijp:
I will still type up a bug report on the item dropping thing though.

DrSmugleaf:
I have emptying storages of non corrodible items when aciding them done for the next update, should fix this

pumpkin-pwr:
> I have emptying storages of non corrodible items when aciding them done for the next update, should fix this

Oh, does this mean we'll be allowed to melt stasis bags since it's fixed?

PR #8022, opened 2025-09-29
Sl Squad Kit recommendation change

CatAndHats:
## About the PR
Changed the squad leader kits from all being Mandatory to having the M4SPR, M54C, and Engineering Supply kits being recommended.
Moved the engineering kit closer to the top so that all the recommended kits were together

## Why / Balance
Recommended items are a useful feature for players new to roles, having a handy pointer for newer SLs is a positive change. It is better for players new to roles to be using simpler weapons that they are more likely to be comfortable with rather than options like the XM88 or M34 - new SLs should be focusing on learning how to lead and control their squad and the recommended options here are best for giving them things that they will have used many times before.

Like all recommended items in the vendors, once people are comfortable they start taking the other options, so this isn't to try and enforce what people are taking - it's simply a handy pointer for newer players.

Feature Discussion:
https://discord.com/channels/1168210010233376858/1421493073552212061

## Technical details
simple yml.

## Media
Old
image

New
image

## Requirements

**Changelog**
:cl:
- tweak: The M4SPR, M54C, and Engi Supply Kits are now shown as recommended in the SL vendor.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/leader.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/leader.yml
@@ -23,7 +23,9 @@
       takeOne: RMCSquadKit
       entries:
       - id: RMCKitRifleM4SPR
+        recommended: true
       - id: RMCKitCM54C
+        recommended: true
       - id: RMCKitPyroMini
     #- id: CMM2CHeavyMachineGun
     #- id: CMM56DHeavyMachineGun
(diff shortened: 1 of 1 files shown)


DrSmugleaf (changes requested):
If we want to recommend any of these, we should just recommend the engineering kit, the m4spr and m54c kits only give you the gun, ap ammo, and attachments

CatAndHats:
> If we want to recommend any of these, we should just recommend the engineering kit, the m4spr and m54c kits only give you the gun, ap ammo, and attachments

Changes made

PR #8064, opened 2025-10-06
Changes to shard gained numbers on slash and damage recieved, also fix a potentional lag causer

yagwog:
## About the PR
The numbers were a bit off for shard gain, I've since corrected them.

I have also specified a rejuv method for the handler that properly resets hedgehog rav's abilities, shard gain and sets shards to 150 to be enable easier testing going forward.

## Why / Balance
n/a

## Technical details
Reinforced timedespawn, even if clean up isn't needed just as a fallback.

XenoProjectileShooterComponent kept references to already deleted projectiles, so when trying to sync the comp state, it tried to reference deleted entities causing mass "Can't resolve MetaDataComponent" errors which caused huge amount of lag with every added rav spike used.

Uncertain if this could be happening with other projectiles but glad I have seemingly fixed this, I might be completely wrong but who knows, it isn't my strong suit.

## Media
n/a

## Requirements

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
@@ -124,6 +124,9 @@ public sealed class XenoProjectileSystem : EntitySystem
 
     private void OnShotRemove<T>(Entity<XenoProjectileShotComponent> ent, ref T args)
     {
+        if (_timing.ApplyingState)
+            return;
+            
         if (ent.Comp.ShooterEnt is not { } shooter)
             return;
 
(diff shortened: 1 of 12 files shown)


Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
```diff
@@ -306,4 +309,26 @@ public bool TryShoot(
...
+            }
+            
+            if (toRemove.Count > 0)
+            {
+                foreach (var shot in toRemove)
+                    comp.Shot.Remove(shot);
+                Dirty(uid, comp);
+            }
+        }
+    }
```
DrSmugleaf: Can't get this to error without this in local
```suggestion
```

Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
```diff
@@ -124,6 +124,9 @@ private void OnShooterRemove<T>(Entity<XenoProjectileShooterComponent> ent, ref
 
     private void OnShotRemove<T>(Entity<XenoProjectileShotComponent> ent, ref T args)
     {
+        if (_timing.ApplyingState)
+            return;
+
```
DrSmugleaf:
```suggestion
```

DrSmugleaf (changes requested):

DrSmugleaf (approved):

PR #8092, opened 2025-10-08
Remove cyanide pill from CL briefcase for real

noctyrnal:
## About the PR

Title

## Why / Balance

Since the update all i've seen it be used for is to be a shitter and poison other people, its the same fiasco as when we had doctors making drugs, or sashimi poisoning people. All it does it cause problems, It can be re-added when people can behave with it.

## Requirements

**Changelog**

:cl: noctyrnal
- remove: Removed the cyanide pill from the liaison's briefcase.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Devices/portable_vendor.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Devices/portable_vendor.yml
@@ -41,7 +41,7 @@
       #  points: 30
       #- id: /obj/item/reagent_container/hypospray/autoinjector/ultrazine/liaison
       #  points: 25
-      #- id: RMCPillCyanide
+      #- id: RMCPillCyanide # https://github.com/RMC-14/RMC-14/pull/8092
       #  name: cyanide pill
       #  points: 20
       #- id: /obj/item/trash/ceramic_plate


yagwog:
I agree it should be removed, even though I wish they weren't.

The pill can create very interesting RP scenarios, as long as the CL has good RP intentions to do it, I.E having a dispute with a we-ya scientist colonist who doesn't agree to disclosure and says they'll tell the world about experiments, then the CL "shares a drink" with them.

But in reality, CL's have been spiking random drinks, leaving the pill out on purpose pretending "oh i didn't think someone would eat it" and just being LRP using it to tempt providence. 

At least the current implementation doesn't work, unless it's time locked behind playtime or something, it's caused nothing but problems / LRP

AverageNotDoingAnythingEnjoyer:
I'd rather it just moved to CL's antag section once we have that, not removed outright as implied here

Resources/Prototypes/_RMC14/Entities/Objects/Devices/portable_vendor.yml
```diff
@@ -41,9 +41,9 @@
       #  points: 30
       #- id: /obj/item/reagent_container/hypospray/autoinjector/ultrazine/liaison
       #  points: 25
-      - id: RMCPillCyanide
-        name: cyanide pill
-        points: 20
+      #- id: RMCPillCyanide
```
DrSmugleaf:
```suggestion
      #- id: RMCPillCyanide # https://github.com/RMC-14/RMC-14/pull/8092
```

DrSmugleaf (dismissed):

PR #8144, opened 2025-10-14
Zombies no longer drop their headsets and gloves on turn

noctyrnal:
title, simple PR

makes spawning zombies en masse x1000 easier as GMs / Admins no longer have to delete dropped gloves and headsets for each spawned zombie.


noctyrnal:
one day we'll have proper zombie spawners and no weird LRP infection pop ups

PR #8150, opened 2025-10-14
xeno default vision to half

Vixting:
## About the PR

xeno default vision to half

## Why / Balance

parity

## Requirements

:cl: Vixting
tweak: xeno default vision to half

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
@@ -239,7 +239,6 @@
   - type: WhitelistPickup
   - type: DropHeldOnIncapacitate
   - type: NightVision
-    state: Half
     alert: XenoNightVision
     overlay: true
   - type: ParalyzeOnPullAttemptImmune


NotInButIn:
Not parity. In cm13, you can set the default vision in the settings.

Vermidia:
It is the default pref, but yeah actual pref settings are parity and prob should be worked on at some point.

NotInButIn:
I don't really want to change my vision every time. It's very inconvenient.

BadaBoomie:
This could be fixed by having your sight mode saved like how ability order is saved now, rather than having it as the default

Vixting:
Erm okay

Vixting:
Added setting, should be fine now

NotInButIn:
Thanks for the setting and localization:3

Content.Client/_RMC14/NightVision/NightVisionSystem.cs
```diff
@@ -34,6 +37,16 @@ public override void Initialize()
 
     private void OnNightVisionAttached(Entity<NightVisionComponent> ent, ref LocalPlayerAttachedEvent args)
     {
+        if (_entity.HasComponent<XenoComponent>(ent))
+        {
+            var preferredMode = _cfg.GetCVar(RMCCVars.RMCPreferredNightVisionMode);
+            if (Enum.TryParse<NightVisionState>(preferredMode, out var mode))
+            {
+                ent.Comp.State = mode;
```
DrSmugleaf, line 45: This is only setting it on the client, you need to do it on the server as well by using INetConfigurationManager and getting the client's CVar
This will also change the state even if its given by an item and not supposed to be changeable, which is not intended if you only mean it to be for xenos

Issue #8156, opened 2025-10-15
Attacking Fake Hosts Increases Acider Reserves


YO0teacher:
The "training dummy" sometimes spawned in medbay by admins is meant for surgery practice, but these can be exploited during hijack. They still increase runner's acider reserves when slashing them, this allows aciders to "farm" the dummy for free acid. 

This isn't worth anybody's time to patch, I just wanted to report it so admins are aware the dummy may be exploited.

<img width="807" height="648" alt="Image" src="https://github.com/user-attachments/assets/952b0bf8-8d57-412e-9402-2909521a87f8" />

PR #8159, opened 2025-10-15
Aciders no longer gain acid from attacking training dummies

YO0teacher:
## About the PR
Slashing anything that has a training dummy component no longer gives the runner acider strain any acid gain. Both the instant acid gain and passive acid gain accounted for.

## Why / Balance
To address my own bug report I made here: https://github.com/RMC-14/RMC-14/issues/8156
Basically, training dummies can now be spawned in medbay by admins without worrying about people exploiting them during hijack as acider.

## Technical details
When hitting an entity, checks if the entity has a "RMCTrainingDummyComponent" and doesn't grant acid gain if it does.

## Media
https://github.com/user-attachments/assets/6fc61843-7803-407c-8203-e8adc611d3d4

## Requirements

:cl:
- fix: Fixed acider runners gaining acid from hitting training dummies.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
@@ -1,5 +1,6 @@
 using Content.Shared._RMC14.Xenonids.Energy;
 using Content.Shared._RMC14.Xenonids.Rest;
+using Content.Shared._RMC14.TrainingDummy;
 using Content.Shared.Mobs;
 using Content.Shared.StatusEffect;
 using Content.Shared.Stunnable;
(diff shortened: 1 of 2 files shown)


PursuitInAshes:
Do synths give acid aswell?
I don't know if parity says they should or shouldn't, but may be wise to double check what it should be doing aswell, fixing if required.

YO0teacher:
> Do synths give acid aswell? I don't know if parity says they should or shouldn't, but may be wise to double check what it should be doing aswell, fixing if required.

Ran a quick test, synths do in fact give acid, but I don't know if its parity. As it stands there would be an inconsistency with "fake hosts" giving acid, might be a good idea to stop synths from giving acid for that reason. Overall, this PR is to prevent helpless dummies being farmed for acid (and encourage putting training dummies in medbay more often), but I can add the fix for synths too if requested.

DrSmugleaf (approved):
Thanks for the PR!

PR #8162, opened 2025-10-16
Golden Sidearms for High Playtime COs

noctyrnal:
## About the PR

COs with 450+ Hours can now get a golden mateba or a golden peregrine handcannon in their loadout.
Could move the regular sidearm selection to loadouts too to prevent dupe weapon selection.

## Why / Balance

Technically parity - CM13 Colonels get the golden sidearms, but Colonels are a senate rank instead of being playtime. So, I chose to make it ruby instead so that it wouldn't just be an extra toy for Colonel rank, and needs a bit more work.

## Media

image
(case has the mateba in it too)

## Requirements

**Changelog**

:cl: noctyrnal
- add: COs with 250+ Hours can now get golden versions of their sidearms in their loadouts.


Aurallianz (changes requested):
failures related

Aurallianz (changes requested):
Error Message:
     Entity RMCMatebaCustomizationCaseLuxuryWeapon has storage-fill item, RMCWeaponRevolverMatebaGold, that is too large
Assert.That(protoMan.Index(entryItem.Size).Weight, Is.LessThanOrEqualTo(protoMan.Index(maxSize.Value).Weight))
  Expected: less than or equal to 2
  But was:  4

QuietlyWhisper:
Put these as an intermediary reward between two medals (ruby and the next tier up) instead of stacking rewards on the same playtime as a medal

noctyrnal:
> Put these as an intermediary reward between two medals (ruby and the next tier up) instead of stacking rewards on the same playtime as a medal

made it 450 hours

Aurallianz (approved):
yep

PR #8163, opened 2025-10-16
Mutineer leader

### LINKED ISSUE #7655: Add mutineer leader

noctyrnal:
Fixes #7655

## About the PR

Revives #7684

## Why / Balance

Mutineer leaders good for coordination, less admin work to manually add everyone to mutiny's. Parity also.

## Technical details

new leader comp that gives a recruit action, if the target has the mutinyeligible component they get a ui that lets them join or decline. also adds localization to it

## Media

image

## Requirements

**Changelog**

:cl: noctyrnal
- admin: Added Mutiny leaders. They have golden skull icons, and can recruit others to mutinies. These should be assigned to mutiny wanters instead of giving everyone the mutineer role.

### DIFF
--- a/Content.Shared/_RMC14/Marines/Mutiny/MutinyEligibleComponent.cs
+++ b/Content.Shared/_RMC14/Marines/Mutiny/MutinyEligibleComponent.cs
@@ -6,7 +6,5 @@ namespace Content.Shared._RMC14.Marines.Mutiny;
 ///     Marks a character as ineligible for mutiny recruitment.
 /// </summary>
 [RegisterComponent, NetworkedComponent]
-public sealed partial class MutinyEligibleComponent : Component
-{
-}
+public sealed partial class MutinyEligibleComponent : Component;
 


AverageNotDoingAnythingEnjoyer:
You forgot to add Parity in Why/Balance

noctyrnal:
todo for myself or someone better at code convert the check for mutiny eligibility into a check function that gives a pop up if returned for player feedback (currently being unable to convert someone just does nothing)

Content.Shared/_RMC14/Marines/Mutiny/MutineerLeaderComponent.cs
```diff
@@ -0,0 +1,18 @@
...
+public sealed partial class MutineerLeaderComponent : Component
+{
+    [DataField("icon"), AutoNetworkedField]
+    public SpriteSpecifier Icon = new SpriteSpecifier.Rsi(new ResPath("_RMC14/Interface/cm_job_icons.rsi"), "hudmutineerleader");
+
+    [DataField("recruitAction"), AutoNetworkedField]
+    public EntProtoId RecruitAction = "ActionMutineerRecruit";
+
+    [DataField("recruitActionEntity"), AutoNetworkedField]
+    public EntityUid? RecruitActionEntity;
```
DrSmugleaf: You don't need to specify the data field tag, it will automatically take the field name camelCased
```suggestion
    [DataField, AutoNetworkedField]
    public SpriteSpecifier Icon = new SpriteSpecifier.Rsi(new ResPath("_RMC14/Interface/cm_job_icons.rsi"), "hudmutineerleader");

    [DataField, AutoNetworkedField]
    public EntProtoId RecruitAction = "ActionMutineerRecruit";

    [DataField, AutoNetworkedField]
    public EntityUid? RecruitActionEntity;
```

Content.Shared/_RMC14/Marines/Mutiny/MutinyEligibleComponent.cs
```diff
@@ -0,0 +1,12 @@
...
+
+namespace Content.Shared._RMC14.Marines.Mutiny;
+
+/// <summary>
+///     Marks a character as ineligible for mutiny recruitment.
+/// </summary>
+[RegisterComponent, NetworkedComponent]
+public sealed partial class MutinyEligibleComponent : Component
+{
+}
```
DrSmugleaf:
```suggestion
public sealed partial class MutinyEligibleComponent : Component;
```

PR #8185, opened 2025-10-19
Tiles

Aurallianz:
Tiles.


Aurallianz:
Do not read too much into this PR, all this really does is ensure that all the plating that will exist in any map is actually cmplating and not plating from basestation.

As for toolshed commands they are pretty self explanatory.

- nearbytile - returns tile refs for a range around entities.
- gridtile - returns tile refs for all tiles on a grid.
- tiletype - sifts through an enum of tilerefs. (Invert-able)
- replacetile - replaces an enum of tilerefs with a separate tile type.

PR #8208, opened 2025-10-22
Fix launch direction for large lifeboats

chavonadelal:
## About the PR

No more sideways boat drifting. I'm tired of watching it.
## Technical details

I haven't tested it in the game, but I believe it should work.
## Requirements

**Changelog**

:cl:to4no_fix
- fix: Fixed launch direction for large lifeboats.


chavonadelal:
I didn't consider the possibility of a collision. In this case, a completely different approach is needed, maybe someone can implement it. Thanks for the review.

PR #8210, opened 2025-10-23
Adding Handshake emote action

CatAndHats:
## About the PR
I've sprited my own handshake emote and made it work like the other interactive emotes.

## Why / Balance
The three existing interactive emotes (fistbump, high five, and hug) are informal friendly emotes, and I think we are missing a more formal emote that would get more use from non-squad roles like the Corporate Liason and Officers.

## Technical details
same as other emote code.

## Media

https://github.com/user-attachments/assets/47c074ac-6af4-4859-9a7f-bd244c446520

## Requirements

**Changelog**
:cl: CatAndHats
- add: Added the handshake as an interactive emote, right-click another player to seal the deal.


Dutch-VanDerLinde:
These should probably be their own hand emotes prototypes defined in YML so we don't have a mega component bloated with every single hand emote in the game

CatAndHats:
> These should probably be their own hand emotes prototypes defined in YML so we don't have a mega component bloated with every single hand emote in the game

Forgot to reply to this at the time, but I do not have the C# knowledge to do this. It would be a very good idea to do since yeah it isn't the best as it's currently done.

Aurallianz:
my future self will regret the non prototyping but its consistent

DrSmugleaf (approved):

PR #8254, opened 2025-10-30
Map Vote Results

CyberTropic:
## About the PR

Map vote now shows more detailed results, like votes per map, new votes per map, and list of tied maps.

## Why / Balance

Parity 

## Technical details

Moved from ftl to building a string, since the message is more complex.
Also cleaned up some hard to read code.

## Media

image
image

## Requirements

**Changelog**

:cl:
- tweak: Changed map vote results display.


Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1833,23 +1833,42 @@ private void StartPlanetVote()
...
+                    totalVotes: newVotes + _carryoverVotes.GetValueOrDefault(planet.Proto.ID)
+                ))
+                .ToList();
+            var maxVotes = adjustedVotes.Max(v => v.totalVotes);
+            var winningMaps = adjustedVotes
+                .Where(v => v.totalVotes == maxVotes)
+                .Select(v => v.planet)
+                .ToList();
+
+            var msg = "Map Vote Results\n";
```
chavonadelal: it would be really cool if you use localization so that this could be translated, especially since it goes straight into the chat.

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
```diff
@@ -1833,23 +1833,42 @@ private void StartPlanetVote()
...
+            foreach (var result in adjustedVotes)
+            {
+                msg += $"  {result.planet.Proto.Name}: {result.totalVotes}"
+                       + (result.newVotes > 0 ? $" ({result.newVotes} new)\n" : "\n");
+            }
 
             if (winningMaps.Count > 1)
             {
+                msg += "Vote tied between:\n";
```
chavonadelal: Too. You should have noticed that the localization in this place was present before your changes

DrSmugleaf (approved):

PR #8259, opened 2025-10-30
Lets xenos upgrade acid on objects.

DoguhanOzgurAkca:
## About the PR

Lets xenos apply stronger acid to objects
## Why / Balance

parity, QoL
## Requirements

**Changelog**

:cl:
- add: Added fun!
- remove: Removed fun!
- tweak: Changed fun!
- fix: Fixed fun!
- admin: Admin related change!
- code: Code related change for contributors!

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
@@ -138,11 +138,15 @@ public abstract class SharedXenoAcidSystem : EntitySystem
         if (args.Handled || args.Cancelled || args.Target is not { } target)
             return;
 
-        // Skip the acid-already-exists check - replacement was already validated during DoAfter start
         if (!TryComp(target, out CorrodibleComponent? corrodible) || !corrodible.IsCorrodible)
             return;
 
         if (!xeno.Comp.CanMeltStructures && corrodible.Structure)
+            return;
+
+        // Re-check if acid can be replaced at DoAfter end to prevent race conditions
+        // (e.g., weak acid downgrading strong acid if both DoAfters were started before any completed)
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
```diff
@@ -303,18 +345,49 @@ public bool IsMelted(EntityUid uid)
...
+        }
+
+        if (existingAcidId == null)
+            return true;
+
+        // Simple string comparison - Strong > Normal > Weak
+        // Strong can replace anything, Normal can replace Weak, Weak can't replace anything
+        return (newAcidId.Id, existingAcidId.Value.Id) switch
+        {
+            ("XenoAcidStrong", "XenoAcidWeak") => true,
```
Vermidia: This would really be better as numbered levels, this will just break if theres ever more acid prototypes

Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
```diff
@@ -138,9 +138,15 @@ private void OnXenoCorrosiveAcidDoAfter(Entity<XenoAcidComponent> xeno, ref Xeno
         if (args.Handled || args.Cancelled || args.Target is not { } target)
             return;
 
-        if (!CheckCorrodiblePopups(xeno, target, out var _, out var mult))
```
DrSmugleaf, line 141: Don't you still need this check if someone acids something with the same strength as you? Or are we relying on the do afters never being mixed up since they are all the same length for the same strength of acid

DoguhanOzgurAkca: We need the check regardless, I just forgot. 
Next commit should fix.

DrSmugleaf (approved):

PR #8275, opened 2025-11-02
Signal Flare color change under cover

ExponentialWizard:
## About the PR
Made the color of signal flares change when they are activated under cover that prevents CAS from targetting them

## Why / Balance
It makes it easier for xenos to see when flares can be ignored, and it makes it easier for marines to see when their flares are being wasted. Overall a quality of life change.

## Technical details
I added an AltColor variable to FlareSignalComponent, which was previously just a marker component. When SharedDropshipSystem activates a flare, it sends out an event which will contain the AltColor if CanCAS fails. ExpendableLightSystem listens to this event and changes its glow color if AltColor is not null. It also now updates the appearance with that new color, and the visualizer now sets the sprite color accordingly.

## Media
Capture

## Requirements

**Changelog**
:cl:
-add: signal flares now change color when CAS is blocked


Timemaster99:
I would consider first adding the parity blinking light on the signal flare stick and using that as an indicator for ceiling, rather than changing the whole flame, which could be seen as a different variety of flare or difficult to parse for colorblind players

ExponentialWizard:
> I would consider first adding the parity blinking light on the signal flare stick and using that as an indicator for ceiling, rather than changing the whole flame, which could be seen as a different variety of flare or difficult to parse for colorblind players

Oh interesting. In that case I'll probably close this PR while I work on that.

PR #8277, opened 2025-11-02
XO, ASO, and SO time requirement change

CatAndHats:
## About the PR
Added 1 hour of requisitions to the SO time requirements.
Added 5 hours of requisitions and 5 hours of auxilliary support to the XO time requirements.
Removed the requisitions time requirement from ASO.

## Why / Balance
Right now XO has no time required in those departments by itself or through the roles in the lead-up to playing XO. I believe these should be added since the XO will in most rounds be the Commander and have direct authority over those departments, with this change XO would now have at least one hour played in each department.

XOs should know how departments function so that they know when things are going wrong and how to get them working again, an XO with no playtime in requisitions or auxiliary is unlikely to know if they are operating correctly, if those departments need help, or whether or not to spend intel on increased budget for either one.

I have also given SO an hour of requisitions as a requirement, since requisitions will often be the shipside department you interact with the most due to calling supply in for your squad(s), SOs having some of an idea of how requisitions functions will be useful for their role.

ASO no longer has any authority over requisitions and so the requisitions time requirement being the same as the departments that ASO does have authority over seems unnecessary. ASO would still inherit the 1 hour of requisitions requirement that SO will gain with this PR, so they would still need to have played some requisitions which should be the case since they still have requisitions access.

## Technical details
basic yml.

## Media
N/A

## Requirements

**Changelog**
:cl: CatAndHats
- tweak: Changed the time requirement for Staff Officer to now include one hour of requisitions.
- tweak: Changed the time requirement for Executive Officer to now include five hours of requisitions and five hours of auxilliary support department.
- tweak: Changed the time requirement for Auxilliary Support Officer to remove their five hours of requisitions time, they will inherit the one hour of requisitions required from Staff Officer.

### DIFF
--- a/Content.Server/_RMC14/Toolshed/StairwellCommands/StairwellCommand.cs
+++ b/Content.Server/_RMC14/Toolshed/StairwellCommands/StairwellCommand.cs
@@ -1,9 +1,6 @@
 using Content.Server.Administration;
 using Content.Shared._RMC14.Teleporter;
 using Content.Shared.Administration;
-using Robust.Server.GameObjects;
-using Robust.Shared.Map;
-using Robust.Shared.Map.Components;
 using Robust.Shared.Toolshed;
 
 namespace Content.Server._RMC14.Toolshed.StairwellCommands;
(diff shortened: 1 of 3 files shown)


CatAndHats:
I no longer agree with my past self about these time requirement changes entirely and as such will close the PR, I will open a different PR for the change that I still agree with which is one hour of requisitions for SO.

PR #8287, opened 2025-11-03
Added charge indicator lights to the defibs

### LINKED ISSUE #8179: Defibs on CM have a light
## Description
The defib in CM has a light on it to show how much charge it has left

**Screenshots**

Image

ExponentialWizard:
## About the PR
Added the parity charge indicator lights to the defib and the compact defib

## Why / Balance
parity, fixes #8179

## Technical details
Added a update charge visuals function to the server defibrillator system that sets the appearance based on the charge level. The generic visualizer then handles the appropriate sprite states. Theoretically this should work out of the box for other defib variants if their charge sprites are named the same.

## Media
DefibLights

## Requirements

**Changelog**
:cl: ExponentialWizard
- add: Added charge indicator lights to defibs


Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -297,4 +301,30 @@ public void Zap(EntityUid uid, EntityUid target, EntityUid user, DefibrillatorCo
         var ev = new TargetDefibrillatedEvent(user, (uid, component));
         RaiseLocalEvent(target, ref ev);
     }
+    private void UpdateChargeVisuals(EntityUid uid)
```
DrSmugleaf: Will this work if the defib is recharged? You might have to subscribe to ChangeChargeEvent and run this there as well, or in general, it should handle it for zapping too

DrSmugleaf (dismissed):

Content.Server/Medical/DefibrillatorSystem.cs
ExponentialWizard: to be honest, I didn't know these could be recharged so I have no clue how to test it. It should be working now though.

DrSmugleaf: You should be able to test with a recharger in game, they are the ones normally in medical

Aurallianz (dismissed):
Tested works

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -53,12 +55,14 @@ public sealed class DefibrillatorSystem : EntitySystem
...
     [Dependency] private readonly RMCDefibrillatorSystem _rmcDefibrillator = default!;
     [Dependency] private readonly SkillsSystem _skills = default!;
+    [Dependency] private readonly SharedAppearanceSystem _appearance = default!;
 
     /// <inheritdoc/>
     public override void Initialize()
     {
         SubscribeLocalEvent<DefibrillatorComponent, AfterInteractEvent>(OnAfterInteract);
         SubscribeLocalEvent<DefibrillatorComponent, DefibrillatorZapDoAfterEvent>(OnDoAfter);
+        SubscribeLocalEvent<DefibrillatorComponent, PowerCellChangedEvent>(OnChangeCharge);
```
Arendian: You can subscribe to the event in the `RMCDefibrillatorSystem`  after moving the methods.

Content.Server/Medical/DefibrillatorSystem.cs
```diff
@@ -299,4 +303,37 @@ public void Zap(EntityUid uid, EntityUid target, EntityUid user, DefibrillatorCo
...
+                charge = DefibrillatorChargeVisuals.Full;
+                break;
+        }
+        if (uses == 0)
+            charge = DefibrillatorChargeVisuals.Empty;
+
+        if (TryComp<AppearanceComponent>(uid, out var appearance))
+            _appearance.SetData(uid, DefibrillatorVisuals.DefibrillatorCharge, charge, appearance);
+    }
+    //RMC14
```
Arendian: These methods should be moved to the RMCDefibrillatorSystem to reduce merge conflicts with upstream. There should also be an empty line between the 2 methods.

Arendian (changes requested):
Code and YAML look good, but some logic should be moved to the RMCDefibrillatorSystem.

Content.Server/Medical/DefibrillatorSystem.cs
ExponentialWizard: hm, the reason I put it in there is that it has access to PowerCellSystem for the battery charge values. I'm not finding any obvious way to get ahold of that information without touching the server code

ExponentialWizard: wait nevermind I just found GetChargeEvent

ExponentialWizard: Hm that one is server only too. I guess I'll just make a server RMCDefibrillatorSystem

Arendian: > Hm that one is server only too. I guess I'll just make a server RMCDefibrillatorSystem

That's the way to do it yes, it all comes down to making merging upstream as easy as possible.

Arendian: In case you didn't notice, you forgot to move this block of code to the server file. You only removed it from the old file.

ExponentialWizard: dammit I forgot I need to mark it in git

Content.Server/_RMC14/Medical/Defibrillator/RMCDefibrillatorSystem.cs
```diff
@@ -0,0 +1,40 @@
+using Content.Server.PowerCell;
+using Content.Shared._RMC14.Medical.Defibrillator;
+using Content.Shared.Medical;
+using Content.Shared.PowerCell;
+using Content.Shared.PowerCell.Components;
+using Content.Shared.Rounding;
+
+namespace Content.Server._RMC14.Medical.Defibrillator;
+public sealed class RMCDefibrillatorSystem : EntitySystem
```
Arendian: You should rename the class in shared to SharedRMCDefibrillatorSystem and make it abstract.  You can then make this class inherit from it like this. Make sure to then also make a client class that inherits from the shared class, that client class can just be empty.

```suggestion
public sealed class RMCDefibrillatorSystem : SharedRMCDefibrillatorSystem
```

Arendian (changes requested):

Content.Server/Medical/DefibrillatorSystem.cs
ExponentialWizard: alright just had to remake it from scratch its working now

Content.Client/_RMC14/Medical/Defibrillator/ClientRMCDefibrillatorSystem.cs
```diff
@@ -0,0 +1,7 @@
+using Content.Shared._RMC14.Medical.Defibrillator;
+
+namespace Content.Client._RMC14.Medical.Defibrillator;
+
+public sealed class ClientRMCDefibrillatorSystem : SharedRMCDefibrillatorSystem
```
Arendian: The client class can have the same name as the server one.
```suggestion
public sealed class RMCDefibrillatorSystem : SharedRMCDefibrillatorSystem
```

Arendian (changes requested):

Arendian (approved):

Arendian:
If you ever need to change the name of something in the future, you can use the "rename" feature of an IDE to modify all references to the previous name automatically.

DrSmugleaf (approved):

PR #8304, opened 2025-11-04
Fix TrashBinBlue missing state

### LINKED ISSUE #8024: 'Trash Bin' Entities missing 'BaseCharging' state (Or need the field removed)
## Description
Shows up when trying to save a map with either trash bin entity present, error message attached but its just a missing state as far as I can tell. Likely caused by weird parenting issues. Map still successfully saves, but this will cause errors when PR'd.

Image

PursuitInAshes:
## About the PR
Added a missing state to the blue trashbin to fix a console error when placed and when saving.

## Why / Balance
Resolves #8024

## Technical details
YAML.
There has to be a way to clean up this YAML as a whole, but I don't want to look into it right now.

## Media
N/A

## Requirements

**Changelog**
No CL.


noctyrnal (approved):
lgtm

PR #8313, opened 2025-11-04
Cades now burn when on the same tile as fire

Vermidia:
## About the PR

Title.

## Why / Balance

Parity!

## Technical details

Base cade damage was tested in CM vs normal barricades and a normal m34, resulting in 60 damage per 2 seconds, so base fire damage here is 3.75.

They work like resin walls pretty much, having the flammable comp and working off of 0.5 firestacks so they unignite quickly without a source.

Had to add an anchored check to tryignite cause otherwise the barricades were getting ignored, unsure if theres a better way to fix that.

## Media

https://github.com/user-attachments/assets/c9e3cfec-8776-43da-8dc6-68bcc5000051

## Requirements

**Changelog**

:cl:
- fix: Fixed barricades not taking damage when on the same tile as fire. Bio and Composite barricades take less.


OmerF-afk:
Does this also include walls and reinforced walls as well? They also get damaged by flames IIRC

Vermidia:
Not right now, no
Dunno how much damage those take in CM.

PR #8333, opened 2025-11-06
Update M5 ATL

noctyrnal:
## About the PR

Updates M5 ATL's sprites, along with making it have a built in magnetic harness.

## Why / Balance

It makes 0 sense that the ATL can only take a magnetic harness but starts without one which made it a noob trap to not take it as its basically mandatory, old sprite was too similar to the SLAW / BRUTE and the inhands were horrible since they were done via a website and not by hand.

## Media

image
image

## Requirements

**Changelog**

:cl:
- tweak: Resprited the M5 ATL.
- tweak: The M5 ATL now has a built-in magnetic harness.


AverageNotDoingAnythingEnjoyer:
Free mag harn is Not parity
It's so over

P.S. SPP RPG should have an integrated mag harness though

noctyrnal:
sod

Vermidia:
It comes with one - its internal
So it has no attachments slots cause of that.

/obj/item/weapon/gun/launcher/rocket/m5
	name = "\improper M5 RPG"
	desc = "The M5 RPG is the primary anti-armor weapon of the USCM. Used to take out light-tanks and enemy structures, the M5 RPG is a dangerous weapon with a variety of combat uses."
	map_specific_decoration = TRUE
	has_empty_icon = TRUE
	auto_retrieval_slot = WEAR_J_STORE
	attachable_allowed = list()

AverageNotDoingAnythingEnjoyer:
> It comes with one - its internal So it has no attachments slots cause of that.

Wot, huh
Why's SPP RPG have a separate field to outline having integrated mag_harn 🗿

Vermidia:
Simple - SPP one is an actual attachment being hidden and stuff, old code. The M5 instead uses the fancy auto-retrieval feature of CM which requires no attachment and acts just like a magharn. Newer code proby.

QuietlyWhisper (approved):

noctyrnal:
i want to make this camo conforming when i can get the energy to do so

DrSmugleaf (approved):

Issue #8233, opened 2025-10-27
UBF/M34 Flamer gives wrong message when empty


Thereisabear:
## Description
Currently when your UBF or incinerator tank is empty, it will tell you 
"Too close to target"

**Reproduction**
Fire UBF/M34 until empty, see resulting message spam.

**Screenshots**
Will attach soon

Abureaucrat:
Good to know this is a bug now.
Didn't before and got me capped in the single most pathetic display I've ever seen.

PR #8350, opened 2025-11-07
Flamethowers out of fuel popup

CatAndHats:
## About the PR
Before when out of fuel they would state "you are too close to your target!", now they will state "You are out of fuel!"

## Why / Balance
QoL. Fixes https://github.com/RMC-14/RMC-14/issues/8233

## Technical details
basic c#

## Media

https://github.com/user-attachments/assets/22f91590-1fe9-41e9-a844-2514c9888611

## Requirements

**Changelog**
:cl: CatAndHats
- fix: Flamethowers now warn you when you are out of fuel


Ian321:
Please just move the logic up. No need to set a global variable.

```cs
    private void OnAttemptShoot(Entity<RMCFlamerAmmoProviderComponent> ent, ref AttemptShootEvent args)
    {
        if (args.ToCoordinates is not { } toCoordinates ||
            CanShootFlamer(ent, args.FromCoordinates, toCoordinates, out _, out var solution, out _, out _))
        {
            return;
        }

        args.Cancelled = true;
        args.ResetCooldown = true;

        var time = _timing.CurTime;
        if (time < ent.Comp.CantShootPopupLast + ent.Comp.CantShootPopupCooldown)
            return;

        ent.Comp.CantShootPopupLast = time;
        Dirty(ent);

        if (solution.Comp.Solution.Volume < ent.Comp.CostPer)
        {
            args.Message = Loc.GetString("rmc-flamer-empty");
            return;
        }

        args.Message = Loc.GetString("rmc-flamer-too-close");
    }
```

CatAndHats:
The current version works correctly, the more elegant solution Ian321 has suggested instantly crashes the game when a flamer reaches zero for some reason, I will come back to this later if the less elegant solution doesn't get merged in the meantime.

CatAndHats:
Better PR is now up as above

PR #8364, opened 2025-11-08
Wooden squad crates

CatAndHats:
## About the PR
Added six new constructable wooden squad crates, in their respective colours.

## Why / Balance
Allows requisitions and the squads themselves to build crates in their colours to hold their stuff, and prevent any crate mix-ups. A simple but useful cosmetic addition.

## Technical details
basic yml

## Media
squad crates
image

construction menu
image

## Requirements

**Changelog**
:cl: CatAndHats
- add: Added six new constructable squad-coloured wooden crates! Press Z with wooden planks in-hand to open the wood construction menu where these will be found.


DrSmugleaf (dismissed):
Forgot to submit the review, this is probably good, but I'm going to hold off on merging it until we do a passover on all content, since Req isn't meant to be able to supply drop craftable wooden crates to begin with, so I'd need to check how much we would need to adjust the budget to compensate once black market and the rest of the Req mechanics are in, with this being one of the few things Engineering can do round start
Keep this PR open though so we can merge it once that happens

PR #8376, opened 2025-11-09
Xeno construction QOL

Tyzemol:
## About the PR
Sticky and speed resin, along with all of the hive structures now can be build directly under yourself
Minimal range for walls and doors have been made _much_ more lenient (it is still big enough to not let you push yourself into a wall)
And also hivelords can now upgrade resin through a corner.

Almost forgot to mention i also made it so the doafter for building does no cancel if you try to build before it ends,
this means you can spam click to build as soon as the previous structure finishes.

## Why / Balance
So in 13 you can straight up build resin right on top of yourself, and you don't collide with it until you move off the wall, but just getting pushed off to the side should be good enough.

## Technical details
While on it caught a tiny bug, `CanOrderConstructionPopup` was passing current selected resin (wall, door etc) structure selected into a `CanSecreteOnTile` check instead of the actual hive structure you are trying to place.

## Media

https://github.com/user-attachments/assets/8b31f436-cd11-4e30-9e46-8c8a6d204b59

## Requirements

**Changelog**
:cl:
- tweak: As a xeno you can now build sticky/speed resin and hive structures on the same tile you are standing on.
- tweak: Minimal range to build walls and doors have been made much more lenient.
- tweak: Hivelords can now upgrade resin on the diagonal corners.
- tweak: Secrete resin ability no longer cancels current resin structure doafter when trying to build a new one too early

### DIFF
--- a/Content.Client/_RMC14/Xenonids/Construction/XenoConstructionGhostSystem.cs
+++ b/Content.Client/_RMC14/Xenonids/Construction/XenoConstructionGhostSystem.cs
@@ -422,7 +422,7 @@ public sealed class XenoConstructionGhostSystem : EntitySystem
             if (buildRange > 0 && !_transform.InRange(origin, target, buildRange))
                 return false;
 
-            if (_transform.InRange(origin, target, 0.75f))
+            if (_transform.InRange(origin, target, _xenoConstruction.GetStructureMinRange(buildChoice).Float()))
                 return false;
 
             if (isRemoteConstruction && !CanDoRemoteConstruction(xeno, target))
@@ -473,7 +473,7 @@ public sealed class XenoConstructionGhostSystem : EntitySystem
 
     private bool CanOrderConstruction(Entity<XenoConstructionComponent> xeno, EntityCoordinates target, EntProtoId? choice)
     {
-        if (!CanSecreteOnTile(xeno, xeno.Comp.BuildChoice, target, false, false))
+        if (!CanSecreteOnTile(xeno, choice, target, false, false))
             return false;
 
         if (_transform.GetGrid(target) is not { } gridId ||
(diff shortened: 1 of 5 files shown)


ExponentialWizard:
I wonder if you can give doors the same tile treatment, it should just automatically open right?

Tyzemol:
I was considering that but it would still push you before opening, so i decided to give it the same minimal range as a wall.

ExponentialWizard:
Also, and this isn't a criticism but an actual question, are empty components like XenoConstruct supposed to remain empty or can you throw properties into them if convenient instead of making a new component?

Tyzemol:
Honestly not sure what `XenoConstructComponent` is for even, have not looked at it, existing code i was touching had a separate component for plasma cost (`XenoConstructionPlasmaCost`) so i just "took inspiration" from that.

Vermidia:
Xeno construct mostly prevents other structures from being built atop one another.

Bobert-Brockson:
With this change is the queen eye still able to cancel construction by starting another construction?

Definitely out of scope but I'm wondering if we want a, "cancel whatever doafter you're doing" button at some point in the far future.

Gallagin:
> With this change is the queen eye still able to cancel construction by starting another construction?
> 
> Definitely out of scope but I'm wondering if we want a, "cancel whatever doafter you're doing" button at some point in the far future.

could probably be like a right click cancels the current construction

Tyzemol:
> With this change is the queen eye still able to cancel construction by starting another construction?

No, but for the sake of all our drone and hivelord mains that's sacrifice i am willing to make.

Bobert-Brockson:
> > With this change is the queen eye still able to cancel construction by starting another construction?
> 
> No, but for the sake of all our drone and hivelord mains that's sacrifice i am willing to make.

I agree its worth. I only brought it up so in the future if someone asks we know it was pondered and decided this route was preferable.

I'd love the master "cancel whatever you're doing" button one day

PR #8380, opened 2025-11-10
Fixes CPR bug that opens the strip menu.

Arykh:
## About the PR

Fixes a CPR bug.
## Why / Balance

Sometimes, when you try to CPR someone, it will instead open the strip menu with a red pop-up saying that CPR is already being performed on that person.
This results in someone savable becoming permadead instead.
## Technical details

This bug is caused by lines 68-69 failing to remove the component. Probably caused by de-sync issues, somehow losing MarineComponent, or getting deleted in the middle of the do-after.

Added a check to remove any stale ReceivingCPRComponent after 7 seconds have passed.
## Requirements

**Changelog**

:cl:
- fix: Fixed not being able to perform CPR after a previous attempt fails, causing the strip menu to open instead.

### DIFF
--- a/Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
+++ b/Content.Shared/_RMC14/Medical/CPR/CPRSystem.cs
@@ -42,7 +42,6 @@ public sealed class CPRSystem : EntitySystem
     {
         base.Initialize();
 
-        // TODO RMC14 use skills
         // TODO RMC14 something more generic than "marine"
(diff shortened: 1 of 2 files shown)


Arykh:
This ain’t gonna work. A second person can easily start CPR in the middle of the do after and throw everything off.

Arykh:
Much better :D

DrSmugleaf (approved):

PR #8404, opened 2025-11-14
Unique Type 71 / 73 gun sounds

noctyrnal:
## About the PR

Title, also fixes the Type 88 gunshot sound not being mono, and moves the PVE Type 71 into the Type 71 file instead of it's own.

## Why / Balance

Host request, makes them more clear compared to m54cs which they shared a gunshot collection with. Other is bugfix and less bloat.

## Media

https://github.com/user-attachments/assets/d574a8dc-0783-438b-a208-b19fe17f9b52

## Requirements

**Changelog**

:cl:
- tweak: Updated the Type 71 and 73 shooting sound.


noctyrnal:
> If these audio files aren't mono I'm going to shoot myself with one

i checked they're mono i swear

PR #8412, opened 2025-11-15
Stealth fixes

YO0teacher:
## About the PR
Parity changes, invisibility bug fixes, and changed code for how IFF with stealthed targets work.

## Why / Balance
Hiding infection shaking popups while in a thermal tarp is parity.
Being able to PB stealthed people in your own faction was an oversight, so you can't do that anymore.

While invisibility giving 1000 friendly evasion is parity, it's redundant as we already have a BlockFriendlyFire bool as part of the thermal cloak component which effectively does the same thing. In-game function is the exact same, and this is better for dev-testing purposes.
## Technical details
I copied BlockFriendlyFire logic over to the ghillie suit, so now you can toggle BlockFriendlyFire on the respective component (ThermalCloakComponent.cs or GhillieSuitComponent.cs) to enable/disable IFF with stealthed people.

Client-side invisibility logic was out of sync with the server, causing visual and audio issues after being decloaked in these ways:
Hit by acid (OnAcidProjectile)
Infected by a parasite (OnParasiteInfect)
Fire extinguisher spray (OnVaporHit)
But this is fixed now, had to dirty the component when cloaked/decloaked.

## Media
https://github.com/user-attachments/assets/4a1b2c2f-0e10-49ef-b1fd-879013a45bae

## Requirements

**Changelog**
:cl:
- tweak: Infected people shaking doesn't show a popup for others when they're inside a thermal tarp.
- fix: Fixed visual and audio issues resulting from client desync with invisibility.
- fix: Fixed being able to PB stealthed people in your own faction.
- code: Stealth IFF now comes from the BlockFriendlyFire bool instead of a 1000 EvasionFriendlyModifier.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
@@ -837,7 +837,7 @@ public abstract partial class SharedXenoParasiteSystem : EntitySystem
 
         _popup.PopupEntity(Loc.GetString("rmc-xeno-infection-shakes-self"), victim, victim, PopupType.MediumCaution);
 
-        if (_container.TryGetContainingContainer(victim, out var container) && HasComp<RMCPassiveStealthComponent>(container.Owner))
+        if (_container.TryGetContainingContainer(victim, out var container) && HasComp<RMCHideParasiteInfectionContainerPopupComponent>(container.Owner))
             return;
 
         _popup.PopupEntity(Loc.GetString("rmc-xeno-infection-shakes", ("victim", victim)), victim, Filter.PvsExcept(victim), true, PopupType.MediumCaution);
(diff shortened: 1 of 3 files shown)


Content.Shared/_RMC14/Xenonids/Parasite/SharedXenoParasiteSystem.cs
```diff
@@ -835,6 +836,10 @@ private void InfectionShakes(EntityUid victim, VictimInfectedComponent infected,
             return;
 
         _popup.PopupEntity(Loc.GetString("rmc-xeno-infection-shakes-self"), victim, victim, PopupType.MediumCaution);
+
+        if (_container.TryGetContainingContainer(victim, out var container) && HasComp<RMCPassiveStealthComponent>(container.Owner))
```
Dutch-VanDerLinde:
```suggestion
        if (_container.TryGetContainingContainer(victim, out var container) && HasComp<RMCPassiveStealthComponent>(container.Owner))
```
Probably should make a new component called ``RMCHideParasiteInfectionPopupContainerComponent`` or something and give this to the stealth bag

So that it isn't hardcoded to the stealth bag

YO0teacher: Ok it works based off the new component now. It's quite a lengthy component name but I can't think of a way to shorten it either without some ambiguity.

PR #8419, opened 2025-11-16
Updated XL armor vest in rfn machine to follow naming conventions

Bobert-Brockson:
## About the PR
The extra light armor aka the armor vest was added to the rfn machine. However it didn't follow the pattern of other armors in the machine where it used a generic name (light armor vs m3-l armor). I updated to use the same naming pattern. Additionally I put it above the light armor in the list so that it is lightest to heaviest option

## Why / Balance
Matches conventions better

### Before
Screenshot From 2025-11-15 19-03-02

### After

Screenshot From 2025-11-15 18-57-00

## Requirements

**Changelog**

:cl:
- tweak: Rifleman's armor vest option now renders like the other armor options.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/rifleman.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/rifleman.yml
@@ -23,6 +23,8 @@
     - name: Armor
       choices: { CMArmour: 1 }
       entries:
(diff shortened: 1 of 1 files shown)


Aurallianz (approved):
You can tell that when I wrote this, I 100% checked it beforehand.

Bobert-Brockson:
> You can tell that when I wrote this, I 100% checked it beforehand.

Psh its more important that its added than looks pretty. It's totally fine without my change ❤️

PR #8441, opened 2025-11-19
Add New Xeno Non-Loop Sound Markers

PursuitInAshes:
## About the PR
(Now with a proper branch name, because it bothered me)
Adds non-loop noise markers for vent crawling, vent passing, and xeno door prying.

## Why / Balance
Useful for PvE, lets us indicate that xenos are doing something without them having to actually be there.

## Technical details
YAML

## Media
N/A

## Requirements

**Changelog**
:cl: PursuitInAshes
- add: [PvE] Three new sound markers have been added for GM use. (Vent Crawl/Pass, Xeno Door Prying)


QuietlyWhisper:
Add a TimedDespawn component, it will save you from having garbage entities

PursuitInAshes:
I believe when they were originally made it was decided to leave off timeddespawn as it would have to be tuned to every sound and not just added to the base entities.
However, I'll test and see if its terrible making them despawn after like 30~ seconds for the non-looping entities as that will likely be okay just to keep things clean, thanks for pointing it out

PursuitInAshes:
Seems as though for some reason they delete on spawn anyways, even without the timeddespawn comp being added. No clue why? (Unless show markers is lying to me and they aren't actually being deleted) Checked on master too, so maybe a non-issue
https://github.com/user-attachments/assets/3bb19958-2fa9-4da7-b0f8-8ef9dff1dc9f

yagwog:
`OnEmitSpawnOnInt` naturally gets cleared with `QueueDel(ent.Owner)`

PR #8442, opened 2025-11-19
Add MSgt rank to CMilitaryWarden

BreakpointDebug1234:
## About the PR

Adds Master Sergeant rank at 175 hours (Platinum) for Military Warden.
## Why / Balance

Since the change for the QM role to the LO role the MSgt. rank has not been used by any roles other than SEA. This would make experienced wardens be above in rank to Squad Leaders. It would also offer a reward for the people which achieve such a number of hours.

@QuietlyWhisper should weigh in if this is something that would be ok for MWs to receive.

## Technical details

yaml

## Requirements

**Changelog**
:cl:
- add: Military Warden can now rank up Master Sergeant


QuietlyWhisper:
4 ranks seems overkill beyond the very high rank - this is used for synths but not used for SEA - SEA uses Master Gunnery Sergeant not Master Sergeant, which is understandable given their position

PR #8443, opened 2025-11-20
Medical Supply Link restocking system

Arykh:
## About the PR

Adds restocking med vendors via medilinks
## Why / Balance

Parity
## Technical details

Adds dylovene autoinjector yaml and adjusted the amount of items in med vendors.

Adds the medi link system that allows valid vendors to restock when anchored to a medilink port as well as giving the ports some animation.
## Media
Note: I renamed the clamped and unclamped sprites around because they did not make any sense. Clamped is now green with the little clamp arms extended while unclamped is red with the clamp arms retracted. 

image

### Here it is roughly 7-8 minutes after anchoring.
image

## Requirements

**Changelog**

:cl:
- add: Added Dylovene autoinjectors to the We-Ya-Med Plus vendor.
- add: Medical vendors can now slowly restock over time if anchored onto a Medical Supply Link port! This will only go into effect after 20 minutes from round start. Each item has a 20% chance to increase by 1 every 30 seconds.
- tweak: Increased the amount of some items found in medical vendors to match parity.


Content.Client/_RMC14/Animations/RMCAnimationSystem.cs
```diff
@@ -57,7 +57,7 @@ private void OnFlick(RMCFlickEvent ev)
             return;
 
         if (_animation.HasRunningAnimation(ent, FlickId))
-            return;
+            _animation.Stop(ent, FlickId);
```
Arykh, line 60: Changed this to stop any running animation and allow the new one to start immediately.

Changing the state too quickly before the animation ends causes a mismatch in the final state.
Example: unanchoring and quickly reanchoring the medi port causes the final state to be unclamped when it should be clamped.

Arykh:
👍

Arykh:
Gonna leave this open in case @DrSmugleaf wants to merge.
This version of the medi link port works without the partial vendor stacking from #8462. 
I’ll close this if I finish #8462 and this hasn’t been merged.

Arykh:
Ugh, nvm. Probably not worth the pain. :agonylimitless:

PR #8486, opened 2025-11-27
Force crit state smite

Vermidia:
## About the PR

Title.

## Why / Balance

GM request.

## Technical details

In smite code. Damages to the desired threshold, if below threshold heals and revives to the threshold. Said threshold depends on which of the 4 smites are chosen, leaving someone at either 100% crit health, 75% crit health, 50% crit health, or 25% crit health.

While doing this I had to fix damage distributed types being broken. The problem was the code expected 'equal' to not be adding types to itself constantly I guess??. So amount no longer sets itself to left, because it doesn't need to do that. Total will always be the damages we already healed so it just works TM. (I'm kinda confused why reverting a change I did forever ago randomly works correctly now, and I don't know why? I did check other things that use it just in case and it seems to work ok still.)

## Media

https://github.com/user-attachments/assets/4a7bc52e-6218-460f-bb38-269403a3bc83

## Requirements

**Changelog**

:cl:
- fix: Fixed some heals not healing enough.


Aurallianz (approved):
yerp

PR #8508, opened 2025-11-30
Allow examining while overwatching

YO0teacher:
## About the PR
Allows examining while viewing Overwatch cameras or watching xenos.

## Why / Balance
Parity. Originally brought up by PR #5079 which wanted to display pronouns on the Overwatch console, but multiple comments (including one from Smug) said it'd be better just to make examining work through cameras and find pronouns that way. 
So, this fixes the pronouns problem and gets us closer to parity.

I also prevented holocard changing while viewing cameras for the sake of parity. If nobody objects, I would like holocard changing to be re-enabled though.
## Technical details
The overwatched entity is treated as the examiner. So whether you can inspect things is determined by its proximity and visibility to the entity you're watching. Same goes for xenos.

For preventing holocard changing, it's just a component check.

## Media
Notice how I cannot examine obscured objects, like the cake while watching the WS's camera.

https://github.com/user-attachments/assets/c930faef-49a7-431a-b0d2-763b149438b0

https://github.com/user-attachments/assets/e0bef652-f674-4371-909c-0a02777883d1

## Requirements

**Changelog**
:cl:
- tweak: Examining now works while watching xenos or viewing Overwatch cameras.

### DIFF
--- a/Content.Shared/_RMC14/Medical/HUD/Systems/HolocardSystem.cs
+++ b/Content.Shared/_RMC14/Medical/HUD/Systems/HolocardSystem.cs
@@ -1,5 +1,4 @@
 using Content.Shared._RMC14.Marines.Skills;
-using Content.Shared._RMC14.Overwatch;
 using Content.Shared._RMC14.Medical.HUD.Components;
 using Content.Shared._RMC14.Medical.HUD.Events;
 using Content.Shared._RMC14.Medical.Scanner;


vorkathbruh:
Holocard changing through OW sounds nice, the SO would be able to triage for the medics since they're often busy doing the actual healing and avoiding death

Content.Shared/_RMC14/Medical/HUD/Systems/HolocardSystem.cs
```diff
@@ -46,6 +47,9 @@ private void ChangeHolocard(Entity<HolocardStateComponent> ent, ref HolocardChan
         if (!_transform.InRange(ent.Owner, viewer.Value, 15f))
             return;
 
+        if (HasComp<OverwatchWatchingComponent>(viewer.Value))
+            return;
+
```
DrSmugleaf:
```suggestion
```

Content.Shared/_RMC14/Medical/HUD/Systems/HolocardSystem.cs
```diff
@@ -1,4 +1,5 @@
 using Content.Shared._RMC14.Marines.Skills;
+using Content.Shared._RMC14.Overwatch;
```
DrSmugleaf:
```suggestion
```

DrSmugleaf (dismissed):
Yeah, holocard changing is fine to re-enable, otherwise the code is good ty

PR #8479, opened 2025-11-26
More order callouts

monomethylhydrazine:
## About the PR

Ported more order lines from https://github.com/cmss13-devs/cmss13/blob/6791e451bc811407b6daf545a6cf8e3535ee4e4c/code/modules/mob/living/carbon/human/powers/issue_order.dm
first time working with cs in case this is egregious shitcode but it does work

## Why / Balance

sovl

## Technical details

minor cs and a bunch of lines in marine_orders.ftl

## Media

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Added more order callouts, ported from CM13.

PR #8509, opened 2025-11-30
callout tweak

BadaBoomie:
## About the PR
addresses a comment from @QuietlyWhisper in https://github.com/RMC-14/RMC-14/pull/8479 
removes some parity'd callouts, some of which would be confusing or counterintuitive in certain situations


QuietlyWhisper:
This is really hard to review one sec

AverageNotDoingAnythingEnjoyer:
Movie soul gone with removal of move-order-callout-6

QuietlyWhisper (changes requested):
Remove  
- hold 5,10,11,12
- focus 10,19

I think those ones are most likely to cause confusion
Rest are probably fine

BadaBoomie:
this is all of what's been removed since it's not easy to read
**Move:** We are on the move!! / Keep up, ladies!! / We are leaving!! / We are Oscar-Mike!! / Get on me!! / On me, men, let's move!! / Stack up on me!! / Fall in!!
**Hold:** Hold position!! / Hold here!! / Hold with me!! / To me, hold!! / Defensive positions!! / No retreat, stand firm!!
**Focus:** Short, controlled bursts!! / On my mark!! / Suppressive fire!! / Lock and load!! / Let's rock!! / Firing lines!!

Aurallianz:
Thanks for the PR!

PR #8559, opened 2025-12-03
Fix acid blood splash chance math and add xeno pain emotes on being shot

Dutch-VanDerLinde:
## About the PR
Fixes acid blood chance math to Cm13 values and adds the emote on shot for xenos

CM13 hissing
image

CM13 code

image

Old Code
image

Huge problem with this. ``BaseTriggerSplashProbability`` is 0.2f, which when added with the damge trigger chance, it always gets to a number above one for most damage sources.

For example -
 ``BaseTriggerSplashProbability`` (0.2) + ExampleDamage (40) * ``DamageTriggerProbabilityMultiplier`` (0.33) = 13.4``.
so it'll always spray acid when possible

In CM13, this value would be 33.2 which is 33.2% chance to spray acid

I changed the code to use the CM13 values and then divide itself by 100 so it becomes a decimal for ``_random.Prob()``

## media

https://github.com/user-attachments/assets/384ed96b-6be8-4a16-90e5-5a5dd0f53d0a

https://github.com/user-attachments/assets/1dd13e0f-e734-428a-a47b-477a48fefdf4

## Requirements

**Changelog**
:cl:
- tweak: Xenos now have a chance to hiss or roar upon being shot
- fix: Fixed damaging a xeno ALWAYS splashing acid onto marines. The chance now scales itself based on the damage as intended.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashComponent.cs
+++ b/Content.Server/_RMC14/Xenonids/AcidBloodSplash/AcidBloodSplashComponent.cs
@@ -17,6 +17,9 @@ public sealed partial class AcidBloodSplashComponent : Component
     public FixedPoint2 MinimalTriggerDamage = 12;
 
     [DataField]
+    public bool WorksWhileDead = true;
+
+    [DataField]
     public bool IsActivateSplashOnGib = true;
 
     [DataField]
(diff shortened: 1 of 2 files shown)


Dutch-VanDerLinde:
Idk what these test fails are

PR #6290, opened 2025-04-28
ARES: Automatic appointment of acting Commander

chavonadelal:
## About the PR

**Saving rounds with low online!**
Added a system for automatically assigning an acting Commander (`MarineCommandOverrideSystem`) in the event of a senior command being unavailable after the round starts. 
ARES controls the process by selecting a suitable Marine based on rank and squad priority, and then announcing the appoint.
Marine Authority Level is set according to the chain of command. The system is intended if CO and XO are not available.
## Why / Balance

Parity. A polished new feature from ss13. Only instead of forcibly assigning without checking, we take everything into account.
https://github.com/cmss13-devs/cmss13/pull/8882
SOP updated to clarify some points.
### Initialization

- When a round starts (`RoundStartingEvent`), a two-minute timer (`_initialTimerEndTime`) is started.

### Check for senior command presence

- 2 minutes after the round starts, the `CheckForSeniorCommandPresence()` method is called.
- The system iterates through all active marines:
   - Ignored:
      - dead,
      - in cryostorage,
      - without an active mind (`Mind`),
      - without a valid ID card,
      - with an invalid rank (Private).
      - Candidates must have a non-zero `MarineAuthorityLevel > 0`.
- If at least one living marine with an authority level (`MarineAuthorityLevel`) **13+** is found:
   - The system stops working - no acting Commander is required.
- If there are no seniors:
   - ARES issues an alert about their absence.
   - An additional timer of 1 minute is started (`_adaptationTimerEndTime`).

### Selecting an acting Commander 

- One minute after the notification, the `CommanderSelection()` method is called. An extra minute is given so that Marines can put on their ID if they want to become a commander or take it off if they don't.
- The system checks for senior commanders again:
   -  If at least one living marine with an authority level (`MarineAuthorityLevel`) **14+** is found:
      - ARES announces this.
      - The system exits.
- If there are still no senior commanders, the system selects candidates, ignoring::
      - dead,
      - in cryostorage,
      - without an active mind (`Mind`),
      - without a valid ID card,
      - with an invalid rank (Private).
      - Candidates must have a non-zero `MarineAuthorityLevel > 0`.

#### Candidate prioritization

1. Marines with the highest `MarineAuthorityLevel` are selected first.
2. Among them, priority is given to those with the highest rank according to `RMCMarineRankHierarchy`.
3. If the ranks are equal, the squad membership according to `RMCMarineSquadHierarchy` is taken into account.
[...]

PR #8564, opened 2025-12-04
move ASO in commander succession, update commander succession procedure

vorkathbruh:
.xml change, shouldn't break anything

ASO being the duty officer to the XO has proven itself to be dubious at best. While SOs may also be unreliable duty officers, the transition of power is a lot more ergonomic due to staff officers already being within the CIC itself, not requiring a long walk nor to be filled in on the state of the operation since they have already had a hand in running it. 

Staff officers however do not have Senior Command access on their ID cards by default, thus requiring their accesses to be modified in order to access the groundside operations console. This problem has actually already existed for some time - for as long as the groundside operations console got it's access locked to Senior Command, nobody outside of the CO, XO, ASO can access it by default and as was previously written, the Commander was not to inherit or gain any accesses. Thus, it makes sense for the procedure to dictate that the Commander be granted Senior Command access if they do not have it already.

:cl:
- tweak: Moved Auxiliary Support Officer below the staff officer in command succession.
- tweak: changed wording of Commander Succession to communicate that the Commander is to be granted Senior Command access.


vorkathbruh:
It just occurred to me that I did not account for the system of automatically appointing the commander as added in https://github.com/RMC-14/RMC-14/pull/6290
From what I remember the system was flawed already, since it triggered with an ASO online which it didn't need to due to their accesses.

AverageNotDoingAnythingEnjoyer:
The system triggered to ASO because at the time there wasn't a way to get extra access that XO has compared to them

AverageNotDoingAnythingEnjoyer:
Also, if this does come through, ASO needs their Command time requirement removed

Aurallianz (dismissed):
Lgtm, @QuietlyWhisper,

I will say this probably goes without saying but honestly it seems the ASO serves less of a purpose and honestly mostly feels like role bloat. 
I believe soon enough we'd rather replace them with something else.
This is good.

vorkathbruh:
> The system triggered to ASO because at the time there wasn't a way to get extra access that XO has compared to them

Yeah that makes sense.

Randomposter12:
> Lgtm, @QuietlyWhisper,
> 
> I will say this probably goes without saying but honestly it seems the ASO serves less of a purpose and honestly mostly feels like role bloat. I believe soon enough we'd rather replace them with something else. This is good.

**CIO - CHIEF INTEL OFFICER. PLEASE I BEG YOU AURA AND MY SOUL IS YOURS**
l2qdcyilml9d1

Aurallianz:
So I'm just gonna remind people that if you'd like to make suggestions, do it on the Discord

QuietlyWhisper (changes requested):
You need to change the order of command succession in job yml files as well

vorkathbruh:
> You need to change the order of command succession in job yml files as well

done

PursuitInAshes (approved):
Code lgtm

Aurallianz (approved):
sovl

Issue #7311, opened 2025-07-29
You are able to reload SLAWs


XxSWAG-MASTER-OVERLORDxX:
## Description
While the gun is now useless, you are still able to put a rocket inside it. this does nothing but remove a rocket from the round since you cant eject the rocket either.

Issue #7194, opened 2025-07-22
Unfolding the SLAW drops it


XxSWAG-MASTER-OVERLORDxX:
## Description
Even if you have an empty hand it drops the SLAW on the ground when unfolding it

PR #8570, opened 2025-12-04
Fix SLAW folding to stay in hand + make single-use

wheelwrightt:
## About the PR
Fixes: [#7311](https://github.com/RMC-14/RMC-14/issues/7311) 
Fixes: [#7194](https://github.com/RMC-14/RMC-14/issues/7194)

Rockets no longer load into an already fired SLAW.
SLAW no longer drops to the floor when folded and unfolded.

## Technical details
2 lines of YML, created whitelist with nothing in it, making rockets unable to load. Maybe a popup might be useful especially for players that might not realise its single shot but the ammo indicator seemed enough at this moment.

The old SLAW entity was still in hand when the pickup check was run causing it to fail and the folded to be dropped to the floor. TryDrop removes the old entitiy and frees the hand up which when PredicitedSpawnNextToOrDrop creates the folded version TryForcePickup succeeds and no more dropped SLAWs.

## Media
[Before.webm](https://github.com/user-attachments/assets/48417b43-6960-4317-9656-079fa1e4dae9)

[After.webm](https://github.com/user-attachments/assets/871593e3-61c5-4104-9a7d-caf2534161b3)

## Requirements

**Changelog**
:cl:
- fix: SLAW can no longer take more rockets after it fires.
- fix: SLAW no longer drops to the ground upon folding/unfolding

### DIFF
--- a/Content.Shared/_RMC14/Weapons/Ranged/Foldable/RMCFoldableGunSystem.cs
+++ b/Content.Shared/_RMC14/Weapons/Ranged/Foldable/RMCFoldableGunSystem.cs
@@ -7,6 +7,9 @@ using Content.Shared.DoAfter;
 using Robust.Shared.Audio.Systems;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Interaction;
+using Content.Shared.Clothing;
+using Content.Shared.Clothing.Components;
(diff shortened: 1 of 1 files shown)


Dutch-VanDerLinde (approved):
Thank you

PR #8577, opened 2025-12-05
Made Neck Slot Items Attachable to Uniforms

wheelwrightt:
## About the PR
Mask slot items such as but not limited to: squad wraps, coloured wraps, torn scarves, smock variants, and all neckerchief variants, now can be attached to uniforms.

## Why / Balance
I believe it's parity but am not 100%.

## Technical details
Added the Uniform Accessory type to all relevant neck items. They initially were layering underneath the uniform so I added a new layer to the .cs file to remedy.
The accessories weren't initially preserving their state when attached to the uniform (ie. a folded neckerchief would be unfolded when attached). So I added a check to the UniformAccessorySystem to make sure it stays consistent.

Tested every one I changed so I *shouldn't* have missed anything.

## Media
neck fold
neck unfold
smock
tactical scarf
wrap down
wrap up

## Requirements

**Changelog**
:cl:
- tweak: Mask slot items (tactical wrap, neckerchief, etc.) now can be attached to uniforms.

### DIFF
--- a/Content.Shared/_RMC14/UniformAccessories/SharedUniformAccessoryEquipBlockSystem.cs
+++ b/Content.Shared/_RMC14/UniformAccessories/SharedUniformAccessoryEquipBlockSystem.cs
@@ -1,0 +1,48 @@
+using Content.Shared._RMC14.Armor;
+using Content.Shared._RMC14.UniformAccessories;
+using Content.Shared.Inventory;
+using Content.Shared.Inventory.Events;
+using Robust.Shared.Containers;
+
+namespace Content.Shared._RMC14.UniformAccessories;
+
+public sealed class SharedUniformAccessoryEquipBlockSystem : EntitySystem
+{
+    [Dependency] private readonly SharedContainerSystem _containers = default!;
+
+    private const string MaskCategory = "Mask";
+
+    public override void Initialize()
+    {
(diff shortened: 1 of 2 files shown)


Dutch-VanDerLinde:
Test if this works while wearing another mask item on the mask slot

You might have to add the layer to humanoids instead of using the inventory layer, like how the ponchos PR added the base layer to it

Dutch-VanDerLinde:
Same with equipping the neck slot item while you have the accessory on your uniform and etc

wheelwrightt:
True, that honestly slipped my mind. Just tested it there and you can wear two at a time which looks pretty goofy lmao will take a look at it

wheelwrightt:
[fix?.webm](https://github.com/user-attachments/assets/614b4d0e-65b3-4d3c-b6df-20b95e089153)

looks good for now. may look into the poncho PR and see how they did the layering later
also tested this with different types and the behavior is consistent

Dutch-VanDerLinde:
That's not what I meant lol, was just making sure it wasn't buggy because ik having two medals on different uniform accessory holders is buggy but nonetheless thanks for the work

wheelwrightt:
> That's not what I meant lol, was just making sure it wasn't buggy because ik having two medals on different uniform accessory holders is buggy but nonetheless thanks for the work

What we thinking about the functionality then, keep it as it originally was in the inital PR or keep the recent changes?

PursuitInAshes:
Is there a specific reason why this was closed? As far as I know this feature was not implemented elsewhere and this PR was just waiting on review?

PR #8593, opened 2025-12-07
Echo and FORECON map icon background change

CatAndHats:
## About the PR
changed the tactical map background icons of the two squads. Echo now uses its colour on the tactical map, FORECON has recieved minor colour tweaks.

## Why / Balance
QOL; to increase map readability.

## Technical details
basic yml

## Media
old
Screenshot 2025-12-08 030036

new
Screenshot 2025-12-08 024532

old
Screenshot 2025-12-08 030014

new
Screenshot 2025-12-08 024607

## Requirements

**Changelog** CatAndHats
:cl:
- tweak: Changed the tactical map colour of Echo to their squad colour.


QuietlyWhisper (dismissed):
Looks ok

noctyrnal (commented):
I'm worried about visibility of the echo icons due to how white they are with white icons. An alternative solution would be keeping the colours as they are, but giving FORECON a different background shape (Circle?), or even just a black background with a green outline?

CatAndHats:
> I'm worried about visibility of the echo icons due to how white they are with white icons. An alternative solution would be keeping the colours as they are, but giving FORECON a different background shape (Circle?), or even just a black background with a green outline?

I have changed the background echo colour to be slightly darker, keeping the correct cyan colour is worth it IMO but the slight darkening of it should make it more readable, pictures updated

PursuitInAshes (dismissed):
lgtm
(I don't know why we don't have this set as a color hex in yaml, but whatever)

Aurallianz (dismissed):

PursuitInAshes (approved):
Yup
We need to change these to just be color hexes as somepoint honestly

DrSmugleaf (approved):

PR #8623, opened 2025-12-09
Bayonet alternatives (Loadout knife cost changes)

august-sun:
## About the PR
Knives too expensive in loadouts

## Why / Balance
shouldn't have to sacrifice drip for the knife

## Requirements

**Changelog**

no cl no fun

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Loadouts/Items/meleeweapons.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Loadouts/Items/meleeweapons.yml
@@ -1,42 +1,42 @@
 # Melee Weapons
 - type: loadout
   id: TantoARMC
-  cost: 1
+  cost: 2
   storage:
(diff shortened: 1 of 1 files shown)


august-sun:
> I believe they are 4 in 13 which might be so not everyone runs around with the other faction knives but we could consider it regardless

Yeah I mean they're limited to one knife right now, just at the moment folks have to not take a lot of other cosmetics in order to claim it, if that makes sense (unless you're saying to keep it to people who REALLY want it)

QuietlyWhisper:
Yeah I think smug has the right idea actually, the intent is based upon less people taking the cosmetic *because* it's expensive, resulting in the items being more rare.

Other than that if they have no mechanical difference it's fine, but I do agree with the point smug brought up

august-sun:
Keeping pricing as-is to keep 'em rare!

PR #8636, opened 2025-12-09
Add name prefix for Ruby rank xenos

blueDev2:
## About the PR

Added name prefixes for Ruby rank xenos
This currently extends prime for regular xenos and merciless for parasites
I have been told that the names might be different, but cm13 wiki is sparse (and likely out of date)

## Requirements

**Changelog**

:cl: blueDev2
- fix: Ruby ranked xenos properly get their prefixes

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Name/XenoRankNamesComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Name/XenoRankNamesComponent.cs
@@ -13,6 +13,6 @@ public sealed partial class XenoRankNamesComponent : Component
         {3, "rmc-xeno-elder"},
         {4, "rmc-xeno-ancient"},
         {5, "rmc-xeno-prime"},
-        {5, "rmc-xeno-ruby"},
+        {6, "rmc-xeno-prime"},
     };
 }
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Xenonids/Name/XenoRankNamesComponent.cs
```diff
@@ -13,5 +13,6 @@ public sealed partial class XenoRankNamesComponent : Component
         {3, "rmc-xeno-elder"},
         {4, "rmc-xeno-ancient"},
         {5, "rmc-xeno-prime"},
+        {5, "rmc-xeno-ruby"},
```
Vermidia: this needs to be 6, repeating 5 won't work

Vermidia: also you can just repeat prime again since its the same name

Vermidia: the loc string I mean

blueDev2: I have been told for larva and para this might be different?
I just don't know
If the rank name for ruby supposed to just be the same as the previous rank, yeah I agree

Vermidia: For larva it would just be their big rank repeated again, for parasites repeating merciless again. For king it'd be I forgot what I called it but should be obvious in the yml.

DrSmugleaf (approved):

PR #8653, opened 2025-12-09
Add skill presets

Dutch-VanDerLinde:
## About the PR
Mainly to reduce the major amounts of shitcode
This exists in cm13 too

Basically, in someone's skill component you can choose to define a ``Preset`` which is a prototype that has a list of skills.
This is so skills can be standardized for survivors. I've noticed people making small mistakes over the tens of survivor role prototypes which is bad

This doesn't add EVERY skill preset but the base and some things like the goons and security so we don't get mistakes across files

Also adds the CO preset

image
skills still work

**Changelog**
No player facing


Content.Shared/_RMC14/Marines/Skills/SkillPresetPrototype.cs
```diff
@@ -0,0 +1,17 @@
+﻿using Robust.Shared.Prototypes;
+
+namespace Content.Shared._RMC14.Marines.Skills;
+
+/// <summary>
+/// A preset of a group of skills. Useful for standardization of skills across roles that have the same skillset.
+/// </summary>
+[Prototype]
+public sealed partial class SkillPresetPrototype : IPrototype
```
DrSmugleaf, line 9: Ideally this should just be an entity prototype with this being its component, so you get name/descriptions for free if we ever want to add commands or UIs for this later, but this is fine for now

PR #8663, opened 2025-12-10
The Infamous ASO Debate, Part Whatever

AverageNotDoingAnythingEnjoyer:
## About the PR

ASO doesnt require any time in Command Department and Req. Dep. as well as Aux Department in general
But they now require 5 hours in Dropship jobs

## Why / Balance

1. As ASO is no longer 3rd in Command and mostly 7th in case of 4 SOs they don't really need a Command timelock. The previous PR that added it justified it by stating that ASO also needs to know overwatch, to which I'd like to say that you don't really need to play SO to learn OW and this only blocks players behind potential extra 10 hours in roles that have next to 0 connection to it right now
2. Req timelock is self-explanatory
3. And replacing Aux Dep timelock with Dropship is motivated by Intel being part of Aux Dep and having the same 5 hour timelock, i.e. you can unlock ASO by only playing Intel which is weird for a head of Pilots too.

## Technical details

YML-maxing

## Media

—

## Requirements

**Changelog**

:cl: J C Denton
- tweak: ASO now doesn't require hours in Command or Requisitions and Aux Dep. generally, but requires 5 hours in Dropship roles.

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Jobs/AuxiliarySupport/auxiliary_support_officer.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Jobs/AuxiliarySupport/auxiliary_support_officer.yml
@@ -17,9 +17,6 @@
   - !type:RoleTimeRequirement
     role: CMJobIntelOfficer
     time: 18000 # 5 hours
-  - !type:DepartmentTimeRequirement
-    department: CMRequisitions
-    time: 18000 # 5 hours
   ranks:
     RMCRankCaptain:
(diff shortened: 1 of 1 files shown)


vorkathbruh:
performing overwatch duties is an expected part of the ASO role so I think just an hour of SO time is fine
removing req playtime is logical
not sure how I feel about the dropship time but seems justified

Aurallianz:
So,
The ASO needs a rework, not patch fixes.
Thanks for putting up a PR, though.

PR #8683, opened 2025-12-11
Add the explosive M20 Claymore anti-personnel mine

### LINKED ISSUE #8074: Any item/nades that spawn a whole bunch of items/shrapnel freeze the server
## Description
Unsure how or why but nades like the M15 or HEFA cause the game to freeze for a moment when they detonate.

### LINKED ISSUE #8147: Shrapnel causing lag?
## Description
I noticed while playing things such as: HEFA, Hornet, Frags and hedgehog sheild have been causing lag spikes. At first i thought it was just my end but others nearby began complaining in LOOC about the spikes, once again during a thunderdome where everyone has M15s when one detonates everyone nearby starts lagging hard.

**Reproduction**
Throw an M15 Grenade or fire a Hornet

Arendian:
## About the PR

- Fixed some errors related to projectile grenades.
- Re-added shrapnel to shrapnel grenades because of the fixed errors.
- Ported the M20 Claymore anti-personnel mine from CM13, it shoots shrapnel in it's facing direction when an entity passes in front of it. The mine only triggers if the entity does not have the same faction as the entity that placed it. It can be picked back up using an access tuner.
- Requisitions can now buy 10 M20 mines for $3000, they also start with some in their vendor.
- All squad roles, except the weapons specialist, can now buy a box of 5 M20 Claymore anti-personnel mines in their vendor for 20(18 for SL) points.

I am not sure if the deploy sound from CM13 can be used, so currently no sound is made when you finish deploying a mine.

## Why / Balance

 I think this Resolves #8147 and Resolves #8074

Mostly parity. Since we have no directional explosions yet, the damage comes from the projectiles instead of a mix of both the explosion and the projectiles.

The mine shoots 12 projectiles in a 120 degree cone, of which a max of 8 can hit the same target.

The projectiles have the following stats:
- Damage: 25
- Armor piercing: 20

if within 3 tiles:
- Stun duration: 1.5 seconds
- Knockback: 1 tile
- Slow duration: 5 second

Xenos can either melee it once, or shoot 2 projectiles(sprite clicked) at it to detonate it.

## Media

https://github.com/user-attachments/assets/3eef10d3-b2dd-456f-a4b0-2a88bc041f1c

## Requirements

**Changelog**

:cl: Dygon
- add: The M20 Claymore anti-personnel mine is now available through requisitions and squad role vendors. It inherits the placer’s IFF signal and detonates when anyone without a matching IFF steps in front of it, releasing shrapnel in a forward cone.
- fix: Shrapnel grenades spawn shrapnel again.


GrigoriyGalintano:
Good

efzapa:
this is up to walance maintainers ultimately but I think having them layer just below mobs/corpses (but above everything else) would be a good idea, primarily to reduce any possible cheesing (especially when OT eventually comes out)

CatoChristopherMrow:
Finally, something to stop the new bravo E1s from leaving cades.

Torma25:
> this is up to walance maintainers ultimately but I think having them layer just below mobs/corpses (but above everything else) would be a good idea, primarily to reduce any possible cheesing (especially when OT eventually comes out)

xeno traps and weed nodes don't get this treatment though, and the former is also a potential round remover.

Arendian:
> Will look at this more later, shrapnel was removed since it caused server freezes, I couldn't replicate those in local though

I did notice some errors that I fixed(I am also the reason they happened since I made the change that caused them in the airburst grenade PR). They were caused by the shrapnel storing a NetEntity to the entity that spawned it on the ProjectileLimitHitsComponent, the entity spawning the shrapnel gets deleted after spawning it's entire payload, so all the shot shrapnel threw errors because their component referenced a deleted entity when they hit something. This means that after a grenade spawned 48 projectiles and deleted itself, the game would throw an error for every single one of them the moment they collided with something. 

Not 100% sure if fixing those errors removed the lag that caused the shrapnel to be removed though, but I didn't notice any possible causes during my testing after fixing them. The errors I fixed were introduced a year ago, and I don't think the grenade lag issue has been a thing for such a long time(or maybe they were but just not reported?), which is why I am not sure if I actually fixed the lag issue.

DrSmugleaf (approved):

PR #8700, opened 2025-12-12
Fixes Bipods staying deployed when moving

### LINKED ISSUE #8377: Deplyoed bipod AND movement? In this Economy? Its likelear than you think!
SO!! A bug I found that is, in my honest opinion, of little to no benefit BUT im still reporting it is: if you deploy your bipod (near a structure like sandbags/cades so its instant) and at the same time hold a direction button, you can basically move into that direction WITH the bipod efffect. You ALSO can stop pressing the movement key to stop in a desired location, retaining your bipod. However, that's the limit of this bug.

https://discord.com/channels/1168210010233376858/1168210011823026274/1437197918443667466

(Discord link to the discord chaneel with an example clip)

Majkl-J:
## About the PR
fixes #8377

## Why / Balance
Fixes a bug

## Technical details
Literally just checks if any movement keys are held after the DoAfter finishes, as the MoveInputEvent only fires when we press new movement buttons

## Media
[Screencast_20251212_131107.webm](https://github.com/user-attachments/assets/3ba440ca-cb73-4ceb-90e7-8ba1f7a2370b)

## Requirements

**Changelog**
:cl:
- fix: Bipods now check if you are moving before deploying

### DIFF
--- a/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
+++ b/Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
@@ -602,11 +602,12 @@ public sealed class AttachableToggleableSystem : EntitySystem
         if (!TryComp(args.Used, out AttachableHolderComponent? holderComponent))
             return;
 
-        if (!attachable.Comp.Active && TryComp<InputMoverComponent>(args.User, out var input )) {
-            if((input.HeldMoveButtons & MoveButtons.AnyDirection) != MoveButtons.None)
-                return;
+        if (!attachable.Comp.Active &&
+            TryComp<InputMoverComponent>(args.User, out var input) &&
+            (input.HeldMoveButtons & MoveButtons.AnyDirection) != MoveButtons.None)
+        {
+            return;
         }
-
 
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Attachable/Systems/AttachableToggleableSystem.cs
```diff
@@ -600,6 +602,12 @@ private void OnAttachableToggleDoAfter(Entity<AttachableToggleableComponent> att
         if (!TryComp(args.Used, out AttachableHolderComponent? holderComponent))
             return;
 
+        if (!attachable.Comp.Active && TryComp<InputMoverComponent>(args.User, out var input )) {
+            if((input.HeldMoveButtons & MoveButtons.AnyDirection) != MoveButtons.None)
+                return;
+        }
+
+
```
DrSmugleaf: Thanks for the PR!
```suggestion
        if (!attachable.Comp.Active &&
            TryComp<InputMoverComponent>(args.User, out var input) &&
            (input.HeldMoveButtons & MoveButtons.AnyDirection) != MoveButtons.None)
        {
            return;
        }

```

DrSmugleaf (dismissed):

PR #8705, opened 2025-12-12
minimum playtime for ghosted queen

Ian321:
## About the PR

title

## Why / Balance

Just had a round where TWO new players took the queen when the original one disconnected.

## Technical details

Set to 10 for now, as that's the minimum required to unlock queen round start. Personally I would even increase it to 24h, that unlocks the 3rd letter.

## Media

image

## Requirements

**Changelog**

:cl:
- tweak: Ghosted queens now require a minimum of 10 hours as xeno to take over.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/queen.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/queen.yml
@@ -15,6 +15,10 @@
     - RMCGuideRoleQueen
   - type: GhostRole
     name: cm-job-name-xeno-queen
+    requirements:
+    - !type:TotalJobsTimeRequirement
+      group: CMJobsXeno
+      time: 36000 # 10 hours
   - type: Sprite
     sprite: _RMC14/Mobs/Xenonids/Queen/queen.rsi
     layers:


AverageNotDoingAnythingEnjoyer:
Just limit it to 1 minute as Queen, maybe
It's done with Forecon that way

Aurallianz:
What I’m going to say may not make complete sense, 
The major issue with this approach is that when a queen SSDs, if there isn’t a 600-minute Xeno ghost, no one else can take on the role of queen.
SO If a queen SSDs for an extended period, it would be better to remove them and free up the queen slot for another xeno, while also returning a larva to the hive.

Ian321:
> SO If a queen SSDs for an extended period, it would be better to remove them and free up the queen slot for another xeno, while also returning a larva to the hive.

That would have the issue of un-maturing the queen. Maybe something that progressively lowers the requirements?

Rehjen:
what if a living xeno could take over the queen (as well as ghosts)? (would probably be a opt in thing like ghost role gamble) - just an idea, depends what the maintainer prefer

Ian321:
> what if a living xeno could take over the queen

we could re-use the king raffle system

PR #8725, opened 2025-12-14
Deployment Restrictions for Nuke Procedure SOP

PursuitInAshes:
## About the PR
Now the MW or CMP, a larva removal surgery qualified person from med (Doctor or CMO), and a member of CIC must remain onboard despite lifted nuclear protocol deployment regulations. These people can be allowed to deploy upon commander discretion. This section mainly exists to encourage commander to think about who they need to stay shipside for things to function, instead of blindly sending everyone down. They still have the authority to overwrite this if the 3~ ish staff are required on the surface.

## Why / Balance
This is just to ensure that all departments can operate properly while still allowing a vast majority of people to deploy.

Engi was skipped from being required to remain as they should be going down to FOB to reinforce and ensure the device actually detonates. OBs can be loaded by the CIC staff if there happens to be an unused OB during nuke (For.. some reason?)
Req was skipped after discussion and reconsideration, as usually their equipment will be sent down with them and requisitions budget will be minimal. If its really required for whatever reason, commander can still require them to stay.

## Technical details
XML

## Media
image

## Requirements

**Changelog**
:cl: PursuitinAshes
- tweak: [SOP] Non-Modifiable Deployment now mandates some staff remain shipside, which can be overwritten if required by the commander.

### DIFF
--- a/Resources/ServerInfo/Guidebook/_RMC14/SOP/RMCMarineSOPNONModDeployment.xml
+++ b/Resources/ServerInfo/Guidebook/_RMC14/SOP/RMCMarineSOPNONModDeployment.xml
@@ -5,7 +5,7 @@
 
   In the event of nuclear ordnance being deployed, all personnel are authorized to deploy to the [bold]FOB[/bold] to protect it during countdown, which overrides all other department SOP preventing deployments. Instead, the following restrictions apply:
 
-  - The following staff must remain shipside, unless stated otherwise by the [color=#008000][bold]Commander[/bold][/color]: The [bold]Military Warden[/bold] or [bold]Chief MP[/bold], a [bold]Doctor or Chief Medical Officer[/bold], and one member of [bold]CIC Personnel[/bold].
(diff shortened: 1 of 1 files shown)


QuietlyWhisper (approved):

PursuitInAshes:
We'll talk about req and MW specifically more, fair on both

MW is mostly there to ensure if someone is arrested they actually get processed in a reasonable time since having to pray someone returns shipside isn't ideal

Req is a fair point, we'll have to talk about more and also maybe see where things settle on it

I'll bring it up again later once I can think on it a tiny bit more

BramvanZijp:
Doesnt MW stand for Military Warden?

CatoChristopherMrow:
You should probably specify at least one member of MP, preferably the warden, instead of specifically THE warden. I say this since there could be a time we don't have a warden and all other MPs simply deploy for the nuke leaving MP empty.

QuietlyWhisper:
Recommended changes were dropping MP requirement unless prisoner but the rest is fine

PursuitInAshes:
@QuietlyWhisper It should be good. Didn't update PR media, but https://github.com/RMC-14/RMC-14/pull/8725/changes/ca3814d06dbbab8d56d5904b3c8511b0c7e40e70 has the adjustment. Let me know if there is anything else

DrSmugleaf (approved):
Probably fine to require a doc since not all maps have close by surgery tables and we don't even have any autodoc stuff yet

PR #8756, opened 2025-12-16
Change the headset in CMD Cabinet

MoonlightHollow:
## About the PR

Break off from my previous PR as I think this simple change has merit. Change out the Command Headset in the Command Tablet Cabinet with a Senior Command Headset. Whoever access the cabinet is the Commander, which is usually CO/XO, but in cases where nether are there then whoever becomes Commander doesn't have a direct line of communication to the MPs. The Commander has a couple of duties under SOP/Mlaw that having a direct way to speak to the MPs would help.

## Media

image

## Requirements

**Changelog**

:cl:
- tweak: Replace the headset in the Command Tablet Cabinet


GoodWheatley:
This is parity

https://github.com/cmss13-devs/cmss13/pull/10832

Issue #2150, opened 2024-05-15
The player that was playing as the hugger that created a larva should get priority on controlling that larva


DrSmugleaf:

Mattshark89:
Implementation idea:
Spectating players who are defibbed receive a prompt to "return to [their] body". Perhaps a similar prompt could appear for the hugger player upon larvae birth, optionally ten or so seconds before the ghost role appears?

sporkyz:
Shouldn't this be closed now that larva queue is merged?

PR #8788, opened 2025-12-19
Larva Queue

### LINKED ISSUE #2651: There should be an automatic queue for observers to become xenos

Aurallianz:
## About the PR

Added Larva Queue...

People who have not observed can still instantly join Burrowed Larva

- Resolves #2651

**Changelog**

:cl:
- add: Added Larva Queue, players will not be able to ghost and join Xeno from round start until after 15 minutes. You must have been dead for at least 5 minutes before you can join the queue.


Nothing98512:
While adding a larva queue is both parity and a very important feature, I believe that making it so no ghosts can take larva before 15 minutes has more consequences than punishing “metagamers” who observe to acquire map knowledge.
It punishes new xenos and unlucky veterans who participate in the oft ordered xeno rushes, making them unable to take burrowed larva and get evo points before 15 minutes- a point around where most queens get up.
It prevents ghosts taking early captures such as monkeys or unlucky survivors.
Any dead or captured survs will also be unable to return for the game until the 15 minutes mark too, facing the same consequences of xenos who died in the rush of being unable to get evo points.
This has an overall negative and punishing effect on xenos that I think will push balance in one direction without any pull in the other direction. Xeno rushes are used so often because the consequences of leaving survs are often greater than rushing them- from tech points to the immediate map control benefits marines acquire.
I would like to restate what I said at the beginning, however, that regardless of the balance effect of this specific implementation, I agree that queue is a very important feature.

jarjarbinksimus:
Could not find code in cm13 relating to not being able to join before 15 minutes, but there is code for instantly able to be a larva after observing from lobby (which is in the game currently) and in experience you can latejoin after observing as well in cm13 and join queue instantly. 
I dont realistically think there is that many ways to meta game roundstart as a ghost, only to find hiding survivors, which should be punished normally, not like this. Most survivors hold anyways, which really cant be metagamed. And if this is supposed to stop drone rushes, it wont, and the reason for drone rushes lies deeply in how the game works survivor vs xeno wise, and adding bandaid timers like this wont fix anything. 

Drone rushes happen because if survivors get to a hold and get it going, it becomes difficult for xenos to do anything to it before tier 3, at which point marines will have landed and saved them. The reason why this is bad is because every alive survives gives marines 0.7 points of intel, totaling to 4.9 points when all 7 survive, and as well survivors may re-deploy and attack the hive, but the intel is more important. 
Since its not parity and a non-parity xeno nerf at that, i think it would be bad for the game.

Bobert-Brockson:
Is 5 minutes being dead the right number?

Content.Server/_RMC14/Rules/CMDistressSignalRuleSystem.cs
MidZik, line 1: This file should be deleted, it was moved under the DistressSignal folder as part of a refactor.

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
```diff
@@ -354,25 +354,29 @@ public bool TryGetFreeSlots(Entity<HiveComponent?> hive, EntProtoId caste, out i
...
 
-    public void IncreaseBurrowedLarva(Entity<HiveComponent> hive, int amount)
+    public void ChangeBurrowedLarva(Entity<HiveComponent> hive, int amount)
     {
         SetHiveBurrowedLarva(hive, hive.Comp.BurrowedLarva + amount);
     }
 
     private void SetHiveBurrowedLarva(Entity<HiveComponent> hive, int larva)
     {
+        var initialValue = hive.Comp.BurrowedLarva;
```
MidZik, line 373: Should probably disallow negative burrowed amounts, and log as an error.
```suggestion
        if (larva < 0)
        {
            Log.Error($"Attempted to set hive burrowed larva to '{larva}', which is less than 0.");
            larva = 0;
        }
        var initialValue = hive.Comp.BurrowedLarva;
```

Content.Shared/_RMC14/Xenonids/JoinXeno/JoinXenoSystem.cs
```diff
@@ -72,21 +72,21 @@ private void OnJoinXenoAction(Entity<JoinXenoComponent> ent, ref JoinXenoActionE
         if (_net.IsClient)
             return;
 
-        var user = args.Performer;
-        if (!CanJoinXeno(user))
+        var denyQueuing = _config.GetCVar(RMCCVars.RMCLarvaQueueRoundstartDelaySeconds);
+        if (_gameTicker.RoundDuration().TotalSeconds <= denyQueuing)
+        {
+            _popup.PopupEntity($"Joining a hive from observer is disabled for {denyQueuing} seconds from the start of the round.", args.Performer);
```
MidZik: Wrong PopupEntity overload, needs to only display to the performer, not everyone in PVS range. Also I think this reads a bit better.

```suggestion
            _popup.PopupEntity($"Joining a hive while observing is disabled for the first {denyQueuing} seconds of the round.", args.Performer, args.Performer);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        var session = actor.PlayerSession.UserId;
+        var actorEntity = actor.PlayerSession.AttachedEntity;
+        var larvaWaitTime = _config.GetCVar(RMCCVars.RMCLarvaQueueWaitSeconds);
+
+        if (actorEntity == null)
+            return;
+
+        if (RemoveFromQueue(actor.PlayerSession.UserId, hive))
+        {
+            _popup.PopupEntity($"You have been removed from the queue", actorEntity.Value);
```
MidZik: Limit who can see the popup to only the user.
```suggestion
            _popup.PopupEntity($"You have been removed from the queue", actorEntity.Value, actorEntity.Value);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+            return;
+        }
+
+        if ((_gameTiming.CurTime - ghost.TimeOfDeath).TotalSeconds >= larvaWaitTime)
+        {
+            Queue.GetOrNew(hive).Add(session);
+            _popup.PopupEntity($"You have been added to the queue in the {Queue.GetOrNew(hive).Count} position",
+                actorEntity.Value);
```
MidZik: Limit who can see the popup to only the user.
```suggestion
                actorEntity.Value, actorEntity.Value);
```

MidZik: Limit who can see the popup to only the user.
```suggestion
                actorEntity.Value, actorEntity.Value);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        if (RemoveFromQueue(actor.PlayerSession.UserId, hive))
+        {
+            _popup.PopupEntity($"You have been removed from the queue", actorEntity.Value);
+            return;
+        }
+
+        if ((_gameTiming.CurTime - ghost.TimeOfDeath).TotalSeconds >= larvaWaitTime)
+        {
+            Queue.GetOrNew(hive).Add(session);
+            _popup.PopupEntity($"You have been added to the queue in the {Queue.GetOrNew(hive).Count} position",
```
MidZik: I think this reads a bit better.
```suggestion
            _popup.PopupEntity($"You have been added to the queue at position {Queue.GetOrNew(hive).Count}",
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                actorEntity.Value);
+        }
+        else
+        {
+            PreQueue.GetOrNew(hive).Add(session, larvaWaitTime + ghost.TimeOfDeath.TotalSeconds);
+            _popup.PopupEntity(
+                $"You have been added to the Pre-queue, until you have waited the required {larvaWaitTime} seconds",
+                actorEntity.Value);
```
MidZik: Limit who can see the popup to only the user.
```suggestion
                actorEntity.Value, actorEntity.Value);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        {
+            Queue.GetOrNew(hive).Add(session);
+            _popup.PopupEntity($"You have been added to the queue in the {Queue.GetOrNew(hive).Count} position",
+                actorEntity.Value);
+        }
+        else
+        {
+            PreQueue.GetOrNew(hive).Add(session, larvaWaitTime + ghost.TimeOfDeath.TotalSeconds);
+            _popup.PopupEntity(
+                $"You have been added to the Pre-queue, until you have waited the required {larvaWaitTime} seconds",
```
MidZik: I think this should display the actual amount of time left. For example, it should read

`$"You died too recently, and will be added to the queue in {timeLeft} seconds."`

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+            return;
+        var hive = (hiveEntity, hiveComp);
+        SpawnLarva(hive);
+    }
+
+    private void SpawnLarva(Entity<HiveComponent> hive)
+    {
+        var list = Queue.GetOrNew(hive);
+        var originalLength = list.Count;
+        if (list.Count < 0)
```
MidZik:
```suggestion
        if (list.Count <= 0)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                _popup.PopupEntity($"You are now in the {Queue.GetOrNew(hive).Count} position for larva queue.",
+                    session.AttachedEntity.Value);
+            }
+        }
+    }
+
+    private void BurstLarvaSpawn(Entity<HiveComponent> hive)
+    {
+        var list = Queue.GetOrNew(hive);
+        if (list.Count < 0)
```
MidZik:
```suggestion
        if (list.Count <= 0)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        var larva = EntityQueryEnumerator<LarvaQueuedComponent>();
+        while (larva.MoveNext(out var ent, out var comp))
+        {
+            if (_mobState.IsDead(ent))
+            {
+                RemComp<LarvaQueuedComponent>(ent);
+                continue;
+            }
+
+            if (!TryComp(ent, out HiveMemberComponent? hiveMember) || hiveMember.Hive != hive.Owner)
```
MidZik:
```suggestion
            if (!_hive.IsMember(ent, hive))
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        var list = Queue.GetOrNew(hive);
+        if (list.Count < 0)
+            return;
+
+        var larva = EntityQueryEnumerator<LarvaQueuedComponent>();
+        while (larva.MoveNext(out var ent, out var comp))
+        {
+            if (_mobState.IsDead(ent))
+            {
+                RemComp<LarvaQueuedComponent>(ent);
```
MidZik: We are in the middle of iterating on this component, probably best to defer removal.
```suggestion
                RemCompDeferred<LarvaQueuedComponent>(ent);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+
+                if (!_player.TryGetSessionById(netId, out var session) || session.AttachedEntity == null)
+                    continue;
+
+                var newMind = _mind.CreateMind(session.UserId,
+                    EntityManager.GetComponent<MetaDataComponent>(ent).EntityName);
+                _mind.TransferTo(newMind, ent, ghostCheckOverride: true);
+                RemComp<LarvaQueuedComponent>(ent);
+                tryFind = true;
+            }
```
MidZik: I see a few issues.

- Players who are disconnected lose their place in line. They should be skipped instead until they return.
- Should use a for loop that iterates by index backwards, instead of a while loop with a flag. For example `for (var i = list.Count - 1; i >= 0; --i) {...}`

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                if (list.Count == 0)
+                    return;
+                var netId = list.Last();
+                list.RemoveAt(list.Count - 1);
+
+                if (!_player.TryGetSessionById(netId, out var session) || session.AttachedEntity == null)
+                    continue;
+
+                var newMind = _mind.CreateMind(session.UserId,
+                    EntityManager.GetComponent<MetaDataComponent>(ent).EntityName);
```
MidZik: Use proxy method instead.
```suggestion
                    MetaData(ent).EntityName);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        var list = Queue.GetOrNew(hive);
+        var originalLength = list.Count;
+        if (list.Count < 0)
+            return;
+
+        BurstLarvaSpawn(hive);
+        BurrowedLarvaSpawn(hive);
+
+        //Popup to people in queue if the queue length changed.
+        if (originalLength >= list.Count)
```
MidZik: Wrong comparison. If queue length changed, `list.Count` will be less than the original length.
```suggestion
        if (list.Count >= originalLength)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+            return;
+
+        foreach (var netId in list)
+        {
+            if (_player.TryGetSessionById(netId, out var session) && session.AttachedEntity != null)
+            {
+                _popup.PopupEntity($"You are now in the {Queue.GetOrNew(hive).Count} position for larva queue.",
+                    session.AttachedEntity.Value);
```
MidZik, line 158: Only the entity should see the popup.
```suggestion
                    session.AttachedEntity.Value, session.AttachedEntity.Value);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        //Popup to people in queue if the queue length changed.
+        if (originalLength >= list.Count)
+            return;
+
+        foreach (var netId in list)
+        {
+            if (_player.TryGetSessionById(netId, out var session) && session.AttachedEntity != null)
+            {
+                _popup.PopupEntity($"You are now in the {Queue.GetOrNew(hive).Count} position for larva queue.",
```
MidZik: Displayed position is incorrect, this will tell everyone their position is the last position. If there are 10 people in the queue, everyone will see "You are now in the 10 position".

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                RemComp<LarvaQueuedComponent>(ent);
+                tryFind = true;
+            }
+        }
+    }
+
+    private void BurrowedLarvaSpawn(Entity<HiveComponent> hive)
+    {
+        var list = Queue.GetOrNew(hive);
+        if (list.Count < 0)
```
MidZik:
```suggestion
        if (list.Count <= 0)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+            return;
+
+        var amount = hive.Comp.BurrowedLarva;
+        if (amount < 1)
```
MidZik: Just for consistency.
```suggestion
        if (amount <= 0)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+
+            if (_player.TryGetSessionById(netId, out var session) && session.AttachedEntity != null)
+            {
+                _hive.JoinBurrowedLarva(hive, session);
+            }
+            else
+            {
+                i--;
+            }
+        }
```
MidZik: As with the earlier code in `BurstLarvaSpawn`, disconnected players lose their place in line. They should keep it until they return.

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    private void OnMindRemoved(Entity<HiveMemberComponent> ent, ref MindRemovedMessage args)
+    {
+        if (!TryComp(ent, out XenoComponent? xeno))
+            return;
+
+        if (HasComp<XenoParasiteComponent>(ent) || !TryPrototype(ent, out var lesserProto) ||
+            lesserProto.ID == Lesser || HasComp<DropshipHijackerComponent>(ent))
+            return;
```
MidZik: See above comment. This code is fragile and there shouldn't be a check like this here. If we add a new `LarvaQueueableComponent` and subscribe to `MindRemovedMessage` on it instead, this check will not be necessary.

Also, see my comment on `base_xeno.yml` for other things `LarvaQueueableComponent` could do here (such as add a ghost role if needed).

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+        }
+    }
+
+    /// <summary>
+    /// This will try and remove them from a queue and return true if they are part of that queue.
+    /// </summary>
+    /// <param name="netUserId">ID of the person we are removing</param>
+    /// <param name="hive">The hive that we are removing them from the queue of</param>
+    /// <returns></returns>
+    private bool RemoveFromQueue(NetUserId netUserId, Entity<HiveComponent> hive)
```
MidZik:
```suggestion
    private bool TryRemoveFromQueue(NetUserId netUserId, Entity<HiveComponent> hive)
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                var netId = list.Last();
+                list.RemoveAt(list.Count - 1);
+
+                if (!_player.TryGetSessionById(netId, out var session) || session.AttachedEntity == null)
+                    continue;
+
+                var newMind = _mind.CreateMind(session.UserId,
+                    EntityManager.GetComponent<MetaDataComponent>(ent).EntityName);
+                _mind.TransferTo(newMind, ent, ghostCheckOverride: true);
+                RemComp<LarvaQueuedComponent>(ent);
```
MidZik: We are in the middle of iterating on this component, probably best to defer removal.
```suggestion
                RemCompDeferred<LarvaQueuedComponent>(ent);
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+    /// </summary>
+    /// <param name="actor">The actor component of the entity</param>
+    private void RemoveFromAllQueues(ActorComponent actor)
+    {
+        var hives = EntityQueryEnumerator<HiveComponent>();
+        while (hives.MoveNext(out var hiveId, out var hive))
+        {
+            RemoveFromQueue(actor.PlayerSession.UserId, (hiveId, hive));
+        }
+    }
```
MidZik, line 246: This function isn't used, and the function below it should be used instead anyway.
```suggestion
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+                    _popup.PopupEntity(
+                        $"You have been added to the queue in the {Queue.GetOrNew(hive.Key).Count} position",
+                        session.AttachedEntity.Value);
+                }
+
+                hive.Value.Remove(user.Key);
+            }
+
+            SpawnLarva(hive.Key);
+        }
```
MidZik: A few changes/fixes.
- Destructure key/value pairs in the foreach loop
- Made popup read a bit better
- Fix popup appearing to everyone instead of only the user
```suggestion
        foreach (var (hive, hivePreQueue) in PreQueue)
        {
            foreach (var (userId, userTime) in hivePreQueue)
            {
                if (userTime >= time.TotalSeconds)
                    continue;
                Queue.GetOrNew(hive).Add(userId);

                if (_player.TryGetSessionById(userId, out var session) && session.AttachedEntity != null)
                {
                    _popup.PopupEntity(
                        $"You have been added to the queue at position {Queue.GetOrNew(hive).Count}",
                        session.AttachedEntity.Value, session.AttachedEntity.Value);
                }

                hivePreQueue.Remove(userId);
            }

            SpawnLarva(hive);
        }
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+    [Dependency] private readonly ISharedPlayerManager _player = default!;
+    [Dependency] private readonly SharedPopupSystem _popup = default!;
+
+    private static readonly EntProtoId Lesser = "CMXenoLesserDrone";
+
+    [ViewVariables]
+    public static readonly Dictionary<Entity<HiveComponent>, List<NetUserId>> Queue = new();
+
+    [ViewVariables]
+    public static readonly Dictionary<Entity<HiveComponent>, Dictionary<NetUserId, double>> PreQueue = new();
```
MidZik: I am very unsure about having `Entity<HiveComponent>` as a key. I think `EntityUid` is better, and then resolve the HiveComponent when it is needed. (This will require more changes in the rest of the file.)
```suggestion
    [ViewVariables]
    public static readonly Dictionary<EntityUid, List<NetUserId>> Queue = new();

    [ViewVariables]
    public static readonly Dictionary<EntityUid, Dictionary<NetUserId, double>> PreQueue = new();
```

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
```diff
@@ -224,15 +224,6 @@
...
   - type: GhostTakeoverAvailable
-  - type: GhostRole
-    makeSentient: true
-    allowSpeech: true
-    allowMovement: true
-    name: cm-xeno-name
-    description: cm-xeno-description
-    raffle:
-      settings: short
-    rules: cm-xeno-rules
```
MidZik, line 235: I have concerns about this general removal of GhostRole from most of the xenos. If we _do_ need to make xenos with ghost roles, data they used to have is now just gone. For example, if an admin is doing thunderdome things, the xenos they make will not be ghosts. Even if an admin manually adds GhostRole to the xeno, it will not be populated with the correct data (name, description, rules, etc).

I think the design should be changed so that xenos still have their GhostRole data in some way. Instead, LarvaQueueSystem should somehow ensure the raffles never appear if the xeno is available for those in the queue, but DOES appear if the xeno isn't part of a queue (for example, when it has no hive, such as in a thunderdome).

I am not currently sure how best to do this. Perhaps a new LarvaQueueableComponent that ALL xenos that could be part of a queue get. This new component can have a component registry defined that has GhostRole component in it. That way, the system will create the appropriate GhostRole component when needed, and no data is lost.

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+
+    public override void Initialize()
+    {
+        SubscribeLocalEvent<JoinXenoComponent, JoinLarvaQueueEvent>(OnJoinLarvaQueue);
+        SubscribeLocalEvent<RoundRestartCleanupEvent>(OnCleanup);
+        SubscribeLocalEvent<BurrowedLarvaAddedEvent>(OnBurrowedLarvaAdded);
+        SubscribeLocalEvent<PlayerAttachedEvent>(OnPlayerAttached);
+        SubscribeLocalEvent<HiveMemberComponent, MindRemovedMessage>(OnMindRemoved);
```
MidZik: Subscribing to `<HiveMemberComponent, MindRemovedMessage>` here doesn't seem great. I suggest making a new component `LarvaQueueableComponent` that xenos who can be queued up for should get.
```suggestion
        SubscribeLocalEvent<LarvaQueueableComponent, MindRemovedMessage>(OnMindRemoved);
```

MidZik (changes requested):
First round of reviews, may need to review again later

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
Aurallianz: uhh it at one point in time could be negative, and its sort of an arbitrary restriction that wouldn't do anything

MidZik: Can you explain when it could or should be negative?

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
Aurallianz: this is a hold over for a future change Ill deal with it at a later date if its not required

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
Aurallianz: Specific ghost role spawners should be added for Thunderdome etc., I don't want to touch ghost role code what so ever.

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
MidZik: Not strictly required, but there is no reason to have it. Wherever you are using this function, it should be calling the other function instead. i.e. `RemoveFromAllQueues(actorComp.PlayerSession.UserId)` instead of `RemoveFromAllQueues(actorComp)`

MidZik: Alternatively if you are going to be calling this function a lot you can do the following, this would also be acceptable.
```c#
private void RemoveFromAllQueues(ActorComponent actor)
{
    RemoveFromAllQueues(actor.PlayerSession.UserId);
}
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
MidZik: Sorry, I put this as a suggestion but it requires other changes in the file as well (renaming all references to this function).

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+
+        var hive = new Entity<HiveComponent>(hiveUid.Value, hiveComp);
+        var session = actor.PlayerSession.UserId;
+        var actorEntity = actor.PlayerSession.AttachedEntity;
+        var larvaWaitTime = _config.GetCVar(RMCCVars.RMCLarvaQueueWaitSeconds);
+
+        if (actorEntity == null)
+            return;
+
+        if (RemoveFromQueue(actor.PlayerSession.UserId, hive))
```
MidZik:
```suggestion
        if (TryRemoveFromQueue(actor.PlayerSession.UserId, hive))
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+            RemoveFromQueue(actor.PlayerSession.UserId, (hiveId, hive));
+        }
+    }
+
+    private void RemoveFromAllQueues(NetUserId netUserId)
+    {
+        var hives = EntityQueryEnumerator<HiveComponent>();
+        while (hives.MoveNext(out var hiveId, out var hive))
+        {
+            RemoveFromQueue(netUserId, (hiveId, hive));
```
MidZik:
```suggestion
            TryRemoveFromQueue(netUserId, (hiveId, hive));
```

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -0,0 +1,299 @@
...
+    /// <summary>
+    /// This will remove them from all queues.
+    /// </summary>
+    /// <param name="actor">The actor component of the entity</param>
+    private void RemoveFromAllQueues(ActorComponent actor)
+    {
+        var hives = EntityQueryEnumerator<HiveComponent>();
+        while (hives.MoveNext(out var hiveId, out var hive))
+        {
+            RemoveFromQueue(actor.PlayerSession.UserId, (hiveId, hive));
```
MidZik:
```suggestion
            TryRemoveFromQueue(actor.PlayerSession.UserId, (hiveId, hive));
```

MidZik (changes requested):
My mistake, I should have included these suggestions as well.

AverageNotDoingAnythingEnjoyer:
Does this account for #2150

Content.Shared/_RMC14/CCVar/RMCCVars.cs
```diff
@@ -561,6 +561,13 @@ public sealed partial class RMCCVars : CVars
...
 
+    public static readonly CVarDef<int> RMCLarvaQueueRoundstartDelaySeconds =
+        CVarDef.Create("rmc.larva_queue_roundstart_delay_seconds", 900, CVar.REPLICATED | CVar.SERVER);
+
+    public static readonly CVarDef<int> RMCLarvaQueueWaitSeconds =
+        CVarDef.Create("rmc.larva_queue_wait_seconds", 300, CVar.REPLICATED | CVar.SERVER);
+
     public static readonly CVarDef<string> RMCChemMasterPresets =
         CVarDef.Create("rmc.chemmaster_presets", "", CVar.CLIENT | CVar.ARCHIVE);
+
```
DrSmugleaf:
```suggestion
```

DrSmugleaf (dismissed):
Might want a big warning on observe
Also asking QA first before I merge this

Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
DrSmugleaf: I noted this down as a TODO now

Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/base_xeno.yml
DrSmugleaf: This is probably best handled with a command for admins to ghost-role nearby entities

DrSmugleaf (approved):

PR #8332, opened 2025-11-06
More Chatsans

noctyrnal:
## About the PR

Title, primarily for xenonids.

## Why / Balance

Mechanical enforcement is good

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

- [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

no cl

PR #8798, opened 2025-12-20
More Chatsans

noctyrnal:
## About the PR

title

## Why / Balance

Chatsan was being bypassed by these words so this covers those.

## Requirements

**Changelog**
!


QuietlyWhisper:
Why:
- Xenonids are also accountable for roleplay standards
- "Tax" is both not at all understandable to someone who isnt aware of the made up term, and is also part of gamer terms we dont want in rmc (roleplay standards). Telling others what you want to do instead of using vague "moba style callouts" is the prefered outcome. Bypassing this will result in admin actions being taken, as this is detrimental to the community for both these reasons.

https://github.com/RMC-14/RMC-14/pull/8332#issuecomment-3511386946

PR #8808, opened 2025-12-20
Disallows transfer of M13 ammo to M63 boxes, adds own box.

### LINKED ISSUE #8802: M13 AP rounds should not be loaded into the 10×20 AP ammo box.
## Description
You can fill the submachine gun box with M13 AP, and from there into the M63.

AverageNotDoingAnythingEnjoyer:
## About the PR

Fixes #8802 
Turns out that M13 ammo, while of 10x20mm caliber has an APC tip, i.e. it's slightly different from the one M63 uses and should use a separate ammo box and stuff
Also added crafting said new box
And potentially fixed some incorrect stats

## Why / Balance

Parity

## Technical details

YML-maxing
For reference, CM stats are here: https://github.com/cmss13-devs/cmss13/blob/master/code/datums/ammo/bullet/pistol.dm#L311.

## Media

Screenshot 2025-12-20 at 20 27 54

## Requirements

**Changelog**

:cl: J C Denton
- fix: Fixed the M13 autopistol having a lower damage multiplier than intended, of 0.65 times instead of 0.75 times. The extended recoil compensator now gives the correct amount of plus 0.05, instead of plus 0.10, for a total of 0.75 without, and 0.8 with, instead of 0.65 without, and 0.75 with.
- fix: Fixed the M13 autopistol having a higher damage fall-off range than intended, falling off at 4 tiles instead of the intended 3 tiles.
- fix: Fixed the M13 autopistol having a higher accurate range than intended, losing accuracy at 4 tiles instead of the intended 2 tiles.
- tweak: M13 ammo is now stored in a pistol ammunition box (10x20-APC) with 984 bullets, instead of the standard smg ammunition box (10x20).
- fix: Fixed being able to transfer ammo between M63 and M13 magazines via (10x20) ammunition box.


DrSmugleaf (changes requested):
This is missing the fix for the damage multiplier, it is lower in 14 than intended, check the values with and without the barrel charger attachment it comes with

AverageNotDoingAnythingEnjoyer:
Should be done, also added a desc_lore bit

AverageNotDoingAnythingEnjoyer:
Unsure whether solid stock wielding mode as well as onehanded movement accuracy penalty are correct tho
Screenshot 2026-01-24 at 00 29 12
Screenshot 2026-01-24 at 00 29 38

PR #8830, opened 2025-12-23
New xeno structure. Plasma tree!

Pgriha:
## About the PR

Xeno hive structure from CM13 (https://github.com/cmss13-devs/cmss13/pull/9926) - **Plasma Tree**!
Literally like Recovery Node, but give plasma instead of healing.

One hive can have up to 3 Plasma Trees. They need the same amount of plasma and space, as nodes. Gives 75 plasma every 3 seconds (25 plasma in a second).
It has the same map icon as recovery node (as in CM13 too).

**Sprite made by Miyugi and used with her confirmation.**

## Why / Balance

Parity.

## Technical details

.yml and .cs made as same for recovery node. Drones, Queen and all T2 Drone evos can build it and of course added to guides.

PlasmaTreeSystem.cs can have some small issues, first time I did something like that

## Media

image
MyServer - Rouny&#39;s Marine Corps 14 23 12 2025 10_54_29

https://github.com/user-attachments/assets/fd636247-d64b-4238-aee8-92eae3a0d9ca

### Guides:
image
image

## Requirements

**Changelog**

:cl: Miyugi and Pgriha
- add: New xenonids structure, Plasma Tree! It replenishes plasma to nearby xenos every few seconds, works like the Recovery Node.


QuietlyWhisper (commented):
This is a good idea for encouraging more building but also conflicts with a few design points that need to be considered

- Healer too strong in current state and this may make the issue worse
- Makes it easier to bypass the resin scaling costs within an area meant to discourage structure spam within the same area

jarjarbinksimus:
This will not affect healer at all, healer will always run out of hp before it will run out of plasma. 
Yes, the plasma tree will make it more effective to build since it... gives you plasma. Denying a parity feature because of the way some xenos build isn't going to get the game closer to parity.

Randomposter19:
< This is a good idea to assist on a mechanic of the game, we just need to consider the design because of other mechanics (Neither confirming or denying, just giving their thoughts)

> Denying a parity feature because of the way some xenos build isn't going to get the game closer to parity.

6f2d1a455d896b7660477d55b034616e

jarjarbinksimus:
< "This is a good idea to assist on a mechanic of the game, we just need to consider the design because of other mechanics"
Plasma tree is not a game breaking feature, it just allows most xenos to build more, trap more and regain their plasma quicker with a limited structure.  Marines have tools to break this resin and these traps, unless you are waiting for even more tools to be added, which is not alot of them. If there is any other feature than the scaling feature and the weedkiller that is limiting xeno building, I don't know them or they are not very impactful.

Pgriha:
Ah, so, as already here said and something from Miyugi:
* Healers don't really have problem with plasma, usually (and i don't understand where they are OP, even prime drones very easy to kill with nades or CAS). And with main xeno tactic "500 rotations" all nodes actually useless. They make sense only in hold one direction for a long time.
* For using it by lords and other T2 drone evos they still need to build it (850 plasma), repeatly come to it and rest (time) and marines still have shittone of methods to destroy all resin, so no really many will change. Imo, lords with actually restrictions useless and need many time to build at least something to hold area.

DrSmugleaf (approved):

PR #8839, opened 2025-12-24
Marines can no longer push Xenos that are standing still.

Thereisabear:
## About the PR
Marines can no longer push Xenonid, if the Xenonid is standing still.
Does not apply to lessers of any kind as they have no mob collision (that I'm aware of)

## Why / Balance
Parity. You cannot push any Xeno off a tile as a Marine in CM.
Also emulates the whole "jostling" over the same tile in CM when both are "moving"...at least I hope so

## Technical details
- A cvar addedto Content Shared rmccvars
- RMCMobCollisionMass Component to base species prototype, to enable not pushing Xenos around.
(_does not appear to affect anything but the calculation push or cancel movement_)
- Emulated and altered the code already present that governs the interaction between T1,T2 Xenos vs T3s.

## Media
https://medal.tv/games/space-station-14/clips/lLHxmRQXYkfmg85x3?invite=cr-MSxadE4sMzU5ODkwMzA4&v=36

## Requirements

**Changelog**
:cl:
- tweak: Sips shared the fruits with the Hive; now that Xenos ate well, they can no longer be pushed around whilst alive by Marines.


Content.Shared/Movement/Systems/SharedMobCollisionSystem.cs
```diff
@@ -334,6 +338,13 @@ protected bool HandleCollisions(Entity<MobCollisionComponent, PhysicsComponent>
             {
                 cancellableDirection += mobMovement;
             }
+
+            if (_xenosCancelMovement &&
+                userIsXeno &&
+                _marineQuery.HasComp(other))
```
Dutch-VanDerLinde, line 344:
```suggestion
                _marineQuery.HasComp(other))
```
Might be better to hook this onto some other component instead of marines, incase you want to make some kind of thing (maybe synths) able to move xenos

Thereisabear: > Might be better to hook this onto some other component instead of marines, incase you want to make some kind of thing (maybe synths) able to move xenos

Synths of any kind cannot push Xenos, they can only grab and drag T1s and some T2s.
Being able to grab and drag them isn't affected by this PR as its handled by another component.

Randomposter12: > Synths of any kind cannot push Xenos, they can only grab and drag T1s and some T2s. Being able to grab and drag them isn't affected by this PR as its handled by another component.

Do note that afaik xenos also cannot push synths if theyre standing still, which CM13 handles that with rules "Body blocking the target’s escape is also to be considered a violation of self-preservation programming."

Thereisabear: > > Synths of any kind cannot push Xenos, they can only grab and drag T1s and some T2s. Being able to grab and drag them isn't affected by this PR as its handled by another component.
> 
> Do note that afaik xenos also cannot push synths if theyre standing still, which CM13 handles that with rules "Body blocking the target’s escape is also to be considered a violation of self-preservation programming."

am pathing through synths as a drone currently.

Randomposter12: > am pathing through synths as a drone currently.

Is the synth on harm mode?

Thereisabear: https://github.com/user-attachments/assets/a0b54b33-896c-48b1-80ed-633ad1135b2c

Dutch-VanDerLinde: I'm not talking about this from a balancing perspective, simply splitting it into components so it isn't hardcoded, this is to make it easier to make changes in the future.

DrSmugleaf (changes requested):
I need to look through this more with the new movement code changes from wizden that we have, marines not pushing xenos is fine, marines not sneaking through xenos doesn't make any sense, for one because we aren't trying to have the hard mob collision of 13 since it is inherently unreliable in a properly networked game, and for two because xenos aren't supposed to be able to freely push through marines either, they bump into them and the marine can walk back into the tile to block them
See this clip from me in 13 to see what I mean

https://github.com/user-attachments/assets/6eb75515-a620-485a-8013-19420f9153ad

Thereisabear:
> I need to look through this more with the new movement code changes from wizden that we have, marines not pushing xenos is fine, marines not sneaking through xenos doesn't make any sense, for one because we aren't trying to have the hard mob collision of 13 since it is inherently unreliable in a properly networked game, and for two because xenos aren't supposed to be able to freely push through marines either, they bump into them and the marine can walk back into the tile to block them See this clip from me in 13 to see what I mean
> lurker.mp4



> I need to look through this more with the new movement code changes from wizden that we have, marines not pushing xenos is fine, marines not sneaking through xenos doesn't make any sense, for one because we aren't trying to have the hard mob collision of 13 since it is inherently unreliable in a properly networked game, and for two because xenos aren't supposed to be able to freely push through marines either, they bump into them and the marine can walk back into the tile to block them See this clip from me in 13 to see what I mean
> lurker.mp4

so..the way i had it prior was correct? because thats what i had previously where marines cannot push the xeno around if theyre stationary but with enough speed they kinda push through (but xeno remains in place)

PR #8843, opened 2025-12-24
Fix: Acid Spray should still cause sizzle damage when hitting 5th stack.

Bobert-Brockson:
## About the PR
Gave the stacking acid spray ability the needed damage aspect of the stacking component

## Why / Balance
This is likely a dramatic praetorian buff.

Praetorian does 30 additional damage when you apply the 5th stack of acid to a target. This however does not occur if the 5th stack is applied from acid spray. 

Acid spray is notably able to penetrate marine blobs to get the 5th stack on a target who is almost able to escape. It also is able to apply as many as 4 stacks to a target if they are between tiles or moving in the direction of the attack.

## Technical details
The stacking acid ball and spit abilities had an acid stacking component that included "damage.type.Heat: 30" but the acid spray only had "amount: 2" meaning that if the last stack is given from the spray it is undefined when the 5th stack is applied. However it is only checked when the 5th stack is applied, causing it to do no additional damage at all.

## Media
I'll add this after the holidays

## Requirements

**Changelog**

:cl:
- fix: Praetorian sizzle no longer does 0 damage if the 5th stack is applied by spray acid.

### DIFF
--- a/Resources/Prototypes/_RMC14/Effects/xeno_acid.yml
+++ b/Resources/Prototypes/_RMC14/Effects/xeno_acid.yml
@@ -107,6 +107,9 @@
     - state: acid2
   - type: ApplyAcidStacks
     amount: 2
+    damage:
+      types:
+        Heat: 30
   - type: DamageOnCollide
     damage:
       types:


Bobert-Brockson:
Seems like this is a bug per discussion in discord
https://discord.com/channels/1168210010233376858/1453342694549098517

PR #8844, opened 2025-12-24
[Add] Added names for certain entities visible to Xenos.

ArZarLordOfMango:
## About the PR

Add FixedIdentity to  certain entity for Xeno(see media).
Also add chatsan for signal flare and tesla

signal flare = green stick
tesla = shock emitter
## Why / Balance

Foremost, it is made for new player, so that when they heard, for example, evil eye, they can understand ingame what is.
Second, for immersion in the game, because when they say green stick and point to the A-17, it little breaks  immersion.
## Media

Screenshot 2026-01-06 182248
Screenshot 2026-01-06 182356
Screenshot 2026-01-06 181732
Screenshot 2026-01-06 181744
Screenshot 2026-01-06 181717

## Requirements

**Changelog**

:cl:
- add: Added names for certain entities visible to Xenos.


Helix-ctrl:
Would be neat if plasteel barricades and such had different names from their normal metal counterpart such as "reinforced metal resin"

JackSage:
auto spitters should be more descriptive based on what type, like shotgun autospitter should be an auto stunspitter, sniper sentry should be auto longspitter, etc...

Nades should also be named based on what they do, so fire nades should be fire boomsticks, etc...

And also yeah, plasteel cades should be reinforced metal resin

I think adding xeno style names for stuff is neat, but it shouldn't reduce the amount of info someone (especially new players - since they'll probably suffer the most from not being able to identify something based on sprite alone and instead using shift click to see the difference between shotgun sentry, standard sentry, and sniper sentry)

CatAndHats:
PR is inactive, has been rolled into https://github.com/RMC-14/RMC-14/pull/10328 with credit

Issue #8033, opened 2025-10-01
Tracker sometimes randomly reset to your teamlead if your in a fireteam


GGlut:
## Description
If your in a fireteam when your tracker got autoset to your teamlead it just randomly resets to your teamlead instead of staying on whatever you have set it.

**Additional context**

https://github.com/user-attachments/assets/dbdd14fe-4b04-42a8-a54e-2390535b0b54

Issue #7605, opened 2025-08-24
All Forecon squad trackers cant track their squadlead


XxSWAG-MASTER-OVERLORDxX:
## Description
Currently it sets itself to any other SL and not the forecon SL

XxSWAG-MASTER-OVERLORDxX:
Just to add, when setting it to CO it works fine, only the SL tracker is broken

Dutch-VanDerLinde:
fixxeed

Issue #7228, opened 2025-07-23
Xeno Tracker does not track the queen by default


Tyzemol:
## Description
Evolving resets it back to tracking nothing

Arendian:
Is this still happening? I am unable to reproduce this.

https://github.com/user-attachments/assets/30401a19-444f-43b4-bce2-d26669d8ac52

Tunguso4ka:
It is on server. If i spawn as lesser, i have to manually select queen.

XxSWAG-MASTER-OVERLORDxX:
This is an issue again, and it breaks every time the queen gets up or goes to ovi, as well as whenever you change strain or evolve/devolve

keegaroo65:
https://github.com/RMC-14/RMC-14/pull/9328 I wasn't aware it was also an issue with queen ovi but my PR fixes it on evolve/devolve. I can investigate queen ovi cases later if it's still happening 👀

PR #8855, opened 2025-12-26
Squad and hive tracker fixes

### LINKED ISSUE #8908: Queen tracker sometimes resets
## Description
Sometimes the queen tracker gets reset and does not track the queen, needing to be set to track the queen again

Dutch-VanDerLinde:
## About the PR
Hive tracker now automatically tracks queen on mapinit
It also sets the tracker when a xeno evolves to the original target

Gave the FTL tracker the sprite from CM13 so you know when you're tracking the FTL
Made the SL the default mode for the FORECON headsets since CO won't always be there, but an SL will

Gave forecon SL the trackable component

resolves #8908

## Media
image

image

https://github.com/user-attachments/assets/b4e13cd5-edc0-4a69-b089-9dd0a73499ea

**Changelog**

:cl:
- add: The FTL tracker now has a special sprite when you switch to that mode.
- fix: Fixed FORECON Squad leader tracker not being functional.
- fix: Fixed Xenos hive tracker resetting each time you evolve. The hive tracker not automatically setting to queen has also been fixed.

### DIFF
--- a/Resources/Prototypes/_RMC14/Trackers/Trackers.yml
+++ b/Resources/Prototypes/_RMC14/Trackers/Trackers.yml
@@ -41,7 +41,7 @@
 
 - type: trackerMode
   id: SquadLeader
-  component: SquadLeader
+  job: CMSquadLeader
 
 - type: trackerMode
   id: SectionSergeant


Arendian:
Were you able to reproduce the tracker not staying on target locally (I know it breaks on the server) before this change, and test if this change fixes the issue?

When I worked on this myself, I wasn’t able to reproduce the issue locally (see my comment in #7228), which made it difficult for me to verify whether an attempted fix actually resolved the problem.
I think the following issues are related to the issue closed by this PR, and can be closed too: #7228, #8033, #7605

PR #8860, opened 2025-12-27
Makes walls no longer eat cas missiles

### LINKED ISSUE #6061: missiles get deleted by walls
## Description

When CAS fires a missile, the green dot can appear on a wall. which results in the missile not exploding on impact.

**Reproduction**

Fire a missile in a area with a lot of (indestructible?) walls.

**Screenshots**

Hard to replicate, have seen it happen twice already.

**Additional context**

Maybe have it look at the tiles around the chosen target to find a free space instead of deleting the missile?

CatoChristopherMrow:
## About the PR

Cas missiles no longer get eaten by walls doing no damage.

## Why / Balance

Fixes: https://github.com/RMC-14/RMC-14/issues/6061

Parity, to an extent. Also general QOL.

## Technical details

For starters, if there's a wall in the way, it just downright deletes it if it's destroyable. Only the wall the missile landed directly on though. Given that it's a missile, I think this is fine. Secondly, if it's going to fire into a wall that's NOT destructible, it finds the closest area that would be valid to fire into, upto 3 tiles away. If there's nothing in 3 tiles, it gives up and fires into the wall anyways.

## Requirements

**Changelog**

:cl:
- tweak: Missiles now destroy any wall that they are directly fired into.
- fix: Missiles can no longer fire into indestructible walls.

### DIFF
--- a/Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
+++ b/Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
@@ -38,6 +38,7 @@ using Content.Shared.Popups;
 using Content.Shared.Projectiles;
 using Content.Shared.Shuttles.Components;
 using Content.Shared.Shuttles.Systems;
+using Content.Shared.Tag;
 using Content.Shared.Throwing;
 using Robust.Shared.Audio;
 using Robust.Shared.Audio.Systems;
@@ -90,6 +91,13 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
     [Dependency] private readonly IGameTiming _timing = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
+    [Dependency] private readonly TagSystem _tagSystem = default!;
+    [Dependency] private readonly ILogManager _log = default!;
+
+    private EntityQuery<RMCWallExplosionDeletableComponent> _wallDeletableQuery;
+    private readonly HashSet<EntityUid> _wallDeleteBuffer = new();
+
+    private readonly HashSet<EntityUid> _tileEnts = new();
 
     private static readonly EntProtoId DropshipTargetMarker = "RMCLaserDropshipTarget";
     private const string SpotlightState = "spotlights_";
@@ -130,6 +138,7 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
         SubscribeLocalEvent<DropshipAmmoComponent, PowerLoaderInteractEvent>(OnAmmoInteract);
 
         SubscribeLocalEvent<ActivateDropshipWeaponOnSpawnComponent, MapInitEvent>(OnDropshipWeaponOnSpawnFire);
+        _wallDeletableQuery = GetEntityQuery<RMCWallExplosionDeletableComponent>();
 
         Subs.BuiEvents<DropshipTerminalWeaponsComponent>(DropshipTerminalWeaponsUi.Key,
             subs =>
@@ -639,6 +648,10 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
         if (spread != 0)
             targetCoords = targetCoords.Offset(_random.NextVector2(-spread, spread + 1));
 
+        if (ev.Explosion != null)
+        {
+            targetCoords = FindAlternateLandingTile(targetCoords, 3);
+        }
         var inFlight = Spawn(null, MapCoordinates.Nullspace);
         var inFlightComp = new AmmoInFlightComponent
         {
@@ -681,7 +694,7 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
         var inFlight = Spawn(null, MapCoordinates.Nullspace);
         var inFlightComp = new AmmoInFlightComponent
         {
-            Target = _transform.GetMoverCoordinates(active).SnapToGrid(EntityManager, _mapManager),
+            Target = FindAlternateLandingTile(_transform.GetMoverCoordinates(active).SnapToGrid(EntityManager, _mapManager),3),
             MarkerAt = time + ammo.TravelTime,
             ShotsLeft = ammo.RoundsPerShot,
             ShotsPerVolley = ammo.ShotsPerVolley,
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -90,6 +91,13 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
     [Dependency] private readonly IGameTiming _timing = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
+    [Dependency] private readonly TagSystem _tagSystem = default!;
+    [Dependency] private readonly ILogManager _log = default!;
```
Vixting: This isn't used anywhere. (the log manager)

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -1390,6 +1407,89 @@ out _
...
+        _entityLookup.GetEntitiesInRange(impactSnap, 0.49f, _wallDeleteBuffer, LookupFlags.Uncontained);
+
+        foreach (var ent in _wallDeleteBuffer)
+        {
+            if (!_wallDeletableQuery.HasComp(ent))
+                continue;
+            var entSnap = Transform(ent).Coordinates.SnapToGrid(EntityManager, _mapManager);
+            if (entSnap != impactSnap)
+                continue;
+            EntityManager.DeleteEntity(ent);
```
Vixting: Use QueueDel(ent) instead

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -1390,6 +1407,89 @@ out _
...
+
+            if (!IsWall(ent))
+                continue;
+
+            foundWall = true;
+            if (_wallDeletableQuery.HasComp(ent))
+                return false;
+        }
+        return foundWall;
+    }
```
Vixting, line 1435:
```suggestion
    private bool HasNonDeletableWallOnTile(EntityCoordinates impactCoords)
    {
        if (_transform.GetGrid(impactCoords) is not { } gridUid ||
            !TryComp(gridUid, out MapGridComponent? grid))
            return false;

        var tile = _map.LocalToTile(gridUid, grid, impactCoords);
        var anchored = _map.GetAnchoredEntitiesEnumerator(gridUid, grid, tile);

        var foundWall = false;
        while (anchored.MoveNext(out var uid))
        {
            if (!IsWall(uid.Value))
                continue;

            foundWall = true;
            if (_wallDeletableQuery.HasComp(uid.Value))
                return false;
        }
        return foundWall;
    }
```

walls are always anchored, don't need to check all entities in a range

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -1390,6 +1407,89 @@ out _
...
+
+        foreach (var ent in _wallDeleteBuffer)
+        {
+            if (!_wallDeletableQuery.HasComp(ent))
+                continue;
+            var entSnap = Transform(ent).Coordinates.SnapToGrid(EntityManager, _mapManager);
+            if (entSnap != impactSnap)
+                continue;
+            EntityManager.DeleteEntity(ent);
+            return true;
```
Vixting, line 1473:
```suggestion
    private bool TryDeleteDestructibleWallAt(EntityCoordinates impactCoords)
    {
        if (_transform.GetGrid(impactCoords) is not { } gridUid ||
            !TryComp(gridUid, out MapGridComponent? grid))
            return false;

        var tile = _map.LocalToTile(gridUid, grid, impactCoords);
        var anchored = _map.GetAnchoredEntitiesEnumerator(gridUid, grid, tile);

        while (anchored.MoveNext(out var uid))
        {
            if (!_wallDeletableQuery.HasComp(uid.Value))
                continue;

            QueueDel(uid.Value);
            return true;
        }

        return false;
    }
```

same thing here

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -639,6 +648,10 @@ private void OnWeaponsFireMsg(Entity<DropshipTerminalWeaponsComponent> ent, ref
         if (spread != 0)
             targetCoords = targetCoords.Offset(_random.NextVector2(-spread, spread + 1));
 
+        if (ev.Explosion != null)
+        {
+            targetCoords = FindAlternateLandingTile(targetCoords, 3);
+        }
```
Vixting:
```suggestion
        if (ev.Explosion != null && HasNonDeletableWallOnTile(targetCoords))
            targetCoords = FindAlternateLandingTile(targetCoords, 3);
```

rather get out of it here then in the FindAlternate

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -1390,6 +1407,89 @@ out _
...
+        return foundWall;
+    }
+
+    private EntityCoordinates FindAlternateLandingTile(EntityCoordinates desired, int maxRadius = 3)
+    {
+        var origin = desired.SnapToGrid(EntityManager, _mapManager);
+
+        if (!HasNonDeletableWallOnTile(origin))
+            return origin;
```
Vixting:
```suggestion
```
Already check for this aove

Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
```diff
@@ -1390,6 +1407,89 @@ out _
...
+
+        for (var r = 1; r <= maxRadius; r++)
+        {
+            var offsets = new List<Vector2i> { new(r,0), new(-r,0), new(0,r), new(0,-r) };
```
Vixting: fairly sure this is redundant, aren't these already produced by the loop?

Vixting (changes requested):
@vixting in Discord if you need any help addressing the issues.

Vixting (approved):
I've addressed the issues, should be fine now.

CatoChristopherMrow:
> I've addressed the issues, should be fine now.

Thank you kindly, I've been in the byond mines all day.

DrSmugleaf (approved):

PR #8882, opened 2025-12-30
New stun baton effects

Dutch-VanDerLinde:
## About the PR
Added the activated sound and people hit by the baton jitter a little, from CM13

Also fixes the bug of the stun baton stunning even when deactivated

https://github.com/user-attachments/assets/6729a85e-d17f-4562-a48b-f721c792fd7a

**Changelog**
:cl:
- tweak: Added sounds and effects for being hit by a stun baton

### DIFF
--- a/Content.Shared/_RMC14/Stamina/RMCStaminaSystem.cs
+++ b/Content.Shared/_RMC14/Stamina/RMCStaminaSystem.cs
@@ -190,7 +190,8 @@ public sealed partial class RMCStaminaSystem : EntitySystem
         foreach (var (hit, comp) in toHit)
         {
             DoStaminaDamage(hit, damage / toHit.Count, true);
-            _jitter.DoJitter(hit, ent.Comp.JitterDuration, true, 7, 5);
+            if (_net.IsServer)
+                _jitter.DoJitter(hit, ent.Comp.JitterDuration, true, 7, 5);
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Stamina/RMCStaminaSystem.cs
```diff
@@ -183,6 +190,7 @@ private void OnStaminaOnHit(Entity<RMCStaminaDamageOnHitComponent> ent, ref Mele
         foreach (var (hit, comp) in toHit)
         {
             DoStaminaDamage(hit, damage / toHit.Count, true);
+            _jitter.DoJitter(hit, ent.Comp.JitterDuration, true, 7, 5);
```
DrSmugleaf: This should be in a INetManager.IsServer check so that the client doesn't repeatedly reset the jitter during prediction

Dutch-VanDerLinde:
```suggestion
            if (_net.IsServer)
                _jitter.DoJitter(hit, ent.Comp.JitterDuration, true, 7, 5);
```

Dutch-VanDerLinde:
Done

DrSmugleaf (approved):

PR #8895, opened 2025-12-31
M890 shotgun YML cleanup and fixes

Dutch-VanDerLinde:
## About the PR
- Made the M890 riot shotgun inherit from the default one, less copypaste
- Improved descriptions from CM13
- M890 now starts with slugs and Honor guard M890 now spawns with buckshot, riot one spawns with beanbags

**Changelog**
:cl:
- fix: Fixed the honor guard M890 not starting with buckshot.
- fix: Fixed the M890 tactical shotgun not starting with slugs.
- fix: Fixed the M890 riot shotgun not starting with beanbags


BreakpointDebug1234:
I think honor guard and riot M890s spawning with their respective ammo is good.

However I think the normal M890 should remain unloaded as people may have a different preference and unloading the shotgun is a painful process.

ghost:
Having forgotten to load my shotgun on one occassion I heartily support the change to make the HG version come preloaded.

I would even go so far as to suggest changing the large general pouch in the HG kit for a shotgun shell pouch (MPs spawn with a large general pouch by default so the second one is kind of useless)

PR #8917, opened 2026-01-03
Overwatch Update: Squad objectives, Confirm buttons

chavonadelal:
## About the PR

This pull request updates Overwatch and the supply drop console by adding squad objectives and replacing the regular buttons with the confirmation buttons in important places like supply drops and orbital cannon firing.
## Why / Balance

Parity. Also, protection against accidental clicks, which I'm already tired of.
missclick
### System Core:
SquadSystem manages squad objectives through two types: Primary Objective and Secondary Objective. All objectives are stored in the SquadTeamComponent and synchronized between the server and clients.

### Objective Display:
Objectives are displayed in several places. In the Overwatch Console, accessible to commanders, they can set and edit objectives through an interface with text fields for Primary and Secondary Objectives, as well as Update and Cancel buttons. When changed, objectives are announced to the entire squad. On the Tactical Map, accessible to all squad members, objectives are displayed as read-only, automatically updated when the map is opened, and presented in the format "Primary Objective: Text" and "Secondary Objective: Text." Furthermore, a player can see their squad's objectives in the Squad Info pop-up window when viewing squad information. There is a cooldown between updates, and each objective change is accompanied by a radio announcement.

Objectives are also announced when joining a round and when transferring from one squad to another.

Localization support has also been added.
## Media

https://github.com/user-attachments/assets/83539cd6-4fb2-4971-905c-c44c226efb16
## Requirements

**Changelog**

:cl: to4no_fix
- add: Squads now have a Primary and Secondary objective, which can be changed through the squad's Overwatch console. Objectives are announced in chat when updating or canceling and are displayed in the squad tracker menu and tactical map. They are also announced when joining a round (if set) and when transferring from one squad to another.
- tweak: The regular supply drop and orbital cannon firing buttons in the Overwatch console and Supply drop console  have been replaced with confirmation buttons that require a confirmation click to prevent accidental pressing.

### DIFF
--- a/Content.Client/_RMC14/TacticalMap/TacticalMapWrapper.xaml
+++ b/Content.Client/_RMC14/TacticalMap/TacticalMapWrapper.xaml
@@ -4,13 +4,14 @@
     xmlns:ui="clr-namespace:Content.Client._RMC14.UserInterface">
 
     <PanelContainer StyleClasses="BorderedWindowPanel">
-        <TabContainer Name="Tabs" Access="Public">
-            <BoxContainer Name="MapTab" Access="Public" Orientation="Vertical" HorizontalExpand="True" VerticalExpand="True">
-                <!-- squad objectives -->
-                <PanelContainer Name="SquadObjectivesContainer" Access="Public" StyleClasses="PanelBackgroundBaseDark" Margin="2" Visible="False">
-                    <BoxContainer Name="SquadObjectivesBox" Access="Public" Orientation="Vertical" Margin="3" />
-                </PanelContainer>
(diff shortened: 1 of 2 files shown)


BadaBoomie:
There should be something that makes it more obvious whether something was *updated* or *cancelled*, at a glance they both look very similar and could possibly be misinterpreted. I'd recommend adding a separate text color to update and cancelled, maybe green for update, red for cancel.

DrSmugleaf:
Changing the button color might be a bit painful since you would need to account for the draw mode so this is fine for now

PR #8929, opened 2026-01-04
Allow department heads to see mentor new player icons and make seeing these icons a toggle in the settings

Dutch-VanDerLinde:
## About the PR
It's good

Also made the new player system use the SpriteSystem for messing with layers instead of sprite component
No more legacy stuff

Given to:
CO, XO, CMP, CMO, CE, LO, and ASO

I want to know if Wardens should get this or not

## Media
image

https://github.com/user-attachments/assets/2ec3dc72-2737-4007-adee-676701bb0ec4

**Changelog**
:cl:
- tweak: Department heads can now see new player icons like Senior enlisted advisors. These icons can be toggled off, check your settings.

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Jobs/Marines/squad_leader.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Jobs/Marines/squad_leader.yml
@@ -75,6 +75,7 @@
         state: leader
     - type: RMCTrackable
     - type: MutinyEligible
+    - type: SeeNewPlayers
 
 - type: startingGear
   id: CMGearSquadLeader


PursuitInAshes:
I am mildly worried about making sure department heads/the people getting this know what each color represents and why we have those the way we do. It may be good to add a guidebook entry along with this or something (I do not know what could be better, and I admit a guidebook entry isn't great for this) to break down what they do.
Explanation is already written and on the wiki, but thats not exactly front and center for someone's first time as a dept head.

QuietlyWhisper:
Guidebook entry would be good, not necessary for the scope of this PR

Requested feature as per 
https://discord.com/channels/1168210010233376858/1457163637381796073

Ian321:
Might I suggest also giving the comp to SLs (and SOs via overwatch) as they are the first ones to see a potential new player and can either directly help them or delegate.

Dutch-VanDerLinde:
> Might I suggest also giving the comp to SLs (and SOs via overwatch) as they are the first ones to see a potential new player and can either directly help them or delegate.

SOs aren't department heads and do not need it, but SLs maybe since they are sort of the department heads of the marine squad "department"

Ian321:
Why would you want it default disabled for SLs? The first thing I would do is turn it on when I spawn to see who might need assistance. I'd imagine they would disable it themself while the Alamo is bringing em down.

Dutch-VanDerLinde:
> Originally I wanted this for SLs too but for it to default disabled for them, and enabled for the other roles you already did, but that would require two options so I'm not sure if it's worth it

Fix would be adding an action for it but uhh there's a lot of actions

Ian321:
> > Originally I wanted this for SLs too but for it to default disabled for them, and enabled for the other roles you already did, but that would require two options so I'm not sure if it's worth it
> 
> Fix would be adding an action for it but uhh there's a lot of actions

Maybe not as action (top bar) but as status (right bar) like xeno night-vision.

DrSmugleaf (changes requested):
Thought about this again and yeah put it on SLs too, then it's good to merge
No need for a separate option

BreakpointDebug1234:
I am late to the party but wardens should get this. Especially on low pop or if the CMP is busy wardens have to train new MPs and sometimes new privates as well.

PR #8930, opened 2026-01-04
Add hyposprays as a pointbuy for corpsmen

YesNoMaybePerhaps:
## About the PR
Added hyposprays as a point buy option for corpsmen for 8 points under Medical Utilities.  Comes filled with tricordizine, similiar to the regular medvendors.

## Why / Balance
Hyposprays are an important aspect of almost every corpsman loadout in some fashion, to the point of being near prohibitive if a corpsman can't get their hands on one.  Latejoining corpsmen, or those going through the req line, could sometimes find themselves without any hyposprays because the quick access ones in medical were empty, and couldn't spare any more time on an already long prep time to get one in a more time consuming but reliable way.  Even then, these more reliable ways need medical to be staffed, and one, the field doctor vendor, requires a CMO to access in the first place, and sometimes neither of these things are the case.  Adding hyposprays as a point buy removes a key restraint to loadouts, allowing for much more freedom to loadout selection, making corpsmen more open to those that want to experiment.  

Of course, with hyposprays being one of the better options for chem delivery, it should be expensive, but not prohibitively so.  Originally I planned on making it 10 points, but went with 8 points after discussions with medical players.  8, 16, even 24 points is a not insignificant number, but still allows for other utilities and gear to be bought.  

## Technical details

N/A
## Media
image

## Requirements

**Changelog**

:cl:
- add: Added hyposprays as a point buy for corpsmen.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/medic.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Squad/medic.yml
@@ -95,7 +95,7 @@
       - id: CMStasisBagFolded
         points: 6
       - id: CMHyposprayFilledTricord
-        points: 10
+        points: 8
     - name: Explosives
       entries:
       - id: CMPacketGrenadeHighExplosiveFilled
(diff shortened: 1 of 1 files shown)


Scyarah:
I think you might still be missing a commit here, currently you have the points still set to 10.

YesNoMaybePerhaps:
That I was.... good catch, should be fixed now, I'll also update the media to match.

PursuitInAshes (approved):
Code for this is fine but it probably won't be considered until post parity
Going to label it accordingly and if for whatever reason we come back to it, it will have an approval

PR #8935, opened 2026-01-04
Surv search SOP changes

QuietlyWhisper:
## About the PR

- Also modified EORG first time text
- Survivors that want to bypass searches may not leave the hangar.
- All survivors (non-military) are entitled to IFF and medical care.
- Survivor equipment that is taken should be moved to the nearest armory, not brig.


DrSmugleaf (dismissed):
Need conflicts from other SOP PR fixed

DrSmugleaf (approved):

PR #8940, opened 2026-01-05
Marines can view crew manifest by right clicking on the tacmap alert, squad-colored crew manifest

Dutch-VanDerLinde:
## About the PR
Parity, marines can press a button in the verbs menu to access the crew manifest in CM13
Put it onto the tacmap here because we dont have CM13s UI

The crew manifest now has colored squad backgrounds like in CM13

## Media

https://github.com/user-attachments/assets/0f8e6ba0-76c8-4258-b0a6-f9f70a5c3a82

Backgrounds
image

**Changelog**
:cl:
- tweak: As a marine, you can now right click on your tacmap to bring up a crew manifest of everyone on your ship.
- tweak: The crew manifest now shows the colored background for squad marines.

### DIFF
--- a/Content.Shared/CrewManifest/SharedCrewManifestSystem.cs
+++ b/Content.Shared/CrewManifest/SharedCrewManifestSystem.cs
@@ -56,6 +56,7 @@ public sealed class CrewManifestEntry
     public string JobPrototype { get; }
 
     public string? Squad { get; }
+
     public Color? SquadColor { get; }
 
 


Content.Shared/CrewManifest/SharedCrewManifestSystem.cs
```diff
@@ -55,6 +56,7 @@ public sealed class CrewManifestEntry
     public string JobPrototype { get; }
 
     public string? Squad { get; }
+    public Color? SquadColor { get; }
```
Dutch-VanDerLinde, line 60:
```suggestion

    public Color? SquadColor { get; }
```

DrSmugleaf (changes requested):
Probably want this in a more obvious place later but this is fine for now
Also, this should check if you are a survivor, it's fine to just do the component check for now until we go through and make that bound to something like the headset

Dutch-VanDerLinde:
> Probably want this in a more obvious place later but this is fine for now Also, this should check if you are a survivor, it's fine to just do the component check for now until we go through and make that bound to something like the headset

Colony headsets don't have the tacmap alert appear actually so it should be fine I believe

PR #8949, opened 2026-01-08
improve rate limit message

QuietlyWhisper:
Might need a comment referencing the change to an upstream file

### DIFF
--- a/Resources/Locale/en-US/connection-messages.ftl
+++ b/Resources/Locale/en-US/connection-messages.ftl
@@ -57,6 +57,7 @@ baby-jail-account-reason-overall = Your overall playtime on the server must be y
 
 generic-misconfigured = The server is misconfigured and is not accepting players. Please contact the server owner and try again later.
 
+# RMC14 Change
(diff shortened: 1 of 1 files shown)


Resources/Locale/en-US/connection-messages.ftl
```diff
@@ -57,7 +57,7 @@ baby-jail-account-reason-overall = Your overall playtime on the server must be y
 
 generic-misconfigured = The server is misconfigured and is not accepting players. Please contact the server owner and try again later.
 
-ipintel-server-ratelimited = This server uses a security system with external verification, which has reached its maximum verification limit. Please contact the administration team of the server for assistance and try again later.
+ipintel-server-ratelimited = You are not banned. This game uses external verification, which has reached its maximum verification limit on new connections. Wait a minute or two, and connect again; no appeal required. If this does not work, rejoin another day or make a ticket.
```
PursuitInAshes, line 61:
```suggestion
# RMC14 Change
ipintel-server-ratelimited = You are not banned. This game uses external verification, which has reached its maximum verification limit on new connections. Wait a minute or two, and connect again; no appeal required. If this does not work, rejoin another day or make a ticket.
```
This should work without formatting issues if you decide a comment is needed

arenaconspiracy:
For what it's worth, shouldn't this be pulled upstream anyway?

DrSmugleaf (approved):

PR #8950, opened 2026-01-09
Give the M85 breech loading characteristics and cycling delay.

august-sun:
## About the PR
Replaced M85 loading characteristics with breech loading characteristics and MOU delay.

## Why / Balance
From an accuracy standpoint, the M85 is literally a breech-loaded weapon; as such, it makes sense for it to behave in accordance with one.

From a gameplay perspective, the M85 has had the capability to spam grenades, which in the right hands has been able to deliver on the same level of performance, if not better, than a grenadier specialist. In fact, some grenadiers eschew the use of their primary weapon in favor of the M85 and its ability to quick reload. This brings the M85 more in-line with other restricted weapons, as well as making the grenadier weapon the better-performing weapon for quick target saturation.

As a caveat, this PR is not a commentary on the adequacy of the performance of the grenadier specialist, or their primary weapon, but just as a means to circumvent the potentially overtuned performance of a non-specialist weapon.

## Media

https://github.com/user-attachments/assets/e8f35983-fa23-4ba9-b525-6624efb8e422

## Requirements

:cl:

- tweak: Converted the M85A1 to a breech-loaded weapon, with a breech delay.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m85a1_grenade_launcher.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Launchers/m85a1_grenade_launcher.yml
@@ -5,16 +5,30 @@
   description: A heavy, low-angle, break-action 40mm grenade launcher. Archaic in core design, inferior to more modern semi automatic M92, M95 grenade launchers and M94 impact launcher, but doesn't require a magnetic armature or an advanced expertice to operate, not to mention near flawless reliability, extremely low cost and low weight due to mostly being made out of polymer materials.
   components:
   - type: Sprite
-    sprite: _RMC14/Objects/Weapons/Guns/GrenadeLaunchers/m85a1/m85a1_inhands.rsi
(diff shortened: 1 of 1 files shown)


kleinerstation13:
Can this be turned into giving the M85 a longer wield delay instead to nerf the individual power? The breech loading mechanic is extremely clunky, makes using weapons unenjoyable and mispredicts quite frequently for people with high ping (using the MOU as an EU player is basically impossible, half the time the breech inputs aren't sent if you are trying to quickly reload), making its' difficulty based on which region you live in instead of, you know, utilizing the weapons' strengths mechanically being the main skill check of using the weapon. I don't think it's a mechanic we should be trying to add to more guns.

RangerXVII:
The M85A1 Is waaay too powerful. If used properly it can be used to kill literally any xeno (except queens, crushers, and ravagers) by spamming a stunning HEDP grenade out so quick it is pretty much permanently stunned, and unable do anything but die.

It's the equivalent to making any rifleman a grenadier specialist which is absolutely terrible balancing.

I really hope this gets merged.

PursuitInAshes:
I believe the current balance adjustment made the M85 covers what you were hoping for here.
Thanks for the PR!

PR #8952, opened 2026-01-09
Tweak: Laser designations End if LOS broken

### LINKED ISSUE #8927: Laser designations should end if LOS broken
## Description

In a situation where a door is opened, a laser is placed past a door, then the door is closed, the designation should no longer be placed, as there is no longer a line of sight.

wheelwrightt:
## About the PR
Resolves: #8927 
Laser designator and rangefinder will now have their targeting interrupted/cancelled when line of sight is broken.

## Technical details
Adds a LOS check when targeting and a check during the doafter, if LOS is obstructed by a structure then ActiveLaserDesignatorComponent is removed.

## Media
[fix.webm](https://github.com/user-attachments/assets/22b4301d-6105-4e8a-942e-933febe11351)

Only LD is shown but functionality works for the rangefinder as well

## Requirements

**Changelog**
:cl:
- tweak: Laser designators and rangefinders will now have their targeting interrupted/cancelled when line of sight is broken.


Dutch-VanDerLinde:
Cancellation for the laser should probably only be run every 5 seconds or so, running a LOS check every update is going to cause a lot of lag

Like meson goggles, functionality was disabled until we get a non laggy version

wheelwrightt:
> Cancellation for the laser should probably only be run every 5 seconds or so, running a LOS check every update is going to cause a lot of lag
> 
> Like meson goggles, functionality was disabled until we get a non laggy version

changed update from every frame to every 5 seconds, tested in game and is working

Dutch-VanDerLinde:
Make sure you can still laser doors that are visible but the laser inside of it

Like directly lasing a wall or a door in your LOS

wheelwrightt:
> Make sure you can still laser doors that are visible but the laser inside of it
> 
> Like directly lasing a wall or a door in your LOS

Yep that still works.

[wall.webm](https://github.com/user-attachments/assets/f6498cd6-9383-4865-9c10-8f36e44635e7)

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -346,4 +351,80 @@ private void TryTarget(Entity<RangefinderComponent> rangefinder, EntityUid user,
...
+                }
+
+                // Check line of sight, ignoring the user and rangefinder
+                SharedInteractionSystem.Ignored predicate = (EntityUid uid) => uid == userUid || uid == doAfter.Args.Used;
+                if (!_examine.InRangeUnOccluded(userCoords, targetCoords, float.MaxValue, predicate))
+                {
+                    _doAfter.Cancel(userUid, doAfter.Index, doAfterComp);
+                }
+            }
+        }
```
Vixting:
```suggestion
        var rangefinderQuery = EntityQueryEnumerator<RangefinderComponent>();
        while (rangefinderQuery.MoveNext(out var rangefinderUid, out var rangefinderComp))
        {
            if (rangefinderComp.DoAfter is not { } doAfter)
                continue;

            if (!_doAfter.IsRunning(doAfter.Id))
                continue;

            if (doAfter.Args.Event is not LaserDesignatorDoAfterEvent laserEvent)
                continue;

            var coordinates = GetCoordinates(laserEvent.Coordinates);
            if (!coordinates.IsValid(EntityManager))
                continue;

            var user = doAfter.Args.User;
            var userCoords = _transform.GetMapCoordinates(user);
            var targetCoords = _transform.ToMapCoordinates(coordinates);

            if (userCoords.MapId != targetCoords.MapId ||
                !_examine.InRangeUnOccluded(userCoords, targetCoords, rangefinderComp.Range, uid => uid == user || uid == rangefinderUid))
            {
                _doAfter.Cancel(doAfter.Id);
            }
        }
```

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -346,4 +351,80 @@ private void TryTarget(Entity<RangefinderComponent> rangefinder, EntityUid user,
...
+
+        var query = EntityQueryEnumerator<ActiveLaserDesignatorComponent, RangefinderComponent>();
+        while (query.MoveNext(out var rangefinderUid, out var active, out var rangefinder))
+        {
+            // Skip if no target
+            if (active.Target == null || !Exists(active.Target.Value))
+                continue;
+
+            // Get origin and target coordinates
+            var originCoords = _transform.ToMapCoordinates(active.Origin);
```
Vixting:
```suggestion
            var originCoords = _transform.GetMapCoordinates(rangefinderUid);
```

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -346,4 +351,80 @@ private void TryTarget(Entity<RangefinderComponent> rangefinder, EntityUid user,
...
+            // Check if on same map
+            if (originCoords.MapId != targetCoords.MapId)
+            {
+                RemCompDeferred<ActiveLaserDesignatorComponent>(rangefinderUid);
+                continue;
+            }
+
+            // Check line of sight, ignoring the rangefinder and target entities
+            SharedInteractionSystem.Ignored predicate = (EntityUid uid) => uid == rangefinderUid || uid == active.Target.Value;
+            if (!_examine.InRangeUnOccluded(originCoords, targetCoords, rangefinder.Range, predicate))
```
Vixting:
```suggestion
            if (!_examine.InRangeUnOccluded(originCoords, targetCoords, rangefinder.Range, uid => uid == rangefinderUid || uid == active.Target.Value))
```

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -39,6 +41,9 @@ public sealed class RangefinderSystem : EntitySystem
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
 
+    private static readonly TimeSpan LosCheckInterval = TimeSpan.FromSeconds(5);
+    private TimeSpan _nextLosCheck = TimeSpan.Zero;
```
Vixting:
```suggestion
```
can be removed after other changes

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -28,6 +29,7 @@ public sealed class RangefinderSystem : EntitySystem
     [Dependency] private readonly SharedDoAfterSystem _doAfter = default!;
     [Dependency] private readonly SharedDropshipWeaponSystem _dropshipWeapon = default!;
     [Dependency] private readonly ExamineSystemShared _examine = default!;
+    [Dependency] private readonly IGameTiming _timing = default!;
```
Vixting:
```suggestion
```

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -16,6 +16,7 @@
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Map;
 using Robust.Shared.Network;
+using Robust.Shared.Timing;
```
Vixting:
```suggestion
```

Vixting (changes requested):

MidZik:
~~Would it be possible to make each do-after or active laser have its own LOS check timer? If the concern is that this will cause lag, iterating over ALL the lasers on the same frame, even if it's only every 5 seconds, is not ideal. Because you will _still_ get lag, it would just be one frame every 5 seconds instead of every frame. Which _is_ an improvement, but far from ideal.~~

~~If each active laser gets its own LOS check timer, the LOS checks will be spread out. Instead of checking 15 lasers on the same frame every 5 seconds, you can reduce the timer to 1 second and only check, on average, one laser every frame, and it's extremely unlikely to have to check more than 5 on a single frame. That would do much better for reducing lag.~~

Edit: I just noticed that the suggested changes above cover this.

Content.Shared/_RMC14/Rangefinder/RangefinderSystem.cs
```diff
@@ -39,6 +39,7 @@ public sealed class RangefinderSystem : EntitySystem
     [Dependency] private readonly SharedUserInterfaceSystem _ui = default!;
     [Dependency] private readonly UseDelaySystem _useDelay = default!;
```
Vixting, line 41:
```suggestion
```
remove blank line

Vixting (changes requested):
Fix the blank line then should be good

Vixting (approved):
LGTM

DrSmugleaf (approved):

DrSmugleaf:
Probably should mark these with an Active component at some point instead of iterating all and checking a field but this is fine for now

PR #8956, opened 2026-01-10
Replace PvI M1984 with a Peregrine Handcannon

PursuitInAshes:
## About the PR
PvIs now spawn with a Peregrine Handcannon instead of a M1984.

## Why / Balance
1984 felt weird given the context of both their position within provost (Below marshals), and above CMP (Who is issued an RSH). This weapon is found already in a few spots depending on map on corpses, like CMB Deputies, so its not unfound in other places. Its still relatively 'exclusive' but not so much I think its unreasonable.

I went and tested and the handcannon with standard ammo (Which this PR would give them) is identical in shots to kill to an M44 with Marksman ammo, which I believe normal MPs can get anyways.

Another option I considered, which I can pursue if its decided thats more preferable, is a resprite for the M1984 with same stats just so they still have something slightly different.

## Technical details
YAML

## Media
N/A - Tested ingame and it works.

## Requirements

**Changelog**
:cl: PursuitinAshes
- tweak: PvIs now spawn with a Peregrine Handcannon instead of a M1984.


noctyrnal:
if you're taking suggestions you could use this as an excuse to port the CO 1911 from cm13 for PvIs + COs

DrSmugleaf (approved):

PR #8964, opened 2026-01-11
Increase the time between All Clear and the End of Round menu appearing by 4 minutes.

BramvanZijp:
## About the PR
After the last Xeno is killed, the time between All Clear and the EOR menu has gone from 3 to 7 minutes. After this, OOC is enabled for the standard 3 minutes and then the round returns to the lobby. Effectively going from 6 to 10 minutes of end of round time for Marine Majors.

## Why / Balance
Because the main times to RP in this game are at the beginning and the end of a round. The beginning is already fairly rushed, as CIC is encouraged to deploy between the 15 and 20 minute mark, while everybody has to get ready. Then at the end of a round, its over fairly quickly, especially when you factor in time to return to the FOB and fly back to the Almayer. 

I hope these 4 extra minutes will allow a bit more closure for rounds. For example, when I play XO and try to do a memorial, the round usually ends right around the time the marines enter the room. But beyond memorials, things like mourning losses with your squad, debriefing, or just cooling down and preparing for the next round are all upsides IMO.

I will also adress the 2 main issues people had when I brought this up a while back:
Q1: Well what about the dead people and xenos?
A1: They can grab a drink, take a bathroom break, or just watch it happen. This is a ROLEPLAY game with rounds about 1-2 hours on average. You can spare 4 minutes.

Q2: Well what if the xenos won? Does everybody just have to sit there and watch them mess around?
A2: No. Read the suggested change. This only impacts the time it takes to go from ALL CLEAR to Round Summary, which only occurs if all xenos are either killed or in vents, etc. while the operation did NOT go to hijack.

## Technical details
Simple change to the component.
I also made it so that the message sent by ARES now actually checks how many minutes it is waiting instead of just assuming the default value of 3.

## Media
image

## Requirements

**Changelog**
:cl: BramvanZijp
- tweak: The time between ARES sounding All Clear and the Round Summary appearing after a major marine victory has been raised from 3 to 7 minutes.


chavonadelal (approved):
This is a good change, I made a big update to the medals, your change will allow CO to spend more time on medals and organize the awards ceremony

GrigoriyGalintano:
This will be very good.

BramvanZijp:
I do not see the merge conflicts?

AndrewFenriz:
> I do not see the merge conflicts?

You don't see a merge conflict because the original CMDistressSignalRuleSystem.cs file was deleted and replaced with new 7 partial classes

BramvanZijp:
PR Updated, tested ingame and does work.

PursuitInAshes:
Posting this before I forget again, I'm worried that this time cause some issues (EORG, which isn't inherently a reason to close this) and would just be used for the same things over and over (Which would result in people just alt tabbing if they are still alive during this time). We'll have to talk about it more as maints

BramvanZijp:
> Posting this before I forget again, I'm worried that this time cause some issues (EORG, which isn't inherently a reason to close this) and would just be used for the same things over and over (Which would result in people just alt tabbing if they are still alive during this time). We'll have to talk about it more as maints

While I get that this might be an issue, I still believe that overall, people will find enough unique things to do. Wether they be debriefs, memorials, winding down with squadmates, visiting the mess hall, etc. And if some people dislike roleplaying in the MRP server, they are more then welcome to alt tab for those 4 minutes, though I suspect most people will find something fun to do.

PR #8965, opened 2026-01-11
Projectiles now ignore machines

Arendian:
## About the PR

Projectiles can now pass through most machines(floodlights, ~~vendors~~, lathes, etc.). 

## Why / Balance

Parity.

## Technical details

Removed `BulletImpassable` from the `MachineLayer`.
Changed some entities `CollisionLayer` to `MachineLayer`.

## Requirements

**Changelog**

:cl: Dygon
- tweak: Projectiles now pass through most machines, except for vending machines.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/base.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/base.yml
@@ -81,3 +81,15 @@
       - !type:DoActsBehavior
         acts: [ "Destruction" ]
   # - type: XenoCrusherChargable TODO knock over by crusher
+  - type: Fixtures #TODO remove this component(making it use the parent's fixtures) after vendor tipping is added.
+    fixtures:
+      fix1:
+        shape:
+          !type:PhysShapeAabb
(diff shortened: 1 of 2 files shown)


QuietlyWhisper:
Video should be attached to show it working

Can you make sure that these objects won't function as makeshift barricades after this change?
- ensure their health is not objectively better than a standard metal barricade
- Ensure they are all destructible by melee damage

Also make sure to list any affected differences in changing these layers if there are any, beyond the bullets

Arendian:
> Can you make sure that these objects won't function as makeshift barricades after this change?
> * ensure their health is not objectively better than a standard metal barricade
> * Ensure they are all destructible by melee damage

It looks like only vendors will be an issue for this reason, I'll look at implementing vendor tipping before making a video. A lazy solution would be keeping vendors block projectile until vendor tipping is added later.

> Also make sure to list any affected differences in changing these layers if there are any, beyond the bullets

I don't think there are any other effects besides projectiles not colliding because of this change.

Arendian:
I re-added vendors blocking projectiles, this can be removed if vendor tipping is added in the future. I didn't find any other machines that could be used as better barricades after this change.

DrSmugleaf (approved):

PR #8968, opened 2026-01-11
Unusable fax machine for mapping

CatAndHats:
## About the PR
added a fax machine that can't fax

## Why / Balance
right now we have to strip the components of faxes manually or make it unpowerable so survs/marines can't fax each other, this adds a pre-made one to save time

## Technical details
yml

## Media
n/a

## Requirements

**Changelog**
no CL


AverageNotDoingAnythingEnjoyer (commented):
Suggest to use `RemoveComponentsComponent` instead, since this allows to parent off the fax entity but just removing the needed components for it to work

CatAndHats:
> Suggest to use `RemoveComponentsComponent` instead, since this allows to parent off the fax entity but just removing the needed components for it to work

changed it to a different solution that a few of the maps use by making it unpowerable.

CatAndHats:
updated to remove components instead

Resources/Prototypes/_RMC14/Entities/Structures/Machines/fax_machine.yml
```diff
@@ -116,3 +116,38 @@
...
+        blacklist:
+          components:
+          - CargoSlip
+    - type: ActivatableUI
+      key: enum.FaxUiKey.Key
+    - type: ActivatableUIRequiresPower
+    - type: UserInterface
+      interfaces:
+        enum.FaxUiKey.Key:
+          type: FaxBoundUi
```
PursuitInAshes, line 149: Unless I'm misremembering, Removecomp just outright removes it, it doesn't care for fields on the comp at all so this list might be able to be shortened?

CatAndHats: without those fields it results in errors since even though they are part of the removecomponent section it still wants the component's fields to be filled.

PursuitInAshes (approved):
lgtm then

PR #8972, opened 2026-01-12
More liaison safe fills, MRE fix

Dutch-VanDerLinde:
## About the PR
Ports bulletproof vest from CM13, which is just an armor vest with more ballistic protection
Fixes the MREs:
- adds back UNMC mre (It was accidentally removed by cro)
- Renames the PMC MRE's ID to WeYaPMC
- Adds a weya MRE for the liaison and corporate security like in CM13

## media
image
image

**Changelog**
:cl:
- fix: Fixed the MRE being given to marines being the WeYa one instead of the UNMC one
- add: Added the bulletproof vest to the liaison's vest


Dutch-VanDerLinde:
Redoing tests

PR #8742, opened 2025-12-15
adds kutjevo mar30s and black market crate

texas1234567889:
## About the PR
3 mar30 and black market crate wooray

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

PR #8992, opened 2026-01-15
Generator tactical map icon

CatAndHats:
## About the PR
generator entity now has a tactical map icon, added new generator entity that doesn't have this.

## Why / Balance
Generators are an important objective, we should highlight where they are.

I made a new generator entity that isn't tracked which can be mapped to reduce the clutter since if all generators have an icon it clutters the tactical map to an insane degree. So for a group of generators only one "normal" generator should be there with all other replaced with the "No Map Icon" variant.

## Technical details
yml

## Media
Screenshot 2026-01-15 151528

Screenshot 2026-01-15 151100

Screenshot 2026-01-15 150704

Screenshot 2026-01-16 085154

Screenshot 2026-01-16 085020

Screenshot 2026-01-16 083656

Screenshot 2026-01-16 083536

Screenshot 2026-01-16 083423

Screenshot 2026-01-16 083310

Screenshot 2026-01-16 083141

## Requirements

**Changelog**
:cl: CatAndHats
- add: Generator locations now have an icon on the tactical map


DrSmugleaf (approved):
This is very good, I wouldn't necessarily bother with separating the tacmap icon so we don't need to remember to do it if/when we add or change generator locations per map
The mapper can remove the components from all but one of the generators in VV, if anything I would add an "untracked" variant and make all current gens tracked by default

DrSmugleaf:
Also your editor seems to be changing every line in the map files when resaving, it shows the entire file as a diff
Check that it isn't accidentally changing the line endings on those maps

CatAndHats:
> Also your editor seems to be changing every line in the map files when resaving, it shows the entire file as a diff Check that it isn't accidentally changing the line endings on those maps

Fixed this, it's an error that can happen sometimes where it wants to replace every ID with the next one higher than it for every entity on the map.

CatAndHats:
> This is very good, I wouldn't necessarily bother with separating the tacmap icon so we don't need to remember to do it if/when we add or change generator locations per map The mapper can remove the components from all but one of the generators in VV, if anything I would add an "untracked" variant and make all current gens tracked by default

Change made, normal generator is tracked, new generator entity that isn't. Mapped them on all 10 maps. Most map changing PRs that were up have been merged now so conflicts shouldn't be that bad, and I don't want to leave any maps out with this version of this idea and leave them with very cluttered tactical maps.

QuietlyWhisper:
Fixed all merge conflicts and added these survivor loot spawns
- #8742

PR #9009, opened 2026-01-17
Fix motion detectors and data detectors turning off when smart-equipped

### LINKED ISSUE #8986: Motion Detectors and Data Detectors turn off when quick stored with a keybind
## Description
Motion Detectors and Data Detectors turn off when quick stored with a keybind. 

**Reproduction**
- bind a button to put an item on your belt or in backpack
- click the button with the MD in your hand
- it turns off

**Additional context**
This is not parity as far as i know and after testing it briefly on cm13

YO0teacher:
## About the PR
Makes sure motion detectors and data detectors stay on after using a keybind to equip them.

## Why / Balance

Fixes #8986 

## Technical details
The issue is with smart-equipping calling TryDrop(), since MDs and data detectors turn off when dropped. But I didn't modify that logic just in case it would break something, instead turning the detectors back on after smart equipping turns them off. This does mean the number on the MD sprite is set back to 0, but it should be fine.

## Media
The video shows detectors only being turned on after smart-equipping if they were previously on.

https://github.com/user-attachments/assets/0456270b-e0c5-4d70-91bd-fd0c7653b679

## Requirements

**Changelog**

:cl:
- fix: Fixed motion detectors and data detectors turning off when equipped using a keybind

### DIFF
--- a/Content.Shared/Interaction/SmartEquipSystem.cs
+++ b/Content.Shared/Interaction/SmartEquipSystem.cs
@@ -1,3 +1,5 @@
+using Content.Shared._RMC14.Intel.Detector;
+using Content.Shared._RMC14.MotionDetector;
 using Content.Shared.ActionBlocker;
 using Content.Shared.Containers.ItemSlots;
 using Content.Shared.Hands.Components;
@@ -28,6 +30,8 @@ public sealed class SmartEquipSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] private readonly ActionBlockerSystem _actionBlocker = default!;
     [Dependency] private readonly EntityWhitelistSystem _whitelistSystem = default!;
+    [Dependency] private readonly MotionDetectorSystem _motionDetectorSystem = default!;
+    [Dependency] private readonly IntelDetectorSystem _intelDetectorSystem = default!;
 
     /// <inheritdoc/>
     public override void Initialize()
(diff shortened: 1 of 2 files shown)


Content.Shared/Interaction/SmartEquipSystem.cs
```diff
@@ -121,6 +125,22 @@ private void HandleSmartEquip(ICommonSession? session, string equipmentSlot)
...
+                return;
+            }
+
+            if (TryComp<IntelDetectorComponent>(handItem.Value, out var intelDetector) && intelDetector.Enabled)
+            {
+                _hands.TryDrop((uid, hands), hands.ActiveHandId!);
+                _inventory.TryEquip(uid, handItem.Value, equipmentSlot, predicted: true, checkDoafter: true);
+                _intelDetectorSystem.Toggle((handItem.Value, intelDetector));
+                return;
+            }
```
DrSmugleaf, line 142: You can move both of these checks below line 145, then only have the Toggle call, since the try drop and try equip are the same

YO0teacher: Well after line 144, because of TryDrop, the detectors will always be off, so the `&& motionDetector.Enabled` check will always be false if moved under line 145. I could remove the `&& motionDetector.Enabled` part, but that would make the detectors always turn on when smart-equipped, even if previously off.

DrSmugleaf: Ah okay I see

DrSmugleaf (approved):

PR #9022, opened 2026-01-19
Change requirements from a checklist to a reminder

QuietlyWhisper:
## About the PR

Makes pull requests more concise by moving the requirements checklist to a comment. Requirements are still in place.

image
image


DrSmugleaf (commented):
Would have to check but having to check those boxes is a more explicit agreement than an implicit one

QuietlyWhisper:
Think we already resolved this

PR #9031, opened 2026-01-20
Adds: Blood Splatter For Humanoids

### LINKED ISSUE #7826: Add marine blood splatters
## Description
Being shot and or stabbed should create the same blood splatters as xenos.
Should probably be limited to attacks that do a certain amount of damage to avoid people being stabbed with a spoon and making a blood splatter

parity

Image

wheelwrightt:
## About the PR
Resolves #7826 . Upon taking sufficient brute damage (12 same as xenos) humanoids will leave behind blood splatter patterns in the same way xenos do. This only triggers on external sources of damage, has a 30% probability to spawn the decal with bonuses for higher damage and will spread in a 0.7 tile radius around. Utilised existing decals in files.

## Why / Balance
parity

## Technical details
Added BloodSlashComponent which handles logic. BloodSplashSystem listens to DamageChangedEvent on entities with BloodSplashComponent to calculate spawn probably based on damage amount and type, spawning blood decals on the damage location with respect to a 3 second cooldown to prevent spam and decal limits.

## Media
[liveleak.webm](https://github.com/user-attachments/assets/dbd5816c-4e46-4881-801e-3952099de72e)

## Requirements

**Changelog**
:cl:
- add: Humanoids will now leave behind blood splatters when sufficient damage is taken


Dutch-VanDerLinde (commented):
Blood splatters are supposed to only happen when a person bleeds or takes woundable damage. I recommend you tie this into the woundable system like CM13 does for soul and cleaner code. I can see you copied some things from xeno acid blood system.

There is:
image

Either a small drip or a big splatter. Located [here](https://github.com/cmss13-devs/cmss13/blob/e5459091831b79c76ce7ec844db1f7eed1632cfa/code/modules/mob/living/blood.dm#L251)

Then, the woundable code, what causes the splatters.
First it checks if it can widen existing wounds. Otherwise, it creates new ones.

image


image

Woundable code is here [here](https://github.com/cmss13-devs/cmss13/blob/e5459091831b79c76ce7ec844db1f7eed1632cfa/code/modules/organs/limbs.dm#L547)

This current implementation is decent but could use some work
Especially the fact that fire direct damage will cause blood splatters. Only brute damage should cause bleeding

There's also supposed to be a cool blood splatter animation when you're wounded but you don't need to add that if you don't want to

wheelwrightt:
Thanks for input as always, will take a look at brushing it up again soon.

wheelwrightt:
don't really intend on working on this any further, gonna close it so others can have at it

PR #9036, opened 2026-01-20
Fix Synth Survival Pouch Whitelist

### LINKED ISSUE #9033: Synth survival pouch can hold plasteel when it shouldn't be able to
## Description
our synth pouches can hold plasteel which is non-parity

**Screenshots**
RMC
Image

CM
Image
Image

PursuitInAshes:
## About the PR
The Synth Survival Pouch should now be correct. The Pouch's metal sheets are now only metal sheets instead of plasteel.. phoron.. etc etc.

## Why / Balance
Fixes #9033 . When I originally PRd the metal I didn't know about the metal sheets ONLY specifically, so this was my fault. Just cleaning up.

## Technical details
Created a new tag and added it to metal sheets, made the synth pouch accept the tag (So it will take all the stack variations of metal)

## Media
N/A

## Requirements

### DIFF
--- a/Content.Shared/_RMC14/Overwatch/SharedOverwatchConsoleSystem.cs
+++ b/Content.Shared/_RMC14/Overwatch/SharedOverwatchConsoleSystem.cs
@@ -11,6 +11,7 @@ using Content.Shared._RMC14.Marines.Squads;
 using Content.Shared._RMC14.OrbitalCannon;
 using Content.Shared._RMC14.Roles;
 using Content.Shared._RMC14.Rules;
+using Content.Shared._RMC14.Scoping;
 using Content.Shared._RMC14.SupplyDrop;
 using Content.Shared._RMC14.TacticalMap;
 using Content.Shared._RMC14.Vendors;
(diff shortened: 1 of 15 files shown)


Resources/Prototypes/_RMC14/Entities/Objects/Materials/Sheets/metal.yml
```diff
@@ -90,6 +90,9 @@
       interfaces:
         enum.RMCConstructionUiKey.Key:
           type: RMCConstructionBui
+    - type: Tag
+      tags:
+      - RMCSheetMetal
```
Dutch-VanDerLinde, line 95: This bulldozes all the other tags, you need to give it the Sheet tag in this too and any other tag the metal parents had before

PursuitInAshes: Ough. Right.
Will fix in next commit

PursuitInAshes: [9b18ab7](https://github.com/RMC-14/RMC-14/pull/9036/commits/9b18ab700043330d54d127ee2cd2111406ce3b8d) should be all of them. I didn't see any further up the parent line

PR #9048, opened 2026-01-22
Fix xeno leaps colliding with boiler fog

shibechef:
## About the PR
Fixes runner/rav/lurker/prae leaps getting stopped by boiler fog

## Technical details
The ApplyLeapingHitEffects part of the code stops a leap if it hits objects from the same hive, but should only be concerned with stunning/damaging mobs instead of hive objects. It now just checks if it hits a mob rather than every individual hive object that isn't a mob

## Requirements

**Changelog**
:cl: shibechef
- fix: Xenonid leaps are no longer stopped by boiler fog

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Leap/XenoLeapSystem.cs
@@ -494,9 +494,7 @@ public sealed class XenoLeapSystem : EntitySystem
         }
 
         if (HasComp<XenoParasiteComponent>(target) ||
-            HasComp<XenoFruitComponent>(target) ||
-            HasComp<XenoEggComponent>(target) ||
-            HasComp<XenoAcidSplatterComponent>(target))
+            !HasComp<MobStateComponent>(target))
         {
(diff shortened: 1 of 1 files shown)


QuietlyWhisper:
Can you add videos of testing here to make sure all the interactions are working?

I would also like to point out that damaging hive structures is an intended mechanic, it's just not used in the current game state

this will affect hostile hives being able to damage hive structures if I'm not mistaken

shibechef:
> Can you add videos of testing here to make sure all the interactions are working?

Yes I tested interactions like ravagers removing window ceils/tables, basic leaps working with enemy hives/marines, and every other collision like with eggs/xeno structures/walls work the same as before. Can add a video soon

> I would also like to point out that damaging hive structures is an intended mechanic, it's just not used in the current game state
> 
> this will affect hostile hives being able to damage hive structures if I'm not mistaken

They didn't collide with it on leaps before and still don't now

shibechef:
and if you tell me exactly what should be collided with for leaping in cm13 relating to enemy hives I can change it to that in this pr but I don't have a cm13 build

PR #9050, opened 2026-01-22
Fix doors becoming ss14 airlocks

QuietlyWhisper:
## About the PR

- Fixes some airlocks not being able to be deconstructed into rmc airlock assemblies, resulting in no metal return and other various exploits
- Double doors still need to get their own construction graphs at some point, we only have single doors.

## Technical details

- Base entity was using ss14 graph, but the other airlocks were updated to use the rmc graph
- All formerly bugged airlock types tested and working
- Removed duplicated YML entry, base entity inherits construction graphs from BaseDoorConstructible

**Changelog**

:cl: Whisper
- fix: All airlocks should now be able to be properly deconstructed into metal.


DrSmugleaf (approved):
Double doors should be a single entity at some point so this is fine

PR #9056, opened 2026-01-23
Fixing flamer's direct hit dealing damage multiple times

RerBreab:
## About the PR
Long story short because of entity overlapping with multiple tiles, one flamer stream could deal up to 4 times more damage with direct hit

## Why / Balance
pixel movement shenanigans

## Technical details
Set a collide chain for flamer chain

## Media
N/A

## Requirements

**Changelog**

:cl:
- fix: Fixed direct hit of the incinerator's stream dealing damage multiple times.


Dutch-VanDerLinde:
Make sure walking onto a new tile fire deals damage as intended

RerBreab:
Ok, it seems like this thing can indeed happen in cm13, yet i think its still should be merged, as damage can go to insane values if it is a napalm-b spray or you are running away

PR #9063, opened 2026-01-24
Allow xenos to see the appearance of their own characters that they bursted out of roundstart

Dutch-VanDerLinde:
## About the PR
This is so the larva can see their own characters
They'll still appear as 'tall host' and such its just the appearance that isnt hidden

## Why/Balance
No balancing effect, you will just be able to see the appearance of roundstart xeno corpse bursts
I believe it's cooler to see your own character when you spawn in as larva and this might prevent possible rulebreaks with xeno corpses, since a player might think that it's a randomly generated corpse when they spawn in, rather then their character profile

**Changelog**
:cl:
- tweak: As a xenonid, you can now see the appearance of your own character that you bursted out of roundstart.


QuietlyWhisper:
This is fine and for good reason, we do definitely get a small count of peoole unaware that the larva does in fact use the character

PR #9068, opened 2026-01-25
Fix SPP Army Reservist ranks and turn their skills into a preset

Dutch-VanDerLinde:
## About the PR
Parity ranks, the current ranks are used by the SPP police, not the military

They rank up to Efreitor at 70 hours (E2) and have Ryadovoy by default (E1)

**Changelog**
:cl:
- fix: Fixed SPP Army Reservist on sorokyne not having the correct ranks. They now have E1 at <70 hours, and E2 at 70+ hours.


noctyrnal (approved):
slowly soro surv becomes more bearable

PR #9088, opened 2026-01-28
ERT Areas!

SharkSnake98:
## About the PR
Adds ERT Areas for CLF, TSE, WEYA, SPP, CMB, Pizza Galaxy, UNMC, & Freelancers, replacing the previous ERT area.

## Why / Balance
Parity. Having distinct, atmospheric areas for each faction that can appear as an ERT makes logical sense. Would allow greater freedom for PVE GM's as well as admins wanting to run events.

## Technical details
YML

## Media
thunderdome-0

## Requirements

**Changelog**
:cl:
- add: Added multiple ERT Bays for the various factions.
- add: Added 2 new unbreakable windows
- remove: Removed the previous ERT area
- tweak: Changed the draw-depth of the coffee machine

### DIFF
--- a/Resources/Maps/_RMC14/thunderdome.yml
+++ b/Resources/Maps/_RMC14/thunderdome.yml
@@ -54060,7 +54060,6 @@ entities:
           occludes: True
           ents:
           - 3370
-          - 3371
           - 3372
         paper_label: !type:ContainerSlot
           showEnts: False


SharkSnake98:
Render's a little fucked up

PursuitInAshes (changes requested):
Do a check through and make sure all the lockers and fridges spawn locked, but you can just remove the access reader comp from them if that is your worry. The molotov in the CLF footlocker (Top right) should probably get removed. The disabler and cuffs in the CMB outpost maybe should be removed just incase, but I highly doubt thats blocking.

General note, we may want later some way for the ERTs themselves to open access to their shuttles, but this is fine for now since we won't see them without admin supervision.

I'm assuming the plan is for shuttles to only be spawned when the ERT is in use, which is fine for now. LGTM besides the above. Really good stuff.

SharkSnake98:
@PursuitInAshes I did map some shuttle spawners, and I have a PR to update the various ones to fit the factions better, they just aren't present in the render.

I didn't make the vessels accessible from the ERT main areas due to if Admins wanted to run little admin events on the bases; having someone run off and launch to the vessel would be pretty unfortunate.

AverageNotDoingAnythingEnjoyer:
I think this should be a separate map
The original Thunderdrome area is simple because of it being an admin level on CM
The real ERT maps are in separate map files but having them bundled together is fine

QuietlyWhisper:
It's fine as is but is also missing the actual thunderdome arena, we use both the thunderdome arena and ert maps on the same location so we only need to teleport to one place

SharkSnake98:
@QuietlyWhisper the Thunderdome should still be in the map, since I didn’t touch it at all, only everything below the big hanger; maybe it didn’t render because it’s seperate from the ERT area by space?

PursuitInAshes:
Yeah, the render doesn't display the thunderdome because its not connected to the main grid
I checked and its there, and the buttons still work.

PursuitInAshes (approved):
Should be fine

PR #8496, opened 2025-11-29
Add M2C and M56D HMGs and dropship deployable emplacements

Arendian:
## About the PR

- Ported the M56D heavy machine gun from CM13, renamed it to ML66D.
- Ported the M2C from CM13
- Combat technicians and squad leaders can now buy the ML66D heavy machine gun in their vendor.
- The requisitions vendor now contains M2C heavy machine gun cases and ML66D gun cases.
- Requisitions can now order M2C and ML66D gun cases and their ammo.
- The dropship part fabricator can now print the  A/A-32-P Sentry Defense System for 200 points and the MTU-4B Door Gunner Hardpoint System for 50 points. These can be installed in a dropship’s weapon or utility slots and deploy either an omnidirectional sentry gun or a scoped heavy machine-gun nest.
- Fuel pump sentries can now be deployed on red alert or higher, and can be undeployed.

If some stats or functions are not parity please tell me so I can correct it.

## Why / Balance

Parity.
Resolves #1443, Resolves #2111, Resolves #1323, Resolves #1128, Resolves #2112, Resolves #1659, Resolves #1444, Resolves #1322, Resolves #1127

**ML66D:**

The ML66D must be fully assembled before use and has a restricted firing arc (160° in front). It can be rotated while deployed by using a wrench.
After assembly it can either be:

- Disassembled back into components, requiring the full assembly process again.
- Folded up, which allows relocation with only the final assembly step needed after redeploying.

Portable version (24 points at the Combat technician vendor, 0 points at the squad leader vendor. Also a few available at the requisitions vendor, extra ones can be bought for $6000 per gun case):

- Magazine capacity: 700 rounds
- Rate of fire: 3.3 rounds per second
- Damage: 36 per shot, projectile loses 1 damage per tile travelled
- Armor penetration: 50
- Accuracy: 105
- Range: starts losing 10 accuracy per tile after 8 tiles
- Grants +2 tiles vision in the mount's facing direction 
- Requires engineering level 1 to shoot (Granted by the pamphlet in the kit)
- Needs to be placed at least 5 tiles away from any other ~~mount~~ ML66D.

Static version (deployed via dropship door-gun hardpoint):

- Magazine capacity: 1500 rounds
- Rate of fire: 5 rounds per second
- Damage: 36 per shot, projectile loses 1 damage per tile travelled
- Armor penetration: 50
- Accuracy: 105
- Range: starts losing 10 accuracy per tile after 8 tiles
- Grants +5 tiles vision in the mount's facing direction 

[...]

PR #9089, opened 2026-01-28
Dropship Sentry launcher

Arendian:
## About the PR

- Ported the LAG-14 internal sentry launcher from CM13.
- Ported the A/C-49-P Air Deployable Sentry from CM13.
- The dropship part fabricator can now print the LAG-14 internal sentry launcher (200 points), this system can be used to deploy the newly added A/C-49-P Air Deployable Sentry (200 points) at the target location. It can be installed in a dropship’s utility slot.
- Supply drops now have a falling animation.
- LZ sentries now have a point light like all the other sentries.
- The paradrop animation now respects the original sprite offset of the dropped entity.
- The "launch" sound of sending a supply drop now also plays for the person that initiated the drop. This was only an issue for echo drop since an SO wouldn't be in the same room as the launch pad when sending a drop.

After this is merged a `RMCDropshipAttachmentAmmoLaunchableSentry` and `RMCDropshipAttachmentLaunchBay` should be mapped in the hangar of the Almayer.

I'd like to suggest adding the small non-parity change, without any real impact on balance, of the sentry launcher being renamed to "LAG-14 internal ordnance launcher"(or something else more generic than sentry launcher), and giving crates the `RMCOrbitalDeployableComponent`, allowing them to be loaded into the deployer.

## Why / Balance

Parity.

Sentries can't be deployed if there already is a sentry within 4 tiles of the targeted location.

The drop pod is supposed to fall for 2 seconds and open 3 seconds after landing. To keep the 3 second falling sound aligned with the drop duration, I made the drop duration 3 seconds and the open delay 2 seconds.

## Technical details

When the pilot presses the deploy button, the system attempts to paradrop the entity currently inserted in the deployer.
If the deployable has a prototype defined on the RMCOrbitalDeployableComponent, that prototype is spawned and paradropped instead.
Dropped entities will be put in a drop pod by default, this can be changed with a bool on the `RMCOrbitalDeployableComponent`.

## Media

https://github.com/user-attachments/assets/75b66a2a-b7b2-44e1-9954-e436f9c5ff84

Spawn location on the Almayer:
image

## Requirements

**Changelog**

:cl: Dygon
- add: The dropship part fabricator can now print the LAG-14 internal sentry launcher (200 points), this system can be used to deploy the newly added A/C-49-P Air Deployable Sentry (200 points) at the target location. It can be installed in a dropship’s utility slot.
- add: Supply drops now have a falling animation.
- fix: LZ sentries now light up the nearby area.


noctyrnal:
This is already covered in https://github.com/RMC-14/RMC-14/pull/8496

BarteG44:
> This is already covered in #8496

This is the same author. The mentioned PR adds sentries mounted TO the dropship, this one allows dropships to deploy sentries on marks on the frontline

noctyrnal:
> > This is already covered in #8496
> 
> This is the same author. The mentioned PR adds sentries mounted TO the dropship, this one allows dropships to deploy sentries on marks on the frontline

OH

Arendian:
I've added the paradrop animation to supply drops. I also updated the video.

PR #9090, opened 2026-01-29
Gives the grenadier satchel an ID lock, creates the Grenade Storage skill

august-sun:
## About the PR
Gives the grenadier satchel an ID lock, created the Grenade Storage skill, assigned said skill to select marine personnel.

## Why / Balance
QoL, an issue Grenadiers can face when needing resupply is their grenades being stolen by other marines. This allows the Grenadiers to be the ones specifically to receive the grenades, with certain select requisitions and command-focused roles also allowed to open the satchels.

## Requirements

**Changelog**
:cl:

- tweak: Grenadier satchels now have an ID lock, accessible by the Grenadier and other select roles.


Oguz0k220:
@august-sun what about a rotten specialist in a xenomorph hive?

august-sun:
> @august-sun what about a rotten specialist in a xenomorph hive?

You'd have to get them back to another member of the marines with the skill to open it. Functionally, it's actually more forgiving than other specialists, as if any other WS dies, their remaining ammo is useless.

Pumkin69:
Does this affect all grenade satchels, and if so this essentially removes the backpack slot grenadier builds from all other roles. I know most people use the rig but i know a few who use the satchel from req and i don't think removing that is necessarily a good idea.

PursuitInAshes:
Thanks for the PR!
This was added somewhere(?) else as an ID lock with a bypass for grenadiers, which is what I would have asked to be changed here instead of a skill

august-sun:
> Thanks for the PR! This was added somewhere(?) else as an ID lock with a bypass for grenadiers, which is what I would have asked to be changed here instead of a skill

Oh good! Glad this got done right, I know I was talking with folks (I think you included) on a better way to implement it

PR #9106, opened 2026-01-31
Slightly increase the size of the fire area created by the AGR-59-I

Arendian:
## About the PR

Slightly increased the size of the fire area created by the AGR-59-I(incendiary mini-mike).

## Why / Balance

Parity.
Instead of a 3x3 square it's now a 3x3 square with an extra tile to the North, East, South and West.

## Media

New fire area shape:
image

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed the AGR-59-I creating a smaller fire area than intended, it now extends one extra tile in each cardinal direction.


DrSmugleaf (approved):

DrSmugleaf:
proper diamond technology if we need it one day

PR #9109, opened 2026-02-01
Barricades now block grab attempts and devouring requires grabbing the target

Arendian:
## About the PR

- Grab attempts are now blocked by barricades between the grabber and the target at the same rate as melee attacks.
- Xenonids must now first grab a marine before they can begin devouring them.
- Changed a few popups related to pulling and devouring to not show the targets actual identity.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed grab attempts not being blocked by barricades between a grabber and its target at the same rate as melee attacks.
- fix: Fixed xenonids being able to start devouring a target without grabbing them first.


Content.Shared/_RMC14/Xenonids/Devour/XenoDevourSystem.cs
```diff
@@ -565,6 +565,9 @@ private bool CanDevour(EntityUid xeno, EntityUid victim, [NotNullWhen(true)] out
             }
         }
 
+        if (!_rmcPulling.IsPulling(xeno, victim))
```
Dutch-VanDerLinde, line 569: Put a popup i think

Arendian: I added the following popup when devouring fails if you're not grabbing the target: We need to get a better grip on {$target}, grab them!

Arendian:
> I originally made it not require a grab so its more ux friendly but now that we have cade blocking we can't really do that

A solution could be making devour attempts also work with the melee delay like grabbing does. We can then leave out the grabbing part from the devouring and do the same thing I did on a grab attempt(simulated attack that turns into a real attack and cancels the grab if intercepted) on a devour attempt.

PR #9019, opened 2026-01-18
Wall / Acid stuff

Vixting:
## About the PR

Acid on guns can now be extinguished, guns have a second wind. Acid now blocks picking up items & splitting / merging stacks of items. Breaking / aciding a wall will now generate a acid hole at 10% health which can be traversed by small xenos. Xenos with sharp enough claws start a doafter to destroy the wall after acid hole created. Acid hole can be repaired with nailgun & sheets. Walls can be repaired with welders (use alt ineract, default alt + left click) & nailgun with material (wood, metal, plasteel, with plasteel repairing the most). When wall is broken it will become girder, girder now turns into damaged girder which can be repaired with welder.

Exact repair values / time are not correct yet.
## Why / Balance

parity

## Technical details

tbd
## Media

https://files.catbox.moe/5gdurx.mp4
## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl: Vixting
- add: Acid on guns can be removed with extinguishers provided the gun still has its second wind.
- add: Guns now have a second wind effect, requiring a 2nd acid to be destroyed.
- add: Acid on items now prevent picking it.
- add: Acid on stackable item prevents splitting or merging.
- add: Walls can now be weld repaired using alt interact
- add: Walls now show integrity in examine text
- add: Xenos now make acid holes when breaking a wall through claw or acid, can be traversed by small xenos. To remove wall, hit or click on acid hole & you will begin destroying it provided you have strong enough claws. Acid holes can be repaired with sheets & nailgun.
- add: Walls can now be repaired with nailgun & plasteel / metal / wood
- add: Girders now have a damaged state that can be repaired with welder.

PR #9136, opened 2026-02-04
Throwing flares and grenades down ladders and through acid holes

Arendian:
## About the PR

Interacting with a ladder or acid hole while holding a flare or grenade now moves the item to the linked ladder's /acid hole exit location after a 1 second do-after. Flares will be lit and grenades will be primed when the do-after finishes.

## Why / Balance

Parity.

## Media

https://github.com/user-attachments/assets/697fc816-3d94-43e1-b3e8-652c9a675834

## Requirements

**Changelog**

:cl: Dygon
- tweak: Interacting with a ladder or acid hole while holding a flare or grenade now moves the item to the other side after a 1-second do-after. Flares are lit and grenades are primed when the do-after finishes.


BarteG44:
this'll be needed for #9019 too i believe

Arendian:
You can now also throw flares and grenades through acid holes.

Content.Server/Light/EntitySystems/RMCExpendableLightSystem.cs
```diff
@@ -40,5 +42,12 @@ private void OnGrenadeContentThrown(Entity<ExpendableLightComponent> ent, ref Gr
                 _physics.SetBodyType(ent, BodyType.Static);
             _light.TryActivate((ent.Owner,ent.Comp));
         }
+
+        private void OnMovedByLadder(Entity<ExpendableLightComponent> ent, ref MovedThroughPassageEvent args)
+        {
+            if(!ent.Comp.PickupWhileOn)
+                _physics.SetBodyType(ent, BodyType.Static);
+            _light.TryActivate((ent.Owner,ent.Comp));
+        }
```
SabreML, line 51: Since this and `OnGrenadeContentThrown()` do the same thing, it'd probably be good to move it into a new single method which gets called by both events.
(I can't do this as a review suggestion so here's a full diff)
```diff
-        /// <summary>
-        ///     Turns on the light and makes it's body type static if enabled in the component.
-        /// </summary>
         private void OnGrenadeContentThrown(Entity<ExpendableLightComponent> ent, ref GrenadeContentThrownEvent args)
         {
-            if(!ent.Comp.PickupWhileOn)
-                _physics.SetBodyType(ent, BodyType.Static);
-            _light.TryActivate((ent.Owner,ent.Comp));
+            ActivateLight(ent);
         }
 
         private void OnMovedByLadder(Entity<ExpendableLightComponent> ent, ref MovedThroughPassageEvent args)
         {
-            if(!ent.Comp.PickupWhileOn)
-                _physics.SetBodyType(ent, BodyType.Static);
-            _light.TryActivate((ent.Owner,ent.Comp));
+            ActivateLight(ent);
         }

+        /// <summary>
+        /// Turns on the light and makes it's body type static if enabled in the component.
+        /// </summary>
+        private void ActivateLight(Entity<ExpendableLightComponent> ent)
+        {
+            if (!ent.Comp.PickupWhileOn)
+                _physics.SetBodyType(ent, BodyType.Static);
+            _light.TryActivate(ent);
+        }
```

Content.Shared/_RMC14/Ladder/LadderComponent.cs
```diff
@@ -30,4 +30,13 @@ public sealed partial class LadderComponent : Component
 
     [DataField, AutoNetworkedField]
     public HashSet<EntityUid> Watching = new();
+
+    [DataField, AutoNetworkedField]
+    public LadderDirection Direction;
+}
+
+public enum LadderDirection
```
SabreML, line 38: Since the enum's being sent over the network:
```suggestion
}

[Serializable, NetSerializable]
public enum LadderDirection
```

Content.Shared/_RMC14/Ladder/PassageThrowableComponent.cs
```diff
@@ -0,0 +1,10 @@
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Ladder;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+public sealed partial class PassageThrowableComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public float DoAfterDuration = 1f;
```
SabreML, line 9: `TimeSpan` is preferred for duration fields afaik
```suggestion
    public TimeSpan DoAfterDuration = TimeSpan.FromSeconds(1);
```

Content.Shared/_RMC14/Ladder/SharedLadderSystem.cs
```diff
@@ -315,6 +387,24 @@ protected bool CanWatchPopup(Entity<LadderComponent> ladder, EntityUid user)
...
+    {
+        if (other is not { } target || TerminatingOrDeleted(target))
+            return false;
+
+        var coordinates = _transform.GetMapCoordinates(target);
+        if (coordinates.MapId == MapId.Nullspace)
+            return false;
+
+        _transform.SetMapCoordinates(moving, coordinates);
+        _popup.PopupPredicted(selfMessage, othersMessage, user, user);
```
SabreML, line 400: If you add a `coordinates` parameter onto this then that would allow it to reuse the variable from `OnLadderDoAfter()` rather than having to fetch it again.
The value being passed in for `EntityUid? other` is also already null checked on L211, so that can be `EntityUid` in this.
Also also `string?` for the message params in order to allow it to be used directly with the messages from `GetThrowablePassageExitEvent` in `OnPassageThrowDoAfter()` below.

```suggestion
    private bool TryMoveEntity(EntityUid target, EntityUid moving, EntityUid user, MapCoordinates coordinates, string? selfMessage, string? othersMessage)
    {
        if (TerminatingOrDeleted(target) || coordinates.MapId == MapId.Nullspace)
            return false;

        _transform.SetMapCoordinates(moving, coordinates.Value);
        _popup.PopupPredicted(selfMessage, othersMessage, user, user);
```

Content.Shared/_RMC14/Ladder/SharedLadderSystem.cs
```diff
@@ -276,6 +279,75 @@ private void OnLadderDragDropDragged(Entity<LadderComponent> ent, ref DragDropDr
...
+            return;
+
+        args.Handled = true;
+
+        _transform.SetMapCoordinates(ent.Owner, ev.Coordinates);
+        _popup.PopupPredicted(ev.SelfMessage, ev.OthersMessage, args.User, args.User);
+
+        var movedEv = new MovedThroughPassageEvent(args.User);
+        RaiseLocalEvent(ent.Owner, ref movedEv);
+    }
```
SabreML, line 322: `TryMoveEntity()` can be called here too to reduce some duplicate logic:

```suggestion
        var ev = new GetThrowablePassageExitEvent(args.User, ent.Owner);
        RaiseLocalEvent(target, ref ev);
        if (!ev.Handled || ev.Coordinates.MapId == MapId.Nullspace)
            return;

        args.Handled = true;
        TryMoveEntity(target, ent.Owner, args.User, ev.Coordinates, ev.SelfMessage, ev.OthersMessage);
    }
```

Content.Shared/_RMC14/Ladder/SharedLadderSystem.cs
```diff
@@ -276,6 +279,75 @@ private void OnLadderDragDropDragged(Entity<LadderComponent> ent, ref DragDropDr
...
+
+        var doAfter = new DoAfterArgs(EntityManager, args.User, throwable.DoAfterDuration, new PassageThrowDoAfterEvent(), args.Used, target: ent)
+        {
+            BreakOnMove = true,
+            NeedHand = true,
+        };
+
+        var directionText = GetDirectionPopup(ent);
+        var selfMessage = Loc.GetString("rmc-ladder-finish-throw-start-self", ("thrown", args.Used), ("direction", directionText), ("ladder", ent));
+        var othersMessage = Loc.GetString("rmc-ladder-finish-throw-start-others", ("user", args.User), ("thrown", args.Used), ("direction", directionText), ("ladder", ent));
```
SabreML, line 295: "rmc-ladder-start-throw-" rather than "rmc-ladder-finish-throw-start"?

```suggestion
        var selfMessage = Loc.GetString("rmc-ladder-start-throw-self", ("thrown", args.Used), ("direction", directionText), ("ladder", ent));
        var othersMessage = Loc.GetString("rmc-ladder-start-throw-others", ("user", args.User), ("thrown", args.Used), ("direction", directionText), ("ladder", ent));
```

Content.Shared/_RMC14/Xenonids/Acid/XenoAcidHoleSystem.cs
```diff
@@ -393,6 +411,18 @@ private void OnHoleBreakDoAfter(Entity<XenoAcidHoleComponent> hole, ref XenoAcid
         ReplaceWallWithDamagedGirder(wall);
     }
 
+    private void OnHoleGetThrowableExit(Entity<XenoAcidHoleComponent> hole, ref GetThrowablePassageExitEvent args)
+    {
+        if (!TryGetCrawlExit(args.User, hole, out var exit))
+            return;
```
SabreML, line 417: Seems to be a problem with this bit spamming the failure messages.

https://github.com/user-attachments/assets/ce442048-ccce-4f96-842e-e1eb4993c694

I don't know if this is the *correct* fix (still learning how prediction works) but this seems to solve it:
```suggestion
    private void OnHoleGetThrowableExit(Entity<XenoAcidHoleComponent> hole, ref GetThrowablePassageExitEvent args)
    {
        if (_net.IsClient || !TryGetCrawlExit(args.User, hole, out var exit))
            return;
```

Content.Shared/_RMC14/Ladder/SharedLadderSystem.cs
SabreML, line 227: following the above `TryMoveEntity()` `coordinates` param thing. Github won't let me put this as an actual "suggestion" so here's just a diff instead:
```diff
-        var coordinates = _transform.GetMapCoordinates(other);
-        if (coordinates.MapId == MapId.Nullspace)
-            return;
-
-        var selfMessage = Loc.GetString("rmc-ladder-finish-climbing-self");
-        var othersMessage = Loc.GetString("rmc-ladder-finish-climbing-others", ("user", user));
-        TryMoveEntity(other, user, user, selfMessage, othersMessage);
-
        ent.Comp.LastDoAfterEnt = null;
        ent.Comp.LastDoAfterId = null;
        Dirty(ent);

-        _rmcTeleporter.HandlePulling(user, coordinates);
+        var coordinates = _transform.GetMapCoordinates(other);
+        var selfMessage = Loc.GetString("rmc-ladder-finish-climbing-self");
+        var othersMessage = Loc.GetString("rmc-ladder-finish-climbing-others", ("user", user));
+        if (TryMoveEntity(other, user, user, coordinates, selfMessage, othersMessage))
+            _rmcTeleporter.HandlePulling(user, coordinates);
    }
```
(Nullspace check removed to use the one in `TryMoveEntity()` instead, `HandlePulling()` only called if it moved successfully, and everything moved to *after* the comp cleanup stuff since that should happen regardless)

Content.Shared/_RMC14/Ladder/SharedLadderSystem.cs
SabreML, line 400: There's an (existing, not caused by the PR) bug with the "finish" messages never appearing when you're climbing through or throwing something through a ladder.

From what I can tell it seems that if the destination ladder is too far away from the player, the coordinates will always point to nullspace and the player won't get their variant of the message. I'm *guessing* this is something PVS related since it's client-side only? (Other players can see the server-side message)

This isn't a problem for acid holes *unless* you add the suggested `_net.IsClient` fix above for the acid hole spam, since that causes the same issue of the client getting kicked out early. Since this is a pre-existing bug though I think that's still worth doing.

Anyway I wasn't able to figure out how to fix this one but I thought it would be good to mention here anyway.

Resources/Locale/en-US/_RMC14/ladder/rmc-ladder.ftl
```diff
@@ -5,3 +5,9 @@ rmc-ladder-finish-climbing-self = You climb up the ladder.
 rmc-ladder-finish-climbing-others = {$user} climbs up the ladder.
 rmc-ladder-someone-else-climbing = Someone else is currently using the ladder.
 rmc-ladder-look-through = Look Through
+rmc-ladder-finish-throw-start-self = You take position to throw the {$thrown} {$direction} the {$ladder}.
+rmc-ladder-finish-throw-start-others = {$user} takes position to throw the {$thrown} {$direction} the {$ladder}.
+rmc-ladder-finish-throw-self = You throw the {$thrown} {$direction} the {$ladder}.
+rmc-ladder-finish-throw-others = {$user} throws the {$thrown} {$direction} the {$ladder}.
```
SabreML, line 11: [`THE()`](https://github.com/space-wizards/RobustToolbox/blob/1d0b581982dac2840943586282d03d633585e28b/Robust.Shared/Localization/LocalizationManager.Functions.cs#L40) to let it handle proper nouns and stuff automatically.
```suggestion
rmc-ladder-start-throw-self = You take position to throw {THE($thrown)} {$direction} {THE($ladder)}.
rmc-ladder-start-throw-others = {$user} takes position to throw {THE($thrown)} {$direction} {THE($ladder)}.
rmc-ladder-finish-throw-self = You throw {THE($thrown)} {$direction} {THE($ladder)}.
rmc-ladder-finish-throw-others = {$user} throws {THE($thrown)} {$direction} {THE($ladder)}.
```

Resources/Locale/en-US/_RMC14/xeno/xeno-abilities.ftl
```diff
@@ -131,6 +131,10 @@ rmc-acid-hole-no-exit = There's no space to crawl through there.
 rmc-acid-hole-blocked = Something is blocking the other side.
 rmc-acid-hole-repair-blocked = The wall is breached and must be patched first.
 rmc-acid-hole-repair-requires-nailgun = You need a nailgun to fix this breach.
+rmc-acid-hole-throw-start-self = You take position to throw the {$thrown} through the {$hole}.
+rmc-acid-hole-throw-start-others = {$user} takes position to throw the {$thrown} through the {$hole}.
+rmc-acid-hole-throw-self = You throw the {$thrown} through the {$hole}.
+rmc-acid-hole-throw-others = {$user} throws the {$thrown} through the {$hole}.
```
SabreML, line 137:
```suggestion
rmc-acid-hole-throw-start-self = You take position to throw {THE($thrown)} through {THE($hole)}.
rmc-acid-hole-throw-start-others = {$user} takes position to throw {THE($thrown)} through {THE($hole)}.
rmc-acid-hole-throw-finish-self = You throw {THE($thrown)} through {THE($hole)}.
rmc-acid-hole-throw-finish-others = {$user} throws {THE($thrown)} through {THE($hole)}.
```

Content.Shared/_RMC14/Xenonids/Acid/XenoAcidHoleSystem.cs
```diff
@@ -393,6 +411,18 @@ private void OnHoleBreakDoAfter(Entity<XenoAcidHoleComponent> hole, ref XenoAcid
...
+
+        args.Coordinates = _transform.ToMapCoordinates(exit);
+        args.SelfMessage = Loc.GetString("rmc-acid-hole-throw-self", ("thrown", args.Throwable), ("hole", hole.Owner));
+        args.OthersMessage = Loc.GetString("rmc-acid-hole-throw-others", ("user", args.User), ("thrown", args.Throwable), ("hole", hole.Owner));
```
SabreML, line 421: just for consistency with the ladder ones
```suggestion
        args.SelfMessage = Loc.GetString("rmc-acid-hole-throw-finish-self", ("thrown", args.Throwable), ("hole", hole.Owner));
        args.OthersMessage = Loc.GetString("rmc-acid-hole-throw-finish-others", ("user", args.User), ("thrown", args.Throwable), ("hole", hole.Owner));
```

Resources/Prototypes/_RMC14/Entities/Structures/rmc_ladder.yml
```diff
@@ -19,6 +19,7 @@
   - type: PlacementReplacement
     key: ladders
   - type: Ladder
+    direction: Down
```
SabreML, line 22: Since these have actual directions now it would be nice to also modify the existing "You start climbing up the ladder." loc strings to change depending on the ladder's direction, like the new throwing popups do.

SabreML (commented):
I honestly thought this was already in the game lol. Definitely makes sense to have it.

PR #9139, opened 2026-02-05
Spam protection reasons for awarding

chavonadelal:
## Why / Balance

Unscrupulous players can sometimes circumvent the minimum character limit by spamming, so I thought it would be a good idea not to punish them and not to burden the administrators by excluding the most conscientious players and closing all obvious abuses. If you circumvented these restrictions and spammed, you did so intentionally.
## Technical details

Take into account the most obvious guaranteed abuses by lazy players (queens in particular) to achieve the minimum length of the reason. Now, spamming more than one space,  3 identical characters (for identical digits 5) in a row is not counted, nor is more than 3 consecutive symbols or punctuation marks that are not necessarily identical (comparison by type rather than direct similarity). So that queens don't suddenly fall in love with exclamation marks mixed with parentheses, for example `!!!)))` - this will be considered a single series of 6 identical punctuation, matching by type.
Also ABAB→ pairs: AB, AB → count = 2 → allowed. ABABAB → pairs: AB, AB, AB→ count = 3 → third pair skipped (both characters). Protection against stupid spam with pairs of letters. I don't know of any legitimate reasons why there might be more than two such pairs.

Added clearer feedback for the player by expanding the hint and coloring the counter red when there are problems.
## Media

image

## Requirements

**Changelog**

:cl: to4no_fix
- add:  Added basic spam protection in the reasons for awarding medals and jelly. A prompt before printing the text and coloring the counter red when problems arise will tell you that you are doing something wrong.


chavonadelal:
As far as I understand, this is unfortunately not in demand.

PR #9141, opened 2026-02-05
hold-click melee / fire

Vixting:
## About the PR

ports https://github.com/space-wizards/space-station-14/pull/42596 Can now hold left click to continue melee / fire
## Why / Balance

reduce carpel tunnel generation 
## Media

[hold.webm](https://github.com/user-attachments/assets/c7bd90e8-a1fe-49a5-934b-cb90f34e0244)

## Requirements

**Changelog**

:cl: Vixting
- add: Can now hold down left click to continually melee / fire. Can be disabled in settings. (ported from upstream)


noctyrnal:
I feel like there should be a delay between shots when using this for semi weapons as its just turning semi-auto weapons into full auto without the consideration of firerate balance and spread and whatnot.

Vixting:
> I feel like there should be a delay between shots when using this for semi weapons as its just turning semi-auto weapons into full auto without the consideration of firerate balance and spread and whatnot.

It doesn't change any fire rate you just don't have to spam click anymore.

noctyrnal:
> > I feel like there should be a delay between shots when using this for semi weapons as its just turning semi-auto weapons into full auto without the consideration of firerate balance and spread and whatnot.
> 
> It doesn't change any fire rate you just don't have to spam click anymore.

my github wont let me check out PRs so i cant show this but i feel there is a difference between a person spam clicking left click to fire a gun and the minor delays between a click, registry and firing, and the game automatically firing a round as soon as it possible can while the button is held down, probably isnt visibly noticeable at a glance but i feel that would affect DPS

redboit:
Imo this should be changed to work like hl2's pistol, faster firerate when spamming, lower firerate when holding. (Edit: weapons with low firerates should maybe keep their full firerate on hold)

BramvanZijp:
> Imo this should be changed to work like hl2's pistol, faster firerate when spamming, lower firerate when holding. (Edit: weapons with low firerates should maybe keep their full firerate on hold)

The entire point of the suggestion is to reduce carpel tunnel, so making using it "sub-optimal" for most weapons is a bad idea. IMO, all melees should not be nerfed by holding. As for guns in semi, those SHOULD be worse IF the gun has an auto mode given otherwise you get the benefits of full auto fire without the drawbacks (spread) of full auto.

PursuitInAshes (dismissed):
Code seems okay and I'd prefer this implementation instead of some mechanical incentive to spam click and this being "inferior" because of that
It can be discussed conceptually though next full review 'cycle'

DrSmugleaf (approved):

PR #9152, opened 2026-02-06
CMO spawns with headset

Vixting:
## About the PR

CMO spawns with headset
## Why / Balance

They are like the only command role that doesn't spawn with their headset for some reason.
## Requirements

**Changelog**

:cl: Vixting
- tweak: CMO now spawns with their headset


BramvanZijp:
> You can remove the headset from the Senior Officer vendor too so we don't have extra command net floating around for no reason

For some reason, all other heads, the xo, etc. All have a spare headset in their vendor. So if thats being removed for cmo, it should be universally.

AverageNotDoingAnythingEnjoyer:
> > You can remove the headset from the Senior Officer vendor too so we don't have extra command net floating around for no reason
> 
> For some reason, all other heads, the xo, etc. All have a spare headset in their vendor. So if thats being removed for cmo, it should be universally.

This should be done in a separate PR

noctyrnal (approved):

PR #9154, opened 2026-02-07
Remove the Chance's Claim All Hostile Gauss Turret

PursuitInAshes:
## About the PR
Removes the premapped active turret in admin on chances claim.

## Why / Balance
All this did was kill new survivors who didn't have ahead of time knowledge that this would attack them for some reason. You _could_ slow down to look at it, but as a Survivor you don't have time to sit and go.

This does _not_ remove the turrets from the fio scavengers ship. After discussion it was decided these do have an intended purpose to prevent Queen's from ovi-ing inside of it for free.

## Technical details
Yaml

## Media
N/A

## Requirements

**Changelog**
:cl: PursuitinAshes
- remove: Removed the All Hostile gauss turret in the Chance's Claim admin building.


PursuitInAshes:
Diff is crazy because its been a bit since things got resaved I believe.

CatAndHats:
> Diff is crazy because its been a bit since things got resaved I believe.

When making map changes if you load the map via mapping command and use Ctrl+S to save it directly to the file location instead of the bin folder, it won't have this issue. You would need to revert the change and delete the turrets again for this to work here however.

PursuitInAshes:
It doesn't matter in any way, I Just know it can be daunting at times so I like to give reassurance in this particular instance

noctyrnal (dismissed):
i trust

DrSmugleaf (dismissed):
Need merge conflicts fixed

PR #9175, opened 2026-02-09
Headset default channel fixes

Dutch-VanDerLinde:
## About the PR
Gives various squad headsets their default channel
Since the channels are applied innately from the squad, we need to give the headset itself the default channel

**Changelog**
:cl:
- fix: Fixed squad headsets not having the correct default channel, they should now default to your squad.


Arendian:
You probably need to make sure it doesn't get reset when the channels get updated in `Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs` 
image

Dutch-VanDerLinde:
> You probably need to make sure it doesn't get reset when the channels get updated in `Content.Shared/Radio/EntitySystems/EncryptionKeySystem.cs` image

Done, added the ``RMCStaticDefaultChannelComponent`` which prevents it being changed on headsets with this component

DrSmugleaf (approved):

PR #9176, opened 2026-02-09
Add barrel charger to req and attachment prep

Dutch-VanDerLinde:
## About the PR
From CM13
It provides value to the game like any other attachment for certain playstyles, making single bullets hit harder. DPS stays the same.
It's horrific on shotguns, so that's not an issue (1.05x damage increase for an extra 0.25 seconds of fire delay)
Req can't buy these so they'll be decently rare

image
image

CM had the attachments vendor be 0.75 and req be 1.5 so I set the attachments vendor to 1 and req to 2

**Changelog**
:cl:
- add: Added the barrel charger to the attachments vendor and requisitions. It decreases rate of fire, but allows the weapon to hit harder.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/requisitions.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/requisitions.yml
@@ -600,6 +600,8 @@
     sections:
     - name: Barrel
       entries:
+      - id: RMCAttachmentBarrelCharger
+        amount: 2
       - id: RMCAttachmentExtendedBarrel
         amount: 7
       - id: RMCM5Bayonet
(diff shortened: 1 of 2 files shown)


QuietlyWhisper:
For consideration this has been out of their gameplay loop for a really long time (since 2023, ~3 years ago), and I'm told the net effect on the majority of firearms results in lower dps- it might be worth waiting to see them test it thoroughly since it's clearly not been balanced anytime recent, and we don't generally want to just throw in experimental cm balance changes, even if it's on paper a way to soft nerf marines if they don't realize this

Dutch-VanDerLinde:
It was fully merged but yes we should wait a bit

QuietlyWhisper:
> It was fully merged

Clarified

PursuitInAshes:
Thanks for the PR!
Spoke with @kleinerstation13 , this seems to be just a straight DPS increase on a lot of weapons and still in a semi-weird state on CM (Forcefully makes time between bursts longer, which feels odd, as an example). We'll revisit the charger post parity if no CM PR changes it significantly. I agree it should exist in some form as a niche(r) attachment, but the balancing will have to happen later

PR #9177, opened 2026-02-09
fix: squad assignment now sets the title correctly

Ian321:
## About the PR

I read the wrong thing leading to things like "Intel Alpha Rifleman" instead of just "Intel Rifleman"

## Why / Balance

bugfix

## Technical details

get the real job title

## Media

https://github.com/user-attachments/assets/0c76c530-57ef-479b-b0e8-15d788dea9db

## Requirements

**Changelog**

:cl:
- fix: Squad assignment via the ID computer now correctly updates the job title.


DrSmugleaf (approved):

BadaBoomie:
Will it also fix non-squad roles not having their names changed? Works in the radio but the inspect title remained the same
image

Ian321:
> Will it also fix non-squad roles not having their names changed? Works in the radio but the inspect title remained the same 

Yes. I was spectating the round and found the bug because of Anne.

Should of done more testing :godo:

PR #9195, opened 2026-02-10
Give Military Warden SEA vision

BreakpointDebug1234:
## About the PR

Gives MW the SEA vision
## Why / Balance

On low population rounds/if the CMP is busy the warden is in charge of showing new MPs around, additionally wardens will ,if the time allows, also train new privates or staff allowed. I think the SEA vision will help them in both of this duties.

Additionally seeing if a player is new may also aid wardens in deciding punishment for them (we don't want new players to be harshly punished for something that may have been an accident, see prep room nades)
## Technical details

1 line yml

## Requirements

**Changelog**

:cl:
- add: Added SEA vision for Military Warden

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Jobs/MilitaryPolice/military_warden.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Jobs/MilitaryPolice/military_warden.yml
@@ -55,6 +55,7 @@
         sprite: /Textures/_RMC14/Interface/map_blips.rsi
         state: background_mp
     - type: RMCTrackable
+    - type: SeeNewPlayers
 
 - type: startingGear
   id: CMGearMilitaryWarden


noctyrnal:
Feels kinda like feature creep giving a non-department head whats meant to be a SEA exclusive thing (heads already got it)

kleinerstation13:
Adding this to warden feels a bit unnecessary

PursuitInAshes (changes requested):
I have conceptual concerns with this mostly just in terms of necessity, primarily that for most roles we still have ranks that display experience and those that don't its usually the case they are high(er) expectation anyways (IE: XO, Dept Heads). In the world where you want to be more lenient you probably want to rope in the CMP anyways so they know whats going on and they can confirm if someone is new or not (Though this second point is far smaller compared to the first)

PursuitInAshes:
I gave it a few days to think on it fairly and get some other opinions (Such as from the other PvIs) and I'm going to close this for the above reasoning. Thanks for the PR

PR #9201, opened 2026-02-10
ASO requires pilot time

BadaBoomie:
## About the PR
ASO requires 5 hours as dropship pilot rather than auxiliary

## Why / Balance
The ASO is the head of auxiliary roles, and currently is not required to play DCC, DP, or GP *at all* to unlock the role. (5h IO and 5h Auxiliary time was required. IO encompasses Auxiliary.)
They are occasionally asked to fill in as either DP or GP due to either being dead, arrested, or not present. This means someone who has never even been inside of the cockpit could be asked or required to replace a pilot.

It makes zero sense for a department head to not require time in the roles they oversee. You can't play CMP without playing Warden, you can't play CMO without playing Doctor, nor can you play CE or LO without playing MT or RT. This was changed so ASO requires 5h DP. (DP requires DCC, both of which the ASO oversees)

changed from GP to DP time after reading comments

## Technical details
one line of YML

## Requirements

- [ X ] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [ X ] I have added media to this PR or it does not require an ingame showcase.

- [ X ] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- tweak: ASO now requires time as Dropship Pilot to unlock


AverageNotDoingAnythingEnjoyer:
Uh..https://github.com/RMC-14/RMC-14/pull/8663
Hopefuly this iteration will go through

BadaBoomie:
> Uh..#8663 Hopefuly this iteration will go through

I searched for PRs but didn't see that, my mistake
I do think this is an issue that needs to be addressed though, so whoever reviews it please consider.

Randomposter19:
I dont think having only GP time is a good idea because i dont think you **understand** how much time it would require for ASOs players to play on other roles just to even unlock it. Starting by DCC you require 5 hours on squads and 1 hour at medical, which is okay, you can reach that normally in a few rounds, then you move on to DP, which requires 5 hours of DCC, which is only _two roles per round_ by the way, but theyre not really that hard to get so thats fineish. But then you get to DP to GP... Which requires more 5 hours of DP to unlock it... A single role... And then you add the 5 hours of GP, a HEAVILLY contested single role, just to unlock ASO. If you discount the requirements of DCC unlock and only apply DCC + DP + GP to unlock ASO... Youre forcing players to play AT LEAST 15 hours of **Three different low number roles**. And lets not count the fact that ASO still requires 5 hours of Engineering, IO and Req + 1 hour of command. ASOs players would have to play AT LEAST 31 hours of DIFFERENT SPECIFIC (Some heavilly contested) ROLES to even unlock ASO.

If we go by the idea that the ASO is a head depart then compare the hours of other head depts/important roles
CE = 20 of only engineering
CMO = 10 medical and 5 Doctor
LO = 10 of only requisitions
SO = 5 of SL and 15 of any department human role
XO = 5 MP and 10 Command

The only important dept head or role that requires near as much hours as ASO would be the CMP, with 15 Military Police, hours, 5 warden and 5 command (25) which is a extremely powerful role on its own department and can literally shape the round depending of the situation. It honestly makes no sense to have that high of a requirement of hours to even be able to play ASO, a role which is being chipped down to a replacement at best of other roles when they are missing for a round. I would argue that if you want to place pilot requirements on ASO, then make it DP focused or lower the hours on GP to 1 round at best.

RosySaturniidae:
the aspects of flight crew duties that would be relevant for an ASO taking over as a pilot are things that can be learned as DCC (either ship) and DP, that being the dropship UI and fultons/medevacs. Marines can live without gunship support (as difficult as it might make pushes) but a lack of medevacs is brutal, and the ASO's duties regarding intel don't *happen* unless fultons are being collected and the Alamo keeps cycling.

All of which to say is that you could get away with 5 hours _Dropship_ Pilot rather than Gunship Pilot for most practical purposes an ASO would be flying a dropship.

BramvanZijp:
Isnt there a playtime group that simply covers both pilot roles which would be good for this?

BadaBoomie:
changed to DP

BramvanZijp:
> changed to DP

Why Dp and not any pilot role? (dp or gp)

kleinerstation13:
> > changed to DP
> 
> Why Dp and not any pilot role? (dp or gp)

gp requires dp

BramvanZijp:
> > > changed to DP
> > 
> > 
> > Why Dp and not any pilot role? (dp or gp)
> 
> gp requires dp

No, gp requires any pilot role, which is what I am suggesting here.
This was done because some people had a lot of GP time before GP started requiring DP, to allow them to keep playing GP which they had 5+ hours in, even when they did not have enough pilot time as a DP.

  - !type:TotalJobsTimeRequirement
    group: CMJobsDropshipPilot
    time: 18000 # 5 hours
    
    Notice how it is a GROUP called CMJob**s**DropshipPilot, which contains DP and GP, heres that group from the code:
    - type: entity
      id: CMJobsDropshipPilot
      categories: [ HideSpawnMenu ]
      components:
      - type: JobGroup
        name: role-timer-dropship-pilot-roles
        color: "#C38312"
        jobs:
        - CMPilotDropship
        - CMPilotGunship

CatAndHats:
The ASO need a full rework and not patch fixes, because of that i'm going to be closing this PR, but thank you for putting this up!

Issue #8892, opened 2025-12-31
Entity names needs to be standartized.


Tunguso4ka:
## Description
- [ ] `Bedroll` -> `RMCBedroll`
- [ ] `ArmorHelmetPMC` -> `RMCArmorHelmetPMC`
- [ ] `XenoWeedsSource` -> `RMCXenoWeedsSource`
- [ ] `WeaponRifleMAR40` -> `RMCWeaponRifleMAR40`
- [ ] `WeaponRifleM16` -> `RMCWeaponRifleM16`
- [ ] `WeaponNailgun` -> `RMCWeaponNailgun`
- [ ] `ArmorHelmetM10` -> `RMCArmorHelmetM10`

Neved4:
To-do:
- [ ] https://github.com/RMC-14/RMC-14/pull/9203
- [ ] https://github.com/RMC-14/RMC-14/pull/9204
- [ ] https://github.com/RMC-14/RMC-14/pull/9205
- [ ] https://github.com/RMC-14/RMC-14/pull/9206
- [ ] https://github.com/RMC-14/RMC-14/pull/9207
- [ ] https://github.com/RMC-14/RMC-14/pull/9208
- [ ] https://github.com/RMC-14/RMC-14/pull/9209
- [ ] https://github.com/RMC-14/RMC-14/pull/9210

PR #9205, opened 2026-02-10
Prefix RMC entities (Factions)

Neved4:
Partially addresses #8892

## About the PR

Standardizes faction-related entity IDs and references under `_RMC14`, using `RMC`-prefixed faction identifiers where applicable.

## Why / Balance

Enforce consistent faction naming:

- Keep `CM` prefixes and `RMC` prefixes.
- Add `RMC` to faction entities without either prefix.

## Technical details

- Renamed faction entity IDs and synchronized ID-card/access/requisition references.
- Updated shared/server systems that reference those faction IDs.
- Added migration mappings for renamed faction entities.

## Media

Not required (naming consistency changes only).

## Requirements

**Changelog**
:cl:

- code: Standardized RMC faction entity IDs and updated access/system references.

### DIFF
--- a/Content.Server/_RMC14/Requisitions/RequisitionsSystem.cs
+++ b/Content.Server/_RMC14/Requisitions/RequisitionsSystem.cs
@@ -47,7 +47,7 @@ public sealed partial class RequisitionsSystem : SharedRequisitionsSystem
 
     private static readonly EntProtoId AccountId = "RMCASRSAccount";
     private static readonly EntProtoId PaperRequisitionInvoice = "RMCPaperRequisitionInvoice";
-    private static readonly EntProtoId<IFFFactionComponent> MarineFaction = "FactionMarine";
+    private static readonly EntProtoId<IFFFactionComponent> MarineFaction = "RMCFactionMarine";
 
     private EntityQuery<ChasmComponent> _chasmQuery;
     private EntityQuery<ChasmFallingComponent> _chasmFallingQuery;
--- a/Content.Shared/_RMC14/Marines/Access/IdModificationConsoleComponent.cs
+++ b/Content.Shared/_RMC14/Marines/Access/IdModificationConsoleComponent.cs
@@ -25,7 +25,7 @@ public sealed partial class IdModificationConsoleComponent : Component
     public bool Authenticated;
 
     [DataField] [AutoNetworkedField]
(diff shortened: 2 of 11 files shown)


Dutch-VanDerLinde:
There's no need for this

Neved4:
@Dutch-VanDerLinde thanks for raising it, I appreciate the feedback 🖤

Don't know if you're referring to this set of changes specifically or to all the other PRs you opened, I'll answer in general terms.

Should have explained the motivation more clearly: this started from #8892, but in follow-up discussion (including with Tung) the point was that the same collision risk isn't limited to just the entities listed there, so the scope should be expanded to all custom entities. To that end, been coordinating https://github.com/RMC-14/RMC-14/issues/8892 with her, which was told was looking good, and decided to split it in several ones for easier review.

Now, the concern is mainly long-term safety/maintenance:
- Local entities can share unprefixed ID/names, and overlap with upstream over time
- That can create avoidable conflicts during upstream merges, easier merge risk, harder to detect breakage.
- Prefixes make custom vs upstream prototypes immediately distinguishable, easier to find in tools also.

Using an `RMC` prefix gives us:
1. Guaranteed namespace separation from upstream
2. Safer future upstream syncs (fewer silent conflicts)
3. Easier filtering/search in mapping tools

Therefore, the PR (and all the related ones) is intended as a consistency pass to reduce future merge/conflict risk, not just a cosmetic rename. It is mainly a correctness/maintenance guardrail.

Ultimately, f the project decision is that it's not a good thing architecture-wise or in general, we should define that explicitly and apply it consistently, including reverting all/similar prior RMC-prefix changes and closing the original issue. If the decision is to standardize, we should apply it consistently as well. Getting approval early and a later rejection, mainly looking for a single stable policy at the moment.

If it's specifically for job IDs that should not be prefixed I'd be more than happy to drop these, and would appreciate that being stated clearly.

Will wait for @Tunguso4ka feedback on this.

Have a great day! 🤍

PursuitInAshes:
These entities are never spawnable and in places they are used its in the RMC namespace already so there won't ever be upstream conflicts with them. Dutch is correct that this is probably unneeded. Also, you seem to be adding in new factions without removing the old ones, see `Resources/Prototypes/_RMC14/Entities/iff_factions.yml`.

PR #9244, opened 2026-02-13
More HMG Fixes

Arendian:
## About the PR

~~**Non-Parity changes:**~~
~~Mostly QOL, if unwanted I can remove it .~~
~~- You can now store the ML66D gun component and the ML66D mount component in the heavy gunner belt.~~
~~- You can store the ML66D gun component in the suit slot. I accidentally already made this possible in the initial PR~~( so ~~it's possible on live), if this is not wanted it has to be removed instead.~~

**Parity changes:**
- Weapon mounts can be ran over by a charging Charger Crusher, dealing damage and undeploying it.
- Projectiles now hit a weapon mount when directly aiming at it.
- A destroyed ML66D now has a 50% chance to drop the gun component instead of it being guaranteed.
- You can now craft a new ML66D mount(the tripod) using 10 plasteel rods, making it possible to reuse a destroyed ML66D that left behind it's gun component.
- Adjusted the damage and accuracy falloff of the M2C.
- Fixed a bug where an undeployed HMG would have a hard fixture if dropped inside another fixture(losing it when no longer colliding with that fixture until dropped again inside a fixture).
- Changed a word in the M2C description.
- Slightly reduced the volume of the M2C's firing sound
- You can now only mount a deployed HMG when standing behind it.
~~- Reduced the M2C disassemble delay from 1.5 seconds to 0.5 seconds.(Disassembly uses the same delay as assembly in CM, We can also keep it as it is at 1.5 right now and mark it as non-parity to make future balancing not take this as a parity value)~~ 
- ~~The doafter duration of undeploying/deconstructing a weapon mount is now reduced by the user's engineering skill.~~(This will be added at a later time)
- All knockdowns, for example from being tackled or from abilities like the Burrower's Tremor, Queen's Neurospit and Sentinel's Scatterspit now dismount and knock down someone who is using a weapon mount instead of just stunning them.
- You can no longer deploy or mount an HMG while restricted from firing your weapon because of invisibility.
- The repair doafter on all repairable entities now breaks if the repaired entity moves
- Prevent rotating the M2C if  the tile behind it's new facing direction is occupied by a wall.
- Having a wall right behind you while mounting an HMG no longer makes you unable to be hit.
- Fixed dismounting from an HMG sometimes making you scope and unscope in quick succesion.
## Why / Balance

~~Non-parity change explanation:~~
~~In CM there is no reason to ever completely disassemble the ML66D after you assemble it. The assembled gun has~~ ~~the exact same storage restrictions as the gun component. Because of this the gunner belt is often tossed away in~~ ~~favor of using the belt slot for something else. With this small, mostly QOL, change there is at least a reason to~~ ~~disassemble the ML66D completely. It also gives people an extra reason to use the heavy gunner belt.~~

**M2C stat adjustment explanation:**
I dove a bit deeper into how the accuracy for the M2C was calculated in CM and noticed a slight difference with what we currently have.

M2C changes:
- The M2C now starts losing damage per unit of distance 1 tile earlier(1 instead of 2). This means it deals 3 damage less at all distances  except within 1 tile.
- Instead of losing 5 accuracy per tile after 6 tiles, the projectile instead loses 15 at once and keeps that  remaining 60 accuracy until 10 tiles. Then it starts losing 10 accuracy per tile.

Accuracy comparison:

| Distance Travelled| Old Accuracy | New Accuracy |
|--------------:|--------------------:|--------------------:|
| 1 | 75 | 75 |
| 2 | 75 | 75 |
| 3 | 75 | 75 |
| 4 | 75 | 75 |
| 5 | 75 | 75 |
| 6 | 75 | 75 |
| 7 | 70 | **60** |
| 8 | 65 | **60** |
| 9 | 60 | 60 |
| 10 | 55 | **60** |
| 11 | 50 | 50 |
| 12 | 45 | **40** |

## Media

https://github.com/user-attachments/assets/3018c80a-1baa-4974-80d1-7e7e75c40493

## Requirements

**Changelog**

:cl: Dygon
- fix: Charger crushers can now run over a deployed HMG, undeploying and heavily damaging it.
- fix: Directly aiming at a deployed HMG when shooting a projectile now makes it hit the HMG.
- fix: When an ML66D is destroyed it now has a 50% chance to drop the gun component instead of a 100% chance.
- fix: You can now craft a new ML66D mount component using 10 plasteel rods.
- fix: Slightly adjusted the damage and accuracy falloff rates of the M2C. This reduces the damage of each projectile by 3 at most distances and makes it slightly less accurate at medium-range.
- fix: Fixed a bug where an undeployed HMG would sometimes collide as if it was deployed.
- fix: You can now only mount a deployed HMG when standing behind it.
- fix: Being knocked down now dismounts you from an HMG instead of only stunning you.
- fix: You can no longer deploy or mount an HMG while invisible.
- fix: Trying to repair something now gets cancelled if the repaired entity moves.
- fix: The M2C cannot rotate in a given direction if the tile behind its new facing direction is blocked by a wall.
- fix: Having a wall right behind you while mounting an HMG no longer makes you unable to be hit.
- fix: Fixed dismounting from an HMG sometimes making your vision flicker.

### DIFF
--- a/Content.Shared/_RMC14/Emplacements/SharedWeaponMountSystem.cs
+++ b/Content.Shared/_RMC14/Emplacements/SharedWeaponMountSystem.cs
@@ -252,7 +252,9 @@ public abstract class SharedWeaponMountSystem : EntitySystem
             return;
         }
 
-        var delay = ent.Comp.DisassembleDelay * _skills.GetSkillDelayMultiplier(args.User, ent.Comp.DisassembleSkill);
+        var delay = ent.Comp.DisassembleDelay;
+        if (ent.Comp.DisassembleSkill != null)
+            delay *= _skills.GetSkillDelayMultiplier(args.User, ent.Comp.DisassembleSkill.Value);
 
         if (ent.Comp.MountedEntity != null)
         {
@@ -914,7 +916,9 @@ public abstract class SharedWeaponMountSystem : EntitySystem
 
     private bool TryUndeployMount(Entity<WeaponMountComponent> ent, EntityUid user, EntityUid? used = null)
     {
-        var delay = ent.Comp.DisassembleDelay * _skills.GetSkillDelayMultiplier(user, ent.Comp.DisassembleSkill);
+        var delay = ent.Comp.DisassembleDelay;
+        if (ent.Comp.DisassembleSkill != null)
+            delay *= _skills.GetSkillDelayMultiplier(user, ent.Comp.DisassembleSkill.Value);
 
         var undeployDoAfterArgs = new DoAfterArgs(EntityManager,
             user,
--- a/Content.Shared/_RMC14/Emplacements/WeaponMountComponent.cs
+++ b/Content.Shared/_RMC14/Emplacements/WeaponMountComponent.cs
@@ -156,7 +156,7 @@ public sealed partial class WeaponMountComponent : Component
     ///     The skill that lowers the duration of the disassemble/undeploy doafter.
     /// </summary>
     [DataField]
-    public EntProtoId<SkillDefinitionComponent> DisassembleSkill = "RMCSkillEngineer";
+    public EntProtoId<SkillDefinitionComponent>? DisassembleSkill; //TODO make this RMCSkillEngineer after rebalancing of the deploy/undeploy doafter durations
 
     [DataField, AutoNetworkedField]
     public string? DeployFixture = "mount";


Resources/Prototypes/_RMC14/Entities/Objects/Materials/parts.yml
```diff
@@ -39,6 +39,7 @@
     buildable:
     - RMCGrilleBuild
     - RMCFenceBuild
+    - RMCWeaponMountBuild
```
Dutch-VanDerLinde: These are normal rods and not plasteel rods, unless that's intended

Arendian: Fixed it, plasteel rods inherited from the base one so when I tested it with plasteel rods it worked.

QuietlyWhisper:
I recommend splitting the parity and non parity changes into different PRs, so we can ensure the intended fixes are expedited and not held back by discussion on non parity changes, though I dont think the proposed changes need much discussion it does make the PR easier to approve

Arendian:
> I recommend splitting the parity and non parity changes into different PRs, so we can ensure the intended fixes are expedited and not held back by discussion on non parity changes, though I dont think the proposed changes need much discussion it does make the PR easier to approve

I've removed the belt change for now

DrSmugleaf (dismissed):

QuietlyWhisper:
image
Another reported bug from discord if you don't mind looking at it

Arendian:
> Another reported bug from discord if you don't mind looking at it

This is fixed now. It was never noticed before because most other repairable entities(sensors, barricades, generators) can't move while being repaired.

PR #9262, opened 2026-02-14
Fix projectiles hitting through objects

Arendian:
## About the PR

In certain situations(a specific distance from the target) projectiles would hit the target behind an entity they should've collided with.
Lowering the projectile speed from 62 to 55 was also needed to make it work at a tickrate of 15, and a physics target minimum tickrate of 50 or higher.

## Why / Balance

Fix.

## Technical details

Projectiles are now instantly deleted instead of queued for deletion when they collide with a target.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed projectiles sometimes hitting through objects.
- fix: Projectiles shot by guns now travel 12% slower.

### DIFF
--- a/Content.Shared/Weapons/Ranged/Components/GunComponent.cs
+++ b/Content.Shared/Weapons/Ranged/Components/GunComponent.cs
@@ -213,7 +213,7 @@ public sealed partial class GunComponent : Component
     /// The base value for how fast the projectile moves.
     /// </summary>
     [DataField]
-    public float ProjectileSpeed = 62f;
+    public float ProjectileSpeed = 55f; //RMC14 changed from 62
 
     /// <summary>
     /// How fast the projectile moves.


DrSmugleaf (dismissed):
I'm not sure if this is enough to fix it but we can try
Try spamming a high firerate, low damage M54CE2 at a wall with a defender behind it hugging the wall, that should be the easiest way to test
Ideally set the physics target update rate cvar to 60 or less, and tickrate to 15

Arendian:
> I'm not sure if this is enough to fix it but we can try Try spamming a high firerate, low damage M54CE2 at a wall with a defender behind it hugging the wall, that should be the easiest way to test Ideally set the physics target update rate cvar to 60 or less, and tickrate to 15

Shooting the M54CE2 at a resin wall with a drone behind it is the test I did to see if the fix worked(it did). I'll go and check after settings those cvars

Arendian:
>Ideally set the physics target update rate cvar to 60 or less, and tickrate to 15

With these cvar settings I had to lower the projectile speed from 62 to 55 to still make it work. Projectiles start potentially passing through again when physics target minimum tickrate gets below 50, so the projectile speed might need to be adjusted again if physics tickrate is ever set lower than 50.


https://github.com/user-attachments/assets/9048b6e4-b588-4ca5-94ea-cef1c7458dd2

PR #9298, opened 2026-02-19
Reduce fire tile damage and duration

Arendian:
## About the PR

- Fire damage taken is no longer multiplied by 2 while standing on a fire tile, instead the result of the following calculation is added to the base fire damage:  intensity * tiledamage / 3 . The /3 is because from my testing this damage is dealt every +/- 30 ticks(3 seconds).
- Reduced the duration of UT fire tiles from 60 to 20 seconds and B-Gel fire from 60 to 50 seconds.

## Why / Balance

Parity.

**Damage While Standing on a tile of UT Napthal**

| Entity| Old Damage (per second) | New Damage (per second) |
|--------|--------------------------|--------------------------|
| Drone  | 30                       | 20                       |
| Boiler | 90                       | 50                       |
| Human  | 12                       | 11                       |

## Media

image

image

image

image

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed fire damage taken while standing on a fire tile being too high.
- fix: Fixed the duration of fire tiles created by UT Naphtal(60>20) and B-Gel(60>50) being too high.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
@@ -530,7 +530,7 @@ namespace Content.Server.Atmos.EntitySystems
                     {
                         if (flammable.TileDamage is { } tileDamage)
                         {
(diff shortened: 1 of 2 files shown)


BarteG44:
Underbarrel flames should last shorter too, even shorter than normal naphtal. It says it has burn duration 10 but from my testing it burned for about 14 seconds

PR #9299, opened 2026-02-19
Smart Scope blocks fire on friendlies

Vixting:
## About the PR

Smart Scope blocks fire on friendlies, instead of allowing you to just shoot over them.

## Why / Balance

parity - idk if its wanted on the smart scope, but can probably be used in the future if you want a weapon to work like this as a balancing thing or for pve sgo
## Technical details

Bassically draws a ray in the direction of fire using a deterministic random spread to prevent mispredicts, if friendlies are found within this ray it cancels the fire.

Added debug overlay to show bullet travel. The bullet still ignores friendlies, because I wouldn't know how to make this work with pvs, just prevents you firing when in pvs range. It didn't seem to have issues with mispredicting in my testing, but I'm also no expert with this stuff.

## Requirements

**Changelog**

:cl: Vixting
- fix: The smart scope now blocks firing on friendlies instead of allowing you to just shoot over them.
- add: Smart scope can be attached to MK2.
- tweak: Removed damage decrease modifier from smart scope.

### DIFF
--- a/Content.Client/_RMC14/Attachable/Systems/AttachableIFFDebugOverlaySystem.cs
+++ b/Content.Client/_RMC14/Attachable/Systems/AttachableIFFDebugOverlaySystem.cs
@@ -1,0 +1,177 @@
+using System.Numerics;
+using Content.Shared._RMC14.Attachable.Events;
+using Robust.Client.Graphics;
+using Robust.Shared.Enums;
+using Robust.Shared.Map;
+using Robust.Shared.Timing;
+
+namespace Content.Client._RMC14.Attachable.Systems;
+
+public sealed class AttachableIFFDebugOverlaySystem : EntitySystem
+{
+    [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly IOverlayManager _overlayManager = default!;
+
+    private static readonly TimeSpan SampleLifetime = TimeSpan.FromSeconds(2);
+    private static readonly TimeSpan MispredictCompareWindow = TimeSpan.FromMilliseconds(500);
+    private const float MispredictPositionTolerance = 0.25f;
+
+    private AttachableIFFDebugOverlay? _overlay;
+    private IFFDebugSample? _clientSample;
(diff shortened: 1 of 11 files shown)


BarteG44:
shouldn't this go along with making it possible to put on a mk2?
https://github.com/RMC-14/RMC-14/pull/6935#issuecomment-3040519921

Vixting:
removed dmg decrease, added to mk2

Torma25:
I understand alt IFF is technically parity, however this change, especially in this form would be extremely unhealthy for the game, specifically for newer players who rely on IFF M4's to even feel like they're doing anything besides being a meatbag on the front. Alt IFF gives zero feedback to the player, your gun simply does nothing if you try to fire it. No sound, not even a text popup, it just makes it look like your gun doesn't work. This is made even worse because of the hectic and chaotic nature of RMC. Simply adding the unaltered CM13 alt iff, while fully in line with the parity vision would not be healthy for the game, and I'd recommend at least adding some form of visual/audio (preferably both) feedback  to alt IFF equipped guns to make using them more new player friendly. 
But that's only if we're really adamant on actually adding this to the game. I'm not even sure alt IFF guns could be used, like genuinely at all because of pixel movement. Getting out of the way of your teammates is a hell of a lot easier in CM13, while in RMC accidentally blocking somebody's line of fire is guaranteed as long there's at least one person behind you. Combat in CM13 is more static, with people practically standing in napoleonic firing lines behind barricades, a thing that's basically never happens on RMC.
And of course, removing IFF M4's is a giant blow to the new players of the game. Smartscope M4's are basically the go-to gun to give to the clueless first timers because it lets them actually feel like they're contributing to their side's effort, while making sure they don't accidentally kill half their squad during an engagement.

Pumkin69:
I completely disapprove of this form of IFF. I believe it to be a horribly designed system with little to no consideration for new players or QOL and genuinely firmly stand that this should not be added despite being parity.

Flameingman2:
This would be useful on PVE, not so much on pvp

PR #9318, opened 2026-02-21
[PVE] [Bugfix] fixes armor webbing visual layering

TadJohnson00:
## About the PR

uniform and armor webbing when applied at the same time cause visual bugs if removed. this fixes that
## Why / Balance

fixes bug
## Technical details

creates a new "Outer" WebbingVisualLayer
creates a function, "GetWebbingLayer", which determines the webbing visual layer to use based on a provided item's WebbingClothingComponent, checking its whitelist to output either WebbingVisualLayers.Outer for WebbingClothingComponent that has a defined whitelist containing the "ArmorWebbing" tag OR to WebbingVisualLayers.Base for anything else. it's not elegant but it Just Works:tm:.
uses the aforementioned function to define variables inside of private voids "OnWebbingClothingEquipmentVisuals" and "OnClothingState", replacing explicit mentions of WebbingVisualLayers.Base and instead referencing the variables.

in short, works on my machine
## Media

insert picture of webbings working as intended
## Requirements

:cl:
- fix: [PVE] fixed a visual error where having uniform and armour webbings equipped simultaneously would result in permanent artifacts upon removing one webbing.


TadJohnson00:
empress' blessings upon squash merge

TadJohnson00:
thank fuck it's just a heisentestfail

PR #9130, opened 2026-02-04
Adds the Sleeper medical pod.

Arykh:
## About the PR

Title

## Why / Balance

Parity

## Technical details

:(
Just needs #9327 

## Media

https://youtu.be/zd5Ss_lhnjc

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

- [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Added the sleeper medical pod! Removes chemicals from the occupant and can inject stabilizing chemicals, with more becoming available as injuries worsen.

PR #9327, opened 2026-02-22
Adds RMCPulseSystem for the heart rate monitor/pulse readings.

Arykh:
## About the PR

Title
Atomized from #9130 

## Why / Balance

Parity

## Technical details

Makes random numbers.

At the moment, this is just fluff numbers.
We could make this work with heart damage in the future.

## Media

image
image
image

## Requirements

**Changelog**

:cl:
- add: Added a heart rate monitor to the health analyzer.


Arykh:
Right now, the only states CM uses are Normal, None, and Thready.

We could make this work with heart damage in the future.
Doing this might make the stethoscope obsolete. Perhaps a niche sidegrade in case you lose your scanner? The stethoscope is constantly on you as a uniform accessory.

Seems like a good idea. Stethoscopes are dated, but reliable tools when compared to a futuristic health scanner.

Perhaps have the pulse readings reflect the combined damage between the heart, lungs, and blood level? That way, there is skill involved in discerning the patient's condition using the pulse.

Arykh:
Oh, just needs these two to fix test fails on main: #9327 #9363

PR #9328, opened 2026-02-22
Hive tracker fixed for evolve/devolve

keegaroo65:
## About the PR

The hive tracker used to keep tracking a target when you evolve. Now it also tracks if you devolve, or if your target evolves or devolves.
## Why / Balance

The hive tracker was recently added, but it loses the target easily when either you devolve, or the target evolves/devolves. This makes it keep the target continuously until being reassigned or death.
## Technical details

1. Moved most of OnNewXenoEvolved's behaviour to a new method: KeepTrackingOnEvolveDevolve
2. Now listens for a XenoDevolvedEvent
## Media

image
Here's an image of my testing set-up. Both drones have their tracker set to the defender. The carrier is set to track the queen. When the defender evolves or devolves, the drones continue to track it as it is the same xeno and leader, with the UID being reassigned by the server. When a drone evolves or devolves, it also continues to track the defender. The carrier is left completely unaffected by anyone else's evolution as expected.

## Requirements

**Changelog**

:cl: keegaroo65
- fix: Fixed Hive Tracker to track targets upon your devolution, and the target's evolution or devolution.

### DIFF
--- a/Content.Shared/_RMC14/Tracker/Xeno/HiveTrackerSystem.cs
+++ b/Content.Shared/_RMC14/Tracker/Xeno/HiveTrackerSystem.cs
@@ -34,6 +34,7 @@ public sealed class HiveTrackerSystem : EntitySystem
     {
         // TODO RMC14 resin tracker
         SubscribeLocalEvent<HiveTrackerComponent, NewXenoEvolvedEvent>(OnNewXenoEvolved);
+        SubscribeLocalEvent<HiveTrackerComponent, XenoDevolvedEvent>(OnXenoDevolved);
         SubscribeLocalEvent<HiveTrackerComponent, MapInitEvent>(OnMapInit);
         SubscribeLocalEvent<HiveTrackerComponent, ComponentRemove>(OnRemove);
         SubscribeLocalEvent<HiveTrackerComponent, HiveTrackerClickedAlertEvent>(OnClickedAlert);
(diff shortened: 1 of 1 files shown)


keegaroo65:
I'm still new to contributing but I've just noticed that the query might be a little bad practice, I see some systems declare it once then reuse it in methods? It's my first time using one so I'm not too sure, let me know thanks! :)

Arendian:
Did you check if you were able to reproduce it not working before you made these changes, and that these changes you made fixed it? The reason I ask this is that during my own testing the tracker works fine locally before any changes and I am unable to find out why it breaks on the server.

keegaroo65:
> Did you check if you were able to reproduce it not working before you made these changes, and that these changes you made fixed it? The reason I ask this is that during my own testing the tracker works fine locally before any changes and I am unable to find out why it breaks on the server.

Yes, the tracker resets to the Queen when devolving before my fix, and if your target evolves or devolves it also resets before my fix. I'll double check tonight

keegaroo65:
> Did you check if you were able to reproduce it not working before you made these changes, and that these changes you made fixed it? The reason I ask this is that during my own testing the tracker works fine locally before any changes and I am unable to find out why it breaks on the server.

I just verified that these are still issues on the latest main branch by testing it again locally. I explain how I tested it in the Media section of the first post :)

PR #9360, opened 2026-02-26
Added Shiva's CO Surv: Research Director

SharkSnake98:
## About the PR
Adds the Research Director CO survivor to Shiva's Snowball

## Why / Balance
Doesn't seem fair that only a few maps have COs. Furthermore, additional non-military survivors would be good for roleplay and gameplay.

## Technical details
YML, FTL, RSI/JSON/PNG

## Media
image

## Requirements

**Changelog**
:cl:
- add: Added the Shiva's Research Director CO surv


Arcticular1:
IMO CO survivors should be strictly military related / law enforcement. Or even Corporate security.
They are whitelisted for a leadership position *within* factions such as the SPP and the TSE. - This would rather be better as a non whitelisted role as it fits in with role play expectations similar to roles like (CL) assistant managers / colony supervisors.

We do not get granted specific provisions for roles outside faction commanders / high ranking officials in those factions.  
Colony supervisors and high ranking WEYA personnel are generally reserved for the CL - I personally believe giving CO direct civilian roles such as researchers and etc is out of place for the current environment. 

I also believe not every map should have a commander survivor where it does not fit. (even though i do play commander survivor and want to experience the gameplay more).

Removing the CO whitelist completely for survivors would be found in better place than giving them civilian survivors that heavily deviate from the skillset of their whitelist.

Arcticular1:
Some possible examples i could see where it fits would be like: 
Heads of Security / law enforcement - CMB
Security Advisors - FOL's relationship with the Assistant Manager
Visiting personnel - Paramarine LtCol.

PR #9391, opened 2026-03-02
Sensors no longer break immediately when turned on, and break more frequently

MidZik:
## About the PR
Sensors now have a 15% chance to break every 50 seconds of being on, and won't break the same instant they're turned on.

They will also now show a popup when they automatically break.

## Why / Balance
Bug fixes + parity.

## Technical details

NextBreakTime is updated every time sensor is powered on.
BreakEvery changed to 50 seconds.
Fixed PopupEntity usage.

## Requirements

**Changelog**

:cl:
- fix: Sensor towers will no longer break immediately after being turned on.
- tweak: Sensor towers will break more frequently.


kleinerstation13:
Is this parity?

MidZik:
The sensor tower not breaking immediately upon being repaired is.

I don't play CM13 and I'm not exactly familiar with DM or SS13's codebase, but looking at their code it _seems_ like it has a 15% chance of breaking every **25 minutes** of being continuously turned on. At least if I'm understanding it correctly, this seems like a bug on _their_ end, because there was a refactor that changed it from 15% every ~50 seconds to 15% every ~1500 seconds.

MidZik:
Fixed to behave as it should in CM13.

Also noticed that popups weren't working so I fixed that too.

DrSmugleaf (approved):

PR #9426, opened 2026-03-04
Announcements

Vixting:
## About the PR

Adds a announcement overlay system. Announcements can have set variation presets, each announcement can be toggled between announcement type or disabled entirely in settings.
image
I will finally finish this
## Why / Balance

soulful
## Technical details

This has been been reworked like 4 times & I still don't really like it.
### Marine announce

Stylized version

https://github.com/user-attachments/assets/afae22ee-0ae8-4ff8-80bf-d92c9ea9a7f8

Default version

https://github.com/user-attachments/assets/995ce857-cdb7-40ff-877b-c5263a7407d4

Simple version

image

### OW 

Stylized version

https://github.com/user-attachments/assets/0bace440-5545-4f83-b1c2-88fe2ec93219

Default version

https://github.com/user-attachments/assets/1f64c52f-58a9-4b82-aba0-b77eb948c001

https://github.com/user-attachments/assets/0e127ff5-b1cd-48fa-9a68-8391659374ed

https://github.com/user-attachments/assets/49663fa5-a8dc-4c0e-bc8d-6d93c38b96ee

### Alerts:

https://github.com/user-attachments/assets/97c60a28-4497-4649-8a46-8975cb4ed72e

https://github.com/user-attachments/assets/e1171b83-2a28-4d74-9524-686d83844337

https://github.com/user-attachments/assets/f8b8cbe6-0448-46d1-9b1b-4a6efe2b9294

https://github.com/user-attachments/assets/1874c0d0-8065-4933-8aea-80e1423fb017

https://github.com/user-attachments/assets/4fa46217-ef62-4d28-8722-17e118ab0b83

image

## Requirements

**Changelog**

:cl: Vixting, GrigoriyGalintano
- add: Added announcements overlays to the tablet, overwatch, queen, alert levels. These can be disabled in the options menu.
- add: Added overlay layout editor, you can customize the scale, position, elements & color of announcements
- add: Added system to create announcements
- tweak: OW now has an 8s cooldown on sending announcements (because it has an overlay now)


BramvanZijp:
@Vixting Once again, great PR, though I will bring up a few things: (Not including the changelog as its still a draft)
1. Do these announcements also still get sent to chat, so you can scroll up for them?
2. Would it be possible to include the name of who sends an overwatch message in the display on screen?
3. The simple version should, IMO, also include the name of who sends the announcement, even if without the picture.
4. Including the rank of the sender for all of them would be good, instead of just the name.

Vixting:
> @Vixting Once again, great PR, though I will bring up a few things: (Not including the changelog as its still a draft)
> 
>     1. Do these announcements also still get sent to chat, so you can scroll up for them?
> 
>     2. Would it be possible to include the name of who sends an overwatch message in the display on screen?
> 
>     3. The simple version should, IMO, also include the name of who sends the announcement, even if without the picture.
> 
>     4. Including the rank of the sender for all of them would be good, instead of just the name.

Yes
Yes but its already in chat, I don't want unnecessary stuff on screen
Yes but its already in chat, I don't want unnecessary stuff on screen for simple version
Yes

XxSWAG-MASTER-OVERLORDxX:
The stuff you code is quite frankly insane 👍

Pumkin69:
this is so insanely cool

GitHubUser53123:
doesnt seem to account for separated UI

Vixting:
> doesnt seem to account for separated UI

Like an announcement on the top right?

GitHubUser53123:
image


the text is also weird

Vixting:
> image
> 
> yup. this isn't with separated and the issue still occurs, which is weird.

Ohhh I kind of forgot that existed. Will fix when I get back to this... I probably need to make it so you can set where announcements are like arma 3 UI settings tbh.

GitHubUser53123:
better screenshot of what i'm talking about.
its not uploading. RIP.

Vixting:
I think you commented before it uploaded or github is just dying as it does everyday

GitHubUser53123:
image

Vixting:
This is going to get entirely redesigned so I wouldn't bother with reviewing it currently, hence it being a draft.

GitHubUser53123:
fifth time's the charm

GrigoriyGalintano:
@Vixting  Hi! I found several issues in the implementation.

### 1. CRT scanline performance

In `CRTOverlay.UpdateAnimations()`, the offset is increased by:

```csharp
Settings.ScanlineSpeed * deltaTime * 60f
```

but only one `ScanlineSpacing` is subtracted afterward.

With the default values, `_scanlineOffset` continues to grow every frame. `DrawScanlines()` then starts iterating from:

```csharp
rect.Top - _scanlineOffset
```

which causes the number of off-screen iterations to continuously increase.

This becomes especially noticeable in the layout editor preview, where the same overlay can run indefinitely.

The offset should probably be wrapped using modulo with the effective scanline spacing.

### 2. Speaker PVS overrides do not work correctly with the announcement queue

`PvsOverrideSystem` stores session overrides as a set rather than as reference-counted entries.

If two queued announcements use the same speaker, both calls to `AddSessionOverrides()` effectively result in only one override being stored. When the first announcement finishes, `AnnouncementPlaybackDoneMsg` removes that override even though the second announcement still needs it.

Additionally, the 30-second fallback timer starts when the announcement is dispatched rather than when it begins playing. Because of this, a queued announcement can lose its speaker PVS override before it is displayed.

This likely needs either a per-delivery identifier or an internal reference count for each speaker/session pair.

### 3. `announcepreset` ignores the target configured in the preset

`AnnouncementPresetPrototype.Target` is configured in YAML, but `AnnouncePresetCommand` does not copy it into `request.Route.Target`.

Since `AnnouncementRoute.Target` defaults to `All`, running a command such as:

```text
announcepreset XenoQueen "message"
```

without an explicit `--target` can send the announcement to every player.

The request should use `resolvedPreset.Target` by default and only override it when `--target` is explicitly supplied.

### 4. `TypewriterAnimation` is frame-rate dependent

The animation resets `_timer` to zero and advances by at most one character per update.

This discards accumulated time, and when the frame time is greater than `PrintSpeed`, the animation is limited to one character per frame. At low FPS, long announcements can therefore take significantly longer than configured.

It should use a bounded loop and subtract `PrintSpeed` from the timer, similar to `GlitchAnimation`.

### 5. `AnnouncementPresetPrototype` should not be network-serializable

The prototype is marked with both `Serializable` and `NetSerializable`.

The Robust analyzers report `RA0037` and `RA0038` because prototypes should not be serialized or sent directly over the network.

The implementation already sends the preset `ProtoId`, so these attributes appear unnecessary.

### 6. Overwatch data is now replicated through the entity component

`OverwatchConsoleComponent` stores squads, objectives, marine locations, and cameras in `AutoNetworkedField`s.

This means the complete Overwatch state is replicated to every client that has the console entity in PVS, rather than only to the player who opened the BUI.

This can expose information to unauthorized clients and also causes a relatively large component to be repeatedly sent whenever the console is refreshed.

It would be safer to keep this information in an addressed BUI state or another session-specific message.

GrigoriyGalintano (changes requested):
@Vixting Hi! There is one last bug that I found.

**PVS override is not reference-counted for queued announcements**
Multiple announcements from the same speaker share one PVS override. When the first announcement finishes or its 30-second fallback timer expires, it removes the override even if another announcement from that speaker is still active or queued. This can make the speaker portrait disappear or fail to load.
Please track overrides per speaker + session using a reference count or announcement ID, and remove them only after all related announcements finish.

PR #9441, opened 2026-03-05
Reworks Alcohol

efzapa:
## About the PR
Changes how alcohol behaves in the body, now working off a scaling system that punishes excessive drinking.

## Why / Balance
With the old system of alcohol, it was very easy for people to abuse the system since you could easily drink bottles upon bottles of vodka, the only actual downside being the occasional vomit slowdown.
This has led to a culture inside of a *concerning* amount of people that play Bravo where they hunt down a booze-mat, drink for an hour, and do nothing else.
**This should not be the standard.**

In the old system:
- Took almost exactly 2 minutes to reach 15u damage threshold when drinking an entire bottle of vodka
- Only did a whopping 3 damage for 90 ethanol.
- Extremely easy to out-heal **literally** any chemical

In the new system:
- [ ] Drunkenness scales with amount of Ethanol in system
- [x] Toxin Damage scales with amount of Ethanol in system
- [x] Vomit Chance scales with amount of Ethanol in system
- [ ] Having high enough amounts of Ethanol in your system can cause you to black out until you return to safe levels
- [ ] Alcohol Ingredients and Mixed Drinks are now properly balanced. You can drink several beers relatively safely, but drink more than a couple swigs of vodka and you'll feel it.

## Technical details
**_WIP_**

## Requirements

**Changelog**

:cl:
- add: Toxin Damage now scales with amount of Ethanol in system
- add: Drunkenness now scales with amount of Ethanol in system
- add: Vomit Chance now scales with amount of Ethanol in system
- add: Having high enough amounts of Ethanol in your system can cause you to black out until you return to safe levels
- tweak: Changed the amount of ethanol that alcoholic drinks give you, proceed with caution.

### DIFF
--- a/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
+++ b/Resources/Prototypes/Reagents/Consumable/Drink/alcohol.yml
@@ -255,9 +255,6 @@
           reagent: Ethanol
           min: 15
           max: 29
-        - !type:OrganType # dwarves aren't actually immune to vomit for some reason
-          type: Dwarf
-          shouldHave: false
       - !type:ChemVomit
         probability: 0.1
         conditions:
@@ -265,24 +262,19 @@
           reagent: Ethanol
           min: 30
           max: 44
-        - !type:OrganType
-          type: Dwarf
-          shouldHave: false
       - !type:ChemVomit
         probability: 0.25
         conditions:
         - !type:ReagentThreshold
           reagent: Ethanol
           min: 45
-        - !type:OrganType
-          type: Dwarf
-          shouldHave: false
       - !type:Drunk # scale up drunk level with ethanol in body
         boozePower: 2
         conditions:
         - !type:ReagentThreshold
           reagent: Ethanol
           min: 1
+# TODO RMC14: make Dwarves actually not vomit with ethanol in their system
 
 - type: reagent
   id: Gin


chavonadelal:
It would be fun, for example, for the military police to have a breathalyzer that players have to blow into to determine their level of alcohol intoxication, and to set a certain threshold for the RP, above which a character is legally considered intoxicated.

efzapa:
> It would be fun, for example, for the military police to have a breathalyzer that players have to blow into to determine their level of alcohol intoxication, and to set a certain threshold for the RP, above which a character is legally considered intoxicated.

does sound cool, but a bit out of scope for this PR. someone can probably do that separately if they want

silencer-pl:
Alcohol doing real damage and actually knocking players out may seem like a cool implement, but in reality its nothing but a headache.

Players shouldn't have an RP "friendly" way of hurting and potentially killing themselves and making themselves a nuisance otherwise, or this will become a future staff issue, see spec chugging a bottle of vodka for a dare. Additionally, new players, not aware of the potential consequences can also be convinced to chug things to death.

All this does is forces both medbay and MPs to deal with the aftermath. The damage. The puking. The sentences. While the offending player lies there, unconscious, or unable to speak straight due to the slurring. If anything, this makes alcohol abuse more of a headache for anyone other than the offending player.

>With the old system of alcohol, it was very easy for people to abuse the system since you could easily drink bottles upon bottles of vodka, the only actual downside being the occasional vomit slowdown.

It is still possible for players to abuse the system, by drinking themselves half of fully to death. This PR does not solve this issue.

If anything, through the vomiting and toxin damage this means that attention seeking players will have MORE, not less opportunities to make themselves a nuisance and force the attention of other players onto them.

>This has led to a culture inside of a concerning amount of people that play Bravo where they hunt down a booze-mat, drink for an hour, and do nothing else.
>This should not be the standard.

Why? Other than the problems this creates for potential rulebreaks (multiple briggings in a short span of time) which this PR does not stop (you can still abuse alcohol for multiple rounds. I doubt the players doing this are frontline combatants, especially if we use the drunk Bravo as an example) 
Also, how does this stop the average Bravo from drinking themselves half to death for memes?

efzapa:
> Additionally, new players, not aware of the potential consequences can also be convinced to chug things to death.

Which this is trying to nip in the bud. You see someone chug a whole bottle of vodka, they die due to cause and effect. Then you do not repeat the same action in the pursuit of not wanting to die. If you do it anyways, you should be liable to be ahelped.

You should never not expect a whole bottle of pure ethanol to debilitate you, much less kill you. That is just a sad side-effect of there never really being substantial damage values on ethanol to begin with and allowing this behavior to fester.

> It is still possible for players to abuse the system, by drinking themselves half of fully to death. This PR does not solve this issue.

> Also, how does this stop the average Bravo from drinking themselves half to death for memes?

Which is another side-effect and aim of this PR, being having a stronger case for people that get ahelped or reported for this behavior.

When it used to be "haha its only funny vision, non-issue, deal with it IC" when you reported someone. You would be hardpressed to find an admin (hopefully) that will not take action against a Marine willingly killing themself on the field. Either you act like a normal person or you get culled with the rest of the LRP shitters.

It's the same as shooting yourself in the foot and then asking a Medic to heal you. We are killing two birds with one stone by both punishing them heavily mechanically, and (hopefully) OOC

PR #9449, opened 2026-03-06
Acid melted walls let weeds spread

keegaroo65:
## About the PR

Now when walls are melted by acid, weeds will spread over where they were
## Why / Balance

Having to plant new source weeds on the small dry spaces left behind from walls, doors, windows, etc, melting when building resin as a xeno is tedious and likely unintended behavior. This takes away that minor annoyance
## Technical details

1. `SharedXenoWeedsSystem` now has a method `GetWeedsNearby` to get an array of all weeds in a 3x3 box (or any distance really) around an entity
2. `SharedXenoAcidSystem` now calls `GetWeedsNearby` just before a melted entity is queued for deletion and ensures any found `XenoWeedsComponent` entities have a `XenoWeedsSpreadingComponent` which allows weeds to spread conforming to their typical rules
## Media

Here's a demonstration with both corrosion time and weed spreading sped up, these are conveniences not included in the PR
https://gyazo.com/96ca4345b531e31b72b521b5a1c63023
## Requirements

**Changelog**

:cl:
- fix: Fixed acid melted walls not letting weeds spread

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
@@ -4,9 +4,11 @@ using Content.Shared._RMC14.Dropship;
 using Content.Shared._RMC14.Dropship.AttachmentPoint;
 using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.Xenonids.Energy;
+using Content.Shared._RMC14.Xenonids.Weeds;
 using Content.Shared.Explosion.EntitySystems;
 using Content.Shared.Weapons.Ranged.Events;
 using Content.Shared.Coordinates;
+using Content.Shared.Coordinates.Helpers;
 using Content.Shared.Damage;
 using Content.Shared.Damage.Prototypes;
 using Content.Shared.DoAfter;
@@ -19,6 +21,8 @@ using Robust.Shared.Prototypes;
 using Robust.Shared.Timing;
 using Robust.Shared.Configuration;
 using Robust.Shared.Containers;
+using Robust.Shared.Map;
+using Robust.Shared.Map.Components;
 
 namespace Content.Shared._RMC14.Xenonids.Acid;
 
@@ -35,7 +39,9 @@ public abstract class SharedXenoAcidSystem : EntitySystem
     [Dependency] private readonly SharedPopupSystem _popup = default!;
     [Dependency] protected readonly IPrototypeManager PrototypeManager = default!;
     [Dependency] private readonly IGameTiming _timing = default!;
+    [Dependency] private readonly IMapManager _map = default!;
     [Dependency] private readonly SharedTransformSystem _transform = default!;
+    [Dependency] private readonly SharedXenoWeedsSystem _weedsSystem = default!;
     [Dependency] private readonly XenoPlasmaSystem _xenoPlasma = default!;
     [Dependency] private readonly XenoEnergySystem _xenoEnergy = default!;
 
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
```diff
@@ -317,11 +325,34 @@ public override void Update(float frameTime)
...
+        if (_transform.GetGrid(coordinates) is not { } gridUid ||
+            !TryComp(gridUid, out MapGridComponent? gridComp))
+            return;
+
+        var nearbyWeeds = _weedsSystem.GetWeedsNearby(
+            new Entity<MapGridComponent>(gridUid, gridComp), coordinates, 1);
+
+        foreach (var weeds in nearbyWeeds)
+        {
+            EnsureComp<XenoWeedsSpreadingComponent>(weeds);
```
Arendian, line 352: If all you want to do is spread a weed onto the tile the destroyed object was on, you should spawn a single weed instead of telling nearby weeds to spread. Your current implementation means that the weeds will spread in any direction and continue spreading until they reach the maximum spread distance from the nearest node.

Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
```diff
@@ -317,11 +325,34 @@ public override void Update(float frameTime)
...
+            return;
+
+        var coordinates = _transform.GetMoverCoordinates(uid).SnapToGrid(EntityManager, _map);
+
+        if (_transform.GetGrid(coordinates) is not { } gridUid ||
+            !TryComp(gridUid, out MapGridComponent? gridComp))
+            return;
+
+        var nearbyWeeds = _weedsSystem.GetWeedsNearby(
+            new Entity<MapGridComponent>(gridUid, gridComp), coordinates, 1);
```
Arendian, line 348: You should make the range value a parameter with a default value like you did in the `GetWeedsNearby` method

Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
keegaroo65: Yes you make an excellent point here. I could spawn a single weed there instead if there are weeds nearby, then my concerns are
1. not spreading diagonally
2. not spreading too far from a source weed
3. making sure it will spread further from this point (as that's intended, as shown in the gif)

As for implementation, I'm not entirely sure how to decide whether the tile should get weeds or not though (think: melting a window vs melting a wall), or how to check whether it's too far from a source or not- so my existing implementation relies more heavily on existing components.

With that in mind, think I should revise it to do it your way?

Arendian: > 1/. not spreading diagonally

You can prevent it from spawning diagonally by checking the direction from the position of the melted entity to the position of the found weeds, then ignore weeds that are NW, NE, SE, SW of the melted object's tile.

> 2/. not spreading too far from a source weed

You can do a `HasWeedsNearby` check using the position of the melted object to check if there is a weed node in range.

> 3/. making sure it will spread further from this point (as that's intended, as shown in the gif)

So if I understand correctly you mean that your change intends to make melting a structure near a tile of weed completely fill all nearby tiles with weeds as long as they are  in range of a weed node? That does sound like a bigger change than what you mentioned in the description:
> Now when walls are melted by acid, weeds will spread over where they were

I personally think making it only spread on the tile the structure occupied can be considered QOL, but spreading it any further than that would move it slightly outside the QOL category, I am just a fellow contributor though, so if you don't agree with this you can keep the behavior as it currently is and wait for a maintainer to review your PR.

> I'm not entirely sure how to decide whether the tile should get weeds or not though (think: melting a window vs melting a wall),

You can check if the melted entity has the structure tag if you want it to happen if any structure is melted. Your current implementation also makes nearby weeds spread when an item is melted, which would be fixed my checking for this tag.

PR #9471, opened 2026-03-10
Queen can't rest anymore while having an attached ovipositor

Arendian:
## About the PR

The Queen can no longer use the rest action while having an attached ovipositor.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix: The Queen can no longer use the rest action while having an attached ovipositor.


DrSmugleaf (approved):
This is good to merge but I want to make it so the plasma tree works while in ovi without having to rest first (not a thing in 13)

Arendian:
> This is good to merge but I want to make it so the plasma tree works while in ovi without having to rest first (not a thing in 13)

Your suggested change sounds like it would keep the queen’s plasma generation higher in Ovi than in 13.

If that’s the goal, what do you think about passively increasing the queen’s plasma generation rate while in Ovi, while keeping her unaffected by plasma trees in Ovi (even if plasma trees no longer require resting)?

That way plasma trees can be balanced independently from the queen’s plasma generation, and it avoids them effectively becoming a construct primarily used by the queen.

rafapecogames:
ive heard (and now seen) that smug wanted to make queen able to regen using trees without resting, which i agree since ovi is really underpowered right now. 

from the suggestion that Arendian is suggesting, it may be too OP to have ovi queen with a HUGE plasma regen, because then they could just focus on heals while hivelords used the actual trees. i would prefer queen having to do this "sacrifice" of plasma trees being used just by queen if she is ovi.

looks fair compared to just let hivelords and queen with huge plasma regen.

Utmanarn:
Make sure to implement a way to cancel queen construction with resting removed.

Currently the only way to cancel accidental build in ovi is to rest.

PursuitInAshes (approved):
Yup
Anything else we'll work out later or separately from this PR

Issue #9070, opened 2026-01-25
Infinite Power Exploit


QuietlyWhisper:
## Description

Batteries recharging in an battery powered APC charge more power than the battery outputs, resulting in an infinite power loop that allows marines to endlessly loop sets of batteries

This results in an exploit in which marines do not need to hold any generators to obtain power for communications, however because SMES is not implemented it would not make a difference.

Dutch-VanDerLinde:
Check if rechargers require power in CM13
It will be much harder to recharge comms if you need power to recharge batteries at the tower in the first place

QuietlyWhisper:
Considering this is legacy content from baystation I would imagine so

QuietlyWhisper:
yep
<img width="320" height="80" alt="Image" src="https://github.com/user-attachments/assets/bc72ab7f-6a70-4f25-b120-b60476b728f9" />

Ian321:
Duplicate of https://github.com/RMC-14/RMC-14/issues/8383. We would need https://github.com/RMC-14/RMC-14/issues/8267 fist.

Dutch-VanDerLinde:
>  We would need [#8267](https://github.com/RMC-14/RMC-14/issues/8267) fist.

The fuel is Phoron, like in base station
You'd need to find phoron on the map or get it supplied from req
OT wouldn't take kindly to stealing their phoron for gens

QuietlyWhisper:
These likely need to be fixed for this to be patched properly
- https://github.com/RMC-14/RMC-14/issues/9072
- https://github.com/RMC-14/RMC-14/issues/9071

BramvanZijp:
@QuietlyWhisper Given this has become a key part of strategies for maps where the better LZ has no nearby power, I have to ask: Would ordering this as CIC be considered bug abuse? Because I have used this in several rounds where getting power early on was not viable given the map layout, but if this IS bug abuse, Ill have to adjust those strategies to just bring a large amount of batteries, and charging those on the alamo whenever it is on the FOB.


I will note though, this does not make nuke holds THEMSELVES easier, as the countdown is interrupted if either tower loses power, which happens if the battery is swapped. But it does help with sieges LEADING UPTO those holds, true.

QuietlyWhisper:
> [@QuietlyWhisper](https://github.com/QuietlyWhisper) Given this has...

After further research with smug and QA members, we found that even with power properly implemented there would be no significant change in power balancing unfortunately, due to portable generators.

[...]

Issue #8267, opened 2025-10-31
P.A.C.M.A.N. portable generator functionality


CyberTropic:
## Description

Pacman/Super Pacman portable generators should use phoron sheets for fuel, and provide power to the powernet, prioritizing power for the APC for the area they're located in

**Reproduction**

**Screenshots**

**Additional context**

PR #9476, opened 2026-03-12
fix: patch infinite power exploit with dynamic charging state

CowboyBingus:
## About the PR

- APCs with power cells now deliver a balanced amount of power when charging another cell.
- As a side effect, chargers now draw from the correct APC when moved to a new location and check if a battery is inside when re-wrenched.

Addresses #9070

## Why / Balance
Currently players can generate unlimited power using two power cells. This obviously isn't intended behavior. This change should only affect charging stations and power cells. A charging station will now draw 1500W but remain at 0W when not in use. This change should make it so that batteries now draw 50% more power than they charge with, preventing the exploit.

When a charger is moved to a new location and wrenched down, it will also update its charging status depending on whether a battery is already present in the charger or not.

## Technical details
Added RMCPowerReceiverComponent to RMCRecharger so it registers as a load on the RMC APC system. The charger draws 1500W while actively charging and 0W when idle by listening to ActiveChargerComponent lifecycle events that ChargerSystem already manages.

OnReceiverRemove subtracts LastLoad in SharedRMCPowerSystem from the area's load on removal and nulls any cached Area reference. 

This also adds an AnchorStateChangedEvent subscription to ChargerSystem so charger status is re-evaluated on anchor changes, fixing re-anchored chargers not resuming when moved.

## Media
N/A

## Requirements

**Changelog**
:cl:
- fix: Fixed rechargers not drawing power from APCs, allowing infinite battery swapping.
- fix: Fixed power load leaking when powered devices are removed from an area.
- fix: Fixed chargers not resuming charging after being re-anchored.
- tweak: Reduced recharger charge rate from 1000 to 10 for power balance.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Power/recharger.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Power/recharger.yml
@@ -32,9 +32,13 @@
       machine_parts: !type:Container
   - type: ApcPowerReceiver
     needsPower: false # TODO RMC14
-    powerLoad: 0 #15 kW
+    powerLoad: 0
+  - type: RMCPowerReceiver
+    idleLoad: 0
+    activeLoad: 1500
+    channel: Equipment
   - type: Charger
-    chargeRate: 1000
+    chargeRate: 10
     chargeLevelSteps: 6
   - type: ItemSlots
     slots:
--- a/Content.Server/Power/EntitySystems/ChargerSystem.cs
+++ b/Content.Server/Power/EntitySystems/ChargerSystem.cs
@@ -28,6 +28,7 @@ internal sealed class ChargerSystem : EntitySystem
     {
         SubscribeLocalEvent<ChargerComponent, ComponentStartup>(OnStartup);
         SubscribeLocalEvent<ChargerComponent, PowerChangedEvent>(OnPowerChanged);
+        SubscribeLocalEvent<ChargerComponent, AnchorStateChangedEvent>(OnAnchorChanged);
         SubscribeLocalEvent<ChargerComponent, EntInsertedIntoContainerMessage>(OnInserted);
         SubscribeLocalEvent<ChargerComponent, EntRemovedFromContainerMessage>(OnRemoved);
         SubscribeLocalEvent<ChargerComponent, ContainerIsInsertingAttemptEvent>(OnInsertAttempt);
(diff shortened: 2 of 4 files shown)


CowboyBingus:
I'm aware #8267 will likely be a requirement to merge this. I can maintain this PR and take a look at parity for the generators as well, assuming it's not currently being worked on.

BramvanZijp:
Quick note: Reducing recharger recharge rate by 100x has impacts WAY beyond the scope of this PR: defibs, nv visors, mp tools, etc.

CowboyBingus:
> Quick note: Reducing recharger recharge rate by 100x has impacts WAY beyond the scope of this PR: defibs, nv visors, mp tools, etc.

I'll double check and test tomorrow since it's late here, but for context the recharge rate was part of some initial testing for the bug fix so I'm pretty sure it can be reverted without issue. 

The core of the patch is the power draw on the APCs outpacing the recharge rate of the battery.

BramvanZijp:
> > Quick note: Reducing recharger recharge rate by 100x has impacts WAY beyond the scope of this PR: defibs, nv visors, mp tools, etc.
> 
> I'll double check and test tomorrow since it's late here, but for context the recharge rate was part of some initial testing for the bug fix so I'm pretty sure it can be reverted without issue.
> 
> The core of the patch is the power draw on the APCs outpacing the recharge rate of the battery.

Yeah I know, just pointing out that the recharge rate is most likely going to add a lot of extra hoops to review which is going to stall the PR significantly.

Though I will note, as an XO player, I am fine with this IF there is a reasonable alternative to using an LZ where we will not immediately have access to planet, generators, be that PACMANs or a feature that means comms are "free" from a power standpoint during the opening of an operation (the same timeframe during which the FOB has turrets)

CowboyBingus:
Generators PR now opened: #9478

Resources/Prototypes/_RMC14/Entities/Structures/Power/recharger.yml
```diff
@@ -32,9 +32,13 @@
       machine_parts: !type:Container
   - type: ApcPowerReceiver
     needsPower: false # TODO RMC14
-    powerLoad: 0 #15 kW
+    powerLoad: 0
+  - type: RMCPowerReceiver
+    idleLoad: 0
+    activeLoad: 1500
```
AverageNotDoingAnythingEnjoyer, line 38:
```suggestion
    activeLoad: 15000
```

PR #9478, opened 2026-03-13
feat: implement (super)pacman generators

CowboyBingus:
## About the PR
Implements PACMAN and Super PACMAN portable generators. Portable generators can be fueled and activated to provide emergency power to area APCs. Includes fuel loading, heat, overheat explosion (won't happen in standard gameplay), xeno interactions, and a GUI for managing power output.

## Why / Balance
These two generators are being implemented in preparation for patching an existing infinite power exploit related to cycling power cells in an APC.

PACMAN outputs 5kW, fueled by phoron, 70 seconds per sheet.
SUPERPACMAN outputs 15kw, fueled by uranium, 120 seconds per sheet.

The SUPERPACMAN's fuel type was set to uranium for balancing reasons, to give both of them a purpose in usage.

## Technical details
- `RMCPortableGeneratorComponent.cs` - New shared component for all generator data fields (watts, fuel type, heat, sheets, power percent, explosion params).
- `SharedRMCPowerSystem.cs` - Shared interaction handlers: fuel loading, xeno sabotage with size gating, power level adjustment, fuel ejection, examine text, and anchor state change shutoff.
- `RMCPowerSystem.cs (Server)` - `UpdatePortableGenerators()` handles per-tick fuel consumption, heating, and power contribution to area APCs via the `_portableGenPower` dictionary.
- `generator.yml` - Prototype definitions for both generators.
- `rmc-portable-generator.ftl` - Locale strings for all generator interactions.
- Client UI - `RMCPortableGeneratorBui.cs`, `RMCPortableGeneratorWindow.xaml/.cs` for the generator interface.

## Media
Content Client_W9euB2GCK7

## Requirements

**Changelog**
:cl:
- add: P.A.C.M.A.N. portable generator (phoron-fueled, 5.0kW power output).
- add: S.U.P.E.R.P.A.C.M.A.N. portable generator (uranium-fueled, 15kW power output).
- add: Xenomorphs can sabotage portable generators, but smaller xenos will be too weak to do so.
- code: Portable generators contribute power to area APCs via the RMC power system.


IceNoobXD:
would resolve #8267

Resources/Prototypes/_RMC14/Entities/Structures/Power/generator.yml
```diff
@@ -194,3 +223,12 @@
       state: portgen1
     - type: Machine
       board: RMCPortableGeneratorSuperPacmanMachineCircuitboard
+    - type: RMCPortableGenerator
+      watts: 15000
+      timePerSheet: 120
+      fuelStackType: Uranium
```
CyberTropic, line 229: This would have to be RMCUranium, which hasn't been added yet so this is as good a time as any

CyberTropic (commented):
I was working on this before I went on hiatus last fall, good to see someone else has done it already. Some things from testing on local:

- The generator should be making noise when it's on, should be fine to use upstream /Audio/Ambience/Objects/engine_hum.ogg
- Eject Fuel button should be deactivated when there is no fuel to eject, it remains clickable if there's partial sheet left but it doesn't do anything when clicked
- Partial sheet should be indicated on the UI somewhere, currently it's not clear when there's a partial sheet left in the generator, so it looks like you can activate it without fuel, but on further testing it works fine, but just not clear.
- There should be a couple more error pop ups in general I feel like, for example when trying to insert incorrect fuel, failing to start because of no fuel, etc

PR #9483, opened 2026-03-13
add: field doctor access

### LINKED ISSUE #9289: Field Doctor parity access
## Description

Currently Field doc doesnt have their own access which is needed to lock some of their stuff away from other medical staff
It'd also be nice if we had a 1 tile wide airlock with their access added for mapping purposes

**Additional context**

https://github.com/cmss13-devs/cmss13/blob/master/code/__DEFINES/access.dm#L47

Ian321:
## About the PR

title

## Why / Balance

- fixes #9289

## Technical details

The following jobs start with the `RMCAccessFieldDoctor` now:

- CO
- XO
- SEA
- ASO
- CMO
- Field Doctor

Voice your objections. (Unsure about ASO)

Also added locked doors and gave the existing locker the access lock.

## Media

Screenshot From 2026-03-13 17-14-10

## Requirements

**Changelog**

Nothing player facing.


BramvanZijp:
Voicing my objection in regards to ASO having it. (Then again, they also have access to areas like research, CE, etc, so might be fine to keep it temporarily to then be later adressed whenever ASO is finally tackled properly.)

DrSmugleaf (approved):
I need to check if the ASO is supposed to have it but this is better for now

PR #9490, opened 2026-03-13
Adds: Coffee Machine Functionality

### LINKED ISSUE #9414: Coffee Machine
## Description

A functional soul prop that allows to brew coffee introduced with Hybrisa map

**Screenshots**

Image

**Additional context**

/obj/structure/machinery/hybrisa/coffee_machine

wheelwrightt:
## About the PR
Coffee machine now dispenses coffee into mugs that are put inside it and activating it. Takes 10s to brew the coffee in that time it cannot be removed. After brewing if the player is near the machine then it goes directly into their hand but if not then it displays a popup notifying of it finishing it brewing. If the player tries to put a full cup it displays a pop up, if the player puts for example a half full cup when it finished some of the coffee will spill with a relevant popup. Resolves: #9414 

## Why / Balance
nonfunctional coffee machine bad. parity

## Technical details
Uses DrinkGlass as whitelist for what can be put in. Displays the generic overlay for the coffee mug inserted, can be added at a later date once proper spriting for individual mugs etc. has been done. Client-side visualizer 
RMCCoffeeMachineVisualizerSystem for sprite layer management. Added and converted coffee1 to mono. Brew time and status managed in RMCCoffeeMachineComponent. YML and relevant localization added.

## Media
https://github.com/user-attachments/assets/6372f041-fbea-4f2d-baf9-507ceb1c8557

[spill.webm](https://github.com/user-attachments/assets/0d9ab563-28ad-4dd4-8c8c-704d731165cf)

## Requirements

**Changelog**
:cl:
- add: Coffee machines now can brew coffee into mugs.


AverageNotDoingAnythingEnjoyer (commented):
You sure it's working as intended? Not a code expert but seems there's some missing lines of text, not to mention it should have a whitelist to what containers it can fit (cups)

In other words, I suggest looking into CM code and probably testing it in their sandbox
https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/structures/hybrisa_props.dm#L1775

AverageNotDoingAnythingEnjoyer:
Also the logic of how it works may be wrong, unsure

wheelwrightt:
> Also the logic of how it works may be wrong, unsure

expand on this

AverageNotDoingAnythingEnjoyer:
> expand on this

Basically you insert a cup/mug (not every container fits) inside of it, the coffee machine fiils it with coffee in a few seconds, if you try to take it out beforehand the system won't let you to. If you try to insert a cup that's full or has any liquid in it, there should be a popup that the coffee is spilt after the brewing is done. 
https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/structures/hybrisa_props.dm#L1775

There's overlays for mugs inserted into the machine in 
https://github.com/cmss13-devs/cmss13/blob/master/icons/obj/structures/machinery/coffee_machine.dmi
Some of them need resprites, for our WY and UNMC mugs are different

While the cup is brewing, there's a sound in 
https://github.com/cmss13-devs/cmss13/blob/master/sound/machines/coffee1.ogg 
Instead of the one you used

After the coffee is done brewing, the cup slides into your hand automatically. If you're too far though, the cup will stay inside the machine and there should be a popup message or something to signal the cup is ready
image
https://github.com/user-attachments/assets/eca97596-536f-401e-8034-a0db8249613c

AverageNotDoingAnythingEnjoyer:
Also, while it could be cool to have resprites of mugs in here included too, having them be just shown with generic mug overlay is fine

Oguz0k220:
I think it would be nice to add a UI, like in the dispenser of alcoholic drinks and the like, with the ability to choose the required number of units

AverageNotDoingAnythingEnjoyer:
> I think it would be nice to add a UI, like in the dispenser of alcoholic drinks and the like, with the ability to choose the required number of units

It'd be better to get parity variant first

wheelwrightt:
Everything but sprites added, uses generic cup sprite as of now.

Updated PR with new info and media.

wheelwrightt:
> I think it would be nice to add a UI, like in the dispenser of alcoholic drinks and the like, with the ability to choose the required number of units

parity first, then comes the mocha matcha latte with relevant non-diary milk options.

AverageNotDoingAnythingEnjoyer:
> Everything but sprites added, uses generic cup sprite as of now.

Looks good, could you add an overlay for coffee cups? There's a sprite for them in cm13 we can use and would be good as an example for future overlays, since the generic one is for all cups/mugs

DrSmugleaf (approved):

PR #9492, opened 2026-03-14
feat: implement proper tall grass collision and occlusion

CowboyBingus:
## About the PR
In its current form, tall grass serves essentially no purpose other than to be removed because it's a hindrance to both teams. This PR makes the grass actually act like grass.

## Why / Balance
The new grass includes following features / balance changes:
- Tall grass no longer blocks placement of resin buildings
- Tall grass now provides full occlusion for small xenos & partial occlusion for larger xenos (mainly for immersion).
- Tall grass slows human movement to 0.65
- Tall grass is now flammable, no longer blocks flames, and no longer blocks xeno ground spits

## Technical details
- TallGrassComponent added to tall grass entity prototypes to identify them for spatial queries.
- HiddenInGrassComponent dynamically added/removed from small xenos when they enter/leave grass. Drives a GetDrawDepthEvent handler that lowers draw depth to SmallMobs, rendering the xeno behind the grass sprite (at Walls depth).
- TallGrassSystem polls every tick, iterating all XenoComponent entities. Filters by RMCSizes (only Small, VerySmallXeno, SmallXeno can hide). Uses EntityLookupSystem.GetEntitiesInRange<TallGrassComponent> to detect proximity to grass and adds/removes HiddenInGrassComponent accordingly, calling UpdateDrawDepth on change.
- Added MapInitEvent handler for initial sync on spawn, and an Update() loop that runs SyncOcclusion for every FloorOcclusionComponent entity each tick.
- SyncOcclusion performs a spatial query, then filters results through RMCWaterSystem.CanCollide, and checks the result against the existing Colliding list. Only calls Dirty()/SetEnabled() when the list actually changes.
- tallgrass.yml: Enabled physics, added TallGrass, FloorOccluder, SpeedModifierContacts (0.65 speed), and Flammable components.
- base_xeno.yml: Added RMCMobStateDrawDepth and FloorOcclusion components to CMXenoBase, giving all xeno castes floor occlusion support.

## Media

flamethrower

spitter

building

Content Client_IlgvCjlHPb

## Requirements

**Changelog**
:cl:
- tweak: Tall grass no longer prevents placement of resin secretions.
- tweak: Tall grass is now flammable.
- tweak: Tall grass now occludes sprites based on size, fully hiding small xenos and partially obscuring larger entities.
- tweak: Tall grass no longer blocks acid spit.
- tweak: Tall grass will now slow marine movement to 0.65.

### DIFF
--- a/Content.Client/Movement/Systems/FloorOcclusionSystem.cs
+++ b/Content.Client/Movement/Systems/FloorOcclusionSystem.cs
@@ -1,3 +1,4 @@
+using Content.Shared._RMC14.TallGrass;
 using Content.Shared.Movement.Components;
 using Content.Shared.Movement.Systems;
 using Robust.Client.GameObjects;
@@ -10,9 +11,13 @@ public sealed class FloorOcclusionSystem : SharedFloorOcclusionSystem
 {
     private static readonly ProtoId<ShaderPrototype> HorizontalCut = "HorizontalCut";
 
+    [Dependency] private readonly EntityLookupSystem _clientLookup = default!;
     [Dependency] private readonly IPrototypeManager _proto = default!;
+    [Dependency] private readonly SharedTransformSystem _clientTransform = default!;
 
     private EntityQuery<SpriteComponent> _spriteQuery;
+    private readonly HashSet<Entity<FloorOccluderComponent>> _clientOccluders = new();
+    private readonly HashSet<EntityUid> _shaderApplied = new();
 
     public override void Initialize()
     {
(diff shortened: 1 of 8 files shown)


CowboyBingus:
Forgot to mention this the other day but this PR may have a potential performance impact associated with it because the occlusion code for the xenos had to be updated. 

This PR probably needs stress testing (specifically with large amounts of xenos) on weaker hardware before merging (which I unfortunately don't have).

Resources/Prototypes/_RMC14/Entities/Structures/Flora/tallgrass.yml
```diff
@@ -5,14 +5,38 @@
   name: tallgrass
   suffix: Centre
   components:
-  # make this slashable by only machete
+  - type: Tag
+    tags: []
```
Vixting, line 9:
```suggestion
```

Content.Client/Movement/Systems/FloorOcclusionSystem.cs
```diff
@@ -10,39 +11,49 @@ public sealed class FloorOcclusionSystem : SharedFloorOcclusionSystem
...
     }
 
     private void OnOcclusionShutdown(Entity<FloorOcclusionComponent> ent, ref ComponentShutdown args)
     {
         SetShader(ent.Owner, false);
     }
 
-    protected override void SetEnabled(Entity<FloorOcclusionComponent> entity)
+#pragma warning disable RA0028
+    public override void Update(float frameTime)
```
Vixting, line 37: this should be event driven instead of running in update loop

Content.Shared/Movement/Systems/SharedFloorOcclusionSystem.cs
Vixting, line 1: this shouldn't need to be changed into a update loop
use the events

Content.Shared/_RMC14/TallGrass/TallGrassSystem.cs
Vixting, line 1: same thing here just use start / end collide events but just make it run after the flooroclusion

Vixting (changes requested):
Use the events instead of running in update loop
If you want I can make the changes myself as I want to look into the new water overlays cm is doing anyway

PR #9531, opened 2026-03-19
Remove Dedicated PvCI Role + Playtime Rewards for PVI

PursuitInAshes:
## About the PR
Removes the dedicated PVCI Role
Adds the PVCI Jacket and Beret as playtime rewards inbetween gold and plat for PVI (120 Hours).
Changes those item descriptions to more clearly be playtime based rather than position/rank.
Fixed a missing suffix on a bag.
Removes the PVCI Rank and ID card.

I feel the need to be clear this is just an idea for what to do with the role. I'm not deadset on it and it should probably be discussed between hosts.

## Why / Balance
Role was woefully outdated and unlikely to be used given we have PvIs now. I didn't want to give the rank as a playtime reward but can easily upon request. I personally don't think its very fitting as there should be very few chief inspectors, or again I can outright remove it and structure wise leave it entirely out of game

## Technical details
N/A

## Media
N/A

## Requirements

**Changelog**
:cl: PursuitinAshes
- tweak: The dedicated PvCI role has been removed, with some of its clothing added as playtime rewards for PvIs.


PursuitInAshes:
The IDs are maybe slop, not sure if its worth the effort to change them though.

PursuitInAshes:
Going to update the IDs to be consistent/better, the names will probably stay "Senior" though

Sifb26:
Throwing in my 2 cents for whoever reviews to maybe take into consideration, In my (somewhat biased, especially since I am already at the standard playtime req for this kind of thing) opinion, PvCI is better off as a playtime rank up reward since they would still be really rare and the title in general would probably not be used on non-inspectors in the first place to avoid confusion

QuietlyWhisper:
I wouldn't mind if CI was cut yeah, unused content with no plans of being used - in regards to any "rank" stuff, senior is fine, but not chief tbh

PursuitInAshes:
Undrafting for tests to run

Removed the PVCI rank
Removed "Senior" from the playtime rewards due to "Senior" being used on some Marshal gear, so I just dropped senior and ID wise added "Blue" to it. Before this PR it used to be "Command" but that really wasn't fitting either, even for PvCIs. So now its just a piece of clothing or a new emblem ICly provost members can get given to represent time in service, which feels fine enough

PursuitInAshes:
Done a final cleanup wave. Anything else related to the rank I think would be best to put on a separate PR (Like if we want a "Senior" rank or not), this should be good for review

DrSmugleaf (dismissed):
yaml good

PursuitInAshes:
I was asked to remove any mentions of "Senior", documenting it here so I don't forget it when I come back to this

PursuitInAshes:
https://github.com/RMC-14/RMC-14/pull/10349 Early merged here so if tests fail I can fix them

PR #9532, opened 2026-03-20
Fix Nutritious Property (again)

Arykh:
## About the PR

I messed up bro 💀 

## Why / Balance

Parity

## Technical details

I used ActualPotency instead of Potency in #8660 

Only use Potency(Level) for adding because chemicals tick every second, not two.
Final result should be 3.5 nutrition/s for **_Nutriment_** as mentioned in #8660 

Because of ActualPotency, it's only restoring 2.5 nutrition/s on live servers right now.

## Requirements

**Changelog**

:cl:
- fix: Fixed Nutritious Property restoring less nutrition than intended. (again)


Arykh:
Tested. All gud.

PR #9534, opened 2026-03-20
tweak: foxtrot scaling

Ian321:
## About the PR

This PR adds scaling to the number of ghosts spawned when purchasing "Wake Up Additional Troops" from the tech console.

## Why / Balance

The current numbers are based on CM13 which has (as far as I know) a player limit of 150. This PR adjusts the number of slots  above that limit.

With the currently configured scale of `301`:

- Base: One leader, medic, ct and two rifleman.
- 151+: One additional rifleman.
- 301+: One additional medic, ct and rifleman.
- 452+: One additional rifleman.
- 602+: One additional medic, ct and rifleman.
- ...

Alternatively we could only scale the number of rifleman, but that is for the hosts to decide.

## Technical details

Additional cvar `rmc.cryo_marines_scale` to configure the scaling with the default value of `301`.

## Media

https://github.com/user-attachments/assets/07528f9c-67e3-4709-8d2f-6a7d6349700f

## Requirements

**Changelog**

:cl:
- tweak: The amount of cryomarines woken up now scales with the amount of players.

### DIFF
--- a/Content.Server/_RMC14/Intel/Tech/TechSystem.cs
+++ b/Content.Server/_RMC14/Intel/Tech/TechSystem.cs
@@ -2,9 +2,12 @@ using Content.Server.GameTicking.Events;
 using Content.Server.Ghost.Roles.Components;
 using Content.Server.Humanoid.Components;
 using Content.Server.Spawners.Components;
+using Content.Shared._RMC14.CCVar;
 using Content.Shared._RMC14.Intel.Tech;
 using Content.Shared.Humanoid.Prototypes;
 using Robust.Server.GameObjects;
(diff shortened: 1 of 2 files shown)


AverageNotDoingAnythingEnjoyer:
CM doesn't have a player limit in general, but considering most of the player scale balancing ends at 200, you could start adding extra slots from there

Ian321:
Having thought a bit about it. The 5 marines waking up is equivalent to the 5 burrowed larvas that the xenos can get. So scaling it doesn't make much sense. Ill set this PR to draft in case someone wants to come back to it.

PR #9541, opened 2026-03-21
Crusher & Charger charge fixes

MarshesAndMeadows:
## About the PR

A number of bug fixes with the charge ability and some for the togglecharge too.

## Why / Balance

Basecrusher has some reliability issues in its charge ability due to various bugs. 
Charger has reliability issues with its charge as well.
Also parity in the case of table-passthrough and knockback

## Technical details
Broadened interactions with charge; before if something did not have an opt-in flag it would simply do nothing and stop the charge, now the charge will always deal damage if the target has a damageablecomponent, and the chargeablecomponent (the opt-in flag) is now used to denote special behavior, like whether the charge should keep going through the item, delete it regardless of health, deal specific damage and so on.
Also fixed the ordering of checks so it should no longer stop the charge when it hits friendly hive structures. (nor will it damage them)

Also huge thanks to TA-NK for meticulously gathering and reporting bugs for the charger and basecrusher.

## Media
https://github.com/user-attachments/assets/18db3dec-d525-4f94-96b7-40af15c70251

https://github.com/user-attachments/assets/6c076865-2e4e-4161-a75b-5cb01692a715

## Requirements

**Changelog**
:cl:
- fix: FIxed crusher charge sometimes phasing through barricades. It should now stop at the barricade and still deal damage. Keep an eye out and report it if the bug returns.
- fix: Fixed handrails being treated like barricades for crushers charge.
- fix: Fixed crushers charge stopping when hitting friendly hive structures like recovery nodes, clusters and so on.
- fix: Fixed crushers charge not dealing damage to many very specific objects (Wey-Ya vendors, certain machines, directional windows, many more)
- fix: Fixed crushers charge stopping on the first table it hits, it now continues through tables while destroying them.
- fix: Fixed crushers charge dealing lower damage than intended to walls and Rwalls, it now deals the correct 750 structural damage.
- fix: Fixed crushers knockback not matching parity (1½ -> 3 tiles)
- fix: Fixed crusher charge not being targetable, causing it to always travel the full length, regardless of targeting.
- fix: Fixed crusher charge range being longer than it should be.
- fix: Fixed crusher charge not correctly destroying window frames.
- add: Added a fallback damage logic path to Charger that should catch interactions that otherwise would take no damage.


Content.Shared/_RMC14/Xenonids/Charge/XenoChargeComponent.cs
```diff
@@ -18,7 +18,7 @@ public sealed partial class XenoChargeComponent : Component
     public DamageSpecifier Damage = new();
 
     [DataField, AutoNetworkedField]
-    public float Range = 8;
+    public float Range = 9;
```
Arendian: You can use RMCMathExtensions.CircleAreaFromSquareAbilityRange to convert the range, 9 is too high.
```suggestion
    public float Range = RMCMathExtensions.CircleAreaFromSquareAbilityRange(7);
```

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
```diff
@@ -378,20 +377,12 @@ private void OnXenoChargeHit(Entity<XenoChargeComponent> xeno, ref ThrowDoHitEve
...
-                    _destruct.DestroyEntity(targetId);
-                return;
-            }
-
         }
 
-        var damage = _damageable.TryChangeDamage(targetId, _xeno.TryApplyXenoSlashDamageMultiplier(targetId, structDamage), origin: xeno, tool: xeno);
-        if (damage?.GetTotal() > FixedPoint2.Zero)
+        var finalDamage = _xeno.TryApplyXenoSlashDamageMultiplier(targetId, structDamage);
+        var damage = _damageable.TryChangeDamage(targetId, finalDamage, origin: xeno, tool: null);
```
Arendian: The tool value should stay "xeno" as it was before, if set to null the damage will ignore armor.
```suggestion
        var damage = _damageable.TryChangeDamage(targetId, finalDamage, origin: xeno, tool: xeno);
```

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
```diff
@@ -404,25 +395,40 @@ private void OnXenoChargeHit(Entity<XenoChargeComponent> xeno, ref ThrowDoHitEve
...
-
                     if (_net.IsClient)
                         _transform.DetachEntity(targetId, Transform(targetId));
                 }
             }
         }
 
-        var range = xeno.Comp.Range;
-
-        if (crush != null && crush.ThrowRange != null)
```
Arendian, line 417: Removing this means that the crusher now always charges the max range it can unless stopped by something else. If you now try to charge to 2 tiles in front of you, you instead charge 9 tiles in that selected direction.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
MarshesAndMeadows: Problem with that was that it was causing the charge damage to be interpreted as a melee slash against walls and Rwalls and having its damage overwritten, the walls have predefined set damage amounts that slashes do to them (5% max hp of the wall in the case of Rwalls) assuming the xeno has the right Claw Sharpness.
So if i correct it back to xeno, the charge goes back to dealing the incorrect 150 dmg to walls and 450 to Rwalls.
Im not quite sure how else to get around this issue for now.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
MarshesAndMeadows: As far as i can tell from my own testing, this was just dead code; the current version of crusher on master also cannot charge less than the full distance.
Not sure whether its parity behavior or not

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
Arendian: This is caused by the `XenoClawsSystem`, and the walls having the `ReceiverXenoClawsComponent`. A way to fix it is adding a new component to the crusher after it's charging DoAfter finishes, removing it again after it stops charging. The `XenoClawSystem` can then return early if the crusher has this new component.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
Arendian: Ah that's my mistake then, I thought it  did charge at the selected tile before your changes. That piece of changed code was the only thing that touched range, so I made a wrong assumption. The parity behavior is that you stop at the location you selected as your target, so if you only want to charge 4 tiles you charge 4 tiles. This can also be done in a later PR.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeComponent.cs
MarshesAndMeadows: image

updated the range based on what i read from the cm13 repo code
not sure if the 7 range is something from localhost testing or what.

either way, ill implement your suggestion until shown otherwise.

Arendian: In CM you can use the charge ability on any tile in range, and their view range is 7 tiles in any direction(this is where the 7 comes from). In CM this means 7 tiles horizontal, but also 7 tiles diagonal. 7 tiles diagonal is however more distance than 7 horizontal/vertical, so we have to do a calculation to get a circle with the same surface area as the CM 7x7 grid range. 

I calculated the surface area using the CircleAreaFromSquareAbilityRange formula, and with how the ability works (You jump to the center of the tile you select, you can't jump to the edge of the tile in CM) it might need to be 6.5 instead of 7. This results in a range value of 7.9(close to the original 8) which has the same surface area as a 7x7 square(49).

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
```diff
@@ -404,25 +397,40 @@ private void OnXenoChargeHit(Entity<XenoChargeComponent> xeno, ref ThrowDoHitEve
...
-            range = crush.ThrowRange.Value;
-
         _rmcPulling.TryStopAllPullsFromAndOn(targetId);
 
         var origin = _transform.GetMapCoordinates(xeno);
 
         _stun.TryParalyze(targetId, xeno.Comp.StunTime, true);
-        _sizeStun.KnockBack(targetId, origin, 2, 2, knockBackSpeed: 10);
+        _sizeStun.KnockBack(targetId, origin, 3, 3, knockBackSpeed: 15);
```
Arendian:
```suggestion
        _sizeStun.KnockBack(targetId, origin, knockBackDistance, knockBackDistance, knockBackSpeed: 15);
```

See the other comment for the reason.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
```diff
@@ -404,25 +397,40 @@ private void OnXenoChargeHit(Entity<XenoChargeComponent> xeno, ref ThrowDoHitEve
...
                         _transform.DetachEntity(targetId, Transform(targetId));
                 }
             }
         }
 
-        var range = xeno.Comp.Range;
-
-        if (crush != null && crush.ThrowRange != null)
-            range = crush.ThrowRange.Value;
-
```
Arendian, line 419: I think this is supposed to be an optional knockback distance overwrite by the entity being hit by the charge.  The person that added it just forgot to replace the hardcoded knockback distance parameter with the range value.

I think making a new datafield `KnockBackDistance` on the XenoChargeComponent that stores the knockback distance, and replacing `var range = xeno.Comp.Range` with `var knockBackDistance = xeno.Comp.KnockBackDistance` is the cleanest option.

Content.Shared/_RMC14/Xenonids/Charge/XenoChargeSystem.cs
MarshesAndMeadows: Appreciated, adjusted.

Content.Shared/_RMC14/Xenonids/ClawSharpness/XenoClawsSystem.cs
```diff
@@ -49,6 +52,10 @@ private void OnReceiverDamageModify(Entity<ReceiverXenoClawsComponent> ent, ref
     {
         var xeno = args.Tool;
         var receiver = ent.Comp;
+
+        if (!_xenoChargingQuery.HasComp(xeno))
```
DrSmugleaf: Is this and the other charging check below intended? This would make anyone but the charger not receive their damage multipliers

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/ClawSharpness/XenoClawsSystem.cs
MarshesAndMeadows: If i recall right, this was done as part of trying to get the charge to do the intended 750 structural to walls instead of the normal xeno claw damage, as the charge damage was getting routed through the claw logic, but i needed that logic to handle armor reductions.
Id definitely appreciate suggestions to better ways to do it, as that is not an intended consequence of that change

DrSmugleaf: You would probably need to pass in more information for that, you could try ignoring resistances if resistances never matter for charge damage, otherwise you would need to add a new argument to the DamageModifyEvent and pass that information through

DrSmugleaf (approved):

PR #9554, opened 2026-03-26
Hive core rebuild cooldown now applies after 15 minute mark

MidZik:
## About the PR

Hive core can currently be destroyed and rebuilt without cooldown any time before the 20 minute mark. This PR changes it to only apply before the 15 minute mark.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl:
- tweak: Hive core cooldown-free rebuild time reduced from 20 minutes after round start to 15 minutes.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Hive/HiveComponent.cs
+++ b/Content.Shared/_RMC14/Xenonids/Hive/HiveComponent.cs
@@ -65,7 +65,7 @@ public sealed partial class HiveComponent : Component
     public TimeSpan NewCoreCooldown = TimeSpan.FromMinutes(5);
 
     [DataField, AutoNetworkedField]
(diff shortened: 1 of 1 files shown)


Arendian (approved):
The change is parity. Parity has it set to 20, but in parity the world time includes the time in lobby.

Hive core replacement without cooldown finishes at 15 minutes(20, but does not add lobby duration to it):

Image
Image

Pre fuel also finishes at 15 minutes in game (set to 15, but adds lobby duration to that value):

Image
Image

PursuitInAshes:
Posting discussion before I forget
Parity has Hivecore restricted to certain locations (I believe by Area?) which is why something like this is viable. We're going to have to look at if we want something https://github.com/RMC-14/RMC-14/pull/9821 first probably before this can be considered

PR #9576, opened 2026-03-30
[PVE] [RCM] Adds minimum RCM time requirement to Troop Commander.

TadJohnson00:
## About the PR

Added an RCM jobs playtime tracker, added one-hour time requirement in any RCM role for the Troop Commander.
## Why / Balance

The structure of the RCM depends more heavily on competent command members due to its unique setup. On multiple occasions now, players with zero experience with the RCM have taken high-impact roles to the detriment of the round.
While I'm of the general opinion that players should *not* take high-impact roles if they're not confident in their ability, *especially if it's their first time playing a faction,* the playtime requirement is very low at only one hour. You could probably get half of the required playtime without even hitting the ground in an RCM operation. You can also roll for every other role in the RCM without needing any experience, so people shouldn't be complaining about missing their rolls.
## Technical details

YAML, locale
## Media

image

## Requirements

**Changelog**

:cl:
- tweak: [PVE] Added a 1hr Royal Crown Marines time requirement to the Whiskey Troop Commander role.

### DIFF
--- a/Resources/Locale/en-US/_RMC14/job/cm-role-timers.ftl
+++ b/Resources/Locale/en-US/_RMC14/job/cm-role-timers.ftl
@@ -6,6 +6,8 @@ role-timer-dropship-pilot-roles = any dropship pilot roles
 role-timer-corporate-roles = any corporate roles
 role-timer-survivor-roles = any survivor roles
 
+role-timer-rcm-roles = any Royal Crown Marines roles
+
 role-timer-total-department-insufficient = You require [color=yellow]{TOSTRING($time, "0")}[/color] more minutes as [color={$rolesColor}]{$roles}[/color] to play this role.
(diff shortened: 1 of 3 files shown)


PursuitInAshes (changes requested):
This probably wont have the intended effect of making the Commander any more competent/skilled and instead just make Commanders less common due to how hard it is to get roles in PVE at times

I'd like to figure out another way that isn't another playtime hurdle if possible. I'll talk with some of the GMs about ideas but you are free to ping me

PursuitInAshes:
Thanks for the PR
Going to opt to close this for the above reasoning, sorry it took so long to give a definitive answer
As a note, so long as GMs/GAs are aware of issues we can be ontop of them and work with players to improve

PR #9580, opened 2026-03-31
Chance's Claim Passover

CatAndHats:
## About the PR
Made parity changes to Chance's Claim. Added pipe network, west lz north comms shortcut, surv loot moved about, missing items, and decals. Made pre-armed claymores for mapping/event purposes. Added miner's equipment secure locker.

Made non-parity changes too. Moved scout cloak and armour out east of the reactor, and added more turrets to the LZs.

## Why / Balance
Parity. Comms shortcut was agreed to be a good change by Smug, Cro, and Vorkath.

## Technical details
Fixes pipes overlaying above some tiles.

## Media
west LZ north comms shortcut
Screenshot 2026-03-31 082136

ML66D added. Guns, ammo, and the mini sentry moved to Tornado. Metal crate removed.
Screenshot 2026-03-31 082732

Items moved here from Typhoon. Sentry laptop, claymores, and metal added. Perimeter claymores added.
Screenshot 2026-04-05 012951

Pipe network, dirt decals and other items added across the map
Screenshot 2026-03-31 082602
Screenshot 2026-03-31 082710

NON-PARITY
scout cloak moved eastward outside of the hive, but still in a dangerous spot
image

turret positions updated for both LZs
old (both LZs had this configuration with only 6 turrets)
Screenshot 2026-04-03 095607

new where turrets cover entrances to the landing zones and match other maps better.
image
Screenshot 2026-04-05 013027

## Requirements

**Changelog**
:cl: CatAndHats
- add: Chance's Claim has recieved a passover: the pipe network has been added which is expansive stretching across the entire map, LZ1 (West LZ) North Comms spawn now has a shortcut to its north giving a new route to the west of A-block, survivor loot has been changed at both dropships and the scout cloak has been moved eastward out of the hive to the reactor entrance, the map is now grimier and has recieved lots of missing prop items, and the landing zones now have much better gauss turret coverage.
- add: Added pre-armed CLF and UNMC IFF claymores for mapping/event purposes.


Thereisabear:
Would this not be the best time to move the scout stuff out of the hive so they don't have to dive through resin ?

Drazamuffin:
> Would this not be the best time to move the scout stuff out of the hive so they don't have to dive through resin ?

I mean why though?
The armor is at least hypothetically intended to be hard to get due to CM13 considering the scout's cloak to be their main weapon/tool, not their rifle. 
The PR that moved the items between the dropships in parity even added sticky resin in that area of the hive specifically to make it harder for scouts to get to their armor before drones unlock

AverageNotDoingAnythingEnjoyer:
Generally speaking non-parity changes are better off being done in a separate PR

QuietlyWhisper:
Forewarning that parts of this apply to all map issues, not just this one

> I mean why though?

Generally, complaints of survivors going through resin to get said loot resulting in xenonid deaths- we're not going to remove the items obviously, and nor should survivors be disallowed from obtaining the items- we've already told said xenonids to avoid loot if they do not want to get into combat to no avail- running out of options to resolve the problem, and making xenonids invincible to the consequences of being nearby/camping/waiting to melt loot or ambush survivors is not a valid solution, nor is relying on moderation to make up for game design flaws.

but more or less the item being in a resin area or not is more or less redundant, it doesn't make it any harder to get, it just puts xenonids at risk and is overall just a community toxicity issue in that neither side is in the wrong but both are trying to rush to these kinds of items (or trying to prevent them from being obtained) at all costs

If it is such a game-breaking balance difference to have this item, that it needs to be shoved into resin areas, it's probably not good for the game and we may need to have a longer discussion about survivor loot items at some point

> Generally speaking non-parity changes are better off being done in a separate PR

Normally yeah, but it would mean being entirely unable to make the PR until this one is merged- if the approval to merge the two changes is given in this case (I'm fine with it but smug is also required to give an opinion) I think this is a fine exception to make. I guess they could fork this PR as well, but if any changes are made to this PR it will also conflict on the second PR

kleinerstation13 (commented):
since you are doing a passover on chance's i'm gonna add some stuff on that would also need to be added/fixed

Image
theres a portable dialysis machine in cargo and i think we have it added now

Image

this should be one 50 plasteel stack instead of 2, SE of cargo

Image

the black market crates are supposed to have a different sprite than the regular BM crate (I.E. the black ones)

Image

Image

powerloader pamphlet NW cargo

Image

some pouches at east LZ (general filled with M63 AP? large shell filled with slugs, and flamer tank pouch

Image

Image

left side of tornado is supposed to have container and no shutters, same w/ right side

Image

there's a terminal prop that you can rename and use and a flight recorder prop you can use, you can look at my SPP bar crashlanding PR if you want to know the entity ID's

CatAndHats:
Converting to draft just to make sure it doesn't get merged early, won't be able to finish it with the extra stuff for a few days.

CatAndHats:
> since you are doing a passover on chance's i'm gonna add some stuff on that would also need to be added/fixed

All of this has been addressed apart from the crate sprites, those can be fixed in another PR.

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Explosives/claymore_mine.yml
```diff
@@ -119,6 +119,44 @@
     - range: 5
       falloff: 35
 
+# Pre-armed claymores for mapping
+
+- type: entity
```
PursuitInAshes, line 124: Make this abstract to serve as a parent, or make its suffix clear this won't have IFF

Resources/Prototypes/_RMC14/Catalog/Fills/Lockers/misc.yml
```diff
@@ -140,3 +140,17 @@
...
+  suffix: Filled, Secure
+  components:
+  - type: StorageFill
+    contents:
+    - id: RMCHandsBlack
+    - id: RMCShoesJackboots
+    - id: RMCFlashlightLantern
+    - id: RMCShovel
+    - id: RMCPickaxe
+    # toadd: meson goggles, analyzer, orebag
```
PursuitInAshes:
```suggestion
    # TODO: Add Meson Goggles, Analyzer, Orebag
```

PursuitInAshes (changes requested):
Ty for the PR!

Turret coverage on both LZs seems fine, we won't know definitively until someone inevitably tries to find a gap as a larva but w/e

The new Scout Gear route is about 10~ seconds shorter for reference, probably closer to 12-14~. While I'm personally fine with this (because if the gear is too strong thats a greater issue than just positioning) I'm mentioning it just so you can consider making it harder or the like, just want the decision to be well informed. If you want, wait for the Hosts to ask you to adjust it if the time different worries them in any meaningful way.

The current route takes about 45-50 seconds, I was able to do this first try in about 37 and I made a few mistakes. So I'm sure there is a larger difference there.

No issues on the pipe net.

Some small yaml cleanup and then this will have map and code approval from me.

kleinerstation13:
Cloak route is probably fine in my opinion. The main choice of doing cloak route is that it prevents you from doing other loot routes and locks you into running back as soon as you've got the cloak. Moving it east doesn't change that, you don't get any extra opportunities to loot stuff while you go for cloak. While the route is indeed faster, the way back to the hold is slower, so it balances out.

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Explosives/claymore_mine.yml
CatAndHats: added suffix

PursuitInAshes (approved):
Cloak spot is fine timing wise, I'm a tiny bit worried that it will be too close and we'll just end up with the current issue again, but that might just be something we'll need to see. 

If a host raises concern about it, just move it inside the first shutter (Where the other armor is and the drum) and that should be fine. The opportunity cost of getting the cloak compared to going to assist with other loot runs evens out its value for the most part.

CatAndHats:
Ty for review

> Cloak spot is fine timing wise, I'm a tiny bit worried that it will be too close and we'll just end up with the current issue again, but that might just be something we'll need to see.

Main thing is that it is now no longer behind what will eventually become indestructible resin, if any of the xenos decide to poke their head out to attack the scout then that is them leaving their protected area and whatever happens is on them, good or bad.

yagwog:
> Main thing is that it is now no longer behind what will eventually become indestructible resin, if any of the xenos decide to poke their head out to attack the scout then that is them leaving their protected area and whatever happens is on them, good or bad.

I think it'll be fine, doesn't mean if there is some issues that a little quick map hotfix can't change anyway, it solves the majority of problems.

## This PR looks fine to ship from my perspective.

DrSmugleaf (approved):

PR #9582, opened 2026-03-31
Tweak: Platform health

### LINKED ISSUE #9482: Platforms have wildly inconsistent health levels.
## Description

Currently the basic 'Platform' takes 5 drone hits, which is a bit tedious but fair enough.
Image

Corners are a bit bigger and take 7, still completely reasonable.
Image

Things get weirder with these microscopic corner pieces also having a full 7 hits, despite ostensibly existing mostly to make the visuals line up better.
Image
This alone means that nearly identical arrangements can take anywhere from 7 to 17 hits to fully remove, depending on which pieces the mapper used to construct it.

Then most strangely, this U piece is completely indestructible??
Image

**Reproduction**
1. Try to prepare any area for xeno build that that has platforms in it.
2. Spend minutes of the build phase attacking tiny ledges.
3. Suffer.

**Additional context**
From what I've seen in other discussions, apparently platforms are half placeholder/meant to indicate a change in elevation. But with them being tediously destructible they're currently more in a weird middle ground of just being weirdly tanky fences.

Ideally it should be standardized one way or the other eventually? If the plan is to keep them breakable, they should hopefully have more consistent health/time scaling. And if the plan is to not be destructible at all, better support for moving past them without fiddly clicking would probably be required.

IceNoobXD:
## About the PR
Platforms are now a bit weaker because breaking them was really tedious.

## Why / Balance
resolves #9482

> Try to prepare any area for xeno build that that has platforms in it.
Spend minutes of the build phase attacking tiny ledges.
Suffer.

grafik

**This is a common spot for xenos to hold**
a drone attacks around once per second so with the old health pool it would take a single drone roughly 365 hits or 6 mins to clear this area
with the new health pool it would "only" take 230 hits or 4 mins
grafik

## Technical details
minor .yml changes

## Media

---
## Requirements

**Changelog**

:cl:

- tweak: all platforms types have now less health and round platforms are now breakable!


BadaBoomie:
would it also be reasonable to make them meltable?

IceNoobXD:
> would it also be reasonable to make them meltable?

i dont think it would make a lot of sense because the broken platform is like a ramp and it needs the broken bits as support,
so if you melt it you are basically deleting the parts that support the ramp or that is how I imagine it in a way
grafik

BarteG44:
why were the small outer corners stronger than full straight pieces anyhow

GitHubUser53123:
ideally autovaulting should be a thing so that platforms can be leaped over automatically

DrSmugleaf (approved):
Eventually these should not block movement (maybe slow you down very very slightly) nor construction in general, but for now this is fine

PR #9623, opened 2026-04-07
add: vesg flamer nozzle

Ian321:
## About the PR

This PR implements the VESG flamer nozzle.

## Why / Balance

parity

## Technical details

Shoot a projectile that spawns the corresponding flame.

## Media

https://github.com/user-attachments/assets/3c44942b-c95a-420d-9bbe-63aa24c6ae48

## Requirements

**Changelog**

Special thanks to @Magicocity1 and their attempt at #8079.

:cl:
- add: The VESG flamer nozzle can be attached to an incinerator and allows it to shoot a ball of fuel at a target.
- fix: Fires set by sentries no longer stack.


PursuitInAshes:
Closes #8079 if merged first.

Magicocity1:
Wow, I was about to fix my vesg PR this weekend, but it seems that this is more complete.

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Flamethrowers/m34_flamer.yml
```diff
@@ -125,6 +126,25 @@
...
+        whitelist:
+          tags:
+          - RMCAttachmentRailFlashlight
+          - RMCAttachmentMagneticHarness
+      rmc-aslot-underbarrel:
+        startingAttachable: RMCAttachmentUnderbarrelExtinguisher
+        whitelist:
+          tags:
+          - RMCAttachmentExtinguisher
+          - RMCAttachmentFlamerNozzle
```
Arendian, line 146: You can add the flamer nozzle tag to the whitelist of the "RMCWeaponFlamerExt" prototype and delete this one, unless there is supposed to be a flamer that can't take the nozzle but can take an extinguisher.

Content.Shared/_RMC14/Weapons/Ranged/Flamer/RMCFlamerNozzleComponent.cs
```diff
@@ -0,0 +1,20 @@
...
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(SharedRMCFlamerSystem))]
+public sealed partial class RMCFlamerNozzleComponent : Component, IShootable
+{
+    [DataField, AutoNetworkedField]
+    public FixedPoint2 CostPer = FixedPoint2.New(3);
+
+    [DataField, AutoNetworkedField]
+    public float MaxRange = 6;
```
Arendian: I think it has 5 tiles range in CM, you can use the following helper to convert it to a circle radius.

```suggestion
    public float MaxRange = RMCMathExtensions.CircleAreaFromSquareAbilityRange(5);
```

Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/under_attachments.yml
```diff
@@ -322,14 +322,55 @@
...
+    offset: -0.125, 0.245
+  - type: AttachableToggleable
+    userOnly: true
+    attachedOnly: true
+    supercedeHolder: true
+    wieldedUseOnly: true
+    activatePopupText: attachable-popup-switch-to-generic
+    deactivatePopupText: attachable-popup-switch-from-generic
+    activateSound:
+      path: /Audio/_RMC14/Attachable/gun_u7_activate.ogg # TODO RMC14
```
Arendian: Please describe what is to be done. I assume it's making it use a different sound but it's still nice to have it described.

Content.Shared/_RMC14/Projectiles/RMCProjectileSystem.cs
```diff
@@ -227,11 +231,26 @@ ent.Comp.Origin is { } origin &&
...
         _hive.SetSameHive(ent.Owner, spawn);
+        if (HasComp<TileFireComponent>(spawn))
+        {
+            var fires = _rmcMap.GetAnchoredEntitiesEnumerator<TileFireComponent>(coordinates);
+            while (fires.MoveNext(out var fire))
+            {
+                if (fire != spawn)
+                    QueueDel(fire);
+            }
+        }
```
Arendian, line 258: Parity behavior is to just not spawn a fire if there already is one on the tile. I think it would probably be fine to "overwrite" an existing fire if they are of a "lower rank" .

Content.Shared/_RMC14/Weapons/Ranged/Flamer/SharedRMCFlamerSystem.cs
```diff
@@ -599,6 +606,74 @@ public void OnBroilerUniqueActionExamine(Entity<RMCCanUseBroilerComponent> ent,
...
+
+        var spawnOnTerminateComp = EnsureComp<SpawnOnTerminateComponent>(ball);
+        _projectile.SetSpawn((ball, spawnOnTerminateComp), reagent.FireEntity);
+
+        var maxRange = nozzle.Comp.MaxRange;
+        if (fromCoordinates.TryDistance(EntityManager, toCoordinates, out var distance) && distance < maxRange)
+        {
+            maxRange = distance;
+        }
+        _projectile.SetMaxRange(ball, maxRange - 1);
```
Arendian: You can remove this -1, I'll make sure to fix that being needed in a different PR.

```suggestion
        _projectile.SetMaxRange(ball, maxRange );
```

Arendian (changes requested):
Looks good. Just a few small things to address.

Content.Shared/_RMC14/Weapons/Ranged/Flamer/RMCFlamerNozzleComponent.cs
Ian321: https://github.com/cmss13-devs/cmss13/blob/master/code/modules/projectiles/gun_attachables.dm#L3728

Content.Shared/_RMC14/Projectiles/RMCProjectileSystem.cs
Ian321: Deleting the old flame is currently the same behavior as when using the flamer normally. I would like to keep them the same.

Content.Shared/_RMC14/Weapons/Ranged/Flamer/RMCFlamerNozzleComponent.cs
Arendian: The actual in game range is 5, if I aim past 5 tiles it stops at 5. If I set range to 1 it creates a fire tile under my feet.
Also fun fact about the DM code, I found out the nozzle range value actually isn't used, the range value on the spawned projectile is (which is also 6, but apparently 5 in practice).

Content.Shared/_RMC14/Projectiles/RMCProjectileSystem.cs
Arendian: Having consistent behavior does sound like the better option.

Arendian (approved):
Code and YAMl looks good to me.
#9802 should be merged together with this.

Content.Shared/_RMC14/Weapons/Ranged/Flamer/RMCFlamerNozzleComponent.cs
```diff
@@ -0,0 +1,21 @@
...
+public sealed partial class RMCFlamerNozzleComponent : Component, IShootable
+{
+    [DataField, AutoNetworkedField]
+    public FixedPoint2 CostPer = FixedPoint2.New(3);
+
+    [DataField, AutoNetworkedField]
+    public float MaxRange = RMCMathExtensions.CircleAreaFromSquareAbilityRange(5);
+
+    [DataField, AutoNetworkedField]
+    public EntProtoId Projectile = "RMCBulletSentryFireProjectile";
```
Arendian: Using RMCBulletSentryFireProjectile can sometimes cause two fire tiles to be spawned. ~~Remove `abstract: true` from `RMCBaseBulletSentryFireProjectile` and add `type: DeleteOnTrigger` so you can use it instead.~~(I've done this in #9802, only the prototype on the nozzle component needs to be changed)

```suggestion
    public EntProtoId Projectile = "RMCBaseBulletSentryFireProjectile";
```

Arendian (dismissed):

Content.Shared/_RMC14/Weapons/Ranged/Flamer/RMCFlamerNozzleComponent.cs
Ian321: Doesn't that mean that fire sentries also sometimes spawn multiple fires?

Arendian: It's not the case for sentries, the duplication comes from the `RMCBulletSentryFireProjectile` using the `DirectionalTileFireOnTriggerComponent` to spawn a fire, the nozzle then also adds the `SpawnOnTerminateComponent` to the projectile to spawn a fire based on the reagent that was in the tank. The only difference between the base sentry fire projectile and the sentry fire projectile is that the second one spawns a tile fire while the first one doesn't, so it shouldn't affect the nozzle's behavior in any way.

Ian321: Ah, I missed that those entities got changed with the WP PR. They used to all use `SpawnOnTerminateComponent`.

Ian321: I endet up having to create a new entity for the fire ball.

Ian321:
> Code and YAMl looks good to me. #9802 should be merged together with this.

#9802 has to be merged fist. After that this PR needs some adjustments:

- removing the `RMCFireProjectileComponent` 
- setting `SpawnOnTerminateComponent.SpawnOffset = 0.5`
- removing the `-1` when setting the max distance

Arendian (approved):

Arendian:
> > Code and YAMl looks good to me. #9802 should be merged together with this.
> 
> #9802 has to be merged fist. After that this PR needs some adjustments:
> 
> * removing the `RMCFireProjectileComponent`
> * setting `SpawnOnTerminateComponent.SpawnOffset = 0.5`
> * removing the `-1` when setting the max distance

#9802 Has been merged, so after pulling master you can apply the required changes.

Arendian (approved):
Requested changes have been applied

Vixting (approved):

PR #9625, opened 2026-04-08
Fix map insert spawning

CatAndHats:
## About the PR
Fixed it so that non-scenario inserts can spawn when a scenario is selected

## Why / Balance
Bugfixing

Currently if a scenario rolls no standard inserts will roll, this most evidently affects Solaris right now where, due to the survivor spawn together scenarios, none of the 30ish standard inserts are spawning whatsoever. This is a bug that has always been in the game, but was never noticed before now since Scenarios were quite rare up until recently.

## Technical details
added an additional check to the insert spawning code to allow for non-scenario inserts to always be able to spawn.

## Media
I tested it 30 times trust me.

## Requirements

**Changelog**
:cl: CatAndHats
- fix: Fixed an error where map inserts wouldn't spawn when a survivor scenario would roll, Solaris is saved

### DIFF
--- a/Content.Server/_RMC14/MapInsert/MapInsertSystem.cs
+++ b/Content.Server/_RMC14/MapInsert/MapInsertSystem.cs
@@ -80,7 +80,8 @@ public sealed class MapInsertSystem : EntitySystem
         {
             cumulativeProbability += variation.Probability;
             if (!forceSpawn &&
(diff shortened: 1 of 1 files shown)


kleinerstation13:
This should be high priority, currently all random inserts other than scenarios on solaris are completely broken and so will Trijent as it also has spawn togethers

DrSmugleaf (approved):

PR #9646, opened 2026-04-11
Tweaks the job rolling "Priority" system to only apply to queen and XO

Daylight2:
## About the PR

When selecting jobs at roundstart, only Queen and XO jobs bypass the normal player-set job preference system. 

## Why / Balance

Currently, we have a whole lot of jobs that bypass player preferences. In order of priority, those jobs are:
Queen > Xeno > CO > XO, CMO, CE, SO, CMP, MW, MP, LO, ASO > Everyone else

I imagine that the intent of this is to make sure "high impact" jobs are filled before other jobs. However, in practice, this leads to the opposite effect. If a player wants to, for example, try rolling for WS, then they NEED to have all of the jobs selected above to be set to "never", or they'll be put into them before they're even considered for WS. This leads to players simply not setting those jobs as "backups", since it sabotages their chances of getting the job they actually want. Many players currently latejoin into xeno, because having xeno on "low" is a 100% guarantee that you will always roll xeno. 

This PR will reduce the "priority" roles to just Queen > XO, as those are the only roles needed for the round to function (and they're high demand anyhow). This will also cut down on confusion, as the only way to know what roles were "priority" before was to codedive, leading to misinformation that roles like SL and survivor were also "priority", discouraging players from having them as backup jobs. 

## Requirements

**Changelog**

:cl:
- tweak: Xeno, CO, CMO, CE, SO, CMP, MW, LO, ASO jobs no longer roll before other roles. Queen and XO retain priority.

### DIFF
--- a/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Spawning.cs
+++ b/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Spawning.cs
@@ -180,6 +180,30 @@ public sealed partial class CMDistressSignalRuleSystem
             return true;
         }
 
+        bool HasHigherPriorityJob(HumanoidCharacterProfile profile, ProtoId<JobPrototype> role, JobPriority priority)
+        {
+            if (!_prototypes.TryIndex(role, out var job))
+                return false;
+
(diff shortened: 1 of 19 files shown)


Daylight2:
Oh whoops, forgot.

PR #9648, opened 2026-04-11
Tweaks the job rolling "Priority" system to only apply to queen, xeno and XO

Daylight2:
## About the PR

When selecting jobs at roundstart, only Queen, xeno and XO jobs bypass the normal player-set job preference system. 

## Why / Balance

Currently, we have a whole lot of jobs that bypass player preferences. In order of priority, those jobs are:
Queen > Xeno > CO > XO, CMO, CE, SO, CMP, MW, MP, LO, ASO > Everyone else

I imagine that the intent of this is to make sure "high impact" jobs are filled before other jobs. However, in practice, this leads to the opposite effect. If a player wants to, for example, try rolling for WS, then they NEED to have all of the jobs listed above to be set to "never", or they'll be put into them before they're even considered for WS. This leads to players simply not setting those jobs as "backups", since it sabotages their chances of getting the job they actually want. 

This PR will reduce the "priority" roles to just Queen > Xeno > XO, as those are the only roles needed for the round to function (and they're high demand anyhow). This will also cut down on confusion, as the only way to know what roles were "priority" before was to codedive, leading to misinformation that roles like SL and survivor were also "priority", discouraging players from having them as backup jobs. I wanted to also remove normal xeno from this list, but they're handled differently by the jobs system and I couldn't get it to work without issues.

## Technical details

The way the job system works is that xeno roles are assigned in CMDistressSignalRuleSystem.Spawning.cs before marine roles are even considered. I tried to get normal xenoes to be of equal weight to other jobs, but the system is complicated enough that I couldn't figure it out without bugs, so scrapped that idea. Others are welcome to give it a shot. 

## Requirements

**Changelog**

:cl:
- tweak: CO, CMO, CE, SO, CMP, MW, MP, LO, ASO jobs no longer roll before other roles. Queen, xeno and XO retain priority.

### DIFF
--- a/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Spawning.cs
+++ b/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Spawning.cs
@@ -180,6 +180,30 @@ public sealed partial class CMDistressSignalRuleSystem
             return true;
         }
 
+        bool HasHigherPriorityJob(HumanoidCharacterProfile profile, ProtoId<JobPrototype> role, JobPriority priority)
+        {
+            if (!_prototypes.TryIndex(role, out var job))
+                return false;
+
+            foreach (var (otherId, otherPriority) in profile.JobPriorities)
+            {
+                if (otherId == role ||
+                    otherPriority <= JobPriority.Never ||
+                    !_prototypes.TryIndex(otherId, out var otherJob))
+                {
+                    continue;
+                }
+
+                if (otherJob.Weight > job.Weight ||
+                    otherJob.Weight == job.Weight && otherPriority > priority)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
         NetUserId? SpawnXeno(List<NetUserId> list, EntProtoId ent, bool doBurst = false)
         {
             var playerId = _random.PickAndTake(list);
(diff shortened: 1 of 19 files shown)


BarteG44:
Could the UI possibly be adjusted to reflect the mechanic? like replacing the never-high scale with a simple yes/no option.
Not sure about 14 but that's been done for ss13 servers a lot before

Daylight2:
> Could the UI possibly be adjusted to reflect the mechanic? like replacing the never-high scale with a simple yes/no option. Not sure about 14 but that's been done for ss13 servers a lot before

A "yes/no" option on the never-high scale seems confusing, given that all of these jobs still have limited slots. Plus, I imagine people who really enjoy playing queen/xo would be quite upset if less "committed" players could have those roles on the equivalent of high whilst still highrolling ws/sg/gp/whatever.

chavonadelal:
This may affect the percentage of rounds without command

rafapecogames:
i agree with this but imo you should change the priority to be like

Queen>CO>XO>Everyone else

reasoning: basically what you said. however youre excluding xeno which just doesn't make sense to me. a lot of xenos are WS rollers that failed, so just let xeno be the same priority as a fall back instead of having 3409284023 burrowed larva or people having to join as larva

not sure on survs though, because although you havent commented, survs have a priority higher than marines (not sure if higher than xenos), so maybe they should keep it? idk tbh

CatAndHats (changes requested):
This touches a lot of the non-RMC roles, which I don't think is necessary. Even if it is, it would require a comment stating that an RMC change was made.

I don't know what to think about this change myself, so I'm not going to approve it, but that code change is still required if it is wanted.

PR #9651, opened 2026-04-11
Add some round stats to the end of round screen

Arendian:
## About the PR

Certain round stats are now tracked and shown in the end of round screen, admin spawned entities don't count towards most stats. The tracker gets reset every round.
## Why / Balance

The ones currently tracked are in the screenshot, not all of them are parity.

## Media

image

### The stats listed in the image below also track per-player and show the player with the highest amount:
image

## Requirements

**Changelog**

:cl: Dygon
- tweak: The round end screen now shows a few round related statistics.


IceNoobXD:
If possible could you maybe add total techpoints gained?

CatAndHats:
> If possible could you maybe add total techpoints gained?

Gained and Spent would be interesting to see

BarteG44:
Parity also has the highscores for singular people like most damage taken and most people revived. would that be in scope?

Arendian:
> > If possible could you maybe add total techpoints gained?
> 
> Gained and Spent would be interesting to see

I've added this now, isn't it already visible through the intel screen though?



> Parity also has the highscores for singular people like most damage taken and most people revived. would that be in scope?

I still need to take a look at the stats tracked in CM, damage received and healed to others is now also tracked. Some stats now also track per-player.

BarteG44:
image
image
image
image
image
image
image

First one is in the chat, rest is in the stats window

Arendian:
> First one is in the chat, rest is in the stats window

Thanks for looking it up for me, that's quite a lot of stats! I might keep it to just adding the general stats tab for now, the other tabs need some changes to how the end screen's UI is filled.

BarteG44:
honestly i think the stats from the "fun facts" section are what i'd personally like to see most. So again, the highest stats in a round for a player

PursuitInAshes (changes requested):
We'll need to go over each of these and trim out the ones that might cause issues regardless of parity or not, whenever you are ready to come back to this between all the other stuff. The hope is to have these be only things that 'encourage good gameplay behaviors' regardless of when they happen in the round.

PR #9658, opened 2026-04-13
xeno doors are now held open

ExponentialWizard:
## About the PR
Xeno doors are now held open for 5 seconds whenever they can't close due to something blocking them

## Why / Balance
Its incredibly annoying when a xeno is holding a door open then leaves at the last second, slamming the door in your face. Theres no way to guarantee 5 seconds unless you close the door and re-open it. Its technically a minor xeno buff but I consider it a QOL change

## Technical details
Added CanHoldOpen flag to doorComponent and set it to true in the xeno doors. Added a HoldOpenDelay variable to track when the door should close. Added a HoldOpenTime variable to set the delay when held open.

The door state update code now resets the HoldOpenDelay when something is blocking it from closing, and prevents the door from closing until HoldOpenDelay is reached.

The airlock system has a check for CanHoldOpen so that it can set the door to update every second

## Media
(https://youtu.be/kcoCVY8_11E)
The right side is with the PR implemented

## Requirements

**Changelog**
:cl:
- tweak: xeno doors are now held open for 5 seconds anytime they are left empty

### DIFF
--- a/Content.Shared/Doors/Components/DoorComponent.cs
+++ b/Content.Shared/Doors/Components/DoorComponent.cs
@@ -44,13 +44,13 @@ public sealed partial class DoorComponent : Component
 
     //RMC14
     /// <summary>
-    /// Closing time when an open door is prevented from closing
+    /// Closing time when an open door is held open
     /// </summary>
     [AutoNetworkedField, ViewVariables]
     public TimeSpan HoldOpenTime;
 
     /// <summary>
-    /// Closing time when an open door is prevented from closing
+    /// Whether or not the door closing time can be delayed when held open
     /// </summary>
     [DataField, ViewVariables(VVAccess.ReadWrite)]
     public bool CanHoldOpen = false;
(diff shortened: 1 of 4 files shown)


PursuitInAshes:
Thanks for the PR!
Consensus for now is to close this and at best revisit it post parity (though unlikely), its important for xeno defenses/resiliance to be able to close doors to prevent marines from simply rushing in with ease when a newer xeno (but really any) falls back into the walls. It is already easy enough for marines to pursue if they meaningfully intend to (M54C W/ Solid Stock + Bayonet), this would make it far to easy regardless

ExponentialWizard:
> Thanks for the PR! Consensus for now is to close this and at best revisit it post parity (though unlikely), its important for xeno defenses/resiliance to be able to close doors to prevent marines from simply rushing in with ease when a newer xeno (but really any) falls back into the walls. It is already easy enough for marines to pursue if they meaningfully intend to (M54C W/ Solid Stock + Bayonet), this would make it far to easy regardless

I'd like to note that in the event of a xeno fleeing through a closed door and leaving the door open behind them, the door close timing is very nearly the same, only increased by less than a second depending on update ticks. This only affects doors that are already open and have a xeno standing in them, which is more common when there are many xenos in one place. 

Its mostly the inconsistency of door closing time that bothers me, but its understandable why you're making this post parity.

PR #9665, opened 2026-04-14
Allow Alt Click Unbuckle.

NotActuallyMarty:
## About the PR

Lets you alt-click buckled ents to unbuckle them regardless of held item.

## Why / Balance

Its annoying to do so via right click -> verb. If we can already do this  via that, I see no reason to not have this as a quick alt-click option

## Technical details

Adds a verb when alt interacting with an entity with Bucklecomponent. You will never see it, but AltInteract handles verbs on alt interact.

Touches upstream code files, but it is just a handler for it, can be moved to _RMC14 but for convenience and clarity i say keep it here. Edits commented.

## Media

Media is compressed to shit because 10 MB github upload limits SMH L bozo i guess. Looks visible on my end.

https://github.com/user-attachments/assets/58927860-5b94-4e25-91b4-8ff54dfed3ad

## Requirements

**Changelog**
:cl:
- fix: You can now unbuckle people by alt-clicking them.

### DIFF
--- a/Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
+++ b/Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
@@ -20,7 +20,6 @@ public abstract partial class SharedBuckleSystem
 
         SubscribeLocalEvent<BuckleComponent, InteractHandEvent>(OnBuckleInteractHand, before: [typeof(InteractionPopupSystem)]);
         SubscribeLocalEvent<BuckleComponent, GetVerbsEvent<InteractionVerb>>(AddUnbuckleVerb);
-        SubscribeLocalEvent<BuckleComponent, GetVerbsEvent<AlternativeVerb>>(AddAltUnbuckleVerb); // RMC 14 Edit.
     }
 
     private void OnCanDropTarget(EntityUid uid, StrapComponent component, ref CanDropTargetEvent args)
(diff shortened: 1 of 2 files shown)


Content.Shared/Buckle/SharedBuckleSystem.Interaction.cs
```diff
@@ -242,4 +243,28 @@ private void AddUnbuckleVerb(EntityUid uid, BuckleComponent component, GetVerbsE
...
+            Act = () => TryUnbuckle(uid, args.User, buckleComp: component),
+            Text = Loc.GetString("verb-categories-unbuckle"),
+            Icon = new SpriteSpecifier.Texture(new("/Textures/Interface/VerbIcons/unbuckle.svg.192dpi.png"))
+        };
+
+        args.Verbs.Add(verb);
+
+    }
+    // RMC 14 Edit end.
+
```
Arendian: You can move this and the subscription to the RMCBuckleSystem. You won't have to touch the upstream file at all that way. It's clearer and easier for pulling upstream if as much RMC code as possible is in the RMC related files.

Arendian (changes requested):
Code looks good, but it needs to be moved to a different file.

NotActuallyMarty:
fair nuff i was just lazy

Arendian (approved):

Vixting (approved):

PR #9669, opened 2026-04-14
Many xenonid abilities will rewind the target to where they got hit

Vixting:
## About the PR

Many xenonid abilities will rewind the target to where they got hit, this prevents the issue for many xenonids where delay between the stun being applied causes the target to be way further away then is intended. This issue is most apparent with runner & lurker.

Lag compensation will now also take into account the clients coordinate sample, this should help with lag compensation. 

## Technical details

its rewind time
### before

https://github.com/user-attachments/assets/0ab379c8-3643-42d4-ae19-7261a33d56ac

### after 

https://github.com/user-attachments/assets/f6dd1049-901b-488e-b889-b9d2f2969d02

## Requirements

**Changelog**

:cl: Vixting
- add: Many xenonid abilities that stun the target will now rewind the target to the point where the stun was applied, instead of it being applied at their current position.

### DIFF
--- a/Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
+++ b/Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
@@ -1,4 +1,5 @@
-﻿using Content.Shared._RMC14.CCVar;
+using System.Numerics;
+using Content.Shared._RMC14.CCVar;
 using Content.Shared.Coordinates;
 using Robust.Shared.Configuration;
 using Robust.Shared.Map;
@@ -20,6 +21,7 @@ public abstract class SharedRMCLagCompensationSystem : EntitySystem
     [Dependency] private readonly SharedTransformSystem _transform = default!;
 
     public float MarginTiles { get; private set; }
+    public float PredictionCoordinateDeviation { get; private set; }
 
     private EntityQuery<ActorComponent> _actorQuery;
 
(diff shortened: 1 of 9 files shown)


Arendian:
Altough it does look/play better for the player applying the crowd control, I don't think lag predicting displacement and crowd control will lead to smooth gameplay. Rewinding movement because of a stun/displacement means that in the video below, the runner would be rewinded and their leap would still be on cooldown for example. 
For the player being rewinded it looks like an even bigger rubberband  than in your example where the runner does the leap, because the rewinded player is also predicting their own movement.

https://github.com/user-attachments/assets/6cfd0f5d-84dd-4792-b739-3bd451920d8e



The video below is an example of what being rewinded by the server looks like to a player with 150 ping. This is only the client's own prediction being "undone".  If I understand it correctly, after this PR it would add another player's prediction on top of this up to a certain cap. Not for claymores of course, since those are serverside, but for stuns predicted by other players.


https://github.com/user-attachments/assets/ee2d4698-237c-4828-b7ff-9d79a97fa3a3



Below is what I think happens before and after this PR, with in my opinion, the current situation being the lesser of 2 evils. Someone please correct me if I am wrong about how this works though.

**Unrewinded stuns(Current):**

- Player A view: I leap at player B while they are at position (0,0). Sees Player B continue walking and getting stunned at (3,0).
- Server view: Player A leaps at (0,0) while Player B is at (3,0). Everyone sees Player A leap at (0,0) and Player B get stunned at (3,0).
- Player B view: Player A leaps at (0,0) while I am at (6,0). Gets teleported to (3,0) and stunned.

**Rewinded stuns(After this PR):**

- Player A view: I leap at player B while they are at position (0,0). Sees Player B continue walking, then get teleported from (3,0) to (0,0) and stunned.
- Server view: Player A leaps at (0,0) while Player B is at (3, 0). Everyone except Player B sees Player B get teleported from (3,0) to (0,0) and stunned
- Player B view: Player A leaps at (0,0) while I am at (6,0). Gets teleported from (6,0) to (0,0) and stunned.

TLDR: I think stuns and displacement locations should be fully server authoritative and not lag compensated. Movement that already happened on the server shouldn't be undone by a client.

HorseCoded:
Is there a list of the 'many' this would apply to? It seems like even if this is too awkward for regular stuns, it would still be incredibly useful for abilities like Tail Seize, where currently the ability fails half the time and applies no effect at all due to the lack of compensation. So this compensation could make that ability actually functional, and the downside(feeling like you've been teleported backwards) is covered up effectively by the ability's normal movement effects.

PR #9708, opened 2026-04-21
Royal Crown Marine cleanup and consistency changes

MoonlightHollow:
## About the PR

Updates various RCM factions(RCM, Paramarines, and Whiskey Troop) cleaning up some of their files and making them consistent with each other as there was some variance for little lore reasons. 

## Why / Balance

First set of changes was making all the icon consistent as these were all over the place before with some roles sharing the same name but not same icons and some were missing icons. My goal wasn't to deviate to much with the existing icons key points I wanted to achieve was keep the shield icon motif and made it so chevrons indicated command ranks. Order goes with chevrons is one strip is lowest then double/thick strip is middle command rank and double with a lower chevron mean highest command. Goal with that was to hopefully make a clear and easy to understand ranking icons that anyone at quick glance can understand. Also a big part of this was unifying tac map icons which before this were kinda dependent on what it had.

Second major change was the unification of ranks with paramarines, this was a hold over from the CM13 port but with new standards it was just combined with normal RCM ranks with a special rifleman rank of Paramarine. Also dropped the ERT MedTech rank down to Lance Corporal to remand consistent with rest of rankings.

PVE are primarily two with GPMG being renamed into the Smart Gun Operator and switch the gun to the ML66C 'Royal' variant as every faction uses their primary SGO name and gun, there is little reason for RCM to be different in this. Also renamed the Medical Assistant to Medical Technician as rest of the RCM uses this name and the original name comes more from their RMC origins which doesn't fit here. Important to note is I didn't touch the decals as that would more than likely require some remapping better leave that to separate thing if possible.

Final change is I decided while I was in the files to skill preset the RCM ERT roles.

## Technical details

Important thing there is plans to finally clean up the cm_job_icon and map_blip .rsi files as they have grown to massive sizes, I decided to split off the new icons and map icons into their own file paths. Even if this set up isn't how we want it in the end I do think worth keep so easier to change at a later date. With this moving to I renamed the files to better represent what they are i.e. rcm at the start.

## Media

image
image

## Requirements

**Changelog**

:cl:
- tweak: RCM Hud Icons and Tac Map Icons
- tweak: Made RCM lore consistent


MoonlightHollow:
Split this PR up into three smaller ones:

#9894
#9895
#9896

PR #9715, opened 2026-04-22
Re-enable the Bulwark Of The Hive lightning effect

Arendian:
## About the PR

Using the Bulwark Of The Hive ability now again draws a lightning beam between the king and every affected xenonid for the shield's duration.

## Why / Balance

King goes BZZZZZZT.

## Technical details

The effect is now drawn through the `RMCTetherOverlay`. Previously it spawned in hundreds of entities that updated their positions every tick, causing severe performance issues.

## Media

Since it's stretched out it doesn't 100% look like the lightning sprite, the convergence points are missing.
The duration of the lightning is shorter than in the video: 4 seconds.

https://github.com/user-attachments/assets/47478c18-cfe7-49dc-8ad2-d979385c72a8

## Requirements

**Changelog**

:cl: Dygon
- fix: The King's Bulwark Of The Hive ability creates lightning again.


DrSmugleaf (approved):

PAFFhassoocks:
this kinda doesnt look good

Arendian:
> this kinda doesnt look good

It's caused by the original sprite not being "stretchable" because of it's shape. To make it look better it would need a sprite that's more uniform(like the oppressor tail sprite),  or it needs to be drawn in a different way.

PR #9731, opened 2026-04-25
Tweak: Xeno eggs abort infection if you leave a 4x4 radius or no other hosts is in that area

IceNoobXD:
## About the PR

When you trigger the egg its affected radius will increase from a 3x3 to 4x4 so that its harder to outrun it in heavy armor. 
If the poor soul who activated leaves the 4x4 radius the egg will search for a new hosts in the 4x4 radius which isnt blocked by a wall.
If the poor soul who activated it re-enters the area in the just the right time they will get infected instead of somebody else
The 3x3 area around an egg is a confirmed hug zone, if you trigger the egg and don't leave that radius, infection is guaranteed. This is relevant in situations where an egg is placed near a wall and a player repeatedly walks in and out of its line of sight, as long as they stay within the 3x3 trigger radius after the initial trigger, they will be infected.
If no host is found it will instantly go back into the egg.
This will make playing around eggs a smoother Marine experience and less of a WTF moment

## Why / Balance

If you activate a Egg, you are guarantee to get infected unless you act fast enough and break it in a 2 second timeframe.
This leads to moments where you walk into a airlock, peek, walk out and still get infected. 
This only leads to more confusion and players instantly asking themself or others in LOOC what just happend or how this was possible. 
Eggs right now are also really punishing with how long they can stay off-weeds. You always fear to get basically round/combat removed by either being dragged off or going shipside for surgery when you peek a door or corners.
to note is that this will heavily weaken xenos like eggsac carrier and oppressors who heavily rely on this "feature" to secure captures, as they now need to make sure that the marine stays inside that radius.

## Technical details

added a few lines of .cs code

## Media

https://github.com/user-attachments/assets/80ddc067-f74e-4712-a2a8-ec8e9326c179

https://github.com/user-attachments/assets/74905a9b-9ace-4bc0-ba2a-f550511c574e

## Requirements

**Changelog**

:cl:
- tweak: Triggering a xeno egg increases its infection radius by one tile. If the host who triggered it leaves this radius, the egg will seek out a new target within range. If none is found, the hugger instantly retreats back into the egg

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Egg/XenoEggSystem.cs
@@ -86,6 +86,9 @@ public sealed class XenoEggSystem : EntitySystem
     private static readonly ProtoId<TagPrototype> AirlockTag = "Airlock";
     private static readonly ProtoId<TagPrototype> StructureTag = "Structure";
 
+    // Expanded infection range once egg is triggered (4x4 area)
+    private const float EggOpeningInfectRange = 2.0f;
+
     private EntityQuery<StepTriggerComponent> _stepTriggerQuery;
 
     public override void Initialize()
@@ -983,12 +986,48 @@ public sealed class XenoEggSystem : EntitySystem
 
                 if (egg.InfectTarget != null)
                 {
-                    if (!_interaction.InRangeUnobstructed(uid, egg.InfectTarget.Value))
+                    if (!_interaction.InRangeUnobstructed(uid, egg.InfectTarget.Value, EggOpeningInfectRange))
                     {
-                        // Target left the radius before the parasite could latch = abort infection
-                        egg.InfectTarget = null;
+                        // Original target left, try to find another valid target nearby
+                        EntityUid? newTarget = null;
+
+                        var infectableQuery = EntityQueryEnumerator<InfectableComponent>();
+                        while (infectableQuery.MoveNext(out var candidate, out var infectable))
+                        {
+                            if (infectable.BeingInfected || _mobState.IsDead(candidate) || HasComp<VictimInfectedComponent>(candidate))
+                                continue;
+
+                            if (!_interaction.InRangeUnobstructed(uid, candidate, EggOpeningInfectRange))
+                                continue;
+
+                            newTarget = candidate;
+                            break;
+                        }
+
+                        if (newTarget != null)
+                        {
+                            egg.InfectTarget = newTarget;
+                        }
+                        else
+                        {
+                            // No valid targets, hugger returns to egg
+                            egg.InfectTarget = null;
+                            if (egg.SpawnedCreature != null)
+                            {
+                                QueueDel(egg.SpawnedCreature.Value);
+                                egg.SpawnedCreature = null;
+                            }
+
+                            egg.OpenAt = null;
+                            SetEggState((uid, egg), XenoEggState.Grown);
+                            Dirty(uid, egg);
+                            continue;
+                        }
                     }
-                    else if (TryComp<XenoParasiteComponent>(egg.SpawnedCreature, out var para))
+
+                    if (egg.InfectTarget != null &&
+                        _interaction.InRangeUnobstructed(uid, egg.InfectTarget.Value, EggOpeningInfectRange) &&
+                        TryComp<XenoParasiteComponent>(egg.SpawnedCreature, out var para))
                     {
                         _parasite.Infect((egg.SpawnedCreature.Value, para), egg.InfectTarget.Value, force: true);
                         _stun.TryParalyze(egg.InfectTarget.Value, egg.KnockdownTime, true);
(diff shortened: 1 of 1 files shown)


kleinerstation13:
Yeah I feel like eggs shouldn't be 3x3 walking radiuses of instant round removal that cement your fate the second you see them, this is a good change.

Smudged:
This is more or less an issue with just us being pixel based. 
Considering on cm13 once you move onto a tile you get hugged because you've already entered the influence of a hugger egg, a change like this only awards marines for inching towards corners and then backing up, in essence this would just make eggs nearly useless and its a balance change.

It might not seem intuitive but this is how it works in cm13 iirc. You either commit to the corner to take out the egg or get hugged because you didn't.

IceNoobXD:
> You either commit to the corner to take out the egg or get hugged because you didn't.

it is nearly impossible to take out the egg with high ping.
And the only gun which can reliably take it out is the shotgun with buck.
As with a Rifle you already need to shot it or know where it exactly is before you walk into the radius because if you start shooting after you entered you wont do enough DPS to take it out
Eggs also layer below iirc every item and get shot over unless pointed at which leads to scenarios where the egg gets placed under a Req supply drop and becomes impossible to see, impossible to shot without a pixel perfect shot and impossible to remove the junk atop it because you need to get close to remove it. 
The only way you could remove this would be with something like nades or fire. But people shouldnt be forced to pack a flamer or many more nades to counter something so commen like a egg. (I know knife throwing is a option but if its placed on a crate the crate will take the hit and you only have one knife with you.)


grafik

MidZik:
Being able to run out of range of the egg _is_ parity. However, I see several issues with this implementation that I think make it unacceptable to merge as is.

1) As we use pixel movement instead of tile movement, the "escape" range should be larger than the "trigger" range. This is to prevent literal edge cases where moving a pixel toward the egg triggers it and then moving a pixel away escapes the hug.
2) The same issue exist for peeking around corners and would need to be fixed somehow. If a marine peeks a corner by a pixel, triggers an egg, and moves back by a pixel, they will not get hugged. There would need to be a secondary check that ensures the marine still gets hugged if they don't move far enough from the original trigger position.
3) In parity the parasite immediately gets back in the egg so that it can trigger again. In this version the parasite stays outside the egg, making the egg unavailable to hug others. This means egg triggers are abusable, where you can run toward the egg on purpose, run away, and then the egg is "disarmed" as the parasite is asleep after leaving the egg.
4) I think with this code, if multiple marines run into the egg, if the original target escapes, none of the other marines will be infected despite also being in range. This is very bad, nonsensical behavior.

BarteG44:
>This is more or less an issue with just us being pixel based.
>Considering on cm13 once you move onto a tile you get hugged because you've already entered the influence of a hugger >egg, a change like this only awards marines for inching towards corners and then backing up, in essence this would just >make eggs nearly useless and its a balance change.

ss13's egg balance is much much different because of how it's sightlines work. You don't walk into an egg because you couldn't see it behind a corner. The only way you walk into it is by making an actual mistake, being pushed in by other marines or being pulled into it by xenos.

-cm13
image

-rmc
image
the egg is in the same corner in both pics but in RMC, even being almost twice as close to the corner, you cannot even see it

Hopie-OHOL:
-1, this is a bad idea and is a significant balance shift. 

First of all, it’s not parity, quickly deploying eggs is how CM13 does it. 

Secondly, the marines have plenty of tools to take down walls that don’t involve them coming close to eggs, such as shotguns, mortars, grenades, CAS, OBs, fire, or even regular bullets. 

Thirdly, if someone DOES run into an egg, nine times out of ten their allies drag them to safety and they get surgery and no round removal occurs. The only times someone gets round removed from an unexpected egg is if they were both too far away from fellow marines for them to drag them to safety, and also close enough to xenos for the xenos to capitalize on the opportunity. That happens when a marine runs off solo or overextends, which is when they SHOULD be punished with a cap. 

Fourthly, if someone with high speed such as a vest user is given the option to walk up to an egg, decide they don’t want to be infected, and just walk away again with no consequences, they’re just going to walk away every time. This allowed vest users to peek corners and push into xeno territory with significantly less risk.

I play eggsac carrier pretty frequently. The vast majority of eggs are shot and killed. Of the few that successfully infect someone, the majority of those talls are dragged to safety. It is only a small minority that lead to a successful unexpected capture, and this change would remove even those few.

yagwog:
In CM you can also press a keybind that warns you if there are traps anywhere.

With pixel movement I do not feel the same reason is applicable to CM.

With the sight lines how it is, you can move from side to side a lot easier than CM to check out an angle.

But the issue to feedback for the player could be addressed by possibly making the animation a bit slower

CatAndHats (changes requested):
This is going to require internal discussions due to how LOS/movement and other differences in RMC compared to CM will affect this. An issue with your current implementation is in this video where moving a few pixels will cancel the infection, this is not the case in CM13 where it will still infect you even if you were a whole tile further north

https://github.com/user-attachments/assets/e84ffcac-19b7-4f56-8078-1f38297014f3

MidZik:
> 2. The same issue exist for peeking around corners and would need to be fixed somehow. If a marine peeks a corner by a pixel, triggers an egg, and moves back by a pixel, they will not get hugged. There would need to be a secondary check that ensures the marine still gets hugged if they don't move far enough from the original trigger position.

I see that my other comments were addressed to some extent, but this one wasn't. That's what the video above shows.

I also recommend making the "escape range" one tile instead of a half tile.

Kenkaitos:
Lets make egg even more useless than it is

PR #9258, opened 2026-02-14
Queen Eye Leashed and Queen Fixes

Thereisabear:
## About the PR

- **Queen eye is now leashed to Weeds +3 tiles.**
(_When no weeds in range, Eye Mode ends._ )
- Plant Weed and Expand Weed(Eye Mode) is now the "same" button.
- Expand weed now incurs a 7 second CD after creating a node via Queen Eye.
 (_Normal Weeding and making a weed tile unaffected._)
- Fixed being able to weed onto tiles barricaded tiles. 
(_still allows for expanding onto barricade if there is an adjacent weed tile "behind/beside" barricaded tile_)
- Fixed Ovi cost being slightly too high.

Fixes #7500 

## Why / Balance
Free flying whilst yes fun was not intended.
Queen has enough buttons 

## Technical details
- Registers an RMCSwappableAction System with the Upstream InstantTarget/WorldTarget comps.
No other upstream edited.
- Plant weed/ExpandWeed(Eye) builds upon the expand weed ability Smug made for Queen eye. 
_Less button is always good._
- Queen Eye weed check popup will only work in game, if you pop out the tactical map, the teleport will work but no visual popup will play if invalid destination. (couldn't use popupentity because if you tried to TP far away, it tries to do some wild _extreme_ range checks)
## Media

https://github.com/user-attachments/assets/6c6aae00-1713-408b-84a5-f5d775b0763c

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**
:cl:
- add: Queen Eye now leashed to edge of weeds + 3 tiles.
- add: Expand Weed now has a 7 second cooldown when planting a node.
- add: Queen Eye tactical map teleport will now check if there is weeds present.
- add: Queen's _Plant Weed_ now dynamically changes to _Expand Weed_ in Queen eye.
- fix: Queen Eye being able to be weed directly on barricades by Arendian.
- fix: Queen Ovi plasma cost being slightly too high.

PR #9735, opened 2026-04-26
Prevent expanding weeds through barricades

Arendian:
## About the PR

- The Queen's "Expand Weeds" action no longer works if the only available nearby weed tile is on the other side of a barricade.

## Why / Balance

Parity.

## Requirements

**Changelog**

:cl: Dygon
- fix: The Queen's "Expand Weeds" ability no longer expands weeds through barricades.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
@@ -10,6 +10,7 @@ using Content.Shared.Atmos;
 using Content.Shared.Coordinates;
 using Content.Shared.Damage;
 using Content.Shared.Maps;
+using Content.Shared.Physics;
 using Content.Shared.Tag;
 using Robust.Server.GameObjects;
 using Robust.Shared.Map.Components;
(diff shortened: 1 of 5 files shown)


yagwog:
[I think this PR tackles this](https://github.com/RMC-14/RMC-14/pull/9258)

GrigoriyGalintano:
I looked into this more and opened a smaller focused PR for the barricade/weeds part: #9817.

From what I can tell, this PR covers two separate things:
1. preventing Expand Weeds / weed spread from going through barricades;
2. adding a 7 second cooldown when Expand Weeds is used to place a weed node.

The first part overlaps with the focused PR. The difference is that the new PR keeps the fix limited to the spread rule itself:
- adds a weeds-specific `CanSpreadWeedsBetween(source, target)` helper;
- uses the same check for manual `Expand Weeds` and automatic weed spread;
- checks manual expand against the center of the target tile, not the clicked coordinate;
- leaves Queen Eye visibility and cooldown behavior unchanged.

The cooldown part looks like a separate parity/balance change and may fit better with #9258 or as its own PR.

So with the focused spread fix plus #9258 handling the Queen Eye/cooldown work, this PR may end up overlapping both. It may be cleaner to either split out the cooldown change or let the smaller barricade-spread PR cover the barricade part.

GrigoriyGalintano:
The main reason I think #9817 is cleaner is that it keeps the barricade behavior as a single spread-path rule, instead of mixing it with action cooldown changes. That makes it easier to review and less likely to affect unrelated Queen Eye / node placement behavior.

Arendian:
> The main reason I think #9817 is cleaner is that it keeps the barricade behavior as a single spread-path rule, instead of mixing it with action cooldown changes. That makes it easier to review and less likely to affect unrelated Queen Eye / node placement behavior.

I personally prefer the way it's implemented by this PR, which is now used in #9258.

PR #9743, opened 2026-04-27
Updates the action descriptions for scopes to be more descriptive

JustSomeGalSS13:
## About the PR
Updated the action descriptions (the text that appears when you hover over the button) for scopes. They now indicate what the button actually does, including whether or not you can move while scoped in.

## Why / Balance
Initially this was just going to change the description for the B8 because when I was new I was confused by it, thinking I had to be scoped in to use the IFF features. But then I noticed that the descriptions for most zooming-in actions was just a repeat of the description of the scope itself, which isn't very useful. So I wrote up some descriptions that actually indicate what hitting the button will do and put them in place. Also tweaked the description of the smart-scope to indicate that the "integrated IFF systems" allow it to shoot past friendlies since new players might not intuit that by reading its old description.

Tl;dr: Descriptions actually showing what the button does is helpful, especially for new players.

## Technical details
Pure YAML string changes

## Media
image

## Requirements

**Changelog**

:cl:
- tweak: Made the action buttons for scopes more descriptive of what they do.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/rail_attachments.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/rail_attachments.yml
@@ -76,7 +76,7 @@
       sprite: _RMC14/Objects/Weapons/Guns/Attachments/rail.rsi
       state: flashlight-on
     actionName: Toggle Rail Flashlight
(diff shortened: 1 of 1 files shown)


PursuitInAshes (approved):
We might have to trim them down later, but the extra info will probably help understanding for new players. TY for the PR

PR #9750, opened 2026-04-29
wall fixes

Vixting:
## About the PR

fix all walls becoming damaged girders when destroyed
add examine text to damaged girder
make breaking acid hold spawn damaged girder
fix acid holes not being blocked by objects
## Requirements

**Changelog**

:cl:
- add: Added examine text to damaged girders.
- fix: Fixed all walls becoming damaged girders when destroyed.
- fix: Fixed acid holes not being blocked by objects.
- tweak: Breaking acid holes now spawns a damaged girder.


MidZik:
I'm testing this right now, it seems acid doesn't do anything to walls. I'm not sure if it's because of this PR or the original addition of the acid hole feature.

Vixting:
Still making some fixes

MidZik:
Alright. I'll just list out some issues I've found just in case. Can't really comment on the code since most of it is in the other PR.

- Acid doesn't do anything to walls (boiler acid completes, wall still at 100% integrity)
- Attacks against walls start AND stop the doafter to break the hole. They should only start it, it's too easy to accidentally cancel it.
- The check that restricts which side you can enter from seems too strict. Image below.
image

- In CM13 objects on your side of the hole don't seem to block it. You can climb into a hole even if a barbed cade is blocking it on your side. This is fine as a non-parity fix.

- Objects which aren't oriented in the right direction are blocking the holes too. Also, although the table is parity behavior, I personally thing any climbable objects shouldn't be blocking the hole. If we are changing the above bullet point to be non-parity, I would appreciate if we make climbable objects not block here as well. Image below.
image

- Tables blocks your own side while you're climbing on them.
image

- A lot of client-side-only damaged girders were made. Trying to acid them crashed the server in debug configuration. Image below.
image

- Getting a lot of console errors client side regarding "predicting the queued deletion of a networked entity: wall/acid hole"

MidZik:
The above issues are mostly fixed. Although I still think the crawl side restriction is too strict, it is not a blocking issue.

Few more things I found after your recent fixes:
- In CM13, expanding a hole only requires that a xeno is Big. It does not care about claw sharpness at all. For example, crest-lowered and fortified defenders can expand holes. So can boilers. This means that if a xeno is too big to crawl through a hole, it is big enough to break it instead.
- This is hard to reproduce but I got an error on the client when destroying a damaged girder. `[ERRO] resolve: Can't resolve "Robust.Shared.GameObjects.TransformComponent" on entity 0!` Did not cause a crash and I did not notice any bad behavior from it, so it might not be a blocking issue.

Vixting:
should all be fixed now, hoepfully

Content.Shared/_RMC14/Xenonids/Acid/XenoAcidHoleSystem.cs
```diff
@@ -261,7 +286,13 @@ private void OnHoleInteractUsing(Entity<XenoAcidHoleComponent> hole, ref Interac
         if (_net.IsClient)
             return;
 
-        if (TryComp(args.Used, out NailgunComponent? _))
+        if (TryComp<WelderComponent>(args.Used, out _))
```
MidZik:
```suggestion
        if (HasComp<WelderComponent>(args.Used))
```

Content.Shared/_RMC14/Xenonids/Acid/XenoAcidHoleSystem.cs
```diff
@@ -261,7 +286,13 @@ private void OnHoleInteractUsing(Entity<XenoAcidHoleComponent> hole, ref Interac
...
+        {
+            _popup.PopupEntity(Loc.GetString("rmc-acid-hole-repair-requires-nailgun"), hole.Owner, args.User, PopupType.MediumCaution);
+            return;
+        }
+
+        if (TryComp<NailgunComponent>(args.Used, out _))
```
MidZik:
```suggestion
        if (HasComp<NailgunComponent>(args.Used))
```

MidZik (changes requested):
Found a couple nits in the code.

Also found a few more issues:

- If you acid a wall and then repair it with a nailgun, the wall health becomes unchangeable:
  - Integrity is stuck at 10%. Repairing it doesn't restore any health. You can spend an entire stack of plasteel and it will do nothing.
  - The wall can no longer be damaged. Guns do nothing. Xenos attacks do nothing.
  - As far as I can tell, nothing will make the wall take damage ever again. Acid will still make a new hole that can be interacted with normally, and a repaired (hole-less) immortal wall can still be deconstructed normally.
- Walls that are rotated display their holes in the wrong direction. Some maps have their walls rotated, so we are likely to see this situation at some point. This is a visual-only glitch. Video below.

https://github.com/user-attachments/assets/e9b746d4-ea93-43f2-98df-38dcc2442ae9

Vixting:
okay should be fixed now @MidZik if you can test would be appreciated (my current temp setup takes like 10 minutes to build xd)

MidZik:
Found more issues:

- Repairing walls damaged from acid seems to work and they can be damaged afterwards, BUT now they will take WAY less damage from hits. A ravager usually deals 5% damage per hit, but after repairing an acided wall a ravager only deals about 2% damage per hit.
- If you drop your nailgun while repairing a wall, the doafter still goes on and completes successfully.
- I'm not sure the intended behavior here, but you can only repair walls with plasteel. Is that correct?

Vixting:
> Found more issues:
> 
>     * Repairing walls damaged from acid seems to work and they can be damaged afterwards, BUT now they will take WAY less damage from hits. A ravager usually deals 5% damage per hit, but after repairing an acided wall a ravager only deals about 2% damage per hit.
> 
>     * If you drop your nailgun while repairing a wall, the doafter still goes on and completes successfully.
> 
>     * I'm not sure the intended behavior here, but you can only repair walls with plasteel. Is that correct?

should be fixed, if you mean to repair the acid holes that is intended. repairing walls worked fine with other means last I tested

MidZik:
Ah I understand. Yes, repairing a wall without a hole works fine with other materials, only the hole needs plasteel.

Last issues I hope:
- Repairing a wall with a blowtorch doesn't require welding protection (it won't blind you).
- Fixing a hole with a nailgun causes a lot of client side errors: `[ERRO] root: Predicting the queued deletion of a networked entity: acid hole`

Vixting:
> Ah I understand. Yes, repairing a wall without a hole works fine with other materials, only the hole needs plasteel.
> 
> Last issues I hope:
> 
>     * Repairing a wall with a blowtorch doesn't require welding protection (it won't blind you).
> 
>     * Fixing a hole with a nailgun causes a lot of client side errors: `[ERRO] root: Predicting the queued deletion of a networked entity: acid hole`

the blind issue is being fixed by https://github.com/RMC-14/RMC-14/pull/9930
error should be fixed

MidZik (approved):
LGTM.

DrSmugleaf (approved):

PR #9751, opened 2026-04-29
Makes alerts on Seperated HUD expand vertically.

### LINKED ISSUE #9342: Xeno HUD drawn too low, cutting off elements.
## Description

The xeno hud is drawn _**far**_ too low on the screen when it really doesn't need to be, causing icons to extend beyond the bottom and become impossible to see depending on resolution. Maybe it's worse or better on other monitors, but I'm on 16:9 so this isn't an artifact of any crazy aspect ratio.

How relevant this is varies from xeno to xeno, but the max icons drawn at my resolution seems to be 7, hiding decently useful info like the muted status&strain specific trackers. 

For lesser drone(and assumedly parasites?), this means muted status on spawn is actually impossible to see, making it seem like these roles just can't talk at all.
For healer, this means being unable to know for sure when you can sacrifice&respawn safely. 
Etc.

**Reproduction**
In my case, I'm playing on a single 1600x900 monitor, windowed, and separated hud layout. This leaves roughly 1200x830 for the gameplay area. With the gameplay window at roughly that scale, evolve into a healer drone and observe that your HUD elements are consistently drawn too low to be fully rendered. Meanwhile there's room for another 6 icons directly above the current list going unused.

**Screenshots**
Image
As lesser drone(Or other newly spawned mid-round xenos), this means you can't see your muted status and just appear to be unable to talk for no reason.

Image

But by resizing, it's clear the elements are simply being drawn too low.

**Additional context**
1600x900 isn't particularly cutting edge in the era of 4k UHD, but it's the exact same aspect ratio as 1080p and only a smidge smaller, so I assume this is also present to a slightly lesser degree there, unless these values were hard-coded assuming 1080p specifically.

Arykh:
## About the PR

Title

## Why / Balance

Alerts get cut off from view.

Resolves #9342 
Well, until we add too many alerts, then we'll have to make it expand into two columns.

## Technical details

Just use SetGrowVertical

## Media

image

## Requirements

**Changelog**

:cl:
- tweak: Alerts on the Seperated HUD layout now expand vertically instead of only downwards to infinity and beyond.

### DIFF
--- a/Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
+++ b/Content.Client/UserInterface/Screens/SeparatedChatGameScreen.xaml.cs
@@ -24,6 +24,9 @@ public sealed partial class SeparatedChatGameScreen : InGameScreen
         SetAnchorAndMarginPreset(Hotbar, LayoutPreset.BottomWide, margin: 5);
         SetAnchorAndMarginPreset(Alerts, LayoutPreset.CenterRight, margin: 10);
 
+        // RMC14
+        SetGrowVertical(Alerts, GrowDirection.Both);
+
         ScreenContainer.OnSplitResizeFinished += () =>
             OnChatResized?.Invoke(new Vector2(ScreenContainer.SplitFraction, 0));
 


Vixting (approved):
LGTM

DrSmugleaf (approved):

PR #9759, opened 2026-05-01
Photo camera

Arendian:
## About the PR

Added a photo camera that you can use to take in-game pictures.

## Why / Balance

Parity.
It can make 1x1, 3x3, 5x5 and 7x7 photos.
The camera only works if there is at least one admin or whitelisted player connected to the server.

## Technical details

To prevent modified clients from sending any valid image to the server, only images sent by admins or whitelisted players are accepted. Instead of the camera user rendering the frame, it is done by any online admin or whitelisted client through the following process:

1. A player uses a camera entity to send a request to the server with coordinates of where the image needs to be taken.
2. Server does some basic checks, then finds any currently connected admin or whitelisted client(priority being: admin>whitelisted in lobby>whitelisted).
3. The server spawns an eye at the coordinates of the camera and sends an event to the found admin/whitelisted client.
4. The admin/whitelisted client renders a frame through the spawned eye, converts this to a PNG, then raises an event to send it to the server as a byte array.
5. The server does some checks, converts the array to a PNG, strips the metadata of the image, then re-encodes it as a PNG.(As long as it's a valid PNG, the server will accept it and display it, abuse should be mitigated by the fact that only admin/whitelisted clients are allowed to send an image).
6. The new byte array is stored on the `RMCPhotoComponent` of the photo entity in a non-networked datafield.
7. When a user interact with a photo, they send a request to the server, the server responds with the byte array.
8. The client that interacted with the photo converts the byte array to a texture, caches the texture on the `RMCCachedPhotoComponent` and displays it in a UI.
9. Until the user disconnects, when the user interacts with the photo they will load the cached texture instead of asking the server again for the data.

The client rendering the image will have UI overlays(like health bars, job icons etc.) disappear for a split second(Only a few frames) on entities in the photo area, this is only noticeable if the client has the photo area within their FOV.

## Media

https://github.com/user-attachments/assets/f34380ab-8915-4d81-8976-9efd8dd1adb4

## Requirements

**Changelog**

:cl: Dygon
- add: Added photo cameras. They are available from the rec-vend, the correspondent vendor and as a loadout option.


BarteG44:
perhaps this could use the same tile highlight system as flamers and such, tho a bit more subtle probably. Still, an amazing feature

PR #9762, opened 2026-05-01
Impenetrable resin - Chances Claim

yagwog:
## About the PR
Added timed impenetrable walls to chances

## Why / Balance
Rather simple, just denied access deeper into reactor for most of it, left certain parts available to push slightly into so marines can somewhat prepare pre drop timer.

Haven't felt the need to change the weeds outside of reactor as it's now just on the larva if they want to take the risk of leaving reactors to evolve if they want to prevent loot from forecon, and possibly deal with getting killed.

## Media
image
image

## Requirements

**Changelog**
:cl:
- add: Added impenetrable walls to Chances Claim


CatAndHats (approved):

PursuitInAshes (changes requested):
Add secondary layers incase somehow people get in, just incase, since its chances you can keep them really close in next to the already existing layer

PursuitInAshes (approved):
Disregard previous, I forgot we made it so you can't go through them anyways if they get held open
This should be fine

PR #9767, opened 2026-05-03
Banerjee's Canal

SharkSnake98:
## About the PR
A large TSE city-map designed for PVE, featuring various civilian buildings!

## Why / Balance
More PVE maps, especially cities, would be nice.

## Technical details
YML, RSI/JSON

## Media
image
image

## Requirements

**Changelog**
:cl:
- add: Added Banerjee's Canal, a large TSE City Map for PVE
- add: Added various props


kleinerstation13:
if you wanted to do something sovl you could make the slot machines make funny coin noises when they're powered

KittenColony:
WE'RE GOIN' TO FALLUJAH BABY

DrSmugleaf (dismissed):
YML good looks amazing check with PvE maints

Shromply (dismissed):
doing baileys bidding and giving this a concept review:

i love it, looks freaking awesome and i know i will get thorough enjoyment out of sending my pveers to banerjees canal for a fun war torn city field trip >:)

GitHubUser53123:
absolutely stunning, although there's not much going on in the south. i'm reluctant to mention that because it's probably in order to differentiate the north and the south part of the map.

Resources/Textures/_RMC14/Structures/hybrisa_misc_props/64x64.rsi/meta.json
PursuitInAshes, line 1: Copyright

PursuitInAshes (changes requested):
Thanks for the PR!

NW Elevator should probably use the actual elevator doors and just be fake instead of having an interior space, we do have real elevators now but I don't want them on PVE maps just yet.

This probably shouldn't be premapped as it can have some decently heavy implications. ` -61, 4` are the coordinates
Image

This safe doesn't need armor and should probably just be a fancy sidearm. At `-40, -130`
Image

Aside from those everything else seems okay. Reactor is really nice, I'm a tiny bit worried about the south feeling to empty but we can see what happens when its used on more rounds.

Resources/Prototypes/_RMC14/Entities/Objects/Misc/lightstick.yml
```diff
@@ -40,3 +40,32 @@
     state: lightstick_blue_variant1
   - type: PointLight
     color: "#2D2DE3"
+
+- type: entity
+  parent: RMCLightStick
```
PursuitInAshes, line 45: Parent and this entity probably shouldn't be in this file? Also space in the name for consistency please

Resources/Prototypes/_RMC14/Entities/Structures/Doors/fence_door.yml
```diff
@@ -108,6 +108,50 @@
     key: fences
     mode: NoSprite
 
+- type: entity
+  parent: RMCFenceDoor
+  id: RMCFenceDoorDark
```
PursuitInAshes, line 113: When this is opening this has weird sprite issues but it may just be my client

Resources/Prototypes/_RMC14/Entities/Objects/Misc/lightstick.yml
SharkSnake98: This entity is here because in operates in the same way as the lightsticks, as opposed to the regular lights.

Resources/Textures/_RMC14/Structures/hybrisa_misc_props/64x64.rsi/meta.json
SharkSnake98: Sprites originate from linked sources.

Resources/Prototypes/_RMC14/Entities/Objects/Misc/lightstick.yml
PursuitInAshes: Is there a reason why you wouldn't just use a normal light stick then? Or name it differently?

PursuitInAshes (approved):
Didn't load the map again, but diff on the latest commit seems to address requested changes
Thanks for the PR!

GitHubUser53123:
-76 -13
image

PursuitInAshes (changes requested):
Will have to discuss with the other maptainers about the lightsticktube and see where its used.

PR #9770, opened 2026-05-03
Made buckled marines infectable by leaps

ExponentialWizard:
## About the PR
Buckled marines in chairs will now be infected by leaps

## Why / Balance
This seems more intuitive, might be a bug fix? I'm not sure if it counts as a bug or not

## Technical details
In SharedXenoParasiteSystem OnParasiteLeap hit now calls Infect with the force flag set to true. the force flag prevents a check to ensure the marine is laying down. On non-buckled marines the knockdown will force them to be laying down, but on buckled marines they stay standing so the force flag seems to be a very simple way to fix the issue

## Requirements

**Changelog**
- fix: Buckled marines are now infectable by parasite leaps.


Scyarah:
Just setting `force = true` might have more unintended side effects? It skips the entire check if the marine is standing or nested.

ExponentialWizard:
> Just setting `force = true` might have more unintended side effects? It skips the entire check if the marine is standing or nested.

Leaping infects standing marines anyway. It would only have unintended consequences if someone added code to force the other checks. Another solution could be to unbuckle a marine when they are hit by a leap, but this solution seems reasonable to me

DrSmugleaf (approved):

PR #9784, opened 2026-05-05
[PVE] [RCM] Replace M96S-B with L144 LRR

TadJohnson00:
## About the PR

Replaces the M96S-B with the L144 Long-Range Rifle, replaces the M96 sight in the Scout Sniper's kits with an L8 HMOD. Reduces the L112's ROF from 3.8 to 2.5.

## Why / Balance

Rationale: replace UNMC gear with appropriate equivalents, balance specialist equipment appropriately.
Balancing notes:
The 10x28mm squash-head round deals 90 damage at 50 AP, the L144 has a damage modifier of 1.5. Using aimed shot deals double damage to the target and slows it for 7.5 seconds. With nine total provided magazines, picking the L144 will give you a theoretical total damage pool of 36,450 if the user only targets enemies with aimed shot. This is counterbalanced by the L144's slow rate of fire, making it a more economical weapon in terms of ammunition-efficiency at the cost of raw damage output. For reference, the DPS of the L144 is 90, whereas the DPS of the L112 (post-nerf) is 150.

## Technical details

YAML, sprites, JSON

## Media

image

https://github.com/user-attachments/assets/bf8881ea-a4fb-434a-9d03-a68c539dc1ea

## Requirements

**Changelog**

:cl:
- add: [PVE] Added the L144 long-range rifle for use by the Royal Crown Marines. Scout Snipers recommended to check their vendors in upcoming operations.


TadJohnson00:
Let it go on record that I'd rather a different gimmick for the L144 than flak ammo. I don't think it's very sovl for a precision weapon to be used as a crowd-clearing tool. Also, just because the rifle has IFF *doesn't* mean the flak does too; marines *will* die if you shoot xenos on top of them.

Shromply:
> Let it go on record that I'd rather a different gimmick for the L144 than flak ammo. I don't think it's very sovl for a precision weapon to be used as a crowd-clearing tool. Also, just because the rifle has IFF *doesn't* mean the flak does too; marines *will* die if you shoot xenos on top of them.

it doesnt need a gimmick, i believe the sniper should be more suited to just single hvt killing, like ravs and shit, and considering it has a 7.5 second slow and triple damage on aimed shot, i think removing the flak effect and just leaving at as a strong single target weapon would be best.

the rcm already has a grenadier for crowds, after all


and tbh, idk why it needs to be balanced differently than the m96s-b. i say just reskin it and keep it mechanically the same as the m96s-b, pawsonally

TadJohnson00:
Flak has been removed. Still pending some balance considerations; as of commit a6e4773 the L144 is still significantly worse than the L112 in terms of damage pool and DPS. For discussion: increasing firerate (so that the optimal usecase for the L144 isn't just quick-switching to a pistol) and/or increasing number of magazines provided

TadJohnson00:
In case there was confusion regarding the last comment, balancing discussions took place in the RMC Discord and were addressed in subsequent commits. This PR has been and is still ready for review.

PR #9802, opened 2026-05-07
Spawn on terminate's spawn coordinates now clamps to a projectile's max range

Arendian:
## About the PR

Fixed projectiles with the `SpawnOnTerminateComponent` spawning the prototype past their max range.

## Why / Balance

Bug fix.

## Technical details

When an entity with the `ProjectileMaxRangeComponent` and the `SpawnOnTerminateComponent` is terminated while past it's max range, the prototype is spawned at the coordinates where it's max range was reached, instead of at it's current coordinates.

## Requirements

**Changelog**

No CL, no prototype with this combination of components exists at the moment.

### DIFF
--- a/Content.Client/_RMC14/Xenonids/Targeting/XenoAbilityPreviewOverlay.cs
+++ b/Content.Client/_RMC14/Xenonids/Targeting/XenoAbilityPreviewOverlay.cs
@@ -642,11 +642,11 @@ public sealed class XenoAbilityPreviewOverlay : Overlay
     {
         if (_prototypes.TryIndex<EntityPrototype>(projectile, out var projectileProto) &&
             projectileProto.TryGetComponent<SpawnOnTerminateComponent>(out var spawn, _componentFactory) &&
-            spawn.ProjectileAdjust)
+            spawn.SpawnOffsetMultiplier > 0)
         {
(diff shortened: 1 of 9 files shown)


Ian321:
The `RMCFireProjectileComponent` exists to move the spawned flame one tile further than the tile it terminates on. Without it the fire does not spawn on the tile the entity is on but rather the tile in front of it.

https://github.com/user-attachments/assets/12775fd2-ac10-497c-b33e-36a045f8c342

The center of the smoke is supposed to be on the warrior, not the tile in front of it.

Arendian:
> The `RMCFireProjectileComponent` exists to move the spawned flame one tile further than the tile it terminates on. Without it the fire does not spawn on the tile the entity is on but rather the tile in front of it.
> 
>  Screencast.From.2026-05-07.20-24-40.mp4 
> The center of the smoke is supposed to be on the warrior, not the tile in front of it.


That behavior is moved to the `SpawnOnTerminateComponent`, I accidentally put the default -0.5 on it. Should work correctly now that it's 0.5. Thanks for reporting it!

Vixting (approved):
LGTM

DrSmugleaf (approved):

PR #9811, opened 2026-05-08
Burst fire stops if a gun is inside a container

### LINKED ISSUE #9810: Weapons with burst mode will shoot you when you stow them while they're shooting
## Description
Title, I got it to happen on my localhost and it still happens with self pb turned off

**Reproduction**
Shoot weapon in burst mode (mk80 works best), quickly stow in belt or bag

For some reason only happens with weapons that are small so are stowable in pistol rigs and backpacks, so just pistols with long burst modes like the m77 and mk80, regular rifles don't appear to hit me on burst

**Screenshots**
Can't get a video right now but I will later

**Additional context**

Arendian:
## About the PR

Putting a gun inside a bag or holster mid-burst now stops it from firing.

## Why / Balance

Because the bag was treated as the gun’s shooter, putting the gun into your bag mid-burst caused the shots to hit yourself.
Resolves #9810 

## Requirements

**Changelog**

:cl: Dygon
- fix:  Putting a gun inside a bag or holster mid-burst now prevents it from shooting.

### DIFF
--- a/Content.Server/Weapons/Ranged/Systems/GunSystem.AutoFire.cs
+++ b/Content.Server/Weapons/Ranged/Systems/GunSystem.AutoFire.cs
@@ -32,6 +32,22 @@ public sealed partial class GunSystem
             else if (gun.BurstActivated)
             {
                 var parent = TransformSystem.GetParentUid(uid);
+
+                //RMC14 Stop burst fire when inside a bag o holster.
+                if (!HasComp<DamageableComponent>(parent) &&
+                    Containers.TryGetOuterContainer(uid, Transform(uid), out var outerParent))
+                {
+                    if (HasComp<DamageableComponent>(outerParent.Owner))
+                    {
+                        gun.BurstActivated = false;
+                        gun.BurstShotsCount = 0;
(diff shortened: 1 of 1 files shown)


PursuitInAshes (approved):
Code Good
(What do you mean my bag wasn't supposed to shoot me?)

Content.Server/Weapons/Ranged/Systems/GunSystem.AutoFire.cs
```diff
@@ -32,6 +32,22 @@ public override void Update(float frameTime)
             else if (gun.BurstActivated)
             {
                 var parent = TransformSystem.GetParentUid(uid);
+
+                //RMC14 Stop burst fire when inside a bag or holster.
+                if (!HasComp<DamageableComponent>(parent) &&
+                    Containers.TryGetOuterContainer(uid, Transform(uid), out var outerParent))
+                {
+                    if (HasComp<DamageableComponent>(outerParent.Owner))
```
DrSmugleaf, line 40: Is this to check just inside a bag or holstered? You should be able to check if its contained by something, then check that container's id to the container owner's hand component container id

DrSmugleaf (changes requested):

Issue #9153, opened 2026-02-06
Put In Hand option on armor


BarteG44:
## Description
Right clicking and choosing the Put In Hand option on armor allows you to take off any armor instantly, while not dropping whatever item you had in the suit storage

**Reproduction**
put on any armor. right click it and click Put In Hand

Ignies:
https://github.com/RMC-14/RMC-14/issues/8848
Duplicated here although mine was poorly explained, also happens with uniforms.

Dutch-VanDerLinde:
<img width="1436" height="624" alt="Image" src="https://github.com/user-attachments/assets/bd29c32a-d2be-4de0-af94-78974823e13d" />

<img width="1430" height="578" alt="Image" src="https://github.com/user-attachments/assets/cb279a36-0c89-45b7-a6e1-76cee4b649de" />

Vermidia:
Fixed by #9814

PR #9814, opened 2026-05-08
Fix equipped armor pickup bypassing unequip delay

GrigoriyGalintano:
## About the PR

Fixes an issue where equipped RMC clothing/armor with an unequip delay could be moved to the hand instantly through the vanilla right-click "pick up / put in hand" interaction, bypassing the intended unequip do-after.

## Why / Balance

This makes the vanilla hand pickup path follow the same behavior as the RMC unequip action. Armor that is meant to take time to remove now consistently uses the unequip delay and can be interrupted by movement.

No balance change beyond enforcing the already intended unequip delay.

## Technical details

Adds an RMC-side interception for delayed equipped clothing in `RMCHandsSystem`.

When delayed clothing is equipped in the user's inventory and the active hand is empty, the RMC interaction starts the normal inventory unequip flow with `checkDoafter: true` instead of allowing the vanilla instant hand pickup path.

Also blocks direct insertion of delayed equipped clothing into hand containers while it is still equipped, preventing the bypass without modifying upstream/shared item behavior.

## Media

Not required, small behavior fix.

## Requirements

**Changelog**

- fix: Fixed some equipped armor being removable instantly through the vanilla hand pickup interaction.

### DIFF
--- a/Content.Shared/_RMC14/Hands/RMCHandsSystem.cs
+++ b/Content.Shared/_RMC14/Hands/RMCHandsSystem.cs
@@ -1,5 +1,6 @@
 using System.Linq;
 using Content.Shared._RMC14.Storage;
+using Content.Shared.Clothing.Components;
 using Content.Shared.Hands.Components;
 using Content.Shared.Hands.EntitySystems;
 using Content.Shared.Interaction;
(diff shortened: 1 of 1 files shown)


BarteG44:
closes #9153

DrSmugleaf (approved):

PR #9817, opened 2026-05-08
Fix weed spread through barricades

GrigoriyGalintano:
## About the PR

Fixes xeno weeds being able to spread through barricade collision when using Queen `Expand Weeds`.

Previously, manual `Expand Weeds` could place weeds onto a barricade tile from the blocked/front side or use nearby weeds through a barricade, effectively letting weeds pass through closed barricade collision.

Manual `Expand Weeds` now uses the same barricade spread rule as automatic weed spreading. Closed folding barricades block weeds from spreading through their closed/front side, while still allowing weeds to spread onto the barricade tile from an unblocked side or when the barricade is open.

Related to #9735 and #9258.

## Why / Balance

Barricades already block automatic weed spread through their impassable collision. Manual `Expand Weeds` could bypass that by spreading onto a barricade tile from the blocked side or by using nearby weeds through a barricade, which let queens extend weeds past intended barricade defenses.

This keeps Queen Eye visibility unchanged: seeing a tile through weeds vision does not by itself mean weeds can spread through barricade collision.

## Technical details

Adds `SharedXenoWeedsSystem.CanSpreadWeedsBetween(source, target)`, which raycasts between the source weeds and target tile using `CollisionGroup.BarricadeImpassable`.

`XenoWeedsSystem.Update` now uses that helper instead of its local duplicated barricade raycast.

`OnXenoExpandWeedsAction` now searches for an adjacent weeds tile that can actually spread into the target tile. If adjacent weeds exist but every path is blocked, the expand fails with the existing generic weeds failure popup.

## Media

N/A, small gameplay fix.

## Requirements

**Changelog**

- fix: Fixed Queen Expand Weeds spreading weeds through barricade collision

### DIFF
--- a/Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
@@ -1,4 +1,3 @@
-using System.Linq;
 using System.Numerics;
 using Content.Server.Atmos.Components;
 using Content.Server.Spreader;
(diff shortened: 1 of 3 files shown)


PursuitInAshes:
Thanks for the PR!
We decided to go with #9258 to fix this, but thank you for being ontop of the issue

PR #9824, opened 2026-05-09
More Larva playtime titles

RangerXVII:
## About the PR
Added new larva playtime titles and moved one around.

It is **now**:
Young: Little
Mature: Mature
Elder: Long
Ancient: Big
Prime: Fat

And **before** it was:
Young: Little
Mature: Mature
Elder: Big
Ancient: Big
Prime: Big

## Why / Balance
There are a lot of ancient larvae. It's only a matter of time until they get prime.

## Requirements

**Changelog**

:cl:
- tweak: Larva now has more titles.

### DIFF
--- a/Resources/Locale/en-US/_RMC14/xeno/xeno-name.ftl
+++ b/Resources/Locale/en-US/_RMC14/xeno/xeno-name.ftl
@@ -1,4 +1,4 @@
-﻿rmc-xeno-name-number = {$baseName} ({$prefix}-{$number}{$postfix})
+rmc-xeno-name-number = {$baseName} ({$prefix}-{$number}{$postfix})
 rmc-xeno-name-leader = Leader
 
 rmc-xeno-name = {$baseName} ({$prefix}{$postfix})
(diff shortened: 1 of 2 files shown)


IceNoobXD:
I belive the reason why Larvas dont have more playtime titles to begin with is because it would reward players with a "cool" title for just camping larva the whole round and wasting a burst or burrowed larva

QuietlyWhisper:
> I belive the reason why Larvas dont have more playtime titles to begin...

Probably means that it should be reading xenonid overall playtime rather than "larva playtime" which is in itself an issue

MidZik:
> > I belive the reason why Larvas dont have more playtime titles to begin...
> 
> Probably means that it should be reading xenonid overall playtime rather than "larva playtime" which is in itself an issue

I was thinking the same, will make a PR for that without including the name changes. I'll leave that to this PR.

If we do that, we will probably need more names though. There are gonna be a lot of apex larvas, maybe some even past apex.

RangerXVII:
> I belive the reason why Larvas dont have more playtime titles to begin with is because it would reward players with a "cool" title for just camping larva the whole round and wasting a burst or burrowed larva

That certainly does make sense... although most people have gotten their ancient larva playtime by skipping tax at the start of rounds since that would be boring. I've never seen a larva just hanging around after the marines land to grind for the current playtime titles.

Completely changing how larva playtime works is kind of out of scope of this PR, and I'm not sure what do about this one right now. I'll just leave it until a maint requests changes or closes it.

CatAndHats (approved):
Code is fine. We might want to wait on this to merge alongside https://github.com/RMC-14/RMC-14/pull/9826 so they're in the same update.

PR #9827, opened 2026-05-10
Limit the amount of lights in an area

Arendian:
## About the PR

The client disables lights if there are too many lights clustered in a certain area.

## Why / Balance

Prevent spamming lights.

- Default max amount of nearby other lights : 4
- Default check range: 2.5 tiles.
- Default check interval: 1 second.

These values can be modified by clients via a CVAR.

It currently counts and disables any nearby light with the `HandHeldLightComponent` or the `ExpendableLightComponent`, I might change it to only counting lights that have a certain brightness.

## Technical details

Once a second the client checks if a light has too many lights nearby, it gives as many of the nearby lights needed the `RMCTemporaryDisabledLightComponent` until the amount of active lights is below the cap. It currently only cares about lights with the `HandHeldLightComponent` or the `ExpendableLightComponent`.
Every frameupdate, all enabled lights with this component are disabled.

## Media

https://github.com/user-attachments/assets/9385991d-c697-4819-a8a4-194a580cd0c6

## Requirements

**Changelog**

:cl: Dygon
- tweak: The amount of light emitted in small areas is now limited.


PursuitInAshes (commented):
I realized I just never replied when you asked me on discord (Whoops) but its probably good to have this as something in the options menu for clients
That or we don't need to let clients edit this and the cvars can stay just incase there are complaints about it even with the adjustment. Not sure which is best so may want to get host opinions aswell

DrSmugleaf (changes requested):
Not sure if there's much value in letting clients set these on their own, but there should be at least a server replicated cvar to disable or enable this behaviour in case it bugs out on live
Not sure how it's going to interact with large amounts of people on the dropship or on the frontline

Arendian:
> Not sure if there's much value in letting clients set these on their own, but there should be at least a server replicated cvar to disable or enable this behaviour in case it bugs out on live Not sure how it's going to interact with large amounts of people on the dropship or on the frontline

I've added the server cvar to enable/disable the behavior. The client CVARs were added in case it was a bit too strict in real game scenarios, allowing people to tweak it themselves if they notice too many lights not working because of this. I can remove them though if you think that's unneeded.

DrSmugleaf (approved):
Should be good ty

PR #9834, opened 2026-05-11
Translate most of the SPP ranks

MoonlightHollow:
## About the PR

Translate the SPP ranks into english while trying to keep some dialect still in the name, mostly in Cs being Ks, etc.

## Why / Balance

A lot of the SPP ranks were very hard to understand and to even write so after asking Whisper translated most of them to be easier to read. I did try to keep some dialect to them, but if someone has suggestions on how to keep the readability but also add some more dialect then feel free to offer.

## Media

image
image
image

## Requirements

**Changelog**

:cl:
- tweak: SPP ranks are translated


kleinerstation13:
I don't like this change, I prefer the anglicized russian

kleinerstation13:
To be clear if you can't read the ranks then the paygrades exist

mmails:
> I don't like this change, I prefer the anglicized russian

I gotta agree with you it takes away from the SPP

CatAndHats (commented):
The issue with the SPP ranks was partly due to a lack of exposure in the regular flow of the game, with the ranks only appearing on Sorokyne (without admemes). Since Trijent is also getting an SPP insert for the next update, we're going to wait and see if this increased SPP presence helps with peoples understand of the ranks.

Drazamuffin:
Isnt the issue of understanding ranks in different languages literally why paygrades of E and O exist IRL?

CatAndHats:
It is the opinion of multiple Maintainers, and also Whisper, that the ranks are fine as they currently are. As Drazamuffin and other people have said: if you need to know their rank you can see the O1/E2/etc. Thank you for the PR but this will be closed.

PR #9840, opened 2026-05-12
[FIX] Superceded by #9866

Thereisabear:
## About the PR
Fixes burrowers being able to acid things whilst burrowed.

## Why / Balance
Bug and unintended mechanic. 

## Technical details
adds an early return in the Acid System which checks if burrower is burrowed.

## Requirements

**Changelog**
:cl:

- fix: Fixed Burrowers being able to acid whilst burrowed.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
@@ -31,6 +31,7 @@ using Robust.Shared.GameObjects;
 using Robust.Shared.Physics;
 using Content.Shared.Chemistry.Components.SolutionManager;
 using Content.Shared._RMC14.Chemistry;
+using Content.Shared._RMC14.Xenonids.Burrow;
 
 namespace Content.Shared._RMC14.Xenonids.Acid;
 
@@ -136,6 +137,9 @@ public abstract class SharedXenoAcidSystem : EntitySystem
                 return;
             }
         }
+
+        if (TryComp<XenoBurrowComponent>(xeno, out var burrow) && burrow.Active)
+            return;
 
         if (xeno.Owner != args.Performer ||
             !CheckCorrodiblePopupsWithReplacement(xeno, target, args.Strength, out var time, out var _))


Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
```diff
@@ -137,6 +138,9 @@ var e when TryComp<DropshipEnginePointComponent>(e, out var engine) => engine.Co
             }
         }
 
+        if (TryComp<XenoBurrowComponent>(xeno, out var burrow) && burrow.Active)
```
MidZik, line 141: I don't think this is the correct way to fix this. Rather than having the acid system check if the xeno is burrowed, the burrowed system should do two things:
- On `BurrowedEvent` (i.e. in `SetBurrow`), if the user just burrowed, cancel all active do-afters.
- On `DoAfterAttemptEvent<XenoCorrosiveAcidDoAfterEvent>`, if the user is currently burrowed, cancel the event.

MidZik (changes requested):

Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
Thereisabear: it was done this way because the yml has CanInteractCheck False.which bypasses the checks in burrowed system
And cancelling all active doafters is not ideal as burrowers need to do after to get out or move in burrow.

Thereisabear: Apologies! I totally misread and thought you were speaking about another interaction. 
I will close this PR since your fix is already up. Thanking ya!

Thereisabear:
See #9866 
Fixes it better.

Content.Shared/_RMC14/Xenonids/Acid/SharedXenoAcidSystem.cs
MidZik: No worries. I ended up writing and testing it myself to confirm so I just opened a new PR with it rather than have you repeat the work.

PR #9847, opened 2026-05-13
Fix gun group penalties not always refreshing

Arendian:
## About the PR

Gun modifiers now refresh on all guns in a user's hands after another gun they were holding is dropped or reparented.

## Why / Balance

Bug fix. It mostly affected M63's with an armbrace, it would keep the dual wield penalty after picking up a gun with the other hand until you undid the brace or did any other action that refreshed modifiers(swapping attachment, wielding/unwielding etc.).

## Technical details

GotUnequipped runs before reparenting, this means the gun that's not dropped refreshed it's modifiers while the other gun was still in the hand.

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed guns not always losing their dual wielding penalty after dropping one of the guns.


PursuitInAshes (approved):
Code should be good

DrSmugleaf (approved):

PR #9849, opened 2026-05-13
underbarrel-flare-gun-v2

ZeD11133:
## About the PR

a sub-barrel attachment from the CM13 has been added, which fires like a flare gun.

## Why / Balance

parity

## Technical details

- Added flare gun attachment prototype to Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/under.yml
- Added flare gun to Cargo Vendor and Squad vendor lists
- M4SPR spawns with flare gun attachment by default

**Issues**
The flare sometimes passes through entities at certain angles. This occurs because `CMFlare` is a `ScatteringGrenade` rather than a standard bullet, and its collision is handled differently from `BaseBullet`.

**Possible solutions:**

1. **Add a stub component** that the `RMCAirShot` system can check to block skyward firing. This would allow using `RMCAirShot` (as on the M82F) without changing the firing logic, preserving correct flare behavior.

2. **Replace `CMFlare` with a `BaseBullet`** that has a `SpawnOnCollide` component which spawns `CMFlare` on impact. This turns the flare into a regular bullet with proper collision, creating a burning flare at the point of impact. Fixes the passthrough issue but requires creating a new projectile prototype.

Not yet sure how to implement this in the most optimized way. Open to suggestion

## Media

https://github.com/user-attachments/assets/f6452f66-69d1-40e5-93ba-1f2303bcec06

## Requirements

**Changelog**

:cl:
- add: Added Underbarrel FlareGun
- remove: 
- tweak: 
- fix: 
- admin: 
- code: new prototype and tag: RMCAttachmentUnderbarrelFlareGun


BarteG44:
the parity underbarrel flare gun has an internal magazine of 3 flares and can shoot much further (as far as you can see with a 4x scope). Also it's meant to be in the automated weapons rack and not the attachments rack, like all the other pre-equipped gun attachments

PR #9850, opened 2026-05-13
Adds: Attachment Tooltips to guns + Menu for all possible attachments

### LINKED ISSUE #2963: what attachments a gun takes isnt communicated
## Description
you need to either do trial and error, rtfsc or use vv to find what attachments a gun can take, or what guns you can use an attachment with

IceNoobXD:
## About the PR

Displays the attachment slots of a gun in the tooltip and if its has a attachment it shows the name of it.
The Tooltip is also only displayed when you are close to the gun. It doesnt display if you are far away or somebody is holding it.
Adds a "view compatible attachments" button for all guns which pulls the data from the gun entity and displays it.
Xenos can access both the examine tooltip and the compatible attachments menu, because the sprite of the attachment on the gun is not obscured, which lead to xenos needing to memorize all of the sprites of attachments to know what a dropped gun has.

## Why / Balance

Tooltip is parity
compatible attachments is QOL

## Technical details

.cs, .ftl and .xaml changes 

Tooltip shows what is currently installed
Tooltip is serversided, uses ExaminedEvent

Menu shows what could be installed
Menu is clientsided, uses GetVerbsEvent<Verb>

Both read from AttachableHolderComponent.Slots

resolves #2963 

The tags RMCAttachmentFlamer and RMCAttachmentExtinguisher did not match their entity IDs, causing the name to display the tag ID instead. I needed to rename it in every gun so that I can match the prototype entity ID and get the entity name (RMCAttachmentMiniFlamethrower and RMCAttachmentUnderbarrelExtinguisher).

## Media

grafik
grafik

https://github.com/user-attachments/assets/6076eed0-4cc6-44ed-9b57-6bac205809d0

## Requirements

**Changelog**

:cl:
- add: Examining a gun now displays its currently equipped attachments per slot in the tooltip!
- add: Added a new QoL feature that displays all compatible attachments for each slot when examining a gun!


PursuitInAshes:
Would be good to make the verb not appear for Xenos if possible to just reduce clutter, the examine is entirely fine though

DrSmugleaf (changes requested):
Thanks for the PR!

The attachment examine should probably go at the very bottom of the tooltip since it's the most dynamic one and would push everything else around otherwise

DrSmugleaf (approved):

Issue #9790, opened 2026-05-05
Calling down dropship as queen should come with announcement in hivemind


AverageNotDoingAnythingEnjoyer:
## Description

Parity

**Screenshots**

<img width="415" height="91" alt="Image" src="https://github.com/user-attachments/assets/321f903b-ff56-48f7-993f-e546719ff433" />

PR #9854, opened 2026-05-14
Hivemind announcement when Queen call the Dropship

Meara1179:
## About the PR

Made it so that a hivemind announcement is sent out when the Queen calls down the dropship to begin hijack, as well as which LZ it was called to. 

Addresses https://github.com/RMC-14/RMC-14/issues/9790 

## Why / Balance

Parity

## Technical details

Added AreaSystem as a dependency to SharedDropshipSystem 
Added SharedXenoAnnounceSystem as a dependency to SharedDropshipSystem

## Media

Content Client_3IzUldqTKd

## Requirements

**Changelog**

:cl:
- add: An announcement is now made to the hive when the Queen calls down the dropship for hijack.

### DIFF
--- a/Content.Shared/_RMC14/Dropship/SharedDropshipSystem.cs
+++ b/Content.Shared/_RMC14/Dropship/SharedDropshipSystem.cs
@@ -1,4 +1,4 @@
-using System.Linq;
+using Content.Shared._RMC14.Areas;
 using Content.Shared._RMC14.CCVar;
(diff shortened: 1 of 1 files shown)


PursuitInAshes (approved):
Code seems fine
I don't know if TyGetArea already has handling for if there is no area, which would make it so you don't have to set "Unknown" but I don't want to check atm

DrSmugleaf (approved):

PR #9864, opened 2026-05-15
CMSignalSystem

kukurua:
## About the PR
Buttons can be linked to doors (and other objects) using a multitool.

## Why / Balance

Buttons can now be linked to doors and other linkable objects using the Multitool (more possibilities in the game—more fun!).

## Technical details

CMSignalSwitchSystem and CMSignalSwitchComponent are copies of SignalSwitchComponent and SignalSwitchSystem from the original SS14, with the addition of access checks.
I decided not to modify the source code, but rather to create modernized copies of it—so that if anyone were to use the original components in the future, they would function exactly as originally intended.
The DeviceLinkSource and CMSignalSwitch components have been added to RMCPodDoorButton.

## Media

A [link](https://drive.google.com/file/d/1enoJEkbYzA0Sgb8lv1HVPIOg3a7JR2Np/view?usp=sharing) with demonstrations of the changes.

## Requirements

**Changelog**

:cl:
- add: You can bind buttons to linkable objects (such as doors) using the Multitool.
- code: Those responsible for the maps will need to link the buttons to the doors using the multitool and release a map update.


PursuitInAshes (changes requested):
Thanks for the PR!

Signal linking is notorious for causing test fails and various mapping issues (Specifically when entities with links are deleted), which is a large reason for why we use our current system of two separate components with strings.

I'm not fond of just copy pasting code from already existing places if we can avoid it.

Conceptually I'm not sure if this is desired, I have some reservations about it. (Is this Parity?) I will check with the others though _if_ you'd like to try to find some other way of implementing door linking, let me know. I wouldn't work on any changes until then, if thats the case.

AverageNotDoingAnythingEnjoyer:
> (Is this Parity?)

It is, but it's uses are niche, like opening doors in operating theaters from inside and CO quarters from the top of my head

kukurua:
> Thanks for the PR!
> 
> Signal linking is notorious for causing test fails and various mapping issues (Specifically when entities with links are deleted), which is a large reason for why we use our current system of two separate components with strings.
> 
> I'm not fond of just copy pasting code from already existing places if we can avoid it.
> 
> Conceptually I'm not sure if this is desired, I have some reservations about it. (Is this Parity?) I will check with the others though _if_ you'd like to try to find some other way of implementing door linking, let me know. I wouldn't work on any changes until then, if thats the case.

Oh. Please forgive me. I originally did this for another project based on this one, where for some reason they forgot to add an RMCPodDoorComponent to doors. Because of this, I thought the button-opening functionality wasn't implemented, so I suggested this solution.
Yes, your solution, which is already in the project, works, and I can understand why you're using it, but it doesn't allow for button binding as part of gameplay (only through the debug panel).

I originally thought this method would be used during map creation or editing, but you have a more painless solution for that. However, I still suggest adding this option for binding buttons to objects in addition to the main one, so that players can make changes during the game (sometimes it's fun to make something yourself), which will be reset every time the round is restarted.
In other words, I propose adding this as something for players, not as a map editing tool.

Regarding the fact that I simply copied the original files and modified them, I don't know how your project handles changes to the original SS14 code, but if it's acceptable, I'll simply modify them and not create extra copies.

PursuitInAshes:
You're all good

The first draft of our coding standards to reference can be found [here](https://github.com/RMC-14/RMC-14/blob/master/coding-standards.md), in this particular instance the addition of access probably isn't enough for the entire system to be copied. I'd defer to one of the C# maintainers though and their judgement.

As for letting players in round use signals, I'm not sure if its conceptually desired. When I mentioned it the general remarks is it may be too niche and just cause "hell chambers" for xenos. If you want I can ask around and get you a definitive answer though, but it may be a bit due to me needing to put my attention elsewhere outside of the project

kukurua:
I don't mind waiting for your decision. This function is not a priority.
I will make changes to avoid copying files.

PR #5813, opened 2025-03-04
Fix wyvern stats

Vermidia:
## About the PR

Title.

## Why / Balance

Compared to what it was ported from, it was *very* wrong. Accuracy was way too high, damage on bullets was wrong, gun was missing a burst mode etc.

## Technical details

YML.

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- fix: Fixed the D50 winter wyvern's stats

PR #9884, opened 2026-05-17
FORECON CO now gets a belt for their choice of sidearm

### LINKED ISSUE #9736: FORECON CO sidearm belt should depend on which sidearm CO has chosen in their preferences
## Description

Parity

This surv role fetches whatever CO has chosen for themselves in personalization of reg CO and adjusts the starting belt

**Additional context**

https://github.com/cmss13-devs/cmss13/blob/master/code/modules/gear_presets/survivors/lv_522/forcon_survivors.dm#L262

Vexerot:
## About the PR

Title. Closes #9736.

## Why / Balance

Parity, though actual implementation differs a little. 

Instead of pulling the preference from the regular CO role, the Survivor Commander role now has has the option of choosing between the Mateba and Winter Wyvern in their loadout. This enable the FORECON CO to spawn with a belt corresponding to this choice. Note that other Survivor Commander roles still spawn with their normal belt.

## Technical details

Added loadouts to the Survivor Commander role. Added an appropriate belt prototype for the Winter Wyvern.

In order to facilitate this, spawning logic was modified to explicitly unequip and delete any item already occupying a slot before equipping the replacement. This is needed because loadout items are spawned before starting gear is applied. Now, the loadout sidearm belt is spawned, then if no belt is specified in the roles starting gear, the player will keep that loadout sidearm belt, whereas if the roles starting gear does specify a belt, it will delete the loadout sidearm belt and replace it with the proper belt. This is what allows for the FORECON CO to get their choice of belt, whereas other Survivor Commanders will not get this choice of belt.

This fix applies universally, so any future job that defines a loadout overlapping with their starting gear will benefit. Simply leave the starting gear for that slot empty and they will get their choice of loadout for that slot.

## Media

https://github.com/user-attachments/assets/3efe7435-8209-4f78-bb3c-a4b5af1e1ba8

## Requirements

**Changelog**

:cl:
- add: FORECON CO now spawns with their choice of sidearm.

### DIFF
--- a/Resources/Prototypes/_RMC14/Catalog/Fills/Clothing/belt.yml
+++ b/Resources/Prototypes/_RMC14/Catalog/Fills/Clothing/belt.yml
@@ -689,6 +689,17 @@
 
 - type: entity
   parent: RMCBeltHolsterPistol
+  id: RMCD50WinterWyvernBeltForeconFilled
+  suffix: Filled, Winter Wyvern, Forecon
+  components:
+  - type: StorageFill
+    contents:
+    - id: RMCMagazinePistolHandcannon
+      amount: 3
+    - id: RMCWeaponPistolHandcannonWinterWyvern
+
+- type: entity
+  parent: RMCBeltHolsterPistol
   id: GoldHandcannonBeltFilled
   suffix: Filled, Gold Handcannon
   components:
--- a/Resources/Prototypes/_RMC14/Roles/Loadouts/role_loadouts.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Loadouts/role_loadouts.yml
@@ -986,6 +986,7 @@
   id: JobRMCSurvivorCommandingOfficer
   points: 7
   groups:
+  - SurvivorCommandingOfficerWeaponRMC
   - EyewearRMC
   - MasksRMC
   - PaperworkRMC
--- a/Content.Shared/Station/SharedStationSpawningSystem.cs
+++ b/Content.Shared/Station/SharedStationSpawningSystem.cs
@@ -88,7 +88,7 @@ public abstract class SharedStationSpawningSystem : EntitySystem
     public void EquipStartingGear(EntityUid entity, LoadoutPrototype loadout, bool raiseEvent = true)
     {
         EquipStartingGear(entity, loadout.StartingGear, raiseEvent);
-        EquipStartingGear(entity, (IEquipmentLoadout) loadout, raiseEvent);
+        EquipStartingGear(entity, (IEquipmentLoadout)loadout, raiseEvent);
     }
 
     /// <summary>
@@ -105,7 +105,7 @@ public abstract class SharedStationSpawningSystem : EntitySystem
     /// </summary>
     public void EquipStartingGear(EntityUid entity, StartingGearPrototype? startingGear, bool raiseEvent = true)
     {
-        EquipStartingGear(entity, (IEquipmentLoadout?) startingGear, raiseEvent);
+        EquipStartingGear(entity, (IEquipmentLoadout?)startingGear, raiseEvent);
     }
 
     /// <summary>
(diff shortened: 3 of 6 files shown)


Resources/Prototypes/_RMC14/Roles/Loadouts/role_loadouts.yml
```diff
@@ -1008,6 +986,7 @@
   id: JobRMCSurvivorCommandingOfficer
   points: 7
   groups:
+  - SurvivorCommandingOfficerWeaponRMC
```
AverageNotDoingAnythingEnjoyer, line 989: It might be better if survCO just used CO sidearm group instead of a custom one, since they're supposed to have access to all the variants anyways

Vexerot: Unfortunately, the way that loadout groups are defined, this would result in the selected sidearm being placed into the players bag, notably not spawning them with the belt nor the appropriate ammo.

Resources/Prototypes/_RMC14/Catalog/Fills/Clothing/belt.yml
```diff
@@ -687,6 +687,17 @@
...
     - id: RMCWeaponPistolHandcannonWinterWyvern
 
+- type: entity
+  parent: RMCBeltHolsterPistol
+  id: RMCD50WinterWyvernBeltForeconFilled
+  suffix: Filled, Winter Wyvern, Forecon
+  components:
+  - type: StorageFill
+    contents:
+    - id: RMCMagazinePistolHandcannon
```
kleinerstation13, line 697: is this parity? this ammo is significantly worse than mateba high impacts

Vexerot: FORECON CO doesn't get access to high impacts, they get instead the basic non-stunning bullets for all their guns. Whether the stats are parity, I'm not entirely sure, but #5813 supposedly fixed the stats of the WW.

kleinerstation13: ok i found out, you need to port heavy .50 ammo for the wyvern belt
obraz

kleinerstation13: obraz

kleinerstation13: as far as i can tell it just has 15 more damage and worse accuracy?

Vexerot: image

--

image

--

AFAIK, looks like the heavy ammo is just the standard ammo for the gun. I haven't really looked much at how stats have been ported over previously, but seems like the damage needs to be increased?

kleinerstation13: yea the damage seems to be non-parity so thats why i thought it was weak, the pistol bullets themselves between the magazines are the same but you could aswell port the heavy mag since it'd be like 10 lines of yml

Vexerot: Actually, after having stared at it a bit more, the heavy mag doesn't seem like a separate thing as it is just what they call the default mag, so no point adding a redundant prototype. Let me know if I'm misreading things though and if the heavy mag would actually be used non-redundantly, then should be quick addition as you said.

In terms of the ammo stats, things don't seem quite right either for the Mateba or the WW actually. But I'll leave any stat changes to a future PR as it's not really within the scope of this one to be micro-balancing.

PursuitInAshes (changes requested):
Thanks for the PR!
I think we're going to divert from parity here for a couple reasons and instead just make it a random gear variant (Like the other surv gear presets) for what belt they spawn with. Favored probably 25% WW, 75% Mateba? You can redo this PR to do that or if you aren't interested you are alright closing and I'll pick it up later.

The reasoning here is mostly:
- There is an objectively better choice in the context of survivor (The WW has a higher ammo capacity), which potentially opens the door to toxicity due to a meta pick being so obvious and it should be preference based. Its a large capability difference between the two.
- No other survivor gets the option to choose their starting gun, and while COs can get a lot of exceptions this probably shouldn't be one of them

PR #9886, opened 2026-05-18
Blood Splatter System

### LINKED ISSUE #7826: Add marine blood splatters

CyberTropic:
## About the PR

When anyone with RMCBloodSpkatterer component receives a brute type wound, a blood splatter is created at their location, up to 4 per tile by default (same as xeno blood). If the wound is below MinimalTriggerDamage then it's just a small blood drip decal. These decals are purely cosmetic, and the upstream puddles are still used for bleeding over time.
When they vomit we also create a vomit decal in the same way. 
Resolves issues brought up in the previous attempt at this in https://github.com/RMC-14/RMC-14/pull/9031

## Why / Balance

Resolves https://github.com/RMC-14/RMC-14/issues/7826
Parity

## Technical details

Decal spawning is based on code in AcidBloodSplashSystem.
Added some events for the splatterer system to listen to: event for wound added, and event for both vanilla and rmc vomit (both are used in gameplay)
In parity there are also supposed to be blood drips and spills from regular bleeding over time, but ~~that would require changes to the puddle system and puddle system changes are out of scope here, so it's a TODO.~~ I think that the use of the upstream puddle system actually makes more sense for regular bleeding, since that's also the same system used for spilled chemical etc.

## Media

sdadsada
image

*Note no blood from being burned, or from mice*

dsadsada

## Requirements

**Changelog**

:cl:
- add: Added blood spill decals when wounded.


GrigoriyGalintano (commented):
@CyberTropic 

Could you also look into the regular Bloodstream blood puddles?

Right now this PR adds cleanable RMC blood/vomit decals for wound and vomit splatters, which is great, but normal bleeding still goes through `SharedBloodstreamSystem -> _puddle.TrySpillAt` and creates `PuddleComponent` blood. That means weather/decal cleanup can clean the new splatter decals, but not the actual blood puddles from bleeding.

Would it make sense to convert or represent those bloodstream blood spills as cleanable RMC decals too, or otherwise route them through the same cleanable decal path? Ideally this should cover normal marine blood and synth blood without hardcoding reagent IDs.

I’m currently finishing the weather system implementation, and for it to behave correctly everything blood/vomit-related should ideally be represented as decals, or at least go through the same cleanable decal pipeline. Otherwise weather cleanup will only affect the new splatter decals while leaving regular bloodstream puddles behind.

CyberTropic:
I didn't touch regular bleeding puddles here, because they shouldn't be handled as decals but rather as entities, and need their own system that's outside the scope of this pr: puddles hold chems which has gameplay effects, and you're supposed to be able to walk through them leaving footprints and getting clothes dirty. A mix of decals for large blood splashes and entities for blood puddles is how other servers handle this as well, we'd just need more randomized sprites for the puddle entities that don't connect into a blood ocean. They currently dry out pretty quick as well, leaving nothing behind, so I don't think it's a big deal that they aren't affected by weather. 
ACTUALLY now that I write this out, I think a good way to handle this would be to have puddles leave decals behind when they dry out, and the decal can be colored the same color as the puddle was. Since the dried decals would be tagged cleanable they can be affected by weather etc without issue.

PR #9896, opened 2026-05-19
RCM Ranks and Jobs consistency changes

MoonlightHollow:
## About the PR

Update the ranks and job names to be consistent with rest of the factions. Splitting from my original PR: #9708.

## Why / Balance

First, was the unification of ranks with paramarines, this was a hold over from the CM13 port but with new standards it was just combined with normal RCM ranks with a special rifleman rank of Paramarine. Also dropped the ERT MedTech rank down to Lance Corporal to remand consistent with rest of rankings.

Second, PVE are primarily two with GPMG being renamed into the Smart Gun Operator and switch the gun to the ML66C 'Royal' variant as every faction uses their primary SGO name and gun, there is little reason for RCM to be different in this. Also renamed the Medical Assistant to Medical Technician as rest of the RCM uses this name and the original name comes more from their RMC origins which doesn't fit here. Important to note is I didn't touch the decals as that would more than likely require some remapping better leave that to separate thing if possible.

## Technical details

Merged Paramarines ranks with normal RCM, they were redundant. Then change the PVE GPMG and Medical Assistant to follow 

## Media

image
image

## Requirements

**Changelog**

:cl:
- tweak: Updated some of RCM ranks and jobs names to follow rest of the faction.


PursuitInAshes:
I believe I resolved conflicts correctly, ping me if something seems weird

PR #9904, opened 2026-05-20
Leap and headbutt no longer affected by xeno's movement

MidZik:
## About the PR

Leap-like abilities (Leap, Pounce, Rush, Dash, Assault) will now always land exactly where you click (or up to max distance towards it, or until you hit something). The landing point is no longer offset in the direction you were moving when you used the ability.

Headbutt will now always land exactly where the target was (or until you hit something). The landing point is no longer offset in the direction you were moving when you used the ability.

## Why / Balance

All the above abilities were affected by your movement at the moment you used them. For example, if you are moving towards the target or position you clicked, you will overshoot by a tile or more. Likewise, if you are moving away from the target or position you clicked, you will undershoot by a tile or more.

Moving perpendicularly to the target also offset where you end up. If your target is to the right, for example, but you are moving up while using the ability, you will end up above where you clicked.

While some of this could be useful as techniques, overall I think it is counterintuitive to how the abilities should work. If you click on a spot or target, you should end up there. Their behavior will be consistent regardless of how you are moving when you use them.

Additionally, this behavior means that some abilities are able to go much farther than intended, past their "max range". In that sense, this is a bugfix as well.

## Technical details

Immediately before applying the impulse to the xeno that's leaping or headbutting, make the xeno stop moving via ResetDynamics.

## Media

**Unfixed headbutt:**

https://github.com/user-attachments/assets/08d36246-ebb6-4926-88df-9c425f02b281

**Fixed headbutt:**

https://github.com/user-attachments/assets/fedf8763-1ed3-4852-b561-82caa2916ef7

**Unfixed leap:**

https://github.com/user-attachments/assets/c0f9caa8-2000-4dca-95a4-0112e337a232

**Fixed leap:**

https://github.com/user-attachments/assets/1d3c5a70-000b-4933-b2b6-abbdfb37ee13

## Requirements

**Changelog**

:cl:
- tweak: Leap, Pounce, Dash, Rush, Assault, and Headbutt no longer overshoot, undershoot, or miss the target position due to your movement at the moment you used them.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Headbutt/XenoHeadbuttSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Headbutt/XenoHeadbuttSystem.cs
@@ -86,7 +86,15 @@ public sealed class XenoHeadbuttSystem : EntitySystem
         Dirty(xeno);
 
         _rmcObstacleSlamming.MakeImmune(xeno);
-        _throwing.TryThrow(xeno, diff);
+
+        if (TryComp<PhysicsComponent>(xeno, out var physics))
+        {
+            // Prevent headbutt from having longer/shorter range or skewed direction
+            // based on the xeno's movement when using it.
+            _physics.ResetDynamics(xeno, physics);
+        }
+
+        _throwing.TryThrow(xeno, diff, doSpin: false);
     }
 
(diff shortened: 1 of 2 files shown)


ThanosDeGraf:
Such small tweaks, but it should heavy reduce the amount of WTH moments!

DrSmugleaf (approved):

PR #9910, opened 2026-05-20
Make water extinguish mobs when resisting fire

GrigoriyGalintano:
## About the PR

Makes fire resist interact with active RMC water.

Burning mobs that use the normal fire resist / stop-drop-roll action while standing in RMC water are extinguished immediately after the standard resist behavior starts. This matches the main CMSS13 behavior where humans and xenos roll/resist first, then `ExtinguishMob()` if they are on river water.

## Why / Balance

**PARITY**. Water now gives burning mobs a reliable way to extinguish themselves, but still requires the normal fire resist action. Simply walking into water does not passively extinguish fire.

Covered water is not treated as active water. The check reuses existing RMC water collision logic, so water covered by catwalks / water cover does not count.

This does not change tile fire behavior: fire can still exist on water, and mobs can still ignite from fire on a water tile until they resist.

## Technical details

Added RMC water helper methods in `RMCWaterSystem`:

- `IsActiveWater(...)`
- `IsInWater(...)`

These helpers reuse `RMCWaterSystem.CanCollide`, so existing RMC water cover behavior is respected.

`FlammableSystem.OnResistFireAlert` now keeps the normal RMC fire resist flow:

1. play stop-drop-roll visuals;
2. start normal `Resist`;
3. if the mob successfully started resisting and is in active RMC water, call `Extinguish`.

## Media

Minor change, no media required.

## Requirements

**Changelog**

- tweak: Burning mobs now extinguish immediately when using fire resist while standing in active RMC water.

### DIFF
--- a/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
@@ -282,10 +282,9 @@ namespace Content.Server.Atmos.EntitySystems
                 return;
 
             // RMC14 use the normal stop-drop-roll resist before active water extinguishes.
-            var wasResisting = ent.Comp.Resisting;
             _rmcFlammable.DoStopDropRollAnimation(ent.Owner);
             Resist(ent, ent);
-            if (!wasResisting && ent.Comp.Resisting)
+            if (ent.Comp.Resisting)
                 TryExtinguishWithWater(ent.Owner, ent.Comp);
             // RMC14 end
 
--- a/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
+++ b/Content.Server/Atmos/EntitySystems/FlammableSystem.cs
@@ -4,6 +4,7 @@ using Content.Server.Damage.Components;
 using Content.Server.Stunnable;
 using Content.Server.Temperature.Systems;
 using Content.Shared._RMC14.Atmos;
+using Content.Shared._RMC14.Water;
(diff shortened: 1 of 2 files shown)


Content.Server/Atmos/EntitySystems/FlammableSystem.cs
```diff
@@ -279,12 +281,37 @@ private void OnResistFireAlert(Entity<FlammableComponent> ent, ref ResistFireAle
             if (args.Handled)
                 return;
 
+            // RMC14 use the normal stop-drop-roll resist before active water extinguishes.
+            var wasResisting = ent.Comp.Resisting;
             _rmcFlammable.DoStopDropRollAnimation(ent.Owner);
             Resist(ent, ent);
+            if (!wasResisting && ent.Comp.Resisting)
+                TryExtinguishWithWater(ent.Owner, ent.Comp);
```
Vixting:
```suggestion
            _rmcFlammable.DoStopDropRollAnimation(ent.Owner);
            Resist(ent, ent);
            if (ent.Comp.Resisting)
                TryExtinguishWithWater(ent.Owner, ent.Comp);
```
Can just do this

Vixting (changes requested):
Very minor issue

AverageNotDoingAnythingEnjoyer (commented):
This should also extinguish mobs with acid on them unless im wrong

DrSmugleaf (approved):
Need to check the acid bit later

PR #9915, opened 2026-05-21
Ore box migration

CatAndHats:
## About the PR
Migrated the incorrect upstream ore boxes for our own

## Why / Balance
A bunch of our pvp and pve maps use the incorrect ore box, this fixes it.

## Requirements

**Changelog**

no CL


CatAndHats:
This won't work because the upstream ore boxes have rotation while ours don't and this will cause errors. Will need to be done manually.

PR #9919, opened 2026-05-21
Port terrain tile mechanics

GrigoriyGalintano:
## About the PR
Ports CMSS13-style terrain dig stages for RMC terrain tiles without map edits. E-tools now use explicit prototype-defined dig graphs for sand, Varadero sand, grass, shale, Hybrisa shale, ice, and snow/brown snow compatibility stages.

Also fixes terrain digging so surface cover entities like grass/bush overlays are removed when the tile underneath is dug down.

## Why / Balance
This makes terrain digging more predictable and closer to CMSS13 behavior while keeping it RMC-native and prototype-driven. Terrain stages now define their own dig target, material yield, and diggability instead of relying on implicit layer fallback.

The surface-cover cleanup prevents confusing visuals where grass remains floating over a newly dug dirt/sand tile.

## Technical details
- Added `rmcTerrainDigGraph` prototypes for supported staged terrain.
- Reworked `RMCTerrainSystem` to use explicit graph stages for dig/lower/raise/damage behavior.
- Added `RMCTerrainClearableComponent` marker for terrain surface cover.
- Digging/lowering terrain now clears anchored entities marked as terrain-clearable.
- Marked RMC bush/grass/tallgrass base prototypes as terrain-clearable.
- Added integration coverage for graph validation, stage behavior, and surface-cover cleanup.

## Media

https://github.com/user-attachments/assets/a95c047d-022f-40cf-8cd2-81186d52e769

## Requirements

**Changelog**
- add: Added staged RMC terrain digging for sand, grass, shale, ice, and snow terrain.


BarteG44:
would be nice if this included stuff like grass getting burnt by flames or explosions

PursuitInAshes (changes requested):
Thanks for the PR!
With us not wanting Shivas Snow mechanics, as discussed on discord, I am unsure if this is required.
I will refrain from closing for now, until I can confirm with the hosts and PVE GMs that they wouldn't want this for some other reason

PursuitInAshes:
Closing per above; thanks for the PR.

PR #9921, opened 2026-05-21
Rechargers No Longer Break Down Into Upstream Machine Frames

### LINKED ISSUE #8204: Recharger uses base ss14 steel
Recharger uses base ss14 steel and cables. When deconstructed with a crowbar, it doesn't drop a board to build it back up again. Which makes rechargers a limited resource and not buildable. And causes a station wide hunt when it's necessary instead of being built by MT or CE.

BlueMoon059:
## About the PR
Rechargers break down into upstreams machine frames right now, and dont give a circuitboard back - meaning they cant be replaced! I totally didnt do this my first time as a comtech and massively grief my team. Trust.
Fixes: https://github.com/RMC-14/RMC-14/issues/8204

## Why / Balance
- Feels bad to decon something you think should be fixable like any machine only to get back materials you cant even use.
- Less feels bad moments for new comstechs.
- Consistent with other machines.

## Technical details
-Adds a new CM circuitboard here Resources/Prototypes/_RMC14/Entities/Objects/Devices/Circuitboards/Machine/production.yml
- Multiple changes: Resources/Prototypes/_RMC14/Entities/Structures/Power/recharger.yml
1. Adds a new parent in the parent field, RMCConstructibleMachine.
2. Adds new construction graph code to overwrite the code from the parent (this is what causes the issue with the recharger deconstruction)
3. adds a new type field to define the circuitboard used in the machine.

## Media

https://github.com/user-attachments/assets/591c05f9-16c0-4564-971e-3fde76a5f632

## Requirements

**Changelog**
:cl:
- fix: The recharger no longer breaks down into an upstream machine frame.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Devices/Circuitboards/Machine/production.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Devices/Circuitboards/Machine/production.yml
@@ -162,3 +162,12 @@
 #   id: RMCElectronicsAirlock
 #   name: airlock electronics
 #   description: A machine printed circuit board for an airlock # TODO RMC14 Airlock Building
+
+- type: entity
+  parent: CMBaseMachineCircuitboard
+  id: RMCRechargerCircuitboard
+  name: recharger machine board
+  description: A machine printed circuit board for a recharger
+  components:
+    - type: MachineBoard
+      prototype: RMCRecharger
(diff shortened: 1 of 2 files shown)


BlueMoon059:
I've not put the circuitboard anywhere (no vendors or anything) as thats a decision above my station. But I can if needed!

DrSmugleaf (approved):

PR #9433, opened 2026-03-04
Trapper Boiler strain

MarshesAndMeadows:
## About the PR

Implemented the Trapper Boiler based on the CM13 version.

## Why / Balance
Xeno parity, featured in a number of issues;
[494#](https://github.com/RMC-14/RMC-14/issues/494)
[345#](https://github.com/RMC-14/RMC-14/issues/345)
[347#](https://github.com/RMC-14/RMC-14/issues/347)
[348#](https://github.com/RMC-14/RMC-14/issues/348)
[433#](https://github.com/RMC-14/RMC-14/issues/433)
[434#](https://github.com/RMC-14/RMC-14/issues/434)
[435#](https://github.com/RMC-14/RMC-14/issues/435)

## Technical details
Added three new active abilities and one passive (AcidMine, DeployTraps, AcidShotgun, Insight)
Each has its own components, systems, events, prototypes, etc.
In the case of DeployTraps it also has a system to handle the behavior of the deployed traps (aptly named DeployedTraps).
Added prototypes for two traps, the base version and the empowered version.
Modified the boiler prototype to allow for the new strain.
Added labels and descriptions to all abilities, strain info, names, etc.

Huge thanks to Hazzulicious for making the custom strain sprites, including new traps.
Also thanks to @CatAndHats for helping get parity sprites for traps.

## Media

https://github.com/user-attachments/assets/058792e3-58e5-447b-8e63-b2953e61195f

https://github.com/user-attachments/assets/987e000a-cdfb-418f-9c5b-4cb0ec9e4197

## Requirements
- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**
:cl:
- add: Added the Trapper boiler strain, a further specialized boiler focusing on ambushes and rapidly chewing through marine static defenses.
- add: Added the Deploy Traps ability, a rapid deployment of snaring traps to snare prey, able to even be deployed right under their feet, though easily destroyed by shooting or with a melee attack. Includes SFX for getting snared. Deploy Traps can also be empowered by Insight, and will in turn empower Acid Mine when cast this way.
[...]

PR #9923, opened 2026-05-22
XenoZoom now increases PVS range

MidZik:
## About the PR

XenoZoomSystem adds some PVS range to the entity, depending on its max zoom.

## Why / Balance

Kind of needed for the Sapper strain's zoom to work (#9433). It has a zoom level of 1.75, which the default PVS range is not suitable for. The sapper encounters frequent pop-out and pop-in several tiles from the viewport's edge.

## Technical details

XenoZoomSystem causes a PVS scale refresh when XenoZoomComponent starts up or shuts down.
XenoZoomSystem adjusts PVS scale depending on what the max zoom is.

## Requirements

**Changelog**

Ideally players will not notice any difference.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Zoom/XenoZoomSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Zoom/XenoZoomSystem.cs
@@ -112,7 +112,7 @@ public sealed class XenoZoomSystem : EntitySystem
             const float pvsOverheadEstimate = 0.2f;
             var scale = (Math.Max(ent.Comp.Zoom.X, ent.Comp.Zoom.Y) + pvsOverheadEstimate) / (1 + pvsOverheadEstimate);
             // args.Scale is added to the scale, not multiplied, so we have to subtract 1 and make sure we don't scale down.
-            args.Scale = Math.Max(0, scale - 1);
+            args.Scale += Math.Max(0, scale - 1);
         }
     }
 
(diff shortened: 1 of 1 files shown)


HorseCoded:
Dumb question, but would this also apply to more moderate zoom levels such as Runner's? If so, you could always changelog it as also being a 'bugfix' for reducing the pop-in they sometimes deal with while nyooming around.

MidZik:
> Dumb question, but would this also apply to more moderate zoom levels such as Runner's? If so, you could always changelog it as also being a 'bugfix' for reducing the pop-in they sometimes deal with while nyooming around.

It does apply to runners, but the increase is small for them because their zoom level is small. I am unsure how much it will help. It will probably be an improvement, but not enough to be noticeable when moving around at top runner speed. I think from my testing I still got pop-in on runner.

As such, I'm not comfortable documenting this as a bug-fix because although it's a slight improvement, it's not a true fix. Runner would need their own PVS radius increase to account for their speed.

Vixting (dismissed):
LGTM

MidZik:
Sorry, I had to update scaling code because it did end up having slight problems when the Sapper's zoom range is extended to 1.9. This calculation should work properly as long as the estimated overhead factor isn't too big.

I measured the default viewport to be 21 tiles wide and the CVars, as best as I can tell, define the PVS to be 25 tiles wide. That is just about a 20% overhead, which I hard-coded for now.

Vixting (dismissed):

DrSmugleaf (approved):
Thanks for the PR!

In the future doing this like binoculars do (spawning an offset entity that you use to load PVS around) would work better, but this is fine for now

MidZik:
Not entirely sure why but it seems on the server the pop-in is still quite bad. Might be that at high pop the server just can't keep up? Or maybe the offset entity might actually perform better.

PR #9924, opened 2026-05-22
Survivors Spawn Together - Fiorina

kleinerstation13:
## About the PR
Added five inserts where survivors spawn all together, Engi Ring insert by SharkSnake98

## Why / Balance
Survivor rework

Fiorina is a very weird map because the spawns are extremely spread apart, which essentially means the place where you spawn decides whatever you're going to be looting, the only way I found to make some kind of difference between the spawns while not locking survivors out of loot due to bad RNG was to cluster them around the middle. Diagonally the difference between them is minimal so the major difference is verticality, which should still vary up the timings enough to make each spawn play uniquely, but generally survivors can now dedicate their loot runs to the 4 major loot spots (Blue dorms, survivor camp, engi ring, souto) no matter where they spawn.

## Technical details
yaml and mapping

## Media
obraz

## Requirements

**Changelog**

:cl:
- tweak: Survivors now spawn together instead of separately on Fiorina. Their holdout locations include the Lowsec Cellblock, Garden, Engineering Ring, and Bobda Promo. Credit to Sharksnake98 for mapping the Engineering Ring insert.


kleinerstation13:
this is not getting the merge conflict tag applied for some reason but i'm unfucking the conflicts and mapping some more inserts to this so. into draft it goes

kleinerstation13:
Yay it works

Spawn locations should be slightly better now but I want @CatAndHats to approve this just incase

Resources/Maps/_RMC14/Inserts/Fiorina/surv_spawn_armory.yml
CatAndHats, line 1: Remove this insert since it is now unused

Resources/Maps/_RMC14/prison.yml
CatAndHats, line 1: When you took over the hive loot moving you did not remove the access lock from the flamer cabinet, survs cannot open it rn without looting an ID (which are really far away). Remove the access lock.

Resources/Prototypes/_RMC14/Entities/Markers/Inserts/fiorina.yml
CatAndHats, line 49: You added this insert so i'm asking you to fix this now.

This will spawn at the same time as Riot In Progress, causing the armoury to double-place walls and other things, this needs to only spawn if the Scenario is "none".

CatAndHats, line 51: the nogear insert you added
Image
Also the insert has two errors, first the APC is meant to be on the outside wall, it is one tile to the east and floating mid-air
There's a also a light in the wall that needs to be placed one tile further south, it also seems to be the old light maybe?

CatAndHats (changes requested):
Some things need fixing, the inserts themselves are great, ty!

Resources/Maps/_RMC14/Inserts/Fiorina/surv_spawn_armory.yml
kleinerstation13: Oh oops

Resources/Prototypes/_RMC14/Entities/Markers/Inserts/fiorina.yml
kleinerstation13: I don't think it's the old light I just didn't know how to place the lights when I mapped this, will be fixed

kleinerstation13:
resolved

CatAndHats (approved):
Looks good to me now

PursuitInAshes (approved):
Mapping seems good
The locations might have to be adjusted after we actually get some testing done on them, but it might just be that this is the best we can do because Fio is also "one of those weird maps"

PR #9930, opened 2026-05-23
Fixes welding not blinding you when repairing cades/weapons

JustMyselfUwU:
## About the PR

Makes welding barricades and mounted weapons without eye protection blind you

## Why / Balance

Parity, you are using a welder, which should blind you

## Technical details

Repairing is now considered a tool use event

## Requirements

**Changelog**

:cl:
- fix: Fixed barricades and mounted weapons not blinding you when welding them

### DIFF
--- a/Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
+++ b/Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
@@ -93,6 +93,9 @@ public sealed class RMCRepairableSystem : EntitySystem
             return;
         }
 
+        var toolUseAttempt = new ToolUseAttemptEvent(user, (float) repairable.Comp.FuelUsed);
+        RaiseLocalEvent(used, toolUseAttempt);
+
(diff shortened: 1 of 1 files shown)


Content.Shared/_RMC14/Repairable/RMCRepairableSystem.cs
```diff
@@ -93,6 +93,9 @@ private void OnRepairableInteractUsing(Entity<RMCRepairableComponent> repairable
             return;
         }
 
+        var toolUseAttempt = new ToolUseAttemptEvent(user, (float) repairable.Comp.FuelUsed);
+        RaiseLocalEvent(used, toolUseAttempt);
```
Thereisabear: This should be moved to after the guards. 
Line 140 on your PR or Line 137/138 on the original.  
After the return on the  If(!usefuel block  and before the  var delay =

Thereisabear (changes requested):

DrSmugleaf (approved):

PR #8823, opened 2025-12-23
Make organs inedible

noctyrnal:
## About the PR

title

## Why / Balance

marines kept being NRP and randomly eating organs, or accidentally eating them. no more NRP BE-baiting slop from my marines!

## Media

https://github.com/user-attachments/assets/f3dfc567-8dbf-409f-91ba-7c240e998e47

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

- [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl: noctyrnal
- add: Organs can no longer be eaten, you monster.

PR #9931, opened 2026-05-23
Fill All Organs with GreyMatter

PursuitInAshes:
## About the PR
Replaces #8823 . We'll leave it open to try this first and then only outright disable it if this still fails.
Gives the organ base 75 units of "Grey Matter" which will do 150 damage (Slightly less, like to 147, because the proteins will make them vomit or more depending on tox from proteins.) which will just barely crit someone or get them close. Grey matter processes faster and because of the quantity it means you have to take around 15~ bites of it to eat it entirely, so there is no "I didn't mean to eat it!" anymore, and if someone force feeds you, you have ample time to walk away.

The organs all get grinded into nutriment by default, so as long as nobody overwrites it later we're not giving out the ability to get "GreyMatter".

The only concern I have is the fact we don't reallyyy have a way to heal cellular damage, but its so small from one bite I am unsure if its a big issue for accidentally and it will stop people from eating them "for funny", probably needs discussed though.

## Why / Balance
Meant to discourage people from doing something that can become an RP issue really quickly.

## Technical details
Increased max vol and then filled it with grey matter, did the same to entities that would have overwritten it.

## Media
N/A

## Requirements

**Changelog**
:cl: PursuitinAshes
- tweak: Organs will now cause 75 units of grey matter to appear in your system, best to avoid eating them now.

### DIFF
--- a/Resources/Prototypes/Body/Organs/human.yml
+++ b/Resources/Prototypes/Body/Organs/human.yml
@@ -16,12 +16,12 @@
         - ReagentId: Nutriment
           Quantity: 10
       food:
-        maxVol: 35 # RMC14 - Increase Max Vol for CarpoToxin
+        maxVol: 80 # RMC14 - Increase Max Vol for GreyMatter
         reagents:
         - ReagentId: UncookedAnimalProteins
           Quantity: 5
-        - ReagentId: CarpoToxin # RMC14 - Add Carpotoxin to mechanically discourage eating organs.
-          Quantity: 30 # This is enough to deal 150 damage, after vomiting this will be slightly lower.
+        - ReagentId: GreyMatter # RMC14 - Add Grey Matter to all organs to mechanically discourage eating organs.
+          Quantity: 75 # This is enough to deal 150 clone damage, along vomiting/tox this will either place them just barely into crit or just below it.
   - type: FlavorProfile
     flavors:
       - people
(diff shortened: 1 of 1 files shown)


titowulk:
I'll just note that I am somewhat concerned this could lead to false positives with MP and the chef's sushi. If someone eats an unwarrantable amount of fish (which people absolutely do) I could see this leading to a false belief that someone may have eaten an organ. Leading to detainment if not outright brigging. Just a few days ago, I pulled someone with 40 carpotoxin in the system to med.

Similarly, if the chef is serving sushi, someone who has earnestly eaten organs could simply blame the chef. Possibly leading to the chef being yelled at by the MP for 'nearly killing someone'

Is there a different toxin source we could use?

AverageNotDoingAnythingEnjoyer:
We're supposed to have food processor which allows to remove toxins from food, such as xeno and carp meat
https://github.com/cmss13-devs/cmss13/blob/master/code/game/machinery/kitchen/processor.dm#L40

PursuitInAshes:
I'll see about changing it or making a different reagent to hopefully avoid the issue until we get the machine added. Not sure which will be easier

Drazamuffin:
Brains deal 5 genetic when eaten already, i am unsure of what reagent causes it though. I think it may just be a hard coded thing upon consumption which gets around being able to remove the toxin from food

CatAndHats (approved):
LGTM

DrSmugleaf (approved):

PR #9943, opened 2026-05-24
fix: prevent escape pods from landing on top of each other

Ian321:
## About the PR

This PR should fix the issue of escape pods and shuttles landing landing at the same position.

## Why / Balance

Happened to me in round 12251 that a pod crashed onto another preventing me from exiting my pod.

## Technical details

Hard to test because its random but the following should happen:

- Check if a blocker exists near the candidate location, try again if one does.
- Create a blocker on the planet on said location so that no other pod or lifeboat can crash land here.

The radius of `10` is definitely enough for pods but could be a bit too small for lifeboats. Might need some tuning if it keep happening.

## Media

https://github.com/user-attachments/assets/0261e572-41c2-4d00-9863-051b8ed613db

## Requirements

**Changelog**

:cl:
- fix: Escape pods and lifeboats should no longer crash land on top of each other.


Ian321:
Test fail unrelated and flaky.

Content.Shared/_RMC14/CrashLand/SharedCrashLandSystem.cs
```diff
@@ -252,6 +263,16 @@ public bool TryGetCrashLandLocation(out EntityCoordinates location)
...
                 location = _mapSystem.GridTileToLocal(grid, gridComp, tile);
+
+                if (blocking)
+                {
+                    var blockers = _entityLookup.GetEntitiesInRange<CrashLandingBlockedComponent>(location, _crashLandingBlockerRadius);
+                    if (blockers.Count > 0)
+                        continue;
+
+                    SpawnAtPosition(_crashLandingBlocker, location);
+                }
```
Vixting, line 276: make it query a reused hashset instead of looking them all up each time

Content.Shared/_RMC14/CrashLand/SharedCrashLandSystem.cs
```diff
@@ -54,6 +55,9 @@ public abstract partial class SharedCrashLandSystem : EntitySystem
 
     private EntityQuery<CrashLandableComponent> _crashLandableQuery;
 
+    private EntProtoId<CrashLandingBlockedComponent> _crashLandingBlocker = "RMCCrashLandingBlocker";
+    private float _crashLandingBlockerRadius = 10;
+
```
Vixting, line 61:
```suggestion
    private readonly HashSet<Entity<CrashLandingBlockedComponent>> _crashLandingBlockers = new();

```
reususe a hashset to store them

Vixting (changes requested):
Thanks for the pr, minor issue.

Content.Shared/_RMC14/CrashLand/SharedCrashLandSystem.cs
Ian321: How would I reset the hashset between rounds? The code is in shared, wouldn't that lead to miss-predictions as only the server would know all the locations?

Ian321: A better way might be to instead get all entities with the component and then check the distance for each.

Ian321: Looking into the implementation of `GetEntitiesInRange`, it already does that. So no need for a hashset because the ECS already does that.

Content.Shared/_RMC14/CrashLand/SharedCrashLandSystem.cs
```diff
@@ -252,6 +263,16 @@ public bool TryGetCrashLandLocation(out EntityCoordinates location)
                     continue;
 
                 location = _mapSystem.GridTileToLocal(grid, gridComp, tile);
+
+                if (blocking)
+                {
+                    var blockers = _entityLookup.GetEntitiesInRange<CrashLandingBlockedComponent>(location, _crashLandingBlockerRadius);
+                    if (blockers.Count > 0)
+                        continue;
```
Vixting:
```suggestion
                    _crashLandingBlockers.Clear();
                    _entityLookup.GetEntitiesInRange(location, _crashLandingBlockerRadius, _crashLandingBlockers);
                    if (_crashLandingBlockers.Count > 0)
```
like this

Ian321: If we always clear the set before filling it again what benefit do we get from having it?

Vixting: its just so you aren't creating a new hashset each time the loop runs

Ian321: Thanks for the explanation. I forgot that Clear only zeroes the memory and does not free it.

Vixting (approved):
LGTM

DrSmugleaf (approved):

PR #9963, opened 2026-05-29
Fixed three serious issues with xeno lag comp

### LINKED ISSUE #9957: Queen neurospit failing to register
Queen neuro shows as having hit on screen as queen but does not affect marines. Example videos in the discord bug thread linked below.

https://discord.com/channels/1168210010233376858/1467379328764346378

MidZik:
## About the PR

Predicted/lag compensated xeno projectiles should hit their targets much more reliably.

Leap-like abilities should hit their targets more reliably.

Fixes #9957.

## Why / Balance

Bugfix.

## Technical details

~~There were two issues I found.~~ Edit: Three now.

**Issue 1: Server was rejecting fair hits**

Clients tell the server that they got a hit on their end at a certain time. The server confirms that by rewinding the position of the target to where the client would have seen them, and then checking if the collision would be valid. However, the server was struggling to confirm the projectile's actual position due to physics substepping. The client was hitting their target at some physics substep, but the server had no knowledge of which substep the hit was on. Although the server tested some substeps, it didn't test all of them, and thus many hits were being rejected.

This is fixed by the client telling the server which substep the hit was on. The server then moves the projectile as if it was substepped that many times and THEN tests for a collision. From my testing, this is significantly more reliable.

**Issue 2: Engine bug regarding StartCollideEvent on clients**

Our version of RT has a bug where StartCollideEvent doesn't trigger on the client if the collision happens while the client is applying server state. This ended up happening VERY reliably, but only in a small band of distances from the target that varies based on your ping. The reason this is a problem is that the client simply DID NOT SEND a predicted hit event. At all. There was no lag comp because the server was never made aware the client hit something by the client itself.

This engine bug is fixed in newer versions of RT (any version released in 2026 or later should work), but in the mean time we have to use a workaround. I decided to use ProjectileHitEvent. This event does fire on StartCollideEvent, but its system ALSO fires it in an update loop which checks for collisions too.

**EDIT**
**Issue 3: Server processes predicted hit message before projectile is spawned**

Fairly simple compared to the above issues. Server processes the `XenoProjectilePredictedHitEvent` it receives before the processing that spawns the projectile occurs. This happens if you are near the target and hit them on the same tick that you made the projectile. For fast projectiles, this range is quite large. The result is that the predicted event is discarded because the projectile wasn't found.

Fixed this by storing the messages and processing them when the shooting actually happens.

## Media

I did not record a video of the fix, it just looks like the abilities working like they should. Instead here is a video of the issue in action, which from my testing no longer occurs:

https://github.com/user-attachments/assets/da0ee49c-0a68-4492-b023-80aeb5a6d856

## Requirements

**Changelog**

:cl:
- fix: Xeno projectile abilities should hit their targets much more reliably. (Spike Shed, Bone Spurs, Tail Seize, and all spit abilities.)
- fix: Leap-like abilities should hit their targets more reliably. (Leap, Pounce, Dash, Rush, Assault)

### DIFF
--- a/Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
+++ b/Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
@@ -1,5 +1,6 @@
-﻿using Content.Shared._RMC14.CCVar;
+using Content.Shared._RMC14.CCVar;
 using Content.Shared.Coordinates;
+using Robust.Shared;
 using Robust.Shared.Configuration;
 using Robust.Shared.Map;
 using Robust.Shared.Network;
@@ -22,8 +23,16 @@ public abstract class SharedRMCLagCompensationSystem : EntitySystem
     public float MarginTiles { get; private set; }
 
     private EntityQuery<ActorComponent> _actorQuery;
+    private int _substeps;
+    private float _substepTime;
+    // This contains _substeps + 1 items and is just an array of ratios
+    // representing how far into a frame each substep is. E.g. if there are four substeps
+    // the array would be [0.0f, 0.25f, 0.5f, 0.75f, 1.0f]
+    private float[] _substepMults = [];
+    private bool _logPrediction = false;
 
     private readonly Dictionary<NetUserId, GameTick> _lastRealTicks = new();
+
 
     public override void Initialize()
     {
@@ -34,11 +43,27 @@ public abstract class SharedRMCLagCompensationSystem : EntitySystem
         SubscribeNetworkEvent<RMCSetLastRealTickEvent>(OnSetLastRealTick);
 
         Subs.CVar(_config, RMCCVars.RMCLagCompensationMarginTiles, v => MarginTiles = v, true);
+        Subs.CVar(_config, CVars.NetTickrate, UpdateSubsteps, true);
+        Subs.CVar(_config, CVars.TargetMinimumTickrate, UpdateSubsteps, true);
     }
 
     private void OnSetLastRealTick(RMCSetLastRealTickEvent msg, EntitySessionEventArgs args)
     {
         SetLastRealTick(args.SenderSession.UserId, msg.Tick - 1);
+    }
+
+    private void UpdateSubsteps(int _)
+    {
+        // This is just ripped out from SharedPhysicsSystem
+        var targetMinTickrate = (float)_config.GetCVar(CVars.TargetMinimumTickrate);
+        var serverTickrate = (float)_config.GetCVar(CVars.NetTickrate);
+        _substeps = (int)Math.Ceiling(targetMinTickrate / serverTickrate);
+        _substepTime = 1.0f / serverTickrate / _substeps;
+
+        // Division is slow so we pre-calculate the multipliers we need here.
+        _substepMults = new float[_substeps + 1];
+        for (var i = 0; i <= _substeps; i++)
+            _substepMults[i] = (float)i / _substeps;
     }
 
     public virtual (EntityCoordinates Coordinates, Angle Angle) GetCoordinatesAngle(EntityUid uid,
@@ -110,7 +135,7 @@ public abstract class SharedRMCLagCompensationSystem : EntitySystem
         RaiseNetworkEvent(new RMCSetLastRealTickEvent(GetLastRealTick(null)));
     }
 
-    public bool Collides(Entity<FixturesComponent?> target, Entity<PhysicsComponent?> projectile, MapCoordinates targetCoordinates)
+    public bool Collides(Entity<FixturesComponent?> target, Entity<PhysicsComponent?> projectile, MapCoordinates targetCoordinates, int substep = 0)
     {
         if (!Resolve(target, ref target.Comp, false) ||
             !Resolve(projectile, ref projectile.Comp, false))
@@ -118,8 +143,12 @@ public abstract class SharedRMCLagCompensationSystem : EntitySystem
             return false;
         }
 
+        substep = Math.Clamp(substep, 0, _substeps);
+
         var projectileCoordinates = _transform.GetMapCoordinates(projectile);
-        var projectilePosition = projectileCoordinates.Position;
+        var projectileVelocity = _physics.GetLinearVelocity(projectile, projectile.Comp.LocalCenter);
+        var substeppedProjectilePos = projectileCoordinates.Position + (projectileVelocity / _timing.TickRate) * _substepMults[substep];
+
         var transform = new Transform(targetCoordinates.Position, 0);
         var bounds = new Box2(transform.Position, transform.Position);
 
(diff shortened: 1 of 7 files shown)


MidZik:
I did some more testing and noticed that there is actually a _third_ issue somewhere that causes close-range prediction to fail. This doesn't appear to be caused by my PR, but rather is an issue that this PR doesn't fix.

I'm going to investigate it and hopefully tackle that in this PR as well.

MidZik:
Root caused and fixed a third issue that caused predicted hits close to your target to not register.

MidZik:
Lag comp is hard. I'm testing some more changes that I'm hoping will be more reliable than what this is right now. Will push them when I'm comfortable with them.

MidZik:
Projectiles should work quite well now, in my testing. There are _still_ edge cases that pop up occasionally, but I can't reliably reproduce them anymore in my dev environment.

I'm working on making the leap lag comp much better now. I'll push those changes here as well at some point. If desired I can make a leap-specific PR instead, though, that depends on this one.

MidZik:
Ugh, I just realized I actually made leaps a bit worse. I need to correct some things for leaps.

MidZik:
Alright, I fixed the leap part. The server should be better about confirming leap hit predictions.

Leaps are still jank as hell as far as lag comp goes, but that is beyond the scope of this PR. Leaps will need some substantial work to reduce the rubberbanding clients get with them.

Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
```diff
@@ -131,28 +164,98 @@ public bool Collides(Entity<FixturesComponent?> target, Entity<PhysicsComponent?
...
+                  Substep:        {substep}
+                  Projectile Pos: {substeppedProjectilePos}
+                  Target Pos:     {targetCoordinates.Position}
+                  Proj AABB:      {projectileBounds.BottomLeft}
+                                  {projectileBounds.TopRight}
+                  Target AABB:    {targetBounds.BottomLeft}
+                                  {targetBounds.TopRight}
+                  AABB Intersect? {targetBounds.Intersects(projectileBounds)}
+                  AABB Distance:  {Math.Sqrt(AABBDistanceSquared(targetBounds, projectileBounds))}
+                """);
```
DrSmugleaf, line 206: For next time I debug this on live, this would be nice if it included the name of the shooter's session
You would have to pass in the session into this method to do that

Content.Shared/_RMC14/Movement/SharedRMCLagCompensationSystem.cs
```diff
@@ -131,28 +164,98 @@ public bool Collides(Entity<FixturesComponent?> target, Entity<PhysicsComponent?
...
 
-        var closest = bounds.ClosestPoint(projectilePosition);
-        if ((closest - projectilePosition).LengthSquared() <= MarginTiles * MarginTiles)
+        if (AABBDistanceSquared(targetBounds, projectileBounds) <= MarginTiles * MarginTiles)
+        {
             return true;
+        }
+
+        if (_logPrediction)
+            Log.Warning("Predicted hit denied.");
```
DrSmugleaf: Same here

Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
```diff
@@ -214,6 +284,39 @@ private void OnClusterSpawned(Entity<XenoProjectileComponent> ent, ref CMCluster
...
+
+        for (var i = 0; i < _earlyMessages.Count; ++i)
+        {
+            var item = _earlyMessages[i];
+            if (item.PredictedHitTick < _timing.CurTick)
+            {
+                if (_logPrediction)
+                    Log.Warning($"Removed expired prediction message: Shooter {item.Shooter}, Shot ID {item.Message.Id}");
+                _earlyMessages[i] = _earlyMessages[_earlyMessages.Count - 1];
+                _earlyMessages.RemoveAt(_earlyMessages.Count - 1);
```
DrSmugleaf: You should be able to use _earlyMessages.RemoveSwap here to do the same thing, though do check if it's actually the same or not
You could also just iterate this in reverse, but that depends on how often you think line 309-310 is going to run

DrSmugleaf (changes requested):
Thanks for the PR!

Content.Shared/_RMC14/Xenonids/Projectile/XenoProjectileSystem.cs
MidZik: RemoveSwap is an extension method that I have small concerns with, enough to prefer leaving this version. Reverse iteration should be fine, probably better since with shotgun spits there will often be contiguous sections that match.

MidZik:
Thanks for the review!

MidZik:
I haven't investigated this fully, but I believe there is one other issue. If the server shoots the projectile late compared to what the client predicted, this code will end up adjusting the projectile in the wrong direction. I think the predicted hit message would also need to store the time that the client predicted the shot began, not just when it hit, so that the server can reconcile it properly.

MidZik:
I refactored it quite a bit, should be much easier to understand what the code is doing now. Added more verbose logging as well.

In my testing, I rarely encounter a strange mispredict. The client somehow "accelerates" its projectile forward. For an example, see the logs below. Shot ID 52 is the offender and I included 51 and 53 for context. The shots I was taking were mostly identical, against a stationary target.

For 51 and 53, the client predicts a hit 2 ticks after the shot. But for 52, the client predicted a hit just 1 tick after the projectile was shot. I think this is actually a mispredict on the client somehow, rather than an issue with this lag comp code.

Server logs:
```
[DEBG] system.xeno_projectile: Received predicted hit:
  Session:  localhost@JoeGenero
  CurTick:  6888
  Target:   2219
  ShotID:   51
  Shot At:  6886
  Hit Tick: 6888
  Substep:  0
[DEBG] system.xeno_projectile: Predicted hit checks passed, will test collision. Details:
  Session Name:   localhost@JoeGenero
  Last Real Tick: 6880
  Shot ID:        51
  During shoot?   False

  Cur Tick:       6888
  Pred Hit Tick:  6888

  Shot Tick:      6886
  Pred Shot Tick: 6886

  Pred Substep:   0
  Adjust Substep: 0
[DEBG] system.r_m_c_lag_compensation: Lag comp collide data:
  Session Name:   localhost@JoeGenero
  Pre-Substep
    Proj Coords:  Map=1, X=0.26, Y=2.14
  CurTick:        6888
  Substep:        0
  Projectile Pos: <0.26002026, 2.141442>
  Target Pos:     <-0.04774299, 2.0958056>
  Proj AABB:      <0.15002026, 2.0314422>
                  <0.37002024, 2.251442>
  Target AABB:    <-0.397743, 1.7458056>
                  <0.302257, 2.4458055>
  AABB Intersect? True
  AABB Distance:  0
[DEBG] system.xeno_projectile: Predicted hit from 'localhost@JoeGenero' ++ CONFIRMED!! ++
[DEBG] system.xeno_projectile: Received predicted hit:
  Session:  localhost@JoeGenero
  CurTick:  6968
  Target:   2219
  ShotID:   52
  Shot At:  6967
  Hit Tick: 6968
  Substep:  0
[DEBG] system.xeno_projectile: Predicted hit checks passed, will test collision. Details:
  Session Name:   localhost@JoeGenero
  Last Real Tick: 6960
  Shot ID:        52
  During shoot?   False

  Cur Tick:       6968
  Pred Hit Tick:  6968

  Shot Tick:      6967
  Pred Shot Tick: 6967

  Pred Substep:   0
  Adjust Substep: 0
[DEBG] system.r_m_c_lag_compensation: Lag comp collide data:
  Session Name:   localhost@JoeGenero
  Pre-Substep
    Proj Coords:  Map=1, X=1.31, Y=0.49
  CurTick:        6968
  Substep:        0
  Projectile Pos: <1.3127991, 0.4916558>
  Target Pos:     <-0.04774299, 2.0958056>
  Proj AABB:      <1.2027991, 0.3816558>
                  <1.4227991, 0.6016558>
  Target AABB:    <-0.397743, 1.7458056>
                  <0.302257, 2.4458055>
  AABB Intersect? False
  AABB Distance:  1.4560407692891966
[WARN] system.r_m_c_lag_compensation: Predicted hit denied for session 'localhost@JoeGenero'
[WARN] system.xeno_projectile: Predicted hit from 'localhost@JoeGenero' -- denied --
[DEBG] system.xeno_projectile: Received predicted hit:
  Session:  localhost@JoeGenero
  CurTick:  7091
  Target:   2219
  ShotID:   53
  Shot At:  7089
  Hit Tick: 7091
  Substep:  0
[DEBG] system.xeno_projectile: Predicted hit checks passed, will test collision. Details:
  Session Name:   localhost@JoeGenero
  Last Real Tick: 7082
  Shot ID:        53
  During shoot?   False

  Cur Tick:       7091
  Pred Hit Tick:  7091

  Shot Tick:      7089
  Pred Shot Tick: 7089

  Pred Substep:   0
  Adjust Substep: 0
[DEBG] system.r_m_c_lag_compensation: Lag comp collide data:
  Session Name:   localhost@JoeGenero
  Pre-Substep
    Proj Coords:  Map=1, X=0.10, Y=2.03
  CurTick:        7091
  Substep:        0
  Projectile Pos: <0.09594649, 2.0308485>
  Target Pos:     <-0.04774299, 2.0958056>
  Proj AABB:      <-0.0140535105, 1.9208485>
                  <0.20594649, 2.1408484>
  Target AABB:    <-0.397743, 1.7458056>
                  <0.302257, 2.4458055>
  AABB Intersect? True
  AABB Distance:  0
[DEBG] system.xeno_projectile: Predicted hit from 'localhost@JoeGenero' ++ CONFIRMED!! ++
```

Client logs:
```
[DEBG] system.xeno_projectile: SENDING PREDICTED PROJECTILE HIT!!
  ShotId:         51
  ShotAtTick:     6886
  CurTick:        6888
  LastRealTick:   6880
  Phys Substep:
  In simulation?  True
  ApplyingState?  False
  FirstTimePred?  True
  PredictedTick:  6888
  Substep:        0
  ShotCoords:     EntId=2, X=0.26, Y=2.14
  Target Coords:  EntId=2, X=-0.05, Y=2.10
[DEBG] system.xeno_projectile: SENDING PREDICTED PROJECTILE HIT!!
  ShotId:         52
  ShotAtTick:     6967
  CurTick:        6968
  LastRealTick:   6960
  Phys Substep:
  In simulation?  True
  ApplyingState?  False
  FirstTimePred?  True
  PredictedTick:  6968
  Substep:        0
  ShotCoords:     EntId=2, X=0.26, Y=2.14
  Target Coords:  EntId=2, X=-0.05, Y=2.10
[DEBG] system.xeno_projectile: SENDING PREDICTED PROJECTILE HIT!!
  ShotId:         53
  ShotAtTick:     7089
  CurTick:        7091
  LastRealTick:   7082
  Phys Substep:
  In simulation?  True
  ApplyingState?  False
  FirstTimePred?  True
  PredictedTick:  7091
  Substep:        0
  ShotCoords:     EntId=2, X=0.10, Y=2.03
  Target Coords:  EntId=2, X=-0.05, Y=2.10
  ```

DrSmugleaf (approved):
God bless thank you

PR #9968, opened 2026-05-29
Fix client mispredicting actions

### LINKED ISSUE #8702: Ghost Actions
## Description

Actions that have cooldowns can be triggerd clientside only causing them to emit sound or emit projectiles that don't really exist. It also removes the action selection (if applicable) making it so you have to select the action again to actually use it.

**Reproduction**

Go into dev environemnt or the main game server.
Use any action that has a cooldown.
Cast it at the very last second before it is ready.

**Video**

https://github.com/user-attachments/assets/8a35eef5-a5b6-42d2-8340-752c0f533be7

MidZik:
## About the PR

Client no longer allows you to use actions before their cooldown is actually up. Actions are now run "in simulation" rather than "out of simulation".

Fixes #8702.

Might require changes in #9963. That PR has a small bit of jank regarding how many substeps the server has to run compared to the client, and I fear that this out-of-simulation action behavior could be the cause of that jank.

~~Edit: I tested these two PRs merged together, and yep, I will need to fix #9963 if this PR is merged. The jank was being caused exactly by this issue, where client actions were predicted out-of-simulation.~~
Edit2: Actually probably not. Testing it is a bit difficult, I now believe this PR doesn't cause issues with it.

## Why / Balance

Bugfix.

## Technical details

Actions are UI. UI is considered out of simulation. Out of simulation timing does not use the same "CurTime" as in simulation timing. Basically every action we have does in-simulation things. The end result is, the client tries to predict a lot of "in simulation" things while the engine says it's "out of simulation". It goes poorly.

The root cause of the ghost action issue is that when the client tries to use an action, it compares CurTime to the action's cooldown EndTime. But because we're "out of simulation", CurTime can actually be a value in the middle of a tick. That means if you use the action late enough in the final tick of the cooldown, the client will let you, because CurTime advanced past the EndTime _during the tick_ that EndTime is in. In simulation that is not possible, you would have to wait for the next tick for CurTime to advance past EndTime.

I am honestly not sure how best to deal with this. This feels like a big oversight by basically all actions in content (both ours and upstream). It seems every action has this problem.

The fix I decided on is very forceful but seems effective. I set `InSimulation` to be true during all the action code, and reset it to its old value when done. This only needs to be done on the client.

This seems to work quite well. A lot of the timing code actually adjusts itself based on the value of `InSimulation`. `CurTick` is calculated based on the value of `InSimulation`. So all the cooldown checks end up working just fine after doing this. All the client code that runs off of an action will also end up having the correct `CurTick` so abilities might actually be predicted much better as a result.

I tested several abilities and they seem to work fine. The root cause is completely dealt with so there are no more ghost actions. However, the effects of this fix could be extensive. This whole time, the client has been predicting actions out-of-simulation. I cannot predict for certain how every action will behave due to this change, but I expect most of the effects should actually be positive.

## Requirements

**Changelog**

:cl:
- fix: Clients will no longer let you use actions before their cooldown is actually over, which resulted in "ghost actions".

### DIFF
--- a/Content.Client/UserInterface/Systems/Actions/ActionUIController.cs
+++ b/Content.Client/UserInterface/Systems/Actions/ActionUIController.cs
@@ -13,6 +13,7 @@ using Content.Client.UserInterface.Systems.Actions.Widgets;
 using Content.Client.UserInterface.Systems.Actions.Windows;
 using Content.Client.UserInterface.Systems.Gameplay;
 using Content.Shared._RMC14.Actions;
+using Content.Shared._RMC14.GameStates;
 using Content.Shared.Actions;
 using Content.Shared.Actions.Components;
 using Content.Shared.Charges.Systems;
@@ -52,6 +53,9 @@ public sealed class ActionUIController : UIController, IOnStateChanged<GameplayS
     [UISystemDependency] private readonly InteractionOutlineSystem? _interactionOutline = default;
     [UISystemDependency] private readonly TargetOutlineSystem? _targetOutline = default;
     [UISystemDependency] private readonly SpriteSystem _spriteSystem = default!;
+
+    // RMC14
+    [UISystemDependency] private readonly SharedRMCGameStateSystem _rmcGameState = default!;
 
     private ActionButtonContainer? _container;
     private readonly List<EntityUid?> _actions = new();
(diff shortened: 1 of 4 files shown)


DrSmugleaf (commented):
I need to look at this more in-depth later, not sure if there's a cleaner way to do this, ty for looking into it

MidZik:
Looked into this more when investigating interactions (patting, tail stabbing), because they suffer from this "out of simulation" problem that is causing slight mispredicts and making lag compensation testing a bit difficult. Simply setting "InSimulation" to true does not help fix the issues. I believe the other player's positions are already interpolated so although the CurTick and CurTime might be correct (and that fixes this ghost action issue) it does not fix using interpolated entity positions.

I think a proper fix is to make InSimulation actions be processed in an Update loop, just like melee actions are. Melee actions do not suffer from this OutOfSimulation problem at all and thus have much more accurate prediction.

I'll mark this as draft for now, since this definitely isn't the correct way to fix this issue as it exists.

PR #9970, opened 2026-05-29
Hijack explosions

GrigoriyGalintano:
## About the PR

Adds randomized Almayer hijack crash destruction.

When a hijacked dropship lands, valid ship structures on the landing map can now be randomly damaged or broken, and RMC pipes start a delayed explosion barrage with warning alerts and local fire.

## Why / Balance

This makes hijack landings feel more destructive without wiping whole structure categories at once.

Damage is percentage-based and random per category, so the result varies between rounds. Indestructible hull structures remain protected: random structural hijack damage only affects entities with normal `Damageable` and `Destructible` behavior.

Pipe explosions are delayed and telegraphed, giving nearby players a short warning before the burst.

## Technical details

Added `RMCHijackRandomDamageSystem`, which listens to `DropshipHijackLandedEvent` and builds target pools only from entities on the landing map.

Added `RMCHijackRandomDamageTargetComponent` with category and damage configuration for:

- walls;
- windows;
- windoors;
- pipes.

Random structural damage uses these ranges:

- walls: 3-6%;
- windows: 30-50%;
- windoors: 50-100%.

Selected wall/window/windoor targets are shuffled and split between damage-only and break outcomes. Damage-only applies only enough damage to reach the configured target amount, while break damage uses the entity's existing destructible flow.

Structural targets are filtered at runtime to require both `DamageableComponent` and `DestructibleComponent`, so indestructible hull walls/windows are not made destructible by hijack damage.

Marked destructible Almayer wall, window, and windoor prototypes as hijack random damage targets. Marked `RMCBasePipe` as a pipe barrage target.

Added pipe barrage behavior:

1. initial barrage selects 3-5% of remaining pipes;
2. later barrages select 1-1.5% every 15 seconds;
3. each selected pipe shows `RMCHijackPipeExplosionWarning`;
4. the pipe plays a hiss warning for 5 seconds;
5. the pipe explodes using an RMC explosion;
6. local `RMCHijackPipeFire` is spawned;
7. the original pipe is replaced with `GasPipeBroken`;
8. used/pending pipes are not selected again.

Added `RMCHijackPipeExplosionWarning` and `RMCHijackPipeFire`.

Updated legacy `RMCReplaceOnHijackLand` handling to only affect the hijack landing map and preserve rotation when replacing entities.

Updated `FireVisualizerSystem` for RMC fire color support and delayed startup initialization, with RMC changes marked in the upstream file.

## Media

https://github.com/user-attachments/assets/0fd2f572-a9d9-4397-b965-15b4bf16df42

## Requirements

**Changelog**

- add: Added randomized Almayer damage after hijacked dropship crashes.
- add: Added delayed hijack pipe explosion barrages with warning alerts and local fire.

### DIFF
--- a/Content.Shared/_RMC14/Hijack/RMCHijackRandomDamageTargetComponent.cs
+++ b/Content.Shared/_RMC14/Hijack/RMCHijackRandomDamageTargetComponent.cs
@@ -33,12 +33,6 @@ public sealed partial class RMCHijackRandomDamageTargetComponent : Component
     /// </summary>
     [DataField]
     public DamageSpecifier? BreakDamage;
-
-    /// <summary>
-    ///     Blocks ordinary damage while still allowing damage applied by the hijack system.
-    /// </summary>
-    [DataField]
-    public bool HijackDamageOnly;
 }
 
 /// <summary>
--- a/Content.Shared/_RMC14/Hijack/RMCHijackRandomDamageTargetComponent.cs
+++ b/Content.Shared/_RMC14/Hijack/RMCHijackRandomDamageTargetComponent.cs
@@ -2,25 +2,48 @@ using Content.Shared.Damage;
 
 namespace Content.Shared._RMC14.Hijack;
 
+/// <summary>
+///     Marks an entity as a possible random damage target when a hijacked dropship lands.
+/// </summary>
 [RegisterComponent]
 public sealed partial class RMCHijackRandomDamageTargetComponent : Component
 {
+    /// <summary>
+    ///     Whether this entity can currently be selected by the hijack damage system.
+    /// </summary>
     [DataField]
     public bool Enabled = true;
 
+    /// <summary>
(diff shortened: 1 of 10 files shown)


GitHubUser53123:
does this account for pipes being unanchored?

Content.Shared/_RMC14/Hijack/RMCHijackRandomDamageTargetComponent.cs
```diff
@@ -0,0 +1,47 @@
+using Content.Shared.Damage;
+
+namespace Content.Shared._RMC14.Hijack;
+
+/// <summary>
+///     Marks an entity as a possible random damage target when a hijacked dropship lands.
+/// </summary>
+[RegisterComponent]
```
DrSmugleaf, line 8: Might want to make this networked in the future but this is fine for now

DrSmugleaf (approved):
Thanks for the PR!

PR #9975, opened 2026-05-30
"Why do you refuse to gain Power? The power of our father Sparda." (I.e. fixing TODO RMC14 Power)

### LINKED ISSUE #9454: A bunch of stuff doesn't behave with power correctly
## Description

Mostly machinery, some consoles dont turn off when unpowered and dont consume power, for example

**Reproduction**

**Screenshots**

**Additional context**

AverageNotDoingAnythingEnjoyer:
## About the PR

Adds power consumption stats to a lot of stuff, computers, scanners, prop computers, cameras, etc. etc.
This fixes most of TODO RMC14 Power comments and also fixes #9454 while also potentially dealing with #9070

Exception is Intercoms, for which i couldn't find power values on CM, dropship fab as it requires modifying C# and also brig timer as we should have our own instead of using upstream which should come in a separate PR.

## Why / Balance

Parity

## Technical details

Makes the instances of upstream ApcPowerReceiver in RMC/CM entities use fixed 250 watts and need power, while adding the correct stats to RmcApcPowerReceiver
Also some entities do not consume power at all but require it in order to function, example is Booze-o-mat
Unfortunately as each machinery may have a unique rsi structure, some may have incorrect nopower visuals, but i tested most of it

## Media

Screenshot 2026-05-30 at 15 08 13

## Requirements

**Changelog**

:cl: J C Denton
- fix: Fixed a lot of entities working off infinite inner power.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/rmc_intel.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/rmc_intel.yml
@@ -133,7 +133,11 @@
   components:
   - type: Sprite
     sprite: _RMC14/Structures/Machines/computer.rsi
-    state: techweb
+    layers:
+    - map: ["base"]
+      state: "techweb0"
+    - map: ["enum.PowerDeviceVisualLayers.Powered"]
+      state: "techweb"
   - type: InteractionOutline
   - type: ActivatableUI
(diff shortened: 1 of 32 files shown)


DrSmugleaf (approved):
Thanks for the PR!

PR #9978, opened 2026-05-30
Allow plastic explosives to stick to mobs

GrigoriyGalintano:
## About the PR

Allows plastic explosives to be planted on mobs, including xenos, using the existing sticky explosives behavior.

Adds mechanical safeguards for planted explosives:
- plastic explosives cannot be planted on mobs from the same faction;
- xenos can dissolve planted C4 with acid;
- active plastic explosives and breaching charges can be disarmed with a multitool.

## Why / Balance

This brings plastic explosives closer to CM13 behavior, where C4 can be planted on mob targets.

The change keeps the existing plastic explosive 5 second stick do-after and 10 second detonation timer, so planting C4 on a mob requires commitment and can be interrupted. To prevent abuse, C4 cannot be attached to same-faction targets, using RMC hive/IFF checks instead of static faction assumptions.

Counterplay was also added. Xenos can remove planted C4 with acid, and engineers can disarm active plastic explosives or breaching charges with a multitool. Breaching charges are still not plantable on mobs.

## Technical details

Expanded the `Sticky` whitelist for `RMCExplosivePlastic` to include `MobState`.

Added `RMCPreventSameFactionStickComponent` and system logic that blocks sticking explosives to same-faction mob targets. The check uses existing RMC hive and IFF faction events so runtime faction/ID changes are respected.

Added `RMCC4DisarmableComponent` and shared disarm logic for active C4-style explosives:
- xeno acid can dissolve planted C4;
- multitools can disarm active plastic explosives and breaching charges;
- disarm interaction works when clicking the charge directly or the mob/wall it is attached to;
- stuck charges override accessibility/range checks so they do not appear unusable while inside the target's contents.

The existing `StickySystem` still handles the planting do-after and attachment. The existing `RMCExplosionSystem` still starts the timer through `EntityStuckEvent`.

`RMCExplosiveDelete` was not expanded to mobs, so planted plastic explosives do not directly delete the target entity. They detonate normally through the existing explosive system.

Breaching charge multitool disarm was set slightly below its 3 second fuse so the SS14 do-after/timer ordering does not make CM13-style disarming impossible.

## Requirements

**Changelog**

- tweak: Plastic explosives can now be planted on mobs.
- add: Active plastic explosives and breaching charges can now be disarmed with a multitool.
- add: Xenos can now dissolve planted C4 with acid.

### DIFF
--- a/Content.Shared/_RMC14/Explosion/RMCC4DisarmableComponent.cs
+++ b/Content.Shared/_RMC14/Explosion/RMCC4DisarmableComponent.cs
@@ -1,0 +1,36 @@
+using Content.Shared._RMC14.Xenonids.Acid;
+using Robust.Shared.GameStates;
+
+namespace Content.Shared._RMC14.Explosion;
+
+[RegisterComponent, NetworkedComponent, AutoGenerateComponentState]
+[Access(typeof(RMCC4DisarmableSystem))]
+public sealed partial class RMCC4DisarmableComponent : Component
+{
+    [DataField, AutoNetworkedField]
+    public TimeSpan AcidDelay = TimeSpan.FromSeconds(4);
+
+    [DataField, AutoNetworkedField]
+    public XenoAcidStrength MinimumAcidStrength = XenoAcidStrength.Weak;
+
+    [DataField, AutoNetworkedField]
+    public TimeSpan MultitoolDelay = TimeSpan.FromSeconds(3);
+
+    [DataField, AutoNetworkedField]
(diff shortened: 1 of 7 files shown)


MidZik:
CM13 has rules to prevent abusing this mechanic. You are not allowed to attach the C4 to a mob or object and then move it around, for example.

I have strong doubts that this PR will even be considered without adding mechanical enforcement for possible abuses, assuming this behavior is even desired in the first place.

GrigoriyGalintano:
finish

PR #9987, opened 2026-06-01
Fix Vent Crawling saving bug

CyberTropic:
## About the PR

Fixes oversight from https://github.com/RMC-14/RMC-14/pull/9803 that caused issues with pipe rotations on saving map.
Tested saving + loading maps + spawning map inserts and vent crawling around. 

## Why / Balance

Bugfix

## Technical details

Moved pipe rotations to MapInit, so travel direction doesn't get overwritten in the map editing mode. Added server only checks for functions, which I think was introducing some multiple rotations issues.

## Requirements

**Changelog**

Not player facing

### DIFF
--- a/Content.Shared/_RMC14/Vents/SharedVentCrawlingSystem.cs
+++ b/Content.Shared/_RMC14/Vents/SharedVentCrawlingSystem.cs
@@ -54,7 +54,7 @@ public abstract class SharedVentCrawlingSystem : EntitySystem
 
         SubscribeLocalEvent<VentExitComponent, VentExitDoafterEvent>(OnVentExitDoafter);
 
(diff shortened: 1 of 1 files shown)


CatAndHats (commented):
High priority due to a current bug where saving a map will cause pipes to choose directions and then subsequently breaking.

I've tested the PR with saving/loading maps and it appears to fix the issue, however I cannot speak for the c#. Ty for the PR cyber

Dogborg (approved):

DrSmugleaf (approved):
Have to check if this fixes it for any existing saved broken maps, or for non-insert pipes, but this is good for now ty

PR #9988, opened 2026-06-01
EZ-11A Smart Tank Pouch

x5tink:
Mainly, This PR Adds a new EZ type Pouch, the Smart Refill Tank Adapter Pouch. Its essentially the same as a Pcan Pouch, but takes a smart can instead, and comes with an E-XS (1x15) needle. This means it stores only 180u, and saves printing needles to dump out the default M needles in the pcan pouches. This pouch requires no medical skill to use, and is aimed at for riflmen and general marines use.

It also adds both the Smart Refill Tank and the Smart Refill Tank Adapter Pouch to the medilathe printer.

Further more it Adjusts the regular Pressurized reagent canister pouch to require Medical Skill 2, so they are still use-able for HM's and other medical staff, but not by regular marines.

## Technical details
New objects and sprites for the Smart Refill Tank Adapter Pouch.
New component to check skill level on item equip.
Largely YAML changes.

## Media
Ignore the Janibucket/Jug stuff in this video, originally it was part of this PR but is now Part of [this](https://github.com/RMC-14/RMC-14/pull/9989)

https://github.com/user-attachments/assets/df94fbbb-777b-4301-8538-4f32e1d86d28

image

## Requirements

**Changelog**
- add: Smart Refill Tank Adapter Pouches
- tweak: Pressurized reagent canister pouch requires med skill


BarteG44:
would be nice if they were made easier to get than begging the CMO to print 1 bajilion of them if they're meant to be specifically for riflemen. generally i do agree that 495u of GC on some average rifleman is a bit too much (speaking as a 400u GC rifleman myself) but the PR might not get merged for the whole "parity first" reason

also a more sovlful name than an "adapter" would be nice

x5tink:
> would be nice if they were made easier to get than begging the CMO to print 1 bajilion of them if they're meant to be specifically for riflemen. generally i do agree that 495u of GC on some average rifleman is a bit too much (speaking as a 400u GC rifleman myself) but the PR might not get merged for the whole "parity first" reason
> 
> also a more sovlful name than an "adapter" would be nice

Another method to obtain these might be possible, maybe like the rfn pill pouch but this instead, with basic chems. as it has a material cost associated i would expect it would need to be a point buy item, but im unsure as to the balance associated with that.

How ever good news, they should be much more available that PCan pouches, because they are cheaper, and you dont need to print the needles separately. Meaning, we can print alot more of them, in the same amount of time. As such i would expect more to be made and put out for use on average.

And yes a better name would be nice, feel free to suggest one. I was originally calling it the Smart Canister pouch, before i realized the they were called Refill tanks... And when i made the sprite, i sort of imagined it to be something you clip onto the top of the bottle, that allows it to work with custom needles, not just the standard auto injectors.

BarteG44:
I'd personally make it like a 15-20 point buy in vendors but that sort of balancing probably needs more discussion

Also the name could be changed to something simple like MS-11 Injector Pouch, since that implies that the two were made for eachother rather than the adapter pouch being made later IC
If you knew what the MS acronym meant, you could possibly change it to a different but similar acronym, that distinguishes it from the tank itself while still having them be related, like PS-11 Injector Pouch or something

x5tink:
> I'd personally make it like a 15-20 point buy in vendors but that sort of balancing probably needs more discussion
> 
> Also the name could be changed to something simple like MS-11 Injector Pouch, since that implies that the two were made for eachother rather than the adapter pouch being made later IC If you knew what the MS acronym meant, you could possibly change it to a different but similar acronym, that distinguishes it from the tank itself while still having them be related, like PS-11 Injector Pouch or something

In terms of points, i think it would be significantly less, like less than 10, feels expensive if its more.

I think we are going with a short name like that, i think it would have to start with EZ, to match the E series needles (EZ for easy/everyone), which would imply we should change the PCan pouch to M something, to match the M series needles ( M for medical training ) Personally in my head lore it was made after the smart can and separately, because the smart can has its own unique use of filling auto needles directly, its not designed for custom needles. So this is like later there were like hmmm lets make it work for all these custom needles ( sorta like Auto injectors are USB-A and Custom injectors are USB-C )

I could see adjusting prefix to EZ-11-A, or E-11A. or some variation of that working.

x5tink:
After discussing with a few people on discord, we all seemed to like the shorter 'EZ-11A Smart Tank Pouch'
Prefix for : EZ for the easy series of needles, 11 for the smart tank, A for its an adapter.
Updated the description to include the longer name to:
`The EZ-11A Smart Refill Tank Adapter Pouch bridges the Gap between the MS-11 Smart Refill Tank and custom injectors, allowing the smart refill tank to be used to refill custom injectors it would otherwise be incompatible with. An MS-11 Tank can clip into the adapter, and can be worn by attaching it to a pocket. Requires no medical training to use.
`
I assume it will just be refered to as the EZ pouch or the Smart can Pouch by marines later.
image

Hopefully this is a bit nicer :D

x5tink:
Typo and rounded the sprite so it looks a little nicer 
image

PR #8316, opened 2025-11-05
Reagent Jugs replace Hi-cap beakers + janitor bucket parity

Vermidia:
## About the PR

Title.

## Why / Balance

Request by Whisper & Xelidayne (the replacement stuff)

Parities janitor buckets, and applies all the hi-cap beaker changes to reagent jugs instead, which parent off the original hi-caps (which can no longer be printed as they've been replaced by the jugs, which are now in the medilathe. I left the original hi-cap yml alone).
All the changes listed:

### Reagent Jugs
Now printed in the medilathe.
Stores 500u
Material cost is 30000 glass instead of 8000 steel (10 glass sheets vs 2.6~ metal sheets)

### Janitor Buckets
Stores 300u
Size changed to huge

### High-Capacity Beaker
No longer printable (also a migration to replace any on maps with jugs)

Also fixed mop bucket size.

## Technical details

YML

## Media

https://github.com/user-attachments/assets/b99335ef-0885-4bc2-94c5-a3fc072d92a7

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- fix: Fixed Janibuckets and Mop buckets not being huge.
- fix: Fixed Janibuckets storing too much (500u->300u)
- tweak: High-Capacity Beakers have been replaced with Reagent Jugs
- fix: Fixed Reagent Jugs storing too little (300u->500u), costs being wrong (2.6~ metal sheets -> 10 glass sheets), and being printable in normal lathes instead of the medilathe

PR #9989, opened 2026-06-01
Large Chems Container adjustment

x5tink:
I've updated the Janitor bucket to not fit in any bags at all, and updated the Reagent Jug to only be storable with medical skill ( the same way the first-aid kits work currently). This is to allow Hypo-Jug based HM's and Shipside staff who use them for chemisty, to still function, but not allow Joe Rifleman to be lugging 2.5ku GC into the field.

## Technical details
YAML changes.

## Media
Ignore the Canister stuff in the video below, originally these were part of the same pr found [here](https://github.com/RMC-14/RMC-14/pull/9988)

https://github.com/user-attachments/assets/4f640c00-4da9-4641-b32b-81ea83bce79b

## Requirements

**Changelog**
:cl:
- tweak: Jani Bucket no longer store-able in bags, Reagent Jugs require med-skill to store in bags.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Tools/bucket.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Tools/bucket.yml
@@ -110,7 +110,6 @@
       visible: false
   - type: Item
     sprite: _RMC14/Objects/Misc/Janitorial/bucket.rsi
-    size: Normal
   - type: SolutionContainerVisuals
     fillBaseName: janibucket-
   - type: SolutionContainerManager
@@ -179,3 +178,10 @@
         enum.OpenableVisuals.Layer:
           True: {visible: false}
           False: {visible: true}
+  - type: StorageNestedOpenSkillRequired
+    skills:
+      all:
+        RMCSkillMedical: 2
+  - type: Tag
+    tags:
+    - CMFirstAidKit


Flameingman2:
I use the jug to carry water with me when I use the flamer, this would affect that

yagwog:
> I use the jug to carry water with me when I use the flamer, this would affect that

Would not be opposed to adding a seperate water only container, like a camelpak or something for this usecase

x5tink:
> > I use the jug to carry water with me when I use the flamer, this would affect that
> 
> Would not be opposed to adding a seperate water only container, like a camelpak or something for this usecase

I actually brought this up with smug prior to the this PR. and they responded with: 
image
So there is an alternative, not quite as much storage, but its a thing. And there are alot of sinks on the ground on the various maps, to refill what ever container you end up using.

QuietlyWhisper:
Superseded by https://github.com/RMC-14/RMC-14/pull/8316 ?

x5tink:
> Superseded by #8316 ?

Sort of, Maybe more... different? with i think more ramifications.

The jugs part yes, the making the Item huge is a better approach.

The changes to reagent jugs and hicaps does a few things from my perspective. 

- It essentially kills the alteration i'd made to allow it to still support ground side medical (ie Hypo Medics). 
- Gating it behind the medilathe (read as CMO) essentially means medical  will be bothered by people wanting jugs instead of being able to get them themselves, to maintain their builds.
- It does bring the container in line with with other beakers, in terms of glass cost ( though seems very expensive comparatively imo )
- Removal of hicaps entirely, means less options, as we've lost an easy to access 300u beaker, and kinda means to maintain the current chemistry methods, we'd now need to use the jani bucket, which feels a bit silly.

PR #10004, opened 2026-06-03
Sorokyne invincible resin fixes

CatAndHats:
## About the PR
Fixes two spots where paradropping into the resin was possible. moved some xenos spawns to be in the invincible resin.

## Why / Balance
polish, bugfixing

## Media

Invincible resin moved back from the red circles since you could paradrop into the blue shaded areas.
image

Invincible resin doors removed from the red area since you could paradrop into the blue area. Xeno spawns in this area moved to either the old wooden building or into deep sec.
image

update. Also removed the fog weeds that were over these containers and instead placed invincible resin doors/walls further back. This is for two reasons: One, xenos should be able to move through this area and not have it be blocked off. Two, the xeno weed fog is possibly getting a re-sprite which looks good when on walls, this is the only place where the fog was used off walls and so i'm removing this now to make that resprite go smoother later on.
image

## Requirements

**Changelog**

no CL


PursuitInAshes (approved):
Good

DrSmugleaf (approved):

PR #10009, opened 2026-06-04
Predicted removal of slow and root effects

MidZik:
## About the PR

Client should now properly predicts when a slow, superslow, or root effect expires. This should reduce rubberbanding that occurs when those effects expire.

## Why / Balance

Bugfix.

## Technical details

The code that handled those effects expiring only ran on the server, so the client simply did not predict them. I changed the update loop to run both on the client and server.

## Media

N/A

## Requirements

**Changelog**

:cl:
- fix: Fixed some rubberbanding that occurred when slow or root effects expired.


MidZik:
I created this PR as a result of investigating an issue with the Sapper's traps (#9433). The traps use a root effect, so the rubberbanding was very frequent.

DrSmugleaf (approved):
Very good ty

PR #10015, opened 2026-06-04
Add dropship fabricator print queue

GrigoriyGalintano:
## About the PR
Adds a pending print queue to the dropship part fabricator. Players can now order multiple CAS weapons, rockets, ammo, and other dropship fabricator items while another item is already printing.

## Why / Balance
This brings the fabricator closer to the CMSS13 behavior and removes the need to wait at the console for every single print. Points are still spent when an item is queued, so this does not allow free over-ordering. Pending orders can be cancelled for a refund, while the currently printing item cannot be cancelled.

## Technical details
Added `DropshipFabricatorQueueEntry` and a networked `Queue` list to `DropshipFabricatorComponent`, plus `MaxQueue = 6` to match the existing lathe/autolathe queue limit.

`DropshipFabricatorSystem` now validates printable prototypes, checks queue capacity, checks the shared dropship fabricator points account, reserves the cost immediately, and appends the item to the pending queue. If the fabricator is idle, `TryStartNextPrint` removes the first pending entry, sets the existing `Printing` field, updates `PrintAt`, and switches the fabricator appearance to fabricating.

When `Update` finishes the active `Printing` item, it spawns the prototype at the existing fabricator offset, clears `Printing`, and immediately attempts to start the next queued entry. If no entry remains, the fabricator returns to idle. Invalid queued prototypes are skipped and their reserved cost is refunded.

Added `DropshipFabricatorCancelQueueMsg` for cancelling pending entries by index. Cancelling removes the pending entry, refunds its stored cost to the shared points account, and dirties all fabricators so open UIs update. The active `Printing` item is still not cancellable.

The client BUI still uses the existing auto-networked component update path through `DropshipFabricatorUISystem`. `DropshipFabricatorBui.Refresh()` now reads `Printing`, `Queue`, `Points`, and `MaxQueue` directly from `DropshipFabricatorComponent`, then redraws the current job label and pending queue rows with cancel buttons. The existing Equipment and Ammo print buttons are unchanged.

Added locale strings for current job, idle state, queue count, empty queue, queue entries, cancel, queue full, and insufficient points.

## Media
image

## Requirements

**Changelog**
- add: Added a queue to the dropship part fabricator.

### DIFF
--- a/Content.Client/_RMC14/Dropship/Fabricator/DropshipFabricatorBui.cs
+++ b/Content.Client/_RMC14/Dropship/Fabricator/DropshipFabricatorBui.cs
@@ -83,10 +83,82 @@ public sealed class DropshipFabricatorBui : BoundUserInterface
         if (_window is not { Disposed: false })
             return;
 
-        if (EntMan.TryGetComponent(Owner, out DropshipFabricatorComponent? fabricator))
+        if (!EntMan.TryGetComponent(Owner, out DropshipFabricatorComponent? fabricator))
+            return;
+
+        _window.PointsLabel.Text = Loc.GetString("rmc-dropship-fabricator-points",
+            ("points", fabricator.Points));
+
+        if (fabricator.Printing is { } printing)
         {
-            _window.PointsLabel.Text = Loc.GetString("rmc-dropship-fabricator-points", 
-                ("points", fabricator.Points));
+            _window.CurrentLabel.SetMarkupPermissive(Loc.GetString("rmc-dropship-fabricator-current",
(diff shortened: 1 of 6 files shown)


Ian321:
Adding something to the queue pushes the menu down.

https://github.com/user-attachments/assets/8db6eb23-e62a-4aa2-9ee3-144e29dc57d3

Could you make it not expand and/or move it so that it doesn't interfere with pressing the fabricate buttons.

DrSmugleaf (approved):

PR #10063, opened 2026-06-07
Fix fiorina podlocks

CatAndHats:
## About the PR
medical pod podlocks on fiorina are openable and you can get into OOB area/space

the solution is I-walls, like the other podlocks on fio, don't trust podlocks, I-wall everything.

## Technical details
image

## Requirements

**Changelog**

:cl: CatAndHats
- fix: Fixed some podlocks on Fiorina being openable allowing access into space


kleinerstation13:
this has to be related to the power PR because they were not unpowered before this patch

AverageNotDoingAnythingEnjoyer:
They were in fact unpowered

kleinerstation13:
ah i think i mightve been dumb and mistook them for the locks on the side since i remember those being unpriable roundstart

DrSmugleaf (approved):

CatAndHats:
> this has to be related to the power PR because they were not unpowered before this patch

to be clear the issue isn't that they were pryable, but that they were meltable. These podlocks can be fixed in future since they are still meltable, however the I-Walls means that there won't be an issue of xenos getting into space ever again at this spot in the map.

PR #10067, opened 2026-06-08
Sapper trap behavior adjustments

MidZik:
## About the PR

Sapper traps no longer trigger if you are standing on them at the moment they spawn.

Sapper trap radius halved from 0.5 to 0.25.

## Why / Balance

Sapper traps not triggering if they spawn under you is parity behavior.

Regarding the radius reduction, there isn't really any "parity" behavior to match here because of the differences between pixel movement and tile movement. This radius change is _not_ necessarily a nerf. If the radius is too large, marines can too easily move around on the traps without triggering them when considering the above change. This reduced radius makes the traps work more like both the marine and the sapper would expect them to.

## Technical details

Added XenoNewlyDeployedTrapsComponent for new traps.
New traps track their initial collisions and exclude them from triggering the trap.
Initial collisions fall off when they step off. ~~(I.e. they can trigger the trap once they step off of it)~~ (I.e. once they step off the trap, they will trigger it if they step on it again)
Slight refactors.

## Requirements

**Changelog**

:cl:
- fix: Sapper traps no longer trigger on marines that were standing on them when they were created.
- fix: Sapper trap radius reduced to prevent odd behavior with the above change.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Construction/DeployedTraps/SharedDeployedTrapsSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Construction/DeployedTraps/SharedDeployedTrapsSystem.cs
@@ -1,10 +1,10 @@
-﻿using Content.Shared._RMC14.Marines;
+using Content.Shared._RMC14.Marines;
 using Content.Shared._RMC14.Slow;
 using Content.Shared._RMC14.Xenonids.Hive;
 using Content.Shared.Mobs.Systems;
 using Content.Shared.StepTrigger.Systems;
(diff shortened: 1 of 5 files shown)


DrSmugleaf (approved):
Thanks for the PR!
It's not supposed to activate if you move out of it but that's probably too hard to replicate for us any more accurately than this, this is good enough

MidZik:
It shouldn't trigger if you step off of it. From my testing it wasn't, anyway.

It's possible there's a mispredict somewhere though, that I missed.

PR #10039, opened 2026-06-07
Update fixes

Arendian:
## About the PR

- Air dropped sentries no longer target resin.
- You're no longer able to set the lowest rank of a role as your rank preference.
- Parasites no longer need to use the verb menu to enter tunnels, vents or to pass through acid holes.
- The sniper overlay no longer shows on a different map than that of the sniper.
- Imaginary friends now have new player vision.
- Flare no longer sometimes show their unlit state while lit.
- Imaginary friends without admin permission can now talk in LOOC, only the friend and other ghosts will see these messages.

Will update this if I find more things to fix.

## Why / Balance

Fixes.
Resolves #10065 

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl: Dygon
- fix: Air dropped sentries no longer target resin.
- fix: You're no longer able to set the lowest rank of a role as your rank preference.
- fix: Parasites no longer need to use the verb menu to enter tunnels, vents or to pass through acid holes.
- fix: The sniper overlay no longer shows on a different map than that of the sniper.
- fix: Imaginary friends now have new player vision.
- fix; Fixed lit flares sometimes having an unlit appearance.
- fix: Imaginary friends can now always use LOOC chat.

PR #10074, opened 2026-06-08
[PVE] Sun Riders Rank Options

ThataKat:
## About the PR
Disallowing the selection of the lowest rank for any given job is now toggleable with a CVar. By default (as on Alamo) it is disallowed. Normandy config now allows the selection of the lowest rank for all jobs.

Added rank preference options to the Sun Rider jobs as follows:

Job: (Old Rank) -> New Lowest Rank - New Highest Rank
RFN: (Private) -> Private - Lance Corporal
HM: (Lance Corporal) -> Private First Class - Corporal
SGO: (Corporal) -> Private First Class - Corporal
SL: (Sergeant) -> Corporal - Sergeant
SecSarge: (Staff Sergeant) -> Staff Sergeant - Gunnery Sergeant
PlatCo: (Second LT) -> Second LT - First LT

After discussion with GMs, the other factions have been left alone due to the various obstructions each has to variable rank.

## Why / Balance
"Parity", enhanced character customization/expression, more opportunity for storytelling in an HRP environment, no longer having 20 people who are all E1s

## Technical details
yaml, CVar + CSharp, toml for Normandy config

Just modified the changes in #10039 to only execute if the `LoadoutLowestRankSelectable` CVar is false.

No time requirements, as in parity.

Auto will still assign highest rank, as on Alamo, but shouldn't be a major issue. Ranks may skew high before people remember to tweak them.

In the rare case that someone picks a low rank on Normandy and takes that preference back to Alamo (where they have timereqs for a higher one), they get set back to auto (in my testing anyways).

## Media
DCC used to demonstrate ability to select lowest rank on standard roles.

https://github.com/user-attachments/assets/f156172d-8eb3-423d-862a-850a13facf86

## Requirements

**Changelog**

:cl:
- add: [PVE] Sun Riders' roles now have rank preference options, chosen in the jobs tab.


ThataKat:
Drafted b/c I've been fighting my devenv and still need to verify this actually works. Preventing accidental overlap of a similar PR.

ThataKat:
A potential note, I could change the yaml rank order to be low -> high (in pve jobs) which would make auto default to the lowest rank.

I don't like that idea much because it breaks the convention of every other job yaml that ranks are listed high to low, but if we don't want the Auto setting players who forget/are new to the highest rank, it's probably the quickest fix.

I'm not sure how I would solve the same problem in the actual C#

ThataKat:
CMPVE has PVT as an option for corpsman and smartgunner as well apparently.

Will add if maints/GMs desire

PursuitInAshes (approved):
Thanks for the PR!
Ranks are fine from a GM perspective, I don't want specalist roles being below private
Implementation will have to be considered by someone else, so long as its easy to add another check in for fields on the job prototypes later this should be okay though

PursuitInAshes (approved):

PR #10077, opened 2026-06-08
Larvae queue improvments

Vixting:
## About the PR

Larvae queue now offers priority to the burst victim & then a player parasite if they infected the burst victim, must be in the queue to be offered these priorities. 
Disconnected players now have a 5 minute grace period before being removed from the queue
New popup & sound effect on queue being popped for you, accept / decline. Follow button to see where you are spawning.
You can now see your position in queue in the menu.

## Why / Balance

qol / parity
## Technical details

The parasite stuff is odd because of the parasite being ghosted on hug, but only infecting them after the delay
## Media

image

image

## Requirements

**Changelog**

:cl: Vixting
- add: Larvae queue now offers priority to the burst victim provided they are within their body when the burst occurs.
- add: Larvae queue priotises a player parasite if they infected the victim being burst (has to be within the queue)
- add: Parasites / lessers will now be offered the larvae (don't have to be a ghost)
- add: Disconnected players now have a 5 minute grace period before being removed from the queue
- add: New popup & sound effect on queue being popped for you, accept / decline. Follow button to see where you are spawning.
- add: You can now see your position in queue in the menu.


Content.Shared/_RMC14/Xenonids/JoinXeno/LarvaQueueSystem.cs
```diff
@@ -27,37 +30,142 @@ public sealed class LarvaQueueSystem : EntitySystem
...
+
+    private sealed class PendingOffer
+    {
+        public required EntityUid? TargetLarva;
+        public required Entity<HiveComponent> Hive;
+        public required double ExpiresAt;
+    }
+
+    private static readonly TimeSpan XenoReconnectGracePeriod = TimeSpan.FromMinutes(5);
+    private static readonly TimeSpan DisconnectGracePeriod = TimeSpan.FromMinutes(5);
```
MidZik: I believe we can use the existing cvar `rmc.disconnected_xeno_ghost_role_time_seconds` for this? It serves the same purpose, and it still applies to xenos that are excluded from the queue like queens. It also has a default of 5 minutes.

MidZik:
I only tested briefly so this isn't a thorough review yet, but I found one issue so far:

- Ghosting from a xeno counts it as "disconnected" so it's not eligible to be taken over until the 5 minutes pass. I think that should only be the case for an actually disconnected player, ghosting should put it in the queue immediately.

Vixting:
Issues addressed also realised that the delay was already handled so moved to just doing       
`  RemComp<CanBeLarvaQueuedComponent>(xeno);` to prevent the for the hive / sacrifice bug
Going to change the host priority to the parity way

Vixting:
Burst victim now only has priority if they are within their body when the burst occurs.
Lesser / parasites are now offered the larvae instead of them having to be a ghost to get it

DrSmugleaf (approved):

PR #10080, opened 2026-06-08
Chance's Claim fixes and some props

CatAndHats:
## About the PR
Fixed a spot having the OOB area instead of the correct one
Fixed a room not being connected to the pipe net correctly
Added MB-6 Composite folding cades, ponchos, and decks of cards
Moved a gun and armour that were inside the invincible resin to where the scout armour is.
Replaces incorrect m1984's with m1911's. Adds custom scout helmet name + description.
Adds vehicle blockers, not many on this map.

## Why / Balance
parity, bugfixing, polish

## Media
fixes this area being OOB instead of the mining area
Screenshot 2026-06-08 173851

Fixed this room not being connected to the pipe net
Screenshot 2026-06-08 191409

composite cades and ponchos
Screenshot 2026-06-08 175143

Screenshot 2026-06-08 175213

yes those cades are meant to be regular ones
Screenshot 2026-06-08 175401

decks of cards here and in some other places
Screenshot 2026-06-08 175745

vehicle blockers. Chance's doesn't seem to stop vehicles going into the caves, so the blockers only exist at these two spots.
Screenshot 2026-06-11 082030
Screenshot 2026-06-11 082040

## Requirements

**Changelog**

:cl: CatAndHats
- add: Added MB-6 folding composite barricades, ponchos, M1911s (replacing M1984s), and decks of cards to Chance's Claim
- fix: Fixed an area filled with metal foam on Chance's Claim that was set as Out Of Bounds area instead of the correct mining area
- fix: Fixed two rooms on Chance's Claim not being connected to the pipe net


kleinerstation13:
there is also a bunch of m1984 guns and ammo where m1911's should be mapped instead, 

obraz

here 

the scout helmet also has a custom name and description so you could change that as well if you want

"battered M3-S light helmet"
"A custom helmet designed for USCM Scouts. Looks like this one has seen some action."

there is also one with 2 mags here at NW sensors
obraz

kleinerstation13:
actually i'm pretty sure you can just find and replace the m1984's and their mags with the m1911 in the map file yml since every instance of an m1984 should be an m1911 on that map

PursuitInAshes (changes requested):
Above

CatAndHats:
M1911's added, scout helmet name and description changed.

Also these pipes circled are not connected to the greater pipe net in CM13. This seems to be a mapping error since there are no explosions/ripped up tiles to suggested these have been removed for an IC reason (unlike in mining), so I have connected them.
image

PursuitInAshes (dismissed):
TY For the PR!

PursuitInAshes (dismissed):

DrSmugleaf (approved):

PR #10082, opened 2026-06-08
Fix non-guns being acid clearable

Vermidia:
## About the PR

Title.

## Why / Balance

Bug. In parity you can't extinguish anything without second wind.

## Technical details

Moved the gun comp check to check if the message should be displayed only when second wind is a comp instead, main part now checks if second wind isn't present or is used.

## Requirements

**Changelog**

:cl:
- fix: Fixed acid being extinguishable off of non-guns


Thereisabear (approved):
LGTM

Vixting (approved):
LGTM

PR #10095, opened 2026-06-09
Adds: Small Tool Webbing (Filled) to Intel Vendor

wheelwrightt:
## About the PR
Adds the small tool webbing filled to the intel vendor.

## Why / Balance
QoL. Most IOs I see carry these tools anyways in their regular webbing. It's a bit tedious to have to take the tools out of the belt and slot them into the webbing every round when we already have a dedicated webbing that could do that.

## Technical details
yml

## Requirements

**Changelog**
:cl:
- add: Small tool webbing (filled) to intel vendor.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/intel.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/intel.yml
@@ -214,6 +214,7 @@
       - id: CMWebbingBlack
       - id: CMWebbingBrown
         recommended: true
+      - id: RMCToolWebbingSmallFilled
       - id: RMCWebbingLegPouch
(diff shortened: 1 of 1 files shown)


xfloodie:
This is somewhat of a balance change, since filled tool webbing comes with 6 slots as opposed to the webbing vest that holds 5.

wheelwrightt:
The listed items are what the tool webbing comes with by default:
imagen

Items that can go into the webbing:
imagen
Of these items the screwdriver is the exception to what we would put in the current webbing which generally would be behind the ear instead, not that big of a difference imo. There would be no need for an IO to carry any of the non-default tools set, maybe the entrenching tool? but you could just put that in your helmet anyways.

CatAndHats (commented):
We can revisit this in the future

PR #10098, opened 2026-06-09
fix: dropship navigation computers now work on all maps again

Ian321:
## About the PR

title

## Why / Balance

bugfix

## Technical details

on some maps the consoles are mapped to an area that doesn't have a APC, so I just removed the power requirement from it. Also cleaned up all the dropship related consoles as they all (for [some](https://github.com/RMC-14/RMC-14/issues/9975) [reason](https://github.com/RMC-14/RMC-14/commit/7b56f2fc9626fc13d6ee006bd5b9fba7c2f08af4#diff-05ce55fd518306f869be42a8a059921866a6aad0a6ab2b69ac75b40b8da28ce8)) inherited a power requirement and then removed it again.

## Media

image

## Requirements

**Changelog**

:cl:
- fix: The planetside dropship navigation computers should now work on all maps again.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/shuttle_console.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/shuttle_console.yml
@@ -11,12 +11,11 @@
   - type: InteractionOutline
   - type: ApcPowerReceiver
     powerLoad: 250
-    needsPower: true
+    needsPower: false
   - type: RMCPowerReceiver
     idleLoad: 300
     activeLoad: 300
     channel: Equipment
-  - type: ActivatableUIRequiresPower
 
 - type: entity
   parent: CMComputerDropship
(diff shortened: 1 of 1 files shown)


AverageNotDoingAnythingEnjoyer:
If the areas dont have any APCs, you might want to check if area stats are correct, there may be a chance of them having infinite power in params
As for redundant power values, RMCComputerDropshipNavigationSmall inherets from BaseStructure, i.e. it doesn't have power values by default

Ian321:
> If the areas dont have any APCs, you might want to check if area stats are correct, there may be a chance of them having infinite power in params As for redundant power values, RMCComputerDropshipNavigationSmall inherets from BaseStructure, i.e. it doesn't have power values by default

I did. I noticed the bug cause we crash landed and even tho I fixed all the APCs around it remained off.

Resources/Prototypes/_RMC14/Entities/Structures/Machines/shuttle_console.yml
```diff
@@ -9,14 +9,6 @@
   - type: Corrodible
     isCorrodible: false
   - type: InteractionOutline
-  - type: ApcPowerReceiver
-    powerLoad: 250
-    needsPower: true
```
AverageNotDoingAnythingEnjoyer, line 14: Nvm, modify this to needsPower: false and it should be done.
They consume power on CM but dont require it, also remove ActivatableUIRequiresPower comp

DrSmugleaf (approved):

PR #10100, opened 2026-06-09
 Healer Drone and Acid Runner fixes 

### LINKED ISSUE #10062: Healer Sacrifice/Acider For the Hive larva gets taken by queue
## Description

Title. The way they work right now is they get gibbed, a burrowed larva is made, and then they're forced to take said larva. Problem is atm I think queue is grabbing the newly created burrowed and giving it to the first person in the queue instead. Might need to directly spawn a new larva and transfer mind instead of what's currently done. This bug should only be affecting xenos that are trying to spawn via hive core, not from their corpse (which already makes a new larva)

**Reproduction**

**Screenshots**

**Additional context**

Bagietens:
## About the PR

Fixes #10062, where the respawn system for xenos that use Sacrifice Heal ability for healer drones and For The Hive ability would prevent the players that used them from respawning as larva when there was a hive core present.

## Why / Balance
Fix is good / makes sacrificing and exploding as acid runner work again

## Technical details

Bypassed the `ChangeBurrowedLarva` in Xeno respawn system so instead of creating a burrowed larva it creates a new larva on the core and transferes the player into it, bypassing the queue

## Media

https://github.com/user-attachments/assets/a2944dec-3594-4a30-acc3-02001ae9180b

https://github.com/user-attachments/assets/01d32237-378c-43fa-b72c-67dc9b084d01

https://github.com/user-attachments/assets/2ed4bbfd-038a-4386-a029-104a991c2876

## Requirements

**Changelog**

:cl:
- fix: Fixed healer drone and acid runner sacrifices being taken by larva queue.

### DIFF
--- a/Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
@@ -3,7 +3,9 @@ using Content.Server.Ghost;
 using Content.Server.Ghost.Roles.Components;
 using Content.Server.Mind;
 using Content.Server.Popups;
+using Content.Shared._RMC14.Xenonids.Construction;
 using Content.Shared._RMC14.Xenonids.Hive;
+using Content.Shared._RMC14.Xenonids.JoinXeno;
 using Content.Shared._RMC14.Xenonids.Respawn;
 using Content.Shared.Mind;
 using Content.Shared.Popups;
@@ -24,12 +26,13 @@ public sealed partial class XenoRespawnSystem : EntitySystem
     [Dependency] private readonly PopupSystem _popup = default!;
     [Dependency] private readonly AudioSystem _audio = default!;
     [Dependency] private readonly TransformSystem _transform = default!;
-    public void RespawnXeno(EntityUid xeno, TimeSpan time, bool atCorpse = false, EntityCoordinates? corpse = null)
+    public void RespawnXeno(EntityUid xeno, TimeSpan time, bool atLocation = false, EntityCoordinates? location = null)
     {
         if (!TryComp(xeno, out ActorComponent? actor))
             return;
 
         RemComp<GhostTakeoverAvailableComponent>(xeno);
+        EnsureComp<IgnoreQueueComponent>(xeno);
 
         var session = actor.PlayerSession;
 
@@ -47,8 +50,8 @@ public sealed partial class XenoRespawnSystem : EntitySystem
             var respawn = EnsureComp<XenoRespawnComponent>(ghost.Value);
             respawn.Hive = _hive.GetHive(xeno);
             respawn.RespawnAt = _timing.CurTime + time;
-            respawn.RespawnAtCorpse = atCorpse;
-            respawn.CorpseLocation = corpse;
+            respawn.RespawnAtLocation = atLocation;
+            respawn.Location = location;
         }
     }
 
@@ -66,15 +69,15 @@ public sealed partial class XenoRespawnSystem : EntitySystem
                 continue;
 
             ActorComponent? actor;
-            if (respawn.RespawnAtCorpse)
+            if (respawn.RespawnAtLocation)
             {
-                if (respawn.CorpseLocation == null)
+                if (respawn.Location == null)
                 {
                     RemCompDeferred<XenoRespawnComponent>(ghost);
                     continue;
                 }
 
-                var spawn = SpawnAtPosition(respawn.Larva, respawn.CorpseLocation.Value);
+                var spawn = SpawnAtPosition(respawn.Larva, respawn.Location.Value);
                 _hive.SetHive(spawn, respawn.Hive);
 
                 if (!TryComp(ghost, out actor))
@@ -87,8 +90,13 @@ public sealed partial class XenoRespawnSystem : EntitySystem
 
                 _mind.TransferTo(mindId, spawn);
 
-                _popup.PopupEntity(Loc.GetString("rmc-xeno-respawn-corpse-self"), spawn, spawn, PopupType.MediumCaution);
-                _popup.PopupEntity(Loc.GetString("rmc-xeno-respawn-corpse-others"), spawn, Filter.PvsExcept(spawn), true, PopupType.MediumCaution);
+
+                if (TryComp<HiveComponent>(respawn.Hive, out var coreComp) &&
+                    !_hive.HasHiveCore((respawn.Hive.Value, coreComp)))
+                {
+                    _popup.PopupEntity(Loc.GetString("rmc-xeno-respawn-corpse-self"), spawn, spawn, PopupType.MediumCaution);
+                    _popup.PopupEntity(Loc.GetString("rmc-xeno-respawn-corpse-others"), spawn, Filter.PvsExcept(spawn), true, PopupType.MediumCaution);
+                }
 
                 _audio.PlayPvs(respawn.CorpseSound, _transform.GetMoverCoordinates(spawn));
 
--- a/Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
+++ b/Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
@@ -90,8 +90,7 @@ public sealed partial class XenoRespawnSystem : EntitySystem
                 _mind.TransferTo(mindId, spawn);
 
 
-                if (TryComp<HiveComponent>(respawn.Hive, out var coreComp) &&
-                    !_hive.HasHiveCore((respawn.Hive.Value, coreComp)))
+                if (!_hive.TryGetHiveCore(spawn, out _))
                 {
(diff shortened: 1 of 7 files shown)


Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
```diff
@@ -24,12 +26,13 @@ public sealed partial class XenoRespawnSystem : EntitySystem
...
     [Dependency] private readonly AudioSystem _audio = default!;
     [Dependency] private readonly TransformSystem _transform = default!;
-    public void RespawnXeno(EntityUid xeno, TimeSpan time, bool atCorpse = false, EntityCoordinates? corpse = null)
+    public void RespawnXeno(EntityUid xeno, TimeSpan time, bool atLocation = false, EntityCoordinates? location = null)
     {
         if (!TryComp(xeno, out ActorComponent? actor))
             return;
 
         RemComp<GhostTakeoverAvailableComponent>(xeno);
+        EnsureComp<IgnoreQueueComponent>(xeno);
```
MidZik: Instead of creating this new component `IgnoreQueueComponent`, shouldn't removing `CanBeLarvaQueuedComponent` be good enough? It would have to be an immediate removal, not deferred.

Vixting:
The issue with for the hive / sacrifice should be resolved by this https://github.com/RMC-14/RMC-14/pull/10077 
(5m delay before a mind is added to the queue on it being removed)

Bagietens:
> The issue with for the hive / sacrifice should be resolved by this #10077 (5m delay before a mind is added to the queue on it being removed)

Alright then! Should I close the pr? Cause I don't think #10077 fixes the part where it adds a burrowed larva when there is core

Vixting:
> > The issue with for the hive / sacrifice should be resolved by this #10077 (5m delay before a mind is added to the queue on it being removed)
> 
> Alright then! Should I close the pr? Cause I don't think #10077 fixes the part where it adds a burrowed larva when there is core

No, just remove the stuff related to the other fix. The thing with the larvae is still needed I think

Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
Bagietens: yeah you're right. I had problems with this before so I added that component. My mistake was instead of adding `RemComp<CanBeLarvaQueuedComponent>` in `RespawnXeno` I tried doing that inside the Xeno Heal system. You're right

Content.Shared/_RMC14/Xenonids/ForTheHive/SharedXenoForTheHiveSystem.cs
```diff
@@ -259,7 +259,7 @@ public override void Update(float frameTime)
         }
     }
 
-    protected bool GetHiveCore(EntityUid xeno, out EntityUid? core)
+    protected bool GetHiveCore(EntityUid xeno, out EntityUid core)
```
MidZik: This should stay nullable. Add this attribute to make the compiler not complain at the call sites.
```suggestion
    protected bool GetHiveCore(EntityUid xeno, [NotNullWhen(true)] out EntityUid? core)
```

Content.Shared/_RMC14/Xenonids/ForTheHive/SharedXenoForTheHiveSystem.cs
```diff
@@ -274,7 +274,7 @@ protected bool GetHiveCore(EntityUid xeno, out EntityUid? core)
             return true;
         }
 
-        core = null;
+        core = default;
```
MidZik: This should stay as a null assignment.
```suggestion
        core = null;
```

Content.Shared/_RMC14/Xenonids/ForTheHive/SharedXenoForTheHiveSystem.cs
```diff
@@ -247,7 +247,7 @@ public override void Update(float frameTime)
                     }
 
                     if (GetHiveCore(xeno, out var core))
-                        ForTheHiveRespawn(xeno, active.CoreSpawnTime);
+                        ForTheHiveRespawn(xeno, active.CoreSpawnTime, true, _transform.GetMoverCoordinates(core));
```
MidZik: Instead of changing the function to return a non-nullable, extract the value via `.Value`. Compiler shouldn't complain if you add the above attribute to the out param.
```suggestion
                        ForTheHiveRespawn(xeno, active.CoreSpawnTime, true, _transform.GetMoverCoordinates(core.Value));
```

Content.Shared/_RMC14/Xenonids/Heal/SharedXenoHealSystem.cs
```diff
@@ -382,7 +382,7 @@ public void CreateHealStacks(EntityUid target, FixedPoint2 healAmount, TimeSpan
         heal.ParallizeHealing = true;
     }
 
-    private bool GetHiveCore(EntityUid xeno)
+    private bool GetHiveCore(EntityUid xeno, out EntityUid core)
```
MidZik: Same as above.
```suggestion
    private bool GetHiveCore(EntityUid xeno, [NotNullWhen(true)] out EntityUid? core)
```

Content.Shared/_RMC14/Xenonids/Heal/SharedXenoHealSystem.cs
```diff
@@ -393,9 +393,11 @@ private bool GetHiveCore(EntityUid xeno)
             if (_mobState.IsDead(uid))
                 continue;
 
+            core = uid;
             return true;
         }
 
+        core = default;
```
MidZik: Same as above.
```suggestion
        core = null;
```

Content.Shared/_RMC14/Xenonids/Heal/SharedXenoHealSystem.cs
```diff
@@ -337,8 +337,8 @@ private void OnXenoSacrificeHealAction(Entity<XenoComponent> ent, ref XenoSacrif
         {
             var corpsePosition = _transform.GetMoverCoordinates(ent);
 
-            if (GetHiveCore(ent))
-                SacrificialHealRespawn(ent, args.RespawnDelay);
+            if (GetHiveCore(ent, out var core))
+                SacrificialHealRespawn(ent, args.RespawnDelay, true, _transform.GetMoverCoordinates(core));
```
MidZik: Same as above.
```suggestion
                SacrificialHealRespawn(ent, args.RespawnDelay, true, _transform.GetMoverCoordinates(core.Value));
```

Content.Shared/_RMC14/Xenonids/Heal/SharedXenoHealSystem.cs
```diff
@@ -393,9 +393,11 @@ private bool GetHiveCore(EntityUid xeno)
             if (_mobState.IsDead(uid))
                 continue;
 
+            core = uid;
             return true;
         }
 
+        core = default;
         return false;
     }
```
MidZik, line 400: This isn't your responsibility and you don't have to fix it if you don't want to, but this function should really not be duplicated in SharedXenoForTheHiveSystem and SharedXenoHealSystem. It should instead exist as a public function in `SharedXenoHiveSystem`, and these systems should use it as a dependency.

Edit: I'm referring to the entire `GetHiveCore` function that exists in both systems.

Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
MidZik, line 116: I'm concerned about this logic existing. You fixed the issue by always setting a location, so I don't think this code is reachable anymore. Maybe the design of XenoRespawnSystem and XenoRespawnComponent should be slightly changed so that setting a location for the respawn isn't optional anymore?

MidZik (changes requested):
Fix seems correct, just have a few comments and suggestions.

Content.Server/_RMC14/Xenonids/Respawn/XenoRespawnSystem.cs
Bagietens: I have no real idea how people will use the `RespawnXeno` method so I left it as a fallback. So for now I want to keep it that way.

Content.Shared/_RMC14/Xenonids/Heal/SharedXenoHealSystem.cs
Bagietens: Good idea, I added a new function `TryGetHiveCore` in the `SharedXenoHiveSystem`

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
```diff
@@ -315,6 +316,19 @@ public bool HasHiveCore(Entity<HiveComponent> hive)
         return GetHiveCore(hive) is not null;
     }
 
+
```
DrSmugleaf:
```suggestion
```

DrSmugleaf (dismissed):

PR #10102, opened 2026-06-10
Dancer Rework

Vermidia:
## About the PR

Title. Ported all the changes from parity, plus some other fixes.

Note someone was working on a new sprite for Tail Lance Slam, its not done yet.

## Why / Balance

Parity!

## Technical details

XenoDodgeOverlay is an overlay for drawing the dancer dodge afterimage effects and slight sprite movement. It does work with other clients (did some testing not shown in the video). It copies dancer's sprite directly like xeno night vision, and displays it on the entities layer, so it is properly hidden by walls etc. The only bug with this is that for some reason the starting offset applies rapidly like a normal jitter before going back to normal. Note that in parity you don't change offsets or create afterimages while still, but it looked cooler and was a good visual indicator the ability was still up, so I kept it.

Dancer's Dodge ability can now be canceled early, like lurker's turn invisible. Other than the overlay, it no longer gives the crowdboost (forgor to remove after mob collisions existed) and also now interacts with RMCGetSwiftStepsThresholdEvent by a value determined in the comp (default -3). It's been seperate into client & server systems, mostly so the client can cleanup the overlay and offset by itself.

There is a new RMCBeforeProjectileAccuracyEvent that is called before accuracy checks happen, it's used to guarantee a bullet dodge if GuarateedMiss is set to true. This is used by the new XenoSwiftSteps system which counts how many bullets have attempted to hit the ent with the comp, with certain leeway based on time, if they were resting/unconscious when hit. It adds projectiles that have been counted to a list so they can be ignored for a certain amount of time. When the counted bullets reach a threshold, the count is cleared and a guaranteed miss is set.

Critical marks are a new type of mark, tracked a bool in XenoMarked. In the XenoFinesse system, when a target that would get marked goes into crit or dies (checked by a seperate comp on damage because MeleeHitEvent happens before damage is calculated) nearby marines will be given a critical mark, which has different visuals from normal marks (which also got updated visually). If a critical mark is detected when a normal mark would be applies, Impale and Tail trip lose cooldown. Critical mark spread is only checked on server because it mispredicted too easily.

XenoImpaleSystem has been slightly overhauled so multiple impales from XenoSecondImpale can happen (if multiple dancers gang up on someone). Also handles critical marks.

XenoTailTripSystem now handles critical marks.

New XenoCappedPheromones component, for dancers and berserkers, which caps certain pheromones to a max multiplier.

New component and a change to the tailstab event to allow for an alternative tailstab. Some tailstab stats now get transferred to a struct to prevent too much code duplicated from tail lance, which uses a different damage, animation, and AP value from a normal tail stab. Tailstab comp can now support AP and only hitting mobs. Technically the alt slam isn't supposed to get damage buffs but...agh its already a mess and probably not the best move 90% of the time anyway.

Cleaned up various descriptions, added some new local. Also fixed impale icon having an orange square on it (I don't think it was intentional, it looked wrong).

## Media

https://github.com/user-attachments/assets/02c2901d-2205-4125-9abf-d5756900544a

## Requirements

**Changelog**

:cl:
- fix: Fixed Impale and Tail Trip miss cooldowns being too long (1.5s -> 0.7s)
- fix: Fixed Impale having less range than Tail Trip. They're both 2 now.
- fix: Fixed Impale and Tail Trip deselecting on hit.
- fix: Fixed Berserk Ravager not capping its applied Frenzy pheros to 2.9, and Dancer Praetorian not capping its Recovery pheros to 3.
- fix: Fixed Dancer Praetorian's general health regen not being higher than a base xeno. It's now 35% better.
- fix: Fixed Dodge still giving a speed boost while in a crowd.
- fix: Fixed Impale rarely not stabbing twice if multiple Dancers are attempting to do so in a short span of time.
- fix: Corrected various ability descriptions (mostly involving tail stabs and Dancer Praetorian)
- add: Critting/killing an enemy (or just attacking them if they're crit) as a Dancer Praetorian now spreads critical marks, which look different and refresh the Dancer's ability cooldowns if hit.
- add: Dancer Praetorian now has Tail Lance Stab and Tail Lance Slam, they share cooldowns. The former is a normal tailstab, while the latter deals much less damage, has less cooldown, but fully pierces armor.
- add: Dancer Praetorians by default now dodge every 6th projectile that would hit them (in a short span of time, and not while resting or crit/unconscious). When dodge is active this is every 3rd projectile instead. This doesn't count rockets or sniper bullets.
- add: Dodge creates afterimages of the praetorian while active.
- fix: Fixed impale having an orange square on the icon


DrSmugleaf (approved):
Thanks for the PR!

PR #10103, opened 2026-06-10
Action prediction and lag comp improvements.

### LINKED ISSUE #8702: Ghost Actions

MidZik:
## About the PR

Actions overall are more accurately predicted. Actions with lag compensation should be more reliable due to this. (Actions are anything that appears on your hotbar.)

Melee hits and interactions (shaking, patting, pulling, etc) are more accurately lag compensated, especially for targets that are moving towards you.

No new lag compensation is being added. Only existing lag-compensated actions/interactions have been made more reliable.

Fixes #8702. Supersedes #9968.

## Why / Balance

Bugfixes/improvements.

The server was rewinding its state a bit too far back when lag compensating for melee hits and interactions. This made interacting with things moving away from you easier than it should be, which isn't generally a problem. However it made hitting things moving _towards_ you harder, with hits or interactions against such objects sometimes being incorrectly rejected by the server.

## Technical details

From #9968:

>Actions are UI. UI is considered out of simulation. Out of simulation timing does not use the same "CurTime" as in simulation timing. Basically every action we have does in-simulation things. The end result is, the client tries to predict a lot of "in simulation" things while the engine says it's "out of simulation". It goes poorly.
>
>The root cause of the ghost action issue is that when the client tries to use an action, it compares CurTime to the action's cooldown EndTime. But because we're "out of simulation", CurTime can actually be a value in the middle of a tick. That means if you use the action late enough in the final tick of the cooldown, the client will let you, because CurTime advanced past the EndTime during the tick that EndTime is in. In simulation that is not possible, you would have to wait for the next tick for CurTime to advance past EndTime.
>
>I am honestly not sure how best to deal with this. This feels like a big oversight by basically all actions in content (both ours and upstream). It seems every action has this problem.

The out-of-simulation action issue was fixed by queuing non-clientside-only actions **when they have the InSimulationOnly flag set**, and only performing those actions during an update loop. This resolves the issue of actions running out-of-simulation, making the client's prediction of actions much more accurate, which in turn makes any lag-compensation that tried using those results more accurate. This is a broad change that affects every xeno action. It might make some actions appear _slightly_ more delayed, but from my own testing (at 15 ticks per second) I could not notice a difference.

Action validation was changed so that the cooldown is no longer compared to an out-of-simulation CurTime, which was causing ghost actions when the client thought an action was off cooldown when it actually was not.

The server no longer sets the client's latest received tick as one less than what the client tells it. This was causing interactions on entities moving towards you to sometimes fail where they should succeed.

The client no longer sends the server its latest received tick during re-predicted actions. That data would be too old to be relevant anyway, and lag compensated actions will send that data when needed. This will reduce the amount of messages each client sends to the server.

## Media

N/A

## Requirements

**Changelog**

:cl:
- fix: Xeno actions/abilties should have fewer mispredicts and hit their targets more reliably.
- fix: Ghost/fake actions should no longer occur.
- fix: Melee hits and interactions (shaking, patting, pulling) should be more reliable. In particular, meleeing or interacting with entities moving towards you should be more reliable.

### DIFF
--- a/Content.Client/Actions/ActionsSystem.cs
+++ b/Content.Client/Actions/ActionsSystem.cs
@@ -5,6 +5,7 @@ using Content.Shared._RMC14.Actions;
 using Content.Shared.Actions;
 using Content.Shared.Actions.Components;
 using Content.Shared.Charges.Systems;
+using Content.Shared.GameTicking;
 using Content.Shared.Mapping;
 using Content.Shared.Maps;
 using JetBrains.Annotations;
@@ -12,6 +13,8 @@ using Robust.Client.Player;
 using Robust.Shared.ContentPack;
 using Robust.Shared.GameStates;
 using Robust.Shared.Input.Binding;
+using Robust.Shared.Map;
+using Robust.Shared.Physics.Systems;
 using Robust.Shared.Player;
 using Robust.Shared.Prototypes;
 using Robust.Shared.Serialization.Manager;
@@ -51,6 +54,7 @@ namespace Content.Client.Actions
 
         // RMC14
         [Dependency] private readonly RMCLagCompensationSystem _rmcLagCompensation = default!;
+        private readonly Queue<ActionRequest> _requested = new();
 
         public override void Initialize()
         {
@@ -64,7 +68,51 @@ namespace Content.Client.Actions
 
             SubscribeLocalEvent<EntityTargetActionComponent, ActionTargetAttemptEvent>(OnEntityTargetAttempt);
             SubscribeLocalEvent<WorldTargetActionComponent, ActionTargetAttemptEvent>(OnWorldTargetAttempt);
+
+            // RMC14
+            SubscribeLocalEvent<RoundRestartCleanupEvent>(OnCleanup);
+            UpdatesOutsidePrediction = true;
+            UpdatesBefore.Add(typeof(SharedPhysicsSystem));
         }
+
+        // RMC14 start
+        public override void Update(float frameTime)
+        {
+            base.Update(frameTime);
+
+            if (!GameTiming.IsFirstTimePredicted)
+                return;
+
+            if (_requested.Count > 0)
+                _rmcLagCompensation.SendLastRealTick();
+
+            RequestPerformActionEvent ev;
+            while (_requested.TryDequeue(out var request))
+            {
+                switch (request)
+                {
+                    case (var actionId, var targetId, { } coords):
+                        // World target case
+                        ev = new RequestPerformActionEvent(actionId, targetId, coords, _rmcLagCompensation.GetLastRealTick(null));
+                        break;
+                    case (var actionId, { } targetId, null):
+                        // Entity target case
+                        ev = new RequestPerformActionEvent(actionId, targetId, _rmcLagCompensation.GetLastRealTick(null));
+                        break;
+                    case (var actionId, null, null):
+                        // Trigger case
+                        ev = new RequestPerformActionEvent(actionId, _rmcLagCompensation.GetLastRealTick(null));
+                        break;
+                }
+                RaisePredictiveEvent(ev);
+            }
+        }
+
+        private void OnCleanup(RoundRestartCleanupEvent ev)
+        {
+            _requested.Clear();
+        }
+        // RMC14 end
 
 
         private void OnActionAutoHandleState(Entity<ActionComponent> ent, ref AfterAutoHandleStateEvent args)
@@ -215,8 +263,8 @@ namespace Content.Client.Actions
             }
             else
             {
-                var request = new RequestPerformActionEvent(GetNetEntity(action), _rmcLagCompensation.GetLastRealTick(null));
-                RaisePredictiveEvent(request);
+                // RMC14
+                _requested.Enqueue(new(GetNetEntity(action), null, null));
             }
         }
 
@@ -332,7 +380,10 @@ namespace Content.Client.Actions
                 PerformAction((user, user.Comp), (uid, action));
             }
             else
-                RaisePredictiveEvent(new RequestPerformActionEvent(GetNetEntity(uid), GetNetEntity(targetEnt), GetNetCoordinates(coords), _rmcLagCompensation.GetLastRealTick(null)));
+            {
+                // RMC14
+                _requested.Enqueue(new(GetNetEntity(uid), GetNetEntity(targetEnt), GetNetCoordinates(coords)));
+            }
 
             args.FoundTarget = true;
         }
@@ -372,7 +423,8 @@ namespace Content.Client.Actions
             }
             else
             {
-                RaisePredictiveEvent(new RequestPerformActionEvent(GetNetEntity(uid), GetNetEntity(entity), _rmcLagCompensation.GetLastRealTick(null)));
+                // RMC14
+                _requested.Enqueue(new(GetNetEntity(uid), GetNetEntity(entity), null));
             }
 
             args.FoundTarget = true;
@@ -385,5 +437,8 @@ namespace Content.Client.Actions
         }
 
         public record struct SlotAssignment(byte Hotbar, byte Slot, EntityUid ActionId);
+
+        // RMC14
+        readonly record struct ActionRequest(NetEntity ActionId, NetEntity? TargetId, NetCoordinates? TargetCoords);
     }
 }
--- a/Content.Client/Actions/ActionsSystem.cs
+++ b/Content.Client/Actions/ActionsSystem.cs
@@ -83,9 +83,6 @@ namespace Content.Client.Actions
             if (!GameTiming.IsFirstTimePredicted)
                 return;
 
-            if (_requested.Count > 0)
-                _rmcLagCompensation.SendLastRealTick();
-
             RequestPerformActionEvent ev;
             while (_requested.TryDequeue(out var request))
             {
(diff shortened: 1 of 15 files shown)


MidZik:
I didn't fix "ghost interactions" yet, such as when you try to wield your gun and it unwields because the the client mispredicted the wield cooldown. I believe the root cause is the same (interactions are handled out of simulation), and interactions would have to also be queued to be handled during the next tick.

DrSmugleaf (commented):
I'm a bit iffy on this because RaisePredictiveEvent should always be reliable, any discrepancy should be from our lag compensation code
I'd need to ask Electro but I'm not sure when he'll be free next

MidZik:
I guess it depends what you mean by "reliable". The server will receive the networked event and process it in-simulation as normal. And the client will have the event re-raised during re-predicted ticks in-simulation as well.

The problem is that the first time the local event is raised, it's raised out of simulation. That will never be completely accurate. CurTime will be some time between ticks and entity positions are going to be interpolated. I understand this is meant to make actions "snappy" and run without delay on the client, and the in-simulation re-predicted ticks should "correct" any discrepancies due to that. But the secondary issue is that this is the only time the event is raised where "IsFirstTimePredicted" is true.

For example, one area I'm a bit familiar with: shooting xeno projectiles. `TryShoot` is called while `InSimulation` is false, as spitting is an action. Then the client creates the clientside projectile because `IsFirstTimePredicted` is true. That's the only time it creates the projectile. But the origin of the projectile is not correct because it was shot out of simulation. Unfortunately, the client informs the server of hits based on in this incorrectly-predicted projectile.

For example, some debug logs of me shooting a projectile **(edit) while I'm moving** without this fix:
Client:
```
[DEBG] system.xeno_projectile: TryShoot:
  CurTick:       524
  LastRealTick:  515
  IsServer?      False
  InSimulation?  False      # <-- The only time the event is processed while InSimulation is false.
  InPrediction?  True
  FirstTimePred? True       # <-- The only time the event is processed while IsFirstTimePredicted is true.
  Xeno Origin:   Map=1, X=-0.82, Y=0.33   # <-- Different origin due to interpolation.
  Target Coords: Map=1, X=-2.40, Y=2.34
[DEBG] system.xeno_projectile: TryShoot:
  CurTick:       524
  LastRealTick:  516
  IsServer?      False
  InSimulation?  True
  InPrediction?  True
  FirstTimePred? False
  Xeno Origin:   Map=1, X=-0.70, Y=0.45
  Target Coords: Map=1, X=-2.40, Y=2.34
[DEBG] system.xeno_projectile: TryShoot:
  CurTick:       524
  LastRealTick:  517
  IsServer?      False
  InSimulation?  True
  InPrediction?  True
  FirstTimePred? False
  Xeno Origin:   Map=1, X=-0.70, Y=0.45
  Target Coords: Map=1, X=-2.40, Y=2.34
...
```
Server: 
```
[DEBG] system.xeno_projectile: TryShoot:
  CurTick:       524
  LastRealTick:  515
  IsServer?      True
  InSimulation?  True
  InPrediction?  False
  FirstTimePred? True
  Xeno Origin:   Map=1, X=-0.70, Y=0.45
  Target Coords: Map=1, X=-2.40, Y=2.34
```

However, **with** this fix, the client's result is often exactly the same as the server's as long as there wasn't a different mispredict elsewhere. The predicted projectile hits that the client sends to the server end up being more accurate.

Now, _many_ actions don't actually care about this detail for their lag compensation to work. For example, tail stabs. The server doesn't rely on the client telling it about any hits like it does for projectiles, so the server will lag compensate those just fine (when the -1 to setting latest tick is removed). The only issue is the client mispredicting if the action will work for a brief moment, because the target might be in range while interpolated, but out of range when finally simulated correctly.

Melee attacks do not suffer from this problem at all because they only get processed during an update loop. In that sense, I am making actions work the same way that melee attacks work.

If you are not confident about this change, I can atomize out the cooldown check changes and the fix to setting the latest real tick. That will still fix the ghost/fake actions issue and make the lag comp more reliable for actions and interactions.

MidZik:
Thinking about it some more, a lower-impact way to fix it would be to explicitly mark actions we want queued for the next update loop somehow, perhaps with a new component or adding some flag to an existing component. Then we can get this run-during-update-loop behavior only for actions we want it for, rather than all actions. It would still need to delay the predictive raising, though.

Another way would be to have each system that handles one of these predictive events to delay the handling in its own system. It would be even more explicit than the above, but more annoying to implement since now each system would need some new code to handle this behavior.

MidZik:
I looked into raising the predictive events as normal and only delaying the handling of the event, but due to the way the client handles predictive events I don't think that's a good avenue. The client gets into what I consider "weird" states during the game loop. 

For example while handling input for tick 100 the last real tick might be 90, but when handling the system updates for tick 100 the client has applied the server's state for tick 91, which means the out of simulation prediction was using a different state than the in-simulation prediction. 

Or the client processes predictive events for tick 100 as if it's "not first time predicted" but then _after that_ the update loop processes tick 100 as "first time predicted". So the client would end up handling the predictive events for tick 100 where IsFirstTimePredicted is false, then my delayed event that's handled during the update loop would be IsFirstTimePredicted true, but at that point the action is "on cooldown" because the predictive event put it on cooldown so my delayed event doesn't end up actually do anything AND all the logic that depends on IsFirstTimePredicted ends up never being called.

So I decided to stick with delaying RaisePredictiveEvent. From my testing, delaying it is reliable. I was re-testing the xeno projectile prediction and lag comp with this change, and I saw the client and the server completely agree on projectile positions.

I made this "delaying" behavior opt-in with a new InSimulationOnly flag for actions, and flagged all xeno actions with it. I suspect all actions that benefit from accurate prediction could use this flag, but better to keep the scope small for now.

MidZik:
Here's some logs I collected for myself with these changes, to double check that client and server were agreeing with each other.

### Tail stab range checks (target moving)
```
# CLIENT
[DEBG] system.interaction: InRangeUnobstructed:
  Session:  localhost@JoeGenero
  CurTick:  5411
  LastReal: 5403
  Target:   Imani Wilkinson
  Origin
    Coords: Map=1, X=0.37, Y=2.41
  Target
    Coords: Map=1, X=-1.92, Y=3.50


# SERVER
[DEBG] system.interaction: InRangeUnobstructed:
  Session:  localhost@JoeGenero
  CurTick:  5411
  LastReal: 5403
  Target:   Imani Wilkinson
  Origin
    Coords: Map=1, X=0.37, Y=2.41
  Target
    Coords: Map=1, X=-1.92, Y=3.50   # the lag-comp'd position
```

### Projectile hit checks (origin and target both moving)
```
# CLIENT
[DEBG] system.xeno_projectile: TryShot
  Session:      Robust.Shared.Player.ActorComponent  # messed up my logging code oh well
  CurTick:      8202
  LastRealTick: 8194
  IsFirstTime?  True
  InSimulation? True
  Origin:       Map=1, X=1.81, Y=0.44
  Target:       Map=1, X=1.18, Y=3.31
[DEBG] system.xeno_projectile: SENDING PREDICTED PROJECTILE HIT!!
  Shot ID:        1
  Cur Tick:       8203
  LastRealTick:   8195
  Phys Substep:   2
  In simulation?  True
  ApplyingState?  False
  FirstTimePred?  True
  Shot At Tick:   8202
  Pred Hit Tick:  8203
  Substep:        2
  Shot Coords:    EntId=2, X=1.18, Y=3.34
  Target Coords:  EntId=2, X=0.83, Y=3.50


# SERVER
[DEBG] system.xeno_projectile: TryShot
  Session:      Robust.Shared.Player.ActorComponent
  CurTick:      8202
  LastRealTick: 8194
  IsFirstTime?  True
  InSimulation? True
  Origin:       Map=1, X=1.81, Y=0.44
  Target:       Map=1, X=1.18, Y=3.31
[DEBG] system.xeno_projectile: Received predicted hit:
  Session:  localhost@JoeGenero
  Cur Tick: 8203
  Target:   2359
  Shot ID:  1
  Shot At:  8202
  Hit Tick: 8203
  Substep:  2
[DEBG] system.xeno_projectile: Predicted hit checks passed, will test collision. Details:
  Session Name:   localhost@JoeGenero
  Last Real Tick: 8195
  Shot ID:        1
  During shoot?   False

  Cur Tick:       8203
  Pred Hit Tick:  8203

  Shot Tick:      8202
  Pred Shot Tick: 8202

  Pred Substep:   2
  Adjust Substep: 2
[DEBG] system.r_m_c_lag_compensation: Lag comp collide data:
  Session Name:   localhost@JoeGenero
  Pre-Substep
    Proj Coords:  Map=1, X=1.39, Y=2.38
  CurTick:        8203
  Substep:        2
  Projectile Pos: <1.1750865, 3.341524>     # EXACT match to client predicted projectile position
  Target Pos:     <0.8333421, 3.5>          # EXACT match to client predicted target position
  Proj AABB:      <1.0650865, 3.231524>
                  <1.2850865, 3.4515238>
  Target AABB:    <0.48334208, 3.15>
                  <1.1833421, 3.85>
  AABB Intersect? True
  AABB Distance:  0
[DEBG] system.xeno_projectile: Predicted hit from 'localhost@JoeGenero' ++ CONFIRMED!! ++
```

PR #10110, opened 2026-06-10
Slightly reduce the Sapper's Long Range Sight view range increase

Arendian:
## About the PR

Reduced the zoom multiplayer from 1.9(20 x 14.25) to 1.8(19x13.5)

## Why / Balance

The view range is currently slightly too high.(13 tiles in parity)

## Requirements

**Changelog**

:cl: Dygon
- fix: Fixed the Sapper Boiler's view range increase from "Long Range Sight" being too high, reduced it by 11.1%.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/boiler.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Mobs/Xeno/boiler.yml
@@ -227,7 +227,7 @@
       590: Critical
       690: Dead
   - type: XenoZoom
-    zoom: 1.9, 1.9
+    zoom: 1.8, 1.8
     offsetLength: 0
     speed: 1
     doAfter: 0


MidZik:
1.9 seems correct to me. Perhaps it can be adjusted to 1.89 to remove that extra pixel.

## CM13

SapperRangeCM13

## RMC14 with 1.9x zoom

SapperRangeRMC14

Arendian:
> 1.9 seems correct to me. Perhaps it can be adjusted to 1.89 to remove that extra pixel.

Your game is not in fullscreen, if you put it in fullscreen you'll see it's actually 14.3 at zoom level 1.9. It also checks out math wise, if you multiply the default vertical view distance(7.5) by the zoom level: 7.5 * 1.9 = 14.25 while 7.5 * 1.8 = 13.5.
image

MidZik:
This whole time I never knew fullscreen adds nearly a tile to my vertical view range, that's my mistake.

MidZik:
1.8x seems correct.

I consider the view range differences between fullscreen and window to be a bug, but I assume that's an engine bug.

Arendian:
> 1.8x seems correct.
> 
> I consider the view range differences between fullscreen and window to be a bug, but I assume that's an engine bug.

It only really happens when you have it windowed and maximized. It appears that the viewport tries to use the whole screen, with the "task bar" part being drawn over the upper part of it. If you make the window slightly smaller you can see the same distance as when playing fullscreen. It's an easy thing to miss, and most likely an engine bug yes.

DrSmugleaf (approved):

PR #10130, opened 2026-06-11
Disabled Hold-To-Fire on Su-6

carccborg:
## About the PR

Added GunClickToFire component to the su-6 yaml.

## Why / Balance

The SU-6 is my (and others') replacement for the M13 now that the M13 has been parity nerfed. Currently, the SU-6 has 300 DPS 20 AP @ 0 tiles and 250 DPS 20 AP @ 5 tiles, which makes it better than marine primaries at close range with the only downside being the small magazine size and somewhat nasty falloff. Ammo is plentiful in squad ammo vendors and in requisitions.
In lieu of deviating from parity and changing its stats, and with the spirit of the M13 parity nerf in mind, I've disabled the hold-to-fire accessibility feature on the SU-6. We can adjust its stats and re-add hold-to-fire post-parity.

## Technical details

Oneline yamlmaxxing

## Media

Small fix, not adding media

## Requirements

**Changelog**

:cl:
- tweak: The SU-6 Smart Pistol can no longer be used with hold-to-fire.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Pistols/su6_pistol.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Pistols/su6_pistol.yml
@@ -27,6 +27,7 @@
     - Burst
     soundGunshot:
       path: /Audio/_RMC14/Weapons/Guns/Gunshots/su6_shoot.ogg
+  - type: GunClickToFire
   - type: RMCSelectiveFire
     baseFireModes:
     - SemiAuto


carccborg:
While I as someone with hand/wrist pain appreciate the feature for existing I think the damage that the SU-6 can spit out is a bit silly and it was a niche pick at best until everyone (including myself) was gifted with the wonders of a perfect ten clicks per second. The accessibility feature makes things like the SPR available as a playstyle to me again, but the SPR isn't any stronger as a result of the feature and the SU-6 is.

kleinerstation13:
Can't this exact DPS still be done if you're fast enough at clicking?

carccborg:
Absolutely. Which is also why I'd imagine the M1984 and MOU were made exempt from the feature.

Vixting (approved):
M1984 was disabled because it has the same insane fire rate

kleinerstation13:
Maybe a feature which caps the firerate on some guns with the autofire enabled to lower the DPS could be implemented at one point in the future so people with wrist pain can still take advantage of these weapons without them being OP in their hands

Vixting:
> Maybe a feature which caps the firerate on some guns with the autofire enabled to lower the DPS could be implemented at one point in the future so people with wrist pain can still take advantage of these weapons without them being OP in their hands

The current disabled guns will have it removed once their fire rate is adjusted, likely post parity

Drazamuffin:
> M1984 was disabled because it has the same insane fire rate

I see what ya mean as M1911 still has it last i checked

kleinerstation13:
M1911 has a much slower firerate tbh I can manually spam it without clicking too much

CatAndHats (approved):
LGTM

PR #10177, opened 2026-06-16
Security colonists on Fiorina can no longer roll for prisoner

kleinerstation13:
## About the PR
Security survivor on Fiorina can no longer roll prisoner, now civilian survivor rolls for it instead

## Why / Balance
Prisoner has abysmal skills, all of which the other secsurv roles have (Medical 0, Engineering 0, Firearms 1) so they have no reason to take up limited slots like RCO's and weya goons do

Gang leader spawns with identical if not better equipment AND better skills and yet they are a civilian role so I see no reason why it shouldn't apply to both

## Technical details
Two lines of yaml

## Requirements

**Changelog**

:cl:
- tweak: Security Colonists can no longer become Prisoners, you must now roll Civilian Colonist to become one

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/rmc_planets.yml
+++ b/Resources/Prototypes/_RMC14/Entities/rmc_planets.yml
@@ -233,6 +233,7 @@
       - CMSurvivor: -1
       - CMSurvivorISR: -1
       - CMSurvivorGangLeader: -1
+      - CMSurvivorPrisoner: -1
       CMSurvivorCorporate:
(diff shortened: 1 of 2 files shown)


MoonlightHollow:
Make sure to change the .yml file for prisoner to inherit off of civilian survivors instead of security.

kleinerstation13:
Good point, also turns out them being a secsurv in preference is parity but this is an extremely minor deviation

DrSmugleaf (approved):

PR #10215, opened 2026-06-19
Lingering acid effect rework

Vermidia:
## About the PR

Acids do a flat 1.25 heat damage per tick, that gets multiplied by x2, x3, and eventually x4 if it stays on long enough. Stronger acids no longer do more damage but instead have a higher max duration, and acid applying moves will prolong acid durations instead of reapplying acid now. Instead of a specific number of resist attempting to extinguish acid removes 27 seconds from the duration.

Added strong acid even though despoiler isn't done yet, with its armor reducing properties aswell.

## Why / Balance

Parity!

## Technical details

Acid part of the spit system seperated out into it's own file. Alot of changes to systems and comps to make them use acid protos and the new functions in the spit system.

Damage numbers for acid are probably right (someone should prob check just in case anyway), for spray acid a bit more suspect since our numbers on that are all over the place,

~~Note: Unsure we can use the weak acid on mob sprites.~~ Seems to be good!

## Media

See development channel, didn't feel like converting them.

## Requirements

**Changelog**

:cl:
- fix: Fixed how spitter/acider/etc lingering acid works. Acid of all tiers now does the same damage, that is multiplied by 2x, x3, and x4 at 21, 30, and 50 respectively. Higher acid tiers have a longer max duration instead, and now acid applying moves generally prolong the duration of the acid instead of replacing it. Attempting to extinguish acid removes 27 seconds from the duration instead of having a resist count. Sprites have changed slightly.
- fix: Fixed non-spitter sprays enhancing acid. Spitter acid spray now does 25 extra damage on successful enhancing/duration prolonging.


Content.Shared/_RMC14/Xenonids/Projectile/Spit/Charge/XenoAcidPrototype.cs
```diff
@@ -0,0 +1,51 @@
+using Content.Shared.Damage;
+using Robust.Shared.Prototypes;
+using Robust.Shared.Serialization.TypeSerializers.Implementations.Custom.Prototype.Array;
+
+namespace Content.Shared._RMC14.Xenonids.Projectile.Spit.Charge;
+
+[Prototype]
+public sealed partial class XenoAcidPrototype : IPrototype, IInheritingPrototype
```
DrSmugleaf: For this it would be better to make this a component on entities, and refer to it by EntProtoId<XenoAcidDefinitionComponent> instead of ProtoId<XenoAcidPrototype>
You don't need to bother with ID/Parents/Abstract that way

https://github.com/RMC-14/RMC-14/blob/master/coding-standards.md#prototypes

DrSmugleaf (changes requested):
Thanks for the PR!

PR #10219, opened 2026-06-19
SOP charging time update

AlexisMoonlight:
## About the PR
Removed an explicit time limit for sentencing.

## Why / Balance
As discussed with Provost Investigators, the spirit of SOP is to make sure arresting and charging done reasonably fast, not to impose a strict time limit on the warden/sentencing MP.  This change brings the text of SOP to be in line with the spirit.

## Technical details
xml change

## Media
N/A

## Requirements

**Changelog**

:cl:
- remove: Removed explicit time limit on arresting/charging. MPs should still sentence a suspect with reasonable speed.

### DIFF
--- a/Resources/ServerInfo/Guidebook/_RMC14/Law/RMCMarineLawArresting.xml
+++ b/Resources/ServerInfo/Guidebook/_RMC14/Law/RMCMarineLawArresting.xml
@@ -21,7 +21,7 @@
   - Take down the Suspect if they resist and securely restrain them.
   - Move the Suspect to a holding cell.
   - Continue to [bold]charging procedure[/bold].
-  - A suspect must be arrested and charged within 10 minutes. Any longer than this without good reason is considered [bold]Neglect of Duty[/bold] by whoever is most responsible for the delay.
+  - A suspect must be arrested and charged within reasonable time. Any significant delay without good reason is considered [bold]Neglect of Duty[/bold] by whoever is most responsible for the delay.
 
   ## Arresting MP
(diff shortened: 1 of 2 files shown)


BreakpointDebug1234:
I believe an actual time limit is needed so as to limit how much someone can spend in holding. I really think 10 minutes is more than enough to gather sufficient evidence to decide on charging someone.

Having no actual limit might encourage wardens to hold people for unnecessarily long amounts of time.

Keep in mind this is taking the time from a real person who might be able to play only a single round that day.

QuietlyWhisper:
Any longer than 10 minutes is absurd, I can't approve of this change. The time limit is explicit for that reason.

AlexisMoonlight:
The intention here to move the time limit to be more informal, wardens/sentencing officers are still strongly encouraged to sentence as fast as possible.
This is intended for situations where there are multiple people to be sentenced, where a person involved is a slow typer, a new warden or a few of the above. I have seen all of these happen and get close to the 10 minute limit.
I also believe that having the limit explicitly stated may encourage some people to take it as alotted time to sentence and take more time than would be necessary otherwise.
I believe having the limit be informal, enforced by more experienced MPs, CMPs and PvIs will prevent people from dragging sentencing out without a good reason, but will allow for the cases where more time is needed to not be dragged out even more by arguing about said time limit.
As I've stated, often enough this time limit is informal already. I've had situations before where it was ignored as a letter of law for Spirit reasons. 
Though I'm not a hardline advocate for this change. For the same reasons listed, Spirit may work just as well. This is only an attempt to bring the letter closer to it.

PursuitInAshes:
Thanks for the PR! Closing per above.
We'll talk about some ideas as PvIs that wouldn't involve law changes, so we can try to leave those for later.

PR #10235, opened 2026-06-21
[Tweak] Watches spawn directly, not in box

x5tink:
## About the PR
I keep forgetting my watch because the boxes spawn on the floor. 

Tweaked so the watches spawn directly, not in a box, so that they go into the inventory directly on spawn, and not get dropped onto the floor and leave behind a tonne of useless boxes around. Side effect, you can now see what the watch looks like in the loadout screen before picking.

As a side note, i've not removed the boxes entirely, as im not sure if its used for other things, or in a vendor of some kind, simply changed what spawns from the loadout screen.

## Technical details
tiny YML changes

## Media
BEFORE:
Screenshot 2026-06-21 152344
Screenshot 2026-06-21 152410

AFTER:
Screenshot 2026-06-21 152456
Screenshot 2026-06-21 152429

## Requirements

**Changelog**
:cl:
- tweak: Watches now spawn directly, not in a box.

### DIFF
--- a/Resources/Prototypes/_RMC14/Roles/Loadouts/Items/accessories.yml
+++ b/Resources/Prototypes/_RMC14/Roles/Loadouts/Items/accessories.yml
@@ -30,7 +30,7 @@
         time: 90000 # 25 hrs
   storage:
     back:
-    - RMCWatchCaseBasic
+    - RMCWatchBasic
 
 - type: loadout
   id: WatchSurvRMC
@@ -42,7 +42,7 @@
         time: 36000 # 10 hrs
   storage:
     back:
-    - RMCWatchCaseSurv
+    - RMCWatchSurv
 
 - type: loadout
   id: WatchPilotRMC
@@ -54,7 +54,7 @@
         time: 43200 # 12 hrs
   storage:
     back:
-    - RMCWatchCasePilot
+    - RMCWatchPilot
 
 - type: loadout
   id: WatchLiaisonRMC
@@ -66,7 +66,7 @@
         time: 126000 # 35 hrs
   storage:
     back:
-    - RMCWatchCaseLiaison
+    - RMCWatchLiaison
 
 - type: loadout
   id: WatchSynthRMC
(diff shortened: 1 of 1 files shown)


AverageNotDoingAnythingEnjoyer:
Why are watches boxes not spawning inside a satchel instead
Actually, some accessories also have this issue where if you take an armband in loadouts for example, it would spawn on the ground

x5tink:
> Why are watches boxes not spawning inside a satchel instead Actually, some accessories also have this issue where if you take an armband in loadouts for example, it would spawn on the ground

The watch cases atleast, can't be stored in the first place, which is why the drop on the floor. I've not looked into why. The boxes contain only 1 item, so they are kind of pointless anyway. ( it looks like its a reused PDT tracker box, which does in fact store more than one item, so it makes sense there )

No clue about the other items your talking about.

AverageNotDoingAnythingEnjoyer:
> The watch cases atleast, can't be stored in the first place, which is why the drop on the floor. I've not looked into why.

Seems they should be of small size, if we go with battle buddy kit parenting
https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/items/devices/personal_data_transmitter.dm#L167
And as for spawning logic, I couldnt find the diff watches we have, but the cheap watch available to marines on CM spawns without a case
https://github.com/cmss13-devs/cmss13/blob/master/code/modules/client/preferences_gear.dm#L1320

CatAndHats:
> Seems they should be of small size, if we go with battle buddy kit parenting https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/items/devices/personal_data_transmitter.dm#L167 And as for spawning logic, I couldnt find the diff watches we have, but the cheap watch available to marines on CM spawns without a case https://github.com/cmss13-devs/cmss13/blob/master/code/modules/client/preferences_gear.dm#L1320

Watches being small fix is now merged.
The different watches are I believe from the battlepass on CM? or something similar. They got the time req locks instead.

CatAndHats (approved):
LGTM. It is correct to how CM13 does it.

Thereisabear (approved):

PR #10242, opened 2026-06-21
Adds: Nightvision Color Options

wheelwrightt:
## About the PR
Added the option to change the color of the nightvision visor. The initial values I tested are what is used in parity, but didn't translate well here so I made my own. Also increased the grain level so its not just a full bright visor. 

## Why / Balance
parity (also the green gives me a headache)

## Technical details
nightvisioncolor mapped to clientside nightvisionvisorcolor cvar, added the dropdown to the accessibility(this can be changed if needed) tab on the option menu and updated the overlays to pass the colour to the shader. Also uncommented the film grain and upped it. CM13 values made the colours too bright so i had to make my own from what little media i could find that nvgs are in.

## Media

https://github.com/user-attachments/assets/1aea741f-efc8-4d64-be17-934ab79498ec

## Requirements

**Changelog**
:cl:
- add: Different color options for the nightvision visor (can be set in the options menu)


AverageNotDoingAnythingEnjoyer:
How parity is white, this seems like negating the color filter altogether

wheelwrightt:
> How parity is white, this seems like negating the color filter altogether

just looked at CM13, rn they have a bug where the nv is just pure fucking white so its hard to test but from old sc all the nvs are darker with even more grain and a vignette on the border of the screen. Will have a look at it. 

If anyone has any solid media showing the colours etc in cm please link it nice one.

https://github.com/user-attachments/assets/20bfc709-2f92-4430-a88f-a0c536e81231 old colours for posterity

wheelwrightt:
Updated media

PR #10290, opened 2026-06-26
Adds a setting to choose a default 'Ghost Hearing' state

SabreML:
## About the PR
Adds a setting to the options menu allowing you to set a default state for 'Ghost Hearing'.

## Why / Balance
It's a bit fiddly having to manually toggle ghost hearing off every time I die to avoid chat spam. This is also already a setting on a number of SS13 servers, so it just makes sense I think.

## Technical details
This removes the `GhostHearing` component from observer yml, and instead adds it if required when a player first "attaches" to a ghost. In practice this works the same way as before, just with an added default setting.

## Media

https://github.com/user-attachments/assets/8b945b74-a9eb-4404-9c38-fba25a1b02ee

## Requirements

**Changelog**

:cl:
- add: Added a setting to the options menu to set a default 'Ghost Hearing' state.


Content.Server/_RMC14/Mobs/CMGhostSystem.cs
```diff
@@ -52,6 +49,29 @@ private void OnCMGhostStartup(EntityUid uid, CMGhostComponent comp, ComponentSta
...
         }
 
+        private void OnPlayerAttached(Entity<CMGhostComponent> ent, ref PlayerAttachedEvent args)
+        {
+            if (!TryComp<GhostComponent>(ent, out var ghost))
+            {
+                // this should never actually happen
+                Log.Error($"{ToPrettyString(ent)} has CMGhostComponent, but not GhostComponent. (Should have both!)");
+                return;
+            }
```
SabreML, line 59: I'm not sure if this check+error is actually required, but it occurred to me as a possible edge case.

SabreML:
Tests all seem to pass locally. I think the failure is unrelated?

SabreML:
Okay this one is *definitely* unrelated (Same fail on master). I'll try pulling again when a fix gets merged.

PursuitInAshes:
Thanks for the PR!
We talked about it on discord, but for others looking:
We're going to close this in favor of #10707 which should have a similar if not identical effect, but mildly cleaner.

PR #10292, opened 2026-06-26
fix being blocked from splitting acided stack

Vixting:
## About the PR

fix being blocked from splitting acided stack
## Why / Balance

parity
## Requirements

**Changelog**

:cl:
- fix: Fixed being blocked from splitting a stack of acided items

### DIFF
--- a/Content.Server/Stack/StackSystem.cs
+++ b/Content.Server/Stack/StackSystem.cs
@@ -42,10 +42,6 @@ namespace Content.Server.Stack
         /// </summary>
         public EntityUid? Split(EntityUid uid, int amount, EntityCoordinates spawnPosition, StackComponent? stack = null)
         {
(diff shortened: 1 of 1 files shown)


ott043:
Iirc, this is intentional as you used to be able to split and combine stacks of sheets to remove the acid. So I think its a balance thing.

Arendian (approved):
Parity

CatAndHats:
https://github.com/cmss13-devs/cmss13/pull/12925 it seems CM13 considers this an exploit and has a PR up to remove it. Entirely up to us if we agree, if we do agree with them then this PR should be closed. We can discuss in Maint chat at some point.

CatAndHats:
As per above, nobody raised objections to following the new CM13 parity (which is what we currently have) so I will close this now

PR #10295, opened 2026-06-27
[Fix] Console Can't resolve on Ent 0!

Thereisabear:
## About the PR
Fixes console spam. Just a defensive check; simple early exit if the thing its referencing is actually not/not yet there (sod ye Ent 0).

## Why / Balance
Fixes are good and I can actually use the console now!
Mayhaps _very minor_ performance bump due to engine not having to log all the errors. 

## Technical details
&& Exists check after the _is null_ bits for those i could find, 3 days worth of launcher logs...

## Requirements

**Changelog**

:cl: Thereisabear

- fix: Console should have less Can't resolve Entity 0 now.


Aurallianz (dismissed):
LGTM, Its odd that these ents are not properly populating for clients, *shrug*

DrSmugleaf (dismissed):

PR #10304, opened 2026-06-27
[PVE] GM Request System

### LINKED ISSUE #8708: [PvE] Gamemaster Request System
### Problem
Right now, we don't have a great way of intaking and processing requests in round. We don't want to use ahelps for these because ideally we should be keeping them as clear as possible and probably about 80% of requests in reality need no reply or ahelp interaction outside of the initial request from the user. For the ones that do then its perfectly fine to use ahelp obvs

### Purposed Fix
BaseSpecies (And Ghosts) on the normandy get a new verb on their entities.
This verb will allow players to input text into a box to send requests to GMs, much like how prayers function upstream (Probably formatted easier to read). With any request there should be a admin log crated and a text warning sent in chat (So admins know to check the panel if they don't have it open).
There might also need to be some kind of way to just.. remove someone's access to this system if we absolutely required, similar to chat bans.

On the GM end, they'll be a new UI/Tab in our `gamemaster` ui + a command to be used for those without access to the UI (Assisting admins) or those who would prefer to just use the command anyway.
Sent in requests should (probably) be formatted like:
``<Time>: <Name> <Username> - <Request/Text> | [TP To] [Open Ahelp] [Delete Request]``
Along with some way to declare that you have "claimed" them (Probably just adding your name to the right of the request or something) that way we don't double up time as much

Which can then be automatically cleared when the round returns back to lobby.

I do not know what CM13 PvE does but this feels like a pretty good solution for what GMs are looking for out of a system to handle requests without adding in a third ahelp type system for people to figure out how to use (Apparently, people have struggled sending the correct thing already with ahelp and mhelp)

ThataKat:
## About the PR
Adds the GM Request system, a streamlined way to collect and handle non-urgent requests for gamemasters. The system is enabled on Normandy by default.

All player controlled entities (Humanoids, ghosts, animals) will find a "GM Request" verb on themselves when the system is enabled, which allows them to send a short request.

The request UI lets GMs toggle the system, claim and hide logs, mute the notification sound, and take a variety of actions to interact with the player who sent the log.

Logs are cleared when a new round starts (logs from the prior round will carry through into the lobby). Player requests are recorded in the admin logger as well, if they need to be referenced for administrative action.

## Why / Balance
Closes #8708

## Technical details
C# (largely EUI), UI code (xaml and cs), toml (Normandy config), ftl (most everything is localized), CVars

`GMRequestManager` handles the creation, storage, and modification of logs on the backend, with help from `GMRequestSystem`. There is no caching or long term storage of logs, the Dictionary in `GMRequestManager` is their "true" location.

`GMRequestVerbSystem` gives out the verb when requesting is enabled, and packages the information to send to the manager.

`GMRequestEui` does Eui things to communicate between client and server, primarily keeping the client up to date with the state of the logs, and relaying any requested log changes to the server.

`GMRequestClientManager` stores and handles the client copy of the logs. The client EUI sends modifications here when it detects updates.

`GMRequestWindow` is the UI panel itself, containing all of the local UI functionality (anything that can be accomplished without speaking to the server Manager). `GMRequestWindowHeader` and `GMRequestWindowEntry` are children separated for convenience, with `Entry` making up the display for the log entries themselves.

Anything in the window that requires communicating with the server Manager is done within the client EUI.

`gmrequestui` and `togglegmrequest` are new commands in their respective server files.

I am relatively new to C#, so expect this to lack some elegance.

## Media
VerbExample
DialogExample
WindowExample

Extended demonstration:

https://youtu.be/qFlTIXwdYOg

## Requirements

**Changelog**

:cl:
- add: [PVE] Added the GM Request system, for handling minor requests. When enabled, look for a new verb on your player entity to send requests.
- admin: [PVE] Added the GM Request system and panel. Access the UI with "gmrequestui", see PR for more details.


Resources/Locale/en-US/_RMC14/gmrequest/gmrequest.ftl
```diff
@@ -0,0 +1,21 @@
...
+rmc-gm-request-subtle-ui-title = Subtle Message
+rmc-gm-request-subtle-ui-message = Message...
+rmc-gm-request-subtle-ui-popup = Popup message...
+rmc-gm-request-subtle-default = You hear the voice of the gamemaster...
+
+rmc-gm-request-command-toggle-description = Toggles the ability to send GM Requests.
+rmc-gm-request-command-toggle-too-many-arguments-error = Too many arguments.
+rmc-gm-request-command-toggle-invalid-argument-error = Invalid argument.
+rmc-gm-request-command-toggle-enabled = GM Request has been enabled.
+rmc-gm-request-command-toggle-disabled = GM Request has been disabled.
```
PursuitInAshes:
```suggestion
rmc-gm-request-command-toggle-enabled = GM Requests have been enabled.
rmc-gm-request-command-toggle-disabled = GM Requests have been disabled.
```

Not a full review, just noticed this while skimming

DrSmugleaf (commented):
I need to check this more on my PC but if this doesn't already have it, there should be a cvar for max message length and message cooldown

ThataKat:
Cooldown and max length added, currently at 10 seconds and 300 characters
image

QuietlyWhisper:
Does the ui handle longer messages well? It seems to get cut off on the usernames as well

ThataKat:
The window can be resized to give them more space, and there's a hover popup with name and message that should be able to fit all of a long message (off the top of my head anyways, been a while since I tested).

I recall trying to get message text to overflow into multiple lines (like Admin Logs) if it was too long but I was having trouble with arcane UI code. Something to do with the StyleClass of ListContainerButton, and the general implementation being different than Logs

Local host usernames are also disproportionately long compared to the vast majority of real player names

PR #10309, opened 2026-06-28
Made spammable xeno announcements use repeated chat system

MidZik:
## About the PR

Xeno announcements that can be spammed now make use of the repeated chat message system to reduce chat spam.

## Why / Balance

Reduce chat spam when many announcements of the same kind appear in succession.

## Technical details

The reason the spammed announcements didn't use the repeated chat system is because each announcement came from a different source entity. Since they came from different sources, they were not eligible to be "repeated". This PR changes the spammable announcements to come from the hive of the source entity instead, making them eligible since they all come from the same source.

I also increased the repeated message history to 5 to increase the reliability of the system working during the announcement spam that happens during an OB, because if too many announcements with different text appear between repeats the repeats will not be detected. It will not be 100% reliable but I hope it is Good Enough™️.

## Media

https://github.com/user-attachments/assets/3dfa8798-e8df-4730-8995-856a75813c60

## Requirements

**Changelog**

:cl:
- tweak: Spammable xeno announcements (such as traps being destroyed) will use the repeated message system when possible to reduce how much the chat window gets flooded with messages, such as during an orbital bombardment.

### DIFF
--- a/Content.Shared/_RMC14/CCVar/RMCCVars.cs
+++ b/Content.Shared/_RMC14/CCVar/RMCCVars.cs
@@ -515,7 +515,7 @@ public sealed partial class RMCCVars : CVars
         CVarDef.Create("rmc.lobby_start_paused", false, CVar.REPLICATED | CVar.SERVER);
 
     public static readonly CVarDef<int> RMCChatRepeatHistory =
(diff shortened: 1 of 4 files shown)


CatAndHats:
Requested by me as xeno QoL, ty for the pr

DrSmugleaf (approved):

PR #10317, opened 2026-06-29
Defender crest toggle popup text

SabreML:
## About the PR
Adds a little popup message when you toggle your crest defense as a defender, to let you know what it's changed to.
I PR'd this for CM13 a few years ago so I figured I'd port it.

## Why / Balance
It can be difficult to tell what crest state you're currently in without moving around to check your movement speed, especially when facing northwards. There is a sprite change and the ability button gets a highlight, but they're both pretty subtle.

An *ideal* solution to this (imo anyway) would be some sort of HUD alert if your crest is lowered, but the popup text at least helps at little.

## Media
fendy

## Requirements

**Changelog**

:cl:
- add: Added a text popup when toggling Crest Defense as a Defender.


SabreML:
Closing in favour of #10475.

PR #10331, opened 2026-06-30
Energy-based castes no longer gain acid from hitting infected hosts

### LINKED ISSUE #7385: Acid runners should not get acid from hitting Hugged marines
## Description
As per the title, It would be better for both xenos and marines if acid runner did not get acid off of hugged marines, forcing the acid runner to actually hit people on the front and not hang around the hive hitting caps until they either die or have been damaged enough for that runners standards. 
This kind of behavior should not be encouraged and should receive no benefit

carccborg:
## About the PR

Title
Closes #7385
Removes the `IgnoreInfectedComponent` member from `XenoEnergyComponent`.

## Why / Balance

Per #7385:
> As per the title, It would be better for both xenos and marines if acid runner did not get acid off of hugged marines, forcing the acid runner to actually hit people on the front and not hang around the hive hitting caps until they either die or have been damaged enough for that runners standards. This kind of behavior should not be encouraged and should receive no benefit

> Wanna note if someone is infected (past stage 4) they get no acid from it. It could, to be fair, probably should check earlier (around when they become dead draggable) instead. (Current value at 4 is due to parity atm).

This PR also prevents _all_ castes with a special resource that uses XenoEnergySystem from gaining their resource upon hitting an infected host.

This change is nonparity. It may not be desired or may need to wait until postparity.

## Technical details

Minor C#. Check if a hit entity has VictimInfectedComponent and if so, continue. Additionally removed `IgnoreInfectedComponent` from `XenoEnergyComponent.cs` since its only use was with acid runners.
## Media

Small fix.

## Requirements

**Changelog**

:cl:
- tweak: Acid runners and valkyrie praetorians no longer gain acid or fury when hitting an infected host

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
@@ -35,7 +35,7 @@ public sealed class XenoAciderGenerationSystem : EntitySystem
                 return;
 
             if (HasComp<VictimInfectedComponent>(hit))
-                return;
+                continue;
 
             startGenerating = true;
             break;
--- a/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/AciderGeneration/XenoAciderGenerationSystem.cs
@@ -1,6 +1,7 @@
 using Content.Shared._RMC14.Xenonids.Energy;
 using Content.Shared._RMC14.Xenonids.Rest;
 using Content.Shared._RMC14.TrainingDummy;
+using Content.Shared._RMC14.Xenonids.Parasite;
 using Content.Shared.Mobs;
 using Content.Shared.StatusEffect;
 using Content.Shared.Stunnable;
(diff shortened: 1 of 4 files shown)


Vermidia (changes requested):
I want to note this doesn't actually stop it fully, as in XenoEnergySystem:

```
            if (xeno.Comp.IgnoreLateInfected && TryComp<VictimInfectedComponent>(hit, out var infect) && infect.CurrentStage >= infect.FinalSymptomsStart)
                continue;
```

was the thing that was talked about in the issue. I will note acid generation system not checking the above or checking for nested IS a bug though.

YO0teacher:
Need to add the same component check to XenoEnergySystem.cs, otherwise acid is still gained on hit.
(like how I did with #8159)
edit: oh, Vermidia got to it first

carccborg:
Oh whoops. I misinterpreted the code lol. I'll fix the PR when I get home from work

carccborg:
Should I deprecate or remove `XenoEnergyComponent.IgnoreLateInfected`? Its only C# usage is in the check that I replaced, and the only strain that uses it in yml is Acider.

Vermidia:
You maybe should rename it? The change in the energy system if applied to all energy caste would also affect valk who gains fury on hit. I forget if hedge also gains shards on hits, but yeah.

carccborg:
Valkyries and Hedgehogs don't have IgnoreLateInfected in yml and it's set to false by default in C# (meaning they could previously get resource from infected hosts), so its removal would prevent them from gaining their respective resources from hitting infected hosts. It'd be an entirely unused member in the XenoEnergyComponent class and its only use in C# would be removed so I think cleaning it up by removing it entirely would be fine? I'm gonna update the PR as well to reflect all castes that this change affects.

Vermidia:
As long as you're aware, just makin sure.

carccborg:
Aye aye! I appreciate the second pair of eyes, tyvm

PR #10333, opened 2026-07-01
Fix flamer direct hits still hitting multiple times

Vermidia:
## About the PR

Title.

## Why / Balance

Bug. The DamageOnCollide comp got deleted early by `RunIgniteOnCollide`, which caused the chain to also, get deleted because its in there, which caused the entire chain system to just ignore it so you still got hit multiple times.

Instead of doing that theres now a disabled field on DamageOnCollide, that just skips the OnCollide part if true. `RunIgniteOnCollide` just disables it now.

## Requirements

**Changelog**

:cl:
- fix: Fixed direct hit of the incinerator's stream dealing damage multiple times, for real this time. This may have also fixed some instances of other fires dealing damage multiple times on spawn.

### DIFF
--- a/Content.Shared/_RMC14/OnCollide/DamageOnCollideComponent.cs
+++ b/Content.Shared/_RMC14/OnCollide/DamageOnCollideComponent.cs
@@ -76,6 +76,6 @@ public sealed partial class DamageOnCollideComponent : Component
     [DataField]
     public bool CanRehit;
 
-    [DataField]
+    [DataField, AutoNetworkedField]
     public bool Disabled = false;
 }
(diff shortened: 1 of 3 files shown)


Content.Shared/_RMC14/OnCollide/DamageOnCollideComponent.cs
```diff
@@ -75,4 +75,7 @@ public sealed partial class DamageOnCollideComponent : Component
 
     [DataField]
     public bool CanRehit;
+
+    [DataField]
```
MidZik: Not sure if it would actually affect any behavior, but since it's a changing field I think it should be networked.
```suggestion
    [DataField, AutoNetworkedField]
```

MidZik (changes requested):
Tested and confirmed it works now. One minor change.

MidZik (approved):

DrSmugleaf (approved):
Thanks for the PR!

PR #10334, opened 2026-07-01
[Tweak] Hijack Explosion Throttle

Thereisabear:
## About the PR
I think the explosions are soul, but as a lot of players have found; they can get quite taxing.
This just limits how many explosions can go off at once.

## Why / Balance
Spreads out the pipe explosions per tick and turns it into staggered rolling explosions.
Should smooth things out slightly more at the start of hijack.

## Technical details
Pain.
Throttles the amount of pipes that can explode per tick,  can be tweaked alongside the % of pipes queued for exploding.
_e.g. if maxpipesburst is 2 and the % of pipe it selects was 5.  Tick 1 would have 2 bursts, tick 2 have 2 ,and tick 3 have1._

Currently its set to 1 to smooth out the whole hijack instead of the current behaviour where the initial crash and subsequent few minutes are a mixture of simultaneous explosions that are unavoidable and teleporting into said explosions because of other explosions.

## Requirements

**Changelog**
:cl:
- tweak: Throttled the explosions on Hijack; behaviour should be a staggered but rolling explosion one at a time now.

### DIFF
--- a/Content.Server/_RMC14/Hijack/RMCHijackRandomDamageSystem.cs
+++ b/Content.Server/_RMC14/Hijack/RMCHijackRandomDamageSystem.cs
@@ -135,21 +135,14 @@ public sealed class RMCHijackRandomDamageSystem : EntitySystem
 
     private void OnPipeRemove<T>(Entity<RMCHijackActivePipeComponent> ent, ref T args)
     {
+        QueueDel(ent.Comp.Warning);
+        ent.Comp.Warning = null;
+
         if (TerminatingOrDeleted(ent.Comp.Map))
             return;
 
         if (TryComp(ent.Comp.Map, out RMCHijackActiveMapComponent? map))
-        {
             map.Pipes.Remove(ent);
-            map.Explode.RemoveAll(pending =>
-            {
-                if (pending.Pipe != ent.Owner)
-                    return false;
-
-                QueueDel(pending.Warning);
-                return true;
-            });
-        }
     }
 
     private void OnPipeAnchorChanged(Entity<RMCHijackActivePipeComponent> ent, ref AnchorStateChangedEvent args)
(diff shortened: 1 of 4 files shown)


Content.Server/_RMC14/Hijack/RMCHijackRandomDamageSystem.cs
```diff
@@ -137,7 +139,17 @@ private void OnPipeRemove<T>(Entity<RMCHijackActivePipeComponent> ent, ref T arg
             return;
 
         if (TryComp(ent.Comp.Map, out RMCHijackActiveMapComponent? map))
+        {
             map.Pipes.Remove(ent);
+            map.Explode.RemoveAll(pending =>
```
DrSmugleaf: This is going to iterate the whole list any time a pipe is removed
It would probably be best if it was a dictionary, though you lose ordering
That's why Pipes is a set, so the removal time is constant
You could also either not bother and let the warning stay behind until it expires, or store the warning  id on RMCHijackActivePipeComponent
The latter is probably best but then you would need to make sure that when the warning expires, the pipe component doesn't hold a deleted entity id

Content.Server/_RMC14/Hijack/RMCHijackRandomDamageSystem.cs
```diff
@@ -310,22 +326,8 @@ public override void Update(float frameTime)
...
-                    {
-                        BurstPipe(explode);
-                    }
-                }
-                finally
-                {
-                    active.Explode.Clear();
-                }
-            }
+            if (active.ExplodeAt is { } explodeAt && time >= explodeAt)
```
DrSmugleaf: The try finally here is so it doesn't get stuck exploding pipes, which with it involving explosions and a lot of other things is not unlikely
You can also try catch ignore every  call of BurstQueuedPipes instead

DrSmugleaf (dismissed):
I need to check if the initial percentage is even correct but this is good for now

PursuitInAshes (approved):
This seems good and will be crucial for larger shipmaps
`RMCHijackActiveMapComponent` is maybe poorly named for its current use but its not your fault, just something to note for later

PR #10352, opened 2026-07-03
Sorokyne Boiling Water

CatAndHats:
## About the PR
Adds the boiling water that deals minor damage to the water on the western half of the map, and to the CLF insert.

## Why / Balance
Parity. The CM13 version is completely indistinguishable from regular river water, however much like how the toxic water on Trijent was done i've added a different name, description, burning sound, and popup to let the player know they are in fact being cooked medium rare.

## Technical details
the cm13 version burns for 0.5/2.5, however our damage over time ticks twice as fast as theirs so it's halved.

There seems to be a lot of position switching happening after the save, I think this may be due to the engine update.

## Requirements

**Changelog**

:cl: CatAndHats
- add: Added the boiling water to Sorokyne Strata, all the water aside from Carp Lake and the water found in the eastern caves will burn to the touch!


kleinerstation13:
The water burns...

CatAndHats:
Converting to draft while I look at the damaging water in CM13 and how best to implement it in RMC. In CM13 they use effect blockers to do the damage rather than the entities themselves, it is likely that it what we'll want to do here

CatAndHats:
The above is fine, ready to review

PursuitInAshes (approved):
Yaml good and Diffs seem okay but I didn't open it ingame or do a parity check on locations

kleinerstation13 (approved):
Parity checked placements, lgtm

PR #10357, opened 2026-07-03
Made weeds look for a new parent node in range when theirs is destroyed

### LINKED ISSUE #3028: Nodes that get resin built on them make the surrounding weeds disappear
## Description

When a resin wall is built on top of a weed node, the weeds the node supports disappear even if there are more nodes nearby.

**Reproduction**

**Screenshots**

**Additional context**

SabreML:
## About the PR
Makes xeno weeds look for a new in-range parent node when their own source is destroyed. (Fixes #3028)

## Why / Balance
QOL improvement for xeno builders, so you don't need to worry about accidentally clearing out multiple rooms worth of weeds by placing a wall.

This does have the effect of making weeds a little bit harder to clear since they won't disappear on their own unless you destroy every node in range, though that is the intended behaviour, at least per CM13. (See [`avoid_orphanage()`](https://github.com/cmss13-devs/cmss13/blob/c46d6585fa520f2589a1fc174116cf0b61ef37b2/code/modules/cm_aliens/weeds.dm#L203-L214))
Weed tiles will also currently link to any source node in range even if it's behind a wall, mainly just because checking for LOS would complicate the code a lot. If that is a balance concern though then I can try to figure out a solution.

## Technical details
As a consequence of getting this to work with wall weeds (which can grow *outside* the range of the weed node if they're spawned by one of the outer weed tiles), wall weeds are now linked to the weed tile which spawned them. The only visible effect of this in practice is that destroying the weed tile will now also destroy the wall weeds adjacent to it.

Also worth noting that this applies to hive weeds from the core or clusters as well. They can't be kept alive by regular nodes, but a second hive cluster will preserve any hive weeds in its range if the first is destroyed.

### Before:

https://github.com/user-attachments/assets/2cef1679-2e44-4014-a280-7e3ca814bd85

### After:

https://github.com/user-attachments/assets/1f00c104-529d-456e-bdf0-800f55ce469c

## Requirements

**Changelog**

:cl:
- add: Xeno weeds won't disappear naturally until every source node in range is destroyed, rather than just the one which spawned them.


Content.Shared/_RMC14/Xenonids/Weeds/XenoWeedNodeComponent.cs
```diff
@@ -0,0 +1,5 @@
+namespace Content.Shared._RMC14.Xenonids.Weeds;
+
+[RegisterComponent]
+[Access(typeof(SharedXenoWeedsSystem))]
+public sealed partial class XenoWeedNodeComponent : Component { }
```
Arendian: Adding this without moving all source-related behavior away from `XenoWeedsComponent.IsSource` means there are two separate ways to determine whether an entity should be treated as a source.

The search should either just check the `IsSource` value on the `XenoWeedsComponent`, or `IsSource` should be removed and all usages should be replaced with checks for `XenoWeedNodeComponent`. The first option is probably the simplest solution.

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
```diff
@@ -198,6 +195,46 @@ private void OnWeedsTerminating(Entity<XenoWeedsComponent> ent, ref EntityTermin
...
+        if (!_weedsQuery.TryComp(weedTile, out var weeds))
+            return false;
+
+        weeds.Source = null;
+
+        _hiveMemberQuery.TryComp(weedTile, out var weedHive);
+
+        // If `weedTile` is within the range of another node, set that as its new source.
+        var weedCoords = _transform.GetMapCoordinates(weedTile);
+        var query = EntityQueryEnumerator<XenoWeedNodeComponent, XenoWeedsComponent>();
```
Arendian: This queries every single node on the map every time a weed tile gets terminated, it should instead query only nearby weeds in an area around it. I recommend looking at the following method ~~and creating a HasWeedNodeNearby method.~~ https://github.com/RMC-14/RMC-14/blob/e9da09f828845f2d7d6176924e4f00fd8957d09f/Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs#L389-L402

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
```diff
@@ -183,11 +183,8 @@ private void OnWeedsTerminating(Entity<XenoWeedsComponent> ent, ref EntityTermin
             if (TerminatingOrDeleted(spread))
                 continue;
 
-            if (_weedsQuery.TryComp(spread, out var weeds))
-            {
-                weeds.Source = null;
-                Dirty(spread, weeds);
-            }
+            if (TryAvoidOrphanage(spread))
```
Arendian: Any reparented weeds should instantly stop spreading the moment it's original source is deleted.

Arendian (changes requested):
In parity a node placed against a wall can sustain up to 2 tiles of weed behind the wall, 1 tile from the wall sustains 1 line on the other side of the wall, and 2 tiles away 0. So it will need some kind of LOS check.

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
Arendian: Oh wait, this method is already the method I had in mind. It only returns true if a weed node is in the specified range, so you can just directly use this one.

SabreML: Yeah I just noticed that as well lol. I'll maybe see if I can modify it to `GetNearbyWeedNode()` with an out parameter for the entity it finds.

SabreML:
Big rewrite which doubled the size of the PR, but it should have addressed everything from the review other than the range thing and fixed a bunch of other things that I hadn't initially considered like wall weeds (which can sometimes be outside the base range of the node due to being adjacent to an outer tile).
I'm looking into the range now, and then I'll go through and double check everything before I undraft.

SabreML:
> In parity a node placed against a wall can sustain up to 2 tiles of weed behind the wall, 1 tile from the wall sustains 1 line on the other side of the wall, and 2 tiles away 0. So it will need some kind of LOS check.

Okay having checked this now, it looks like the re-parenting size difference on CM is just due to the overall range being smaller. The `avoid_orphanage()` proc there doesn't seem to be shrinking it deliberately (it just uses `node_range`), so I think it might actually be unintentional. Either way though just changing the `XenoWeedsComponent` search range should work to match it.

Comparing the weed ranges for just placing a node versus placing a node on an already weeded area, deleting the others, and letting it take over:

<details><summary><b>CM-SS13</b></summary>
<h3>Natural spread range: (9x9 with no corners)</h3>

<h3>"Re-parenting" range: (7x7)</h3>

</details>

<details><summary><b>RMC-SS14</b></summary>
<h3>Natural spread range: (9x9 with corners)</h3>

<h3>"Re-parenting" range: (9x9 with corners)</h3>

</details>

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
```diff
@@ -386,19 +400,27 @@ private void WeedsOvipositorChanged(Entity<AffectableByWeedsComponent> ent, ref
...
+            }
         }
 
-        return false;
+        return nodeList;
+    }
+
+    public bool HasWeedNodeNearby(Entity<MapGridComponent> grid, EntityCoordinates coordinates, int range = 5)
+    {
+        return GetNearbyWeeds(grid, coordinates, range).Count > 1;
```
Thereisabear: rang).Count > 1;  
Kills lessies ability to weed if there is only 1 node.
Needs to either be .Any() or a short circut method like the above 405 onwards minus a few bits. Should be ok with .Any() up to you.

Content.Shared/_RMC14/Xenonids/Weeds/XenoWeedsDecayingComponent.cs
```diff
@@ -0,0 +1,13 @@
...
+/// <summary>
+/// This is just a copy of <see cref="Robust.Shared.Spawners.TimedDespawnComponent"/> made for xeno weeds,
+/// so that they have a chance to cancel the despawn if a new parent node is found.
+/// </summary>
+[RegisterComponent]
+[Access(typeof(SharedXenoWeedsSystem))]
+public sealed partial class XenoWeedsDecayingComponent : Component
+{
+    [DataField]
+    public float Lifetime;
```
Thereisabear: Use TimeSpan instead of a float for the Decay

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
```diff
@@ -173,25 +175,34 @@ private void OnWeedsTerminating(Entity<XenoWeedsComponent> ent, ref EntityTermin
...
             {
                 weeds.Source = null;
                 Dirty(spread, weeds);
             }
 
-            var timed = EnsureComp<TimedDespawnComponent>(spread);
+            var decaying = EnsureComp<XenoWeedsDecayingComponent>(spread);
             var offset = _random.Next(ent.Comp.MinRandomDelete, ent.Comp.MaxRandomDelete);
-            timed.Lifetime = (float)offset.TotalSeconds;
+            decaying.Lifetime = (float)offset.TotalSeconds;
```
Thereisabear: then change from floatoffset.Totalseconds: to _timing.CurTime+ offset.

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
```diff
@@ -208,21 +207,32 @@ public override void Update(float frameTime)
...
+        var decayingQuery = EntityQueryEnumerator<XenoWeedsDecayingComponent>();
+        while (decayingQuery.MoveNext(out var uid, out var comp))
+        {
+            comp.Lifetime -= frameTime;
+
+            // Before despawning the weed entity, check to see if there's a different node in range which could take over as its parent.
+            if (comp.Lifetime <= 0 && !TryAvoidOrphanage(uid))
+            {
+                QueueDel(uid);
+            }
```
Thereisabear, line 253: then also change if(comp.Lifetime<= 0 &&
to  If (time>=comp.DecayAt &&

Content.Shared/_RMC14/Xenonids/Weeds/XenoWallWeedsComponent.cs
```diff
@@ -1,11 +1,20 @@
...
+    /// </summary>
     [DataField, AutoNetworkedField]
-    public EntityUid? Weeds;
+    public EntityUid Weeds;
+
+    /// <summary>
+    /// The entity that the wall weeds are attached to.
+    /// </summary>
+    [DataField, AutoNetworkedField]
+    public EntityUid AttachedTo;
```
Thereisabear: probably best to make these nullable EntityUid?
Then update the blocs to suppor tthe nullage.

Thereisabear (changes requested):

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
SabreML: Honestly good job spotting that, I completely forgot lesser drones existed lol.

`.Any()` seems a lot better but it gives me a popup saying that apparently '[CA1860](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1860)' recommends avoiding `.Any()` when `.Count` is available for performance reasons (LINQ stuff). Performance isn't exactly a problem here but I'll use `.Count != 0` just to satisfy the popup.

SabreML:
Can't reproduce the test fail locally and nothing in the last commit should have caused it, so I'm just going to do a quick draft to re-run.

Content.Shared/_RMC14/Xenonids/Weeds/SharedXenoWeedsSystem.cs
Thereisabear: yeah those are the easy way outs, if we really wanted to ensure minimal footprint given how finicky weed performance can be, we'd short circuit it  while avoiding a list allocation.

```
public bool HasWeedNodeNearby(Entity<MapGridComponent> grid, EntityCoordinates coordinates, int range = 5)
    {
        var position = _mapSystem.LocalToTile(grid, grid, coordinates);
        var checkArea = new Box2(position.X - range + 1, position.Y - range + 1, position.X + range, position.Y + range);

        foreach (var anchored in _mapSystem.GetLocalAnchoredEntities(grid, grid, checkArea))
        {
            if (!TerminatingOrDeleted(anchored) && WeedsQuery.TryComp(anchored, out var weeds) && weeds.IsSource)
                return true;
        }

        return false;
    }
```

SabreML: Ohh right yeah I see what you mean now. I think in this case it should be okay as it is just because `HasWeedNodeNearby()` is only called when a player tries to place a weed (as opposed to in an update loop or something).

SabreML: I figured out a solution!
`GetNearbyWeeds()` now returns the `IEnumerable` from `GetLocalAnchoredEntities()` with some extra linq stuff added on top, rather than looping through the whole enumerable and putting everything into a list even if only one element was needed.

This should now mean that anything using `GetNearbyWeeds()` will only search as much as it needs to. (Deferred execution/lazy evaluation)

SabreML:
There we go it reran them this time.

PR #10359, opened 2026-07-03
Allows players to see the other faction's speech as garbled text

SabreML:
## About the PR
Added the ability for marines to hear xenos speaking and vice versa, with the speech of each getting uniquely garbled/obfuscated for players of the other faction. 
(All of the code making this work was added in the languages PR, not by me. I'm only implementing it here.)

The "translated" speech is shown above players' heads as normal but doesn't get added to the chat log in order to not clog the whole thing up with meaningless noises.
The exception to this is Synthetics (or anyone with `LanguageLearningComponent`), who *do* get the messages added to their chatlog since it's actually useful for them.
(Observers see everything as normal)

## Why / Balance
It's a really fun feature to have IMO. Especially for scenarios like a xeno "talking" with a nested host, since they can do more than make noises at each other (noises *and* text!).

## Media
English translated into random unga noises for xenonids: (Not shown in the chat log)

https://github.com/user-attachments/assets/9903ca01-cc2d-4f68-83bc-984a8cdbf8a4

More examples, including from the perspective of a Synth with `LanguageLearningComponent` (All speech shown in chat):

https://github.com/user-attachments/assets/6ff6f19c-63f0-4c68-b820-b000980b4ff6

## Requirements

**Changelog**

:cl:
- add: Marines and Xenonids can now ""understand"" each other's speech.


sporkyz:
This rules and I hope it gets added

Beansies:
How will this work with cloaked scout and sniper specialists? I feel like they should have an exception.

Vermidia:
I have checked real quick, in parity from up to 2 tiles you can see whisper popups from using the radio, aka whispering, no matter if you're invisible or not. And you can tell someone used the radio from much farther away, but not where.

The whispering range thing should probably be a seperate PR, as upstream's default range is 5 (2 is used for the distance where you can hear a whisper and fully get the words from it, parity would be making these the same I guess so once out of distance of a whisper you can't tell someone tried at all I guess)

So yes you can see a scout saying stuff if they're close enough to you and they aren't blocked by, say a wall, as you need to be in LOS to hear a whisper.

Beansies:
I was thinking more about just the pop up at all. If Xenia can see the popup the stealth is easily defeated if they speak at the wrong time. I know chat messages above your head linger and a Xeno coming from off screen could see a pop up even if they didn’t know what it said. Maybe make the scout and snipers chat box have the same transparency that they do to Xenos? But I agree it should be a separate PR.

PursuitInAshes (changes requested):
Thanks for the PR!

After talking with some others and thinking on it more, I think it would still be good to filter out text bubbles displaying if you aren't learning xeno (Or already know it) and vice versa, just to prevent a mess visually in large inter-faction blobs (Like queen scream) or outing things like lurker/paras/scouts since its just a "new player check" which isn't something we want if possible

So, if its a "Humanoid" language, we can still display a garbled message bubble and put it in the chat box if the reader also has a "Humanoid" language (IE: English speaker hearing German).
Then, if the reader understands/or is learning Xeno, we can display the bubbles for them just so they can still idly listlen, and vise versa (Its not like it will out synths, health UI already does that).

Feel free to ping me on discord if this wasn't super understandable

SabreML:
@PursuitInAshes Ok I've got a new version mostly working now, which I think I'll put up as a separate PR since the scope has changed a lot.

Just to double check that I've got the right idea from your review before I post it:
With the new version, rather than being based entirely on `MarineComponent` and `XenoComponent`, a chat message will instead only be visible to a player if one of these conditions is met:
<ol type="A">
<li>The player is an observer.</li>
<li>The player can understand the spoken language.</li>
<li>The player can understand a <i>different</i> language which is in the same "Category" as the spoken language.<br>(Currently, this is either 'Humanoid' or 'Xenonid')<br>(The message will be gibberish but they can see it)</li>
<li>The player has <code>LanguageLearningComponent</code>, and is able to learn the spoken language.<br>(Message will be gibberish here too)</li>
<li>The message has no language at all (e.g. LOOC), and the player is in the same faction as the speaker.<br>(standard <code>MarineComponent</code>/<code>XenoComponent</code> check)</li>
</ol>

So in practice:
- Marine and Xeno players won't see any difference at all.
- Synthetic players will be able to see and understand messages from 'Humanoid' players, as before, since they fully understand those languages, and can now also see garbled messages from Xenonids (although currently not Xenonid LOOC).

I haven't been able to figure out a good method of doing the "vise versa" part yet of xenos being able to see (gibberish) messages from synths, but I can keep working on that.

PursuitInAshes:
Responded on Discord because I only just now got to my computer, but yes thats correct (This is mostly for if anyone is looking at this later so it doesn't seem like you closed things randomly)

SabreML:
Closing in favour of #10764!

PR #10379, opened 2026-07-04
fix admin spawned synths 

Vixting:
## About the PR

fix admin spawned synths having ice skates & other missing comps
## Technical details

The cleargeneration was wiping all components, causing default values to be used because most things aren't defined on the generation but the actual base. There is probably a much simpler way to do this, but tldr is to just not remove everything if it shouldn't be removed
## Requirements

**Changelog**

:cl:
- fix: Fixed admin spawned synths not having the correct components assigned


Aurallianz:
#10381

PR #10380, opened 2026-07-05
[Fix] Queen Eye Fix 2.0

Thereisabear:
## About the PR
**Edit: The math wasn't mathing properly but post fix it now absolutely uses the cached anchor which means its practically free within weeds and when at the "edge" it only does a cheap vector calc instead of the expensive lookup. 
HALF THE COST DOUBLE THE VALUE.tm**

- Reverts the logic change to Queen Eye. 
- Fixes Queen Eye's leash breaking after the lag/crash which was caused by the below.
- Fixes the leak from stale entries not being cleared, it just grows exponentially while in eye as the round goes on.

Added an early out to the expensive look up so should equal out to the same as the perf fixes commit.

## Why / Balance
Fixes the new bugs and keeps the performance saving ideal.

## Technical details
**Edit: Fixed the eye trying to continuously use the GetEntitiesInRange at the edge of the +3 range off weeds it now only does a cheap vector calc from its cached anchorweeds which ends up being near or at 0 cost when on weeds and within anchor** 

_Due to prediction the leash calculations needing to live inline with the onQueenEyeMove or we get a jittery mess at the edges just due to timings._

_Adds a AnchorWeed cache on QueenEyeComp as an early out and cheap vector check instead of the expensive GetEntitiesInRange._

_Essentially throttles the expensive check to only run when its needed._ 

_Ive tried everything from tweaking the anchors, coords but nothing resolves the jitters and stutters.
Only way was to keep it inline so the timings line up._

## Requirements

**Changelog**

:cl:

- fix: Fixed Queen Eye's leash breaking.
- fix: Fixed various Queen Eye performance issues.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Eye/QueenEyeSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Eye/QueenEyeSystem.cs
@@ -319,16 +319,16 @@ public sealed class QueenEyeSystem : EntitySystem
 
     public override void Update(float frameTime)
     {
+        if (_timing.ApplyingState || _isRevertingMove)
+        {
+            _movedQueenEyes.Clear();
+            return;
+        }
+
         foreach (var (ent, oldCoords, newCoords) in _movedQueenEyes)
         {
-            if (_timing.ApplyingState)
-                return;
-
-            if (_isRevertingMove)
-                return;
-
             if (TerminatingOrDeleted(ent))
-                return;
+                continue;
 
             _nearbyWeeds.Clear();
             _entityLookup.GetEntitiesInRange(newCoords, ent.Comp.SoftWeedDistance, _nearbyWeeds);
@@ -382,6 +382,7 @@ public sealed class QueenEyeSystem : EntitySystem
                 _isRevertingMove = false;
             }
         }
+        _movedQueenEyes.Clear();
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


Arendian (dismissed):
Code LGTM

Content.Shared/_RMC14/Xenonids/Eye/QueenEyeSystem.cs
```diff
@@ -177,7 +175,95 @@ private void OnQueenEyeMove(Entity<QueenEyeComponent> ent, ref MoveEvent args)
...
+    {
+        EntityUid? closest = null;
+        closestWeedPos = worldPos;
+        var closestDistSq = float.MaxValue;
+        foreach (var weed in weeds)
+        {
+            var weedPos = _transform.GetWorldPosition(weed.Owner);
+            var distSq = Vector2.DistanceSquared(worldPos, weedPos);
+            if (distSq >= closestDistSq)
+                continue;
```
metalgearsloth, line 266: TryDistance is going to be faster on same-grid than continuously cycling through to map transform.

Thereisabear: 500 WS tokens upon you.
for legal purposes that is a joke but thanks a bunch!

DrSmugleaf (approved):

PR #10395, opened 2026-07-05
Fix access for fridges

BreakpointDebug1234:
## About the PR

Removed access requirements for fridges, this will allow more roles to use the fridges such as MPs, SOs, etc.
## Why / Balance

Parity, QoL
## Technical details

yaml

## Requirements

**Changelog**

:cl:
- tweak: Fridges are now all-access

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Storage/fridge.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Storage/fridge.yml
@@ -93,3 +93,12 @@
       amount: 14
     - id: RMCFoodMeatFish
       amount: 14
+
+- type: entity
+  parent: CMLockerFridge
+  id: CMLockerFridgeColony
+  name: fridge
+  suffix: colony
+  components:
+  - type: AccessReader
+    access: [["CMAccessColonyPublic"]]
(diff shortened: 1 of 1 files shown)


PursuitInAshes (dismissed):
Code good, needs a parity check but regardless I don't mind this

AverageNotDoingAnythingEnjoyer (commented):
Might be a diff fridge ofc, but seems CM doesn't have an access limit on fridges at all

https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/structures/crates_lockers/closets/secure/kitchen.dm#L3

AverageNotDoingAnythingEnjoyer:
Correction, there's also colony freezers which have an AccessPublic limit for some colony mapping and OT freezer

https://github.com/cmss13-devs/cmss13/blob/master/code/game/objects/structures/crates_lockers/closets/secure/freezer.dm#L1

BreakpointDebug1234:
It seems all shipside fridges are all access now. We do not have an OT fridge so it's not an issue.

I'll have to look in regards to which fridges are colony access on CM before editing maps but I have added an proto for it if anyone wants to cover them in a map passover.

Vixting (approved):

CatAndHats (approved):
Yep. Ta!

PR #10406, opened 2026-07-06
Wall weed visual improvements

SabreML:
## About the PR
This PR does a few things, all related to wall weeds:
1. Added wall weeds to windows and window frames.
2. Fixed wall weed smoothing so that they takes into account adjacent walls. (See the left walls in the before/after)
3. Changed the melee sound for wall weeds to be the same as regular weeds.
4. Added spawn menu suffixes for weed entities, since most of them were just "weeds".

## Why / Balance
It looks fancy. :)

## Technical details
I removed the 'weedframeX' sprites since they're just 'weedwindowX' sprites with added weeds on the upper half of the tile, "behind" the window. Weeds on RMC *can* actually can grow under window frames and fill that part of the tile, so they're unnecessary.

Also the nine new identical 'weedwindowX' I added are copies of 'weedwindow0.png' to make the `CardinalFlags` smoothing system work, since it needs exactly 16 icon states. From looking at existing window RSIs it doesn't seem to be possible to assign multiple directions to the same sprite, so duplicates it is.

### Before:
Before

### After:
After

## Requirements

**Changelog**

:cl:
- add: Added wall weeds for windows and window frames.
- fix: Fixed wall weed smoothing.


SabreML:
I can't reproduce the test fail by running it locally and it seems to be caused by something in Xenoarchaeology, so I think it's unrelated?

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
```diff
@@ -278,6 +278,7 @@
 - type: entity
   parent: XenoHiveWeeds
   id: XenoHiveWeedsCluster
+  suffix: Cluster, 6 Range
```
Thereisabear:
```suggestion
  categories: [ HideSpawnMenu ]
```

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
```diff
@@ -287,6 +288,7 @@
 - type: entity
   parent: XenoHiveWeeds
   id: XenoHiveWeedsCore
+  suffix: Core, 8 Range
```
Thereisabear:
```suggestion
  categories: [ HideSpawnMenu ]
```

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
```diff
@@ -407,10 +417,28 @@
...
+    additionalKeys:
+    - windows
+    base: weedwindow
+    mode: CardinalFlags
+
 # Gardener's weeds
 - type: entity # TODO RMC14 sound
   parent: XenoWeeds
   id: XenoHardyWeeds
+  suffix: Gardener
```
Thereisabear:
```suggestion
  categories: [ HideSpawnMenu ]
```

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
```diff
@@ -472,7 +500,7 @@
 - type: entity
   parent: XenoHardyWeeds
   id: XenoHardyWeedsSourceHidden
-  suffix: Hidden Node
+  suffix: Hidden Node, Gardener
```
Thereisabear:
```suggestion
  id: XenoHardyWeedsSourceHidden
  name: hardy weed node
  suffix: Hidden
```

Thereisabear (changes requested):

Thereisabear:
This introduces an issue where you can now nest someone on a window but breaking the window does not break them out.
If you wanted to clean up the spawn menu for the weeds specifically it should just be the different weed nodes and their hidden variants. All others can be safely hidden from the menu with    categories: [ HideSpawnMenu ]

SabreML:
> This introduces an issue where you can now nest someone on a window but breaking the window does not break them out. (...)

Oh shoot I hadn't considered that. Thank you.

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/xeno_weeds.yml
SabreML: It makes sense to keep the variant name in the suffix rather than changing the actual name, I think. Since otherwise you can tell which tile is the hidden node by right clicking and checking the name instead of it being completely "hidden".

Everything else sounds good though so I'll get on that. 👍

Edit: Spawn menu after changes
image

SabreML:
I changed the wall weed prototypes so that they're all children of a base abstract `XenoWeedCovering` prototype (which can in theory then be used to add new weedable things in the future).

`XenoWeedsWall` has `XenoNestSurfaceComponent` added to it and `XenoWeedsWindow` inherits that as before. `XenoWeedsWindowFrame` on the other hand is a direct child of `XenoWeedCovering` and therefore doesn't get the nest component.
Because of this, when a window gets changed into a window frame the `XenoNestSurfaceComponent`'s `Weedable` field won't be transferred over in `OnConstructionChangeEntity()`, and the nest gets deleted.

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
```diff
@@ -236,4 +227,55 @@ public override void Update(float frameTime)
...
+                nestSurface.Weedable = args.New;
+                Dirty(newWeedable.Entity.Value, nestSurface);
+            }
+        }
+        // else make a new one.
+        else
+        {
+            SpawnWallWeeds((args.New, newWeedable), oldWallWeeds.Weeds);
+            // Removal of the old weeds is handled separately when `args.Old` is swapped out and deleted.
+        }
```
Thereisabear:
```suggestion
        else
        {
            var spawned = SpawnWallWeeds((args.New, newWeedable), oldWallWeeds.Weeds);
            if (spawned is { } spawnedWeeds &&
                _xenoWeedsQuery.TryComp(oldWallWeeds.Weeds, out var srcWeeds))
            {
                srcWeeds.Spread.Add(spawnedWeeds.Owner);
                Dirty(oldWallWeeds.Weeds.Value, srcWeeds);
            }
        }
```
otherwise weeds on frames get orphaned and don't disappear when the source goes.

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
```diff
@@ -236,4 +227,55 @@ public override void Update(float frameTime)
...
+
+    // Transfer wall weeds over when one `XenoWeedable` entity gets turned into another. (E.g. Window into window frame)
+    private void OnConstructionChangeEntity(Entity<XenoWeedableComponent> ent, ref ConstructionChangeEntityEvent args)
+    {
+        if (!_xenoWeedableQuery.TryComp(args.Old, out var oldWeedable) ||
+            !TryComp<XenoWallWeedsComponent>(oldWeedable.Entity, out var oldWallWeeds))
+            return;
+
+        if (!_xenoWeedableQuery.TryComp(args.New, out var newWeedable) || newWeedable.Entity != null)
+            return;
```
Thereisabear, line 264:
```suggestion
        if (ent.Owner != args.New)
            return;

        if (!_xenoWeedableQuery.TryComp(args.Old, out var oldWeedable) ||
            !TryComp<XenoWallWeedsComponent>(oldWeedable.Entity, out var oldWallWeeds))
            return;

        var newWeedable = ent.Comp;
        if (newWeedable.Entity != null)
            return;
```

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
```diff
@@ -236,4 +227,55 @@ public override void Update(float frameTime)
...
+
+        // If the new and old entities both spawn the same thing when weeded, just transfer the old one over.
+        if (oldWeedable.Spawn == newWeedable.Spawn)
+        {
+            newWeedable.Entity = oldWeedable.Entity;
+            oldWeedable.Entity = null;
+
+            if (_xenoNestSurfaceQuery.TryComp(newWeedable.Entity, out var nestSurface))
```
Thereisabear, line 274:
```suggestion
        if (oldWeedable.Spawn == newWeedable.Spawn)
        {
            newWeedable.Entity = oldWeedable.Entity;
            oldWeedable.Entity = null;
            Dirty(args.New, newWeedable);
            Dirty(args.Old, oldWeedable);

            if (_xenoNestSurfaceQuery.TryComp(newWeedable.Entity, out var nestSurface))
```

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
Thereisabear: fixes the doubles handling

Thereisabear (changes requested):
Someone else will have to check the attribution if they still correct.

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
Thereisabear: dirtied, pretty sure this was a bug in the previous original underlying code or at least something very similar.

Content.Server/_RMC14/Xenonids/Weeds/XenoWeedsSystem.cs
SabreML: Oh yep very true, thank you

PR #10409, opened 2026-07-06
Add xeno last stand system

Vermidia:
## About the PR

Title.

## Why / Balance

parity.

## Technical details

If round doesn't end distress signal will check for min alive xenos (currently, 1) that count for slots, and if they're under that all get a new comp (if they have laststandable).

This new comp shows a popup, rejuvenates, and prevents ANY further heals from the rest of the xeno systems, for the most part.
I tried to catch everything that still didn't use CanHeal, and cleaned up some other code that was using it's own healing stuff. Some of these still show particles even without healing taking place - this can be cleaned up later.

This comp persists through evos. I haven't tested if it actually works in the game rule, but it *should*.

## Media

https://github.com/user-attachments/assets/18d3ae97-76b6-4b6e-9a4a-f3e35993dabf

## Requirements

**Changelog**

:cl:
- add: Added Xeno Last Stand. When there is one xeno left they get told this, fully healed, and can no longer heal from any sources, unless they're a queen.

### DIFF
--- a/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.RoundEnd.cs
+++ b/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.RoundEnd.cs
@@ -1,4 +1,4 @@
-﻿using System.Linq;
+using System.Linq;
 using Content.Server.GameTicking;
 using Content.Server.Ghost.Roles.Components;
 using Content.Shared._RMC14.Armor.Ghillie;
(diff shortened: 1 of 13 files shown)


OnsenCapy:
So all this does is not let them heal? Why not buff their combat capabilities like increasing armor and attack damage as well?

Vermidia:
This is just the stuff from parity. Later on it could be changed to have more significant effects, if wanted.

Beansies:
So what happens when it gets down to one xeno, then a capture bursts or someone joins as buried larva? Can they then heal? Will the event trigger again when it drops to one xeno left alive a second time?

Vermidia:
Once applied, last stand is permanent. If other xenos are born then don't get said effect, and then it goes back down to one xeno, that new xeno would get last stand. As far as I'm aware thats how it is in parity too if that somehow happens.

QuietlyWhisper (approved):
Concept is fine, we can talk about adding more buffs to the last stand xenonid later on

PR #6777, opened 2025-06-17
Adds specialization kit to CO's equipment vendor (NOT PARITY) 

Arcticular1:
## About the PR
Added specialization kits to CO's equipment vendor
## Why / Balance
Executive officers and staff officers get such kits, however the CO does not.

Not having the option to get a medical kit or a engineering kit as CO makes getting equipped a very slow process, usually in hijack and other combat situations you need to visit medical just to get *basic* healing supplies, which is very annoying and something that you cant really do as CO on hijack due to needing to get orders out, activating evacuation, and getting to a secure area by the time the ship crashes.

(TLDR very inconvenient) 
## Technical details
YML
## Media
image

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [x] I have added media to this PR or it does not require an ingame showcase.

- [x] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Added medical essential kit and engineering essential kit to the CO vendor.

PR #10412, opened 2026-07-07
Add the same specialization kits that SOs and the XO get to the CO's equipment vendor.

BramvanZijp:
## About the PR
Title.

## Why / Balance
The XO, and even SOs, get this option. And it is great to have.
Wether it be hijack or deployment, being able to quickly get either kit can be important, but even if not, it is just a general quality of life which, for some reason, is given to the XO and SOs, but not the CO.

Furthermore, I would argue that a CO is MORE likely to need this then an XO, since CO can often be late joined, including in situations where they then need to act very quickly, such as with 30 seconds remaining before a dropship hits the almayer.

I know this is not parity, but I see no reason for a CO to not get this if the XO/SO's do.

## Technical details
I shamelessly copied #6777 (Which got closed as the person who opened that left RMC.), and yes I tested that it still works.

## Media
image

## Requirements

**Changelog**
:cl: Arcticular1, BramvanZijp
- add: Added the medical & engineering specialization kits that XOs & SOs get to the CO's equipment vendor.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Machines/Vending/Crew/command.yml
@@ -468,6 +468,11 @@
       - id: CMArmorHelmetM11CCO
       - id: RMCHandsDeluxeMarine
       - id: CMBootsBlackFilled
+    - name: Specialisation Kit
+      takeOne: CMBundle
+      entries:
+      - id: CMVendorBundleCombatTechnicianEssentials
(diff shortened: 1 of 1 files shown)


PursuitInAshes (approved):
Code wise this is fine, conceptually it will have to go through Whisper in time
I only see the engineering kit taken even, but thats moreso a greater design thing than an issue with this

BramvanZijp:
> Code wise this is fine, conceptually it will have to go through Whisper in time I only see the engineering kit taken even, but thats moreso a greater design thing than an issue with this

I will note, I only ever see the medical kit taken / medical supplies asked for by COs, but who knows.

PR #10426, opened 2026-07-09
Fixed the muzzle flash effect sometimes looping

SabreML:
## About the PR
Shortens the lifetime of the muzzle flash effect in order to prevent it from sometimes looping. (See media for details)

The new lifetime value *is* shorter than the theoretical length of the animation ([0.06 * 6](https://github.com/RMC-14/RMC-14/blob/72742ed9b18def419d04e8aec47894da64633994/Resources/Textures/_RMC14/Objects/Weapons/Guns/Projectiles/projectiles.rsi/meta.json#L25-L34) = 0.36), but setting it to that value still causes the animation to loop for whatever reason. Maybe a floating point thing since they're such small numbers?

## Why / Balance
This isn't a particularly massive issue of course but I do notice it occasionally, so worth fixing. :)

### Before:
(First two shots are fine, next three show the bug)
Before

<hr>

#### Frame-by-frame:
Frames

### After:
After

## Requirements

**Changelog**

:cl:
- fix: Fixed the muzzle flash effect sometimes looping back to the start on the last frame.

### DIFF
--- a/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
+++ b/Resources/Prototypes/Entities/Objects/Weapons/Guns/Projectiles/projectiles.yml
@@ -3,14 +3,14 @@
   categories: [ HideSpawnMenu ]
   components:
   - type: TimedDespawn
-    lifetime: 0.3
+    lifetime: 0.3 # RMC
   - type: Sprite
     drawdepth: BelowMobs
-    offset: 0.33, 0
+    offset: 0.33, 0 # RMC
     layers:
     - shader: unshaded
       map: ["enum.EffectLayers.Unshaded"]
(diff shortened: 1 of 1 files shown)


kleinerstation13 (dismissed):
It worked on my machine

kleinerstation13:
Your other PR deals with that change touching an upstream line and needing a comment but I'm assuming you'll solve the conflicts in that PR once they appear when this is merged so I won't touch on it here

SabreML:
That's what I was planning to do yeah. Although now that I think about it if I just make the same change here then that would resolve any conflicts pre-emptively.

SabreML:
Oh whoops that dismissed it, sorry.

CatAndHats (approved):
LGTM. TY for the PR

kleinerstation13 (approved):
Looks ok

PR #10428, opened 2026-07-09
[Fix] A rare bug in PlayTimeTrackingSystem.

Thereisabear:
## About the PR
Fixes a rare bug in the Playtime tracking system using the existing methods that should hopefully close this crash loop

## Why / Balance
Should prevent the same rare start crash from happening due to invalid things.

## Technical details
Changed _GetPreferences_ to _GetPreferencesOrNull_ from IServerPreferencesManager.  
Swapped it out on the IsAllowed GetDisallowed RemovedDisallowed  blocks. 
_Stack trace was only IsAllowed but each one of the above calls GetPreferences and has the same potential._
Method was implemented 2 years ago but doesn't seem to be called anywhere.

## Requirements

**Changelog**
No changelog as with this fix hopefully there's no player facing crash again. (or at least even rarer)

### DIFF
--- a/Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
+++ b/Content.Server/Players/PlayTimeTracking/PlayTimeTrackingSystem.cs
@@ -200,7 +200,9 @@ public sealed class PlayTimeTrackingSystem : EntitySystem
             playTimes = new Dictionary<string, TimeSpan>();
         }
 
+        // RMC14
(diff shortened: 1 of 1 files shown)


Arendian (approved):
LGTM

PursuitInAshes:
Is there a reason we can't just put the code comments in line?

Thereisabear:
> Is there a reason we can't just put the code comments in line?

more so for ease of visual clarity during merging and potential conflicts and is per the coding standard documentation.
And sometimes comments at the end doesnt show very well and you might now realise the merge deleted something important. 
Saves guess work

PursuitInAshes (approved):
Alright
Its kinda hard to reproduce the bug so we'll have to just wait and see, its a more precise check anyways so its for the best

PR #10436, opened 2026-07-10
King music

Vermidia:
## About the PR

retribution now plays when King spawns, and no longer during hijack.

## Why / Balance

It's cool.

## Technical details

Removed retribution from the hijack sound collection, added a path in the cocoon comp, make cocoon play the audio before it gets terminated.

## Media

https://github.com/user-attachments/assets/bc1727e0-4601-4a10-b87f-1b4d013bdb0e

## Requirements

**Changelog**

:cl:
- add: Added music when King spawns from a cocoon. Said track that plays no longer plays during hijack.


kleinerstation13:
Won't the music be a bit odd when the king spawns as the marines leave/have already left the planet? I have seen the king spawn in when evacuation already completed a few times.

Vermidia:
It may, but it wouldn't overlap with hijack music at all due to dropship travelling. There is a function for checking if there's enough alive marines, used to quicken the timer so if marines hijack xenos auto get queen. So I could make it not play if marines already left.

rafapecogames:
*feels the aura* bro got a whole terror radius...

honestly, i think hijack should have a different music if there is a King onboard, and it might be harder, maybe undoable, that they actually get more agitated when king is close (like DBD/among us hide in seek yes im giving this example terror radius)

but just a more metal theme when there is a King would be cool. a music that plays when king spawns but it doesnt when its on a hijack wouldnt make sense. in all of the rounds since nuke and king got in, there was NEVER an occasion that King was on ground with marines (unless admemes)

PursuitInAshes (changes requested):
File path should move to something that isn't in the hijack folder
The ogg itself is 3:15 so we maybe want a shorter version and keep the full length in hijack's list? Unsure.
I'm trying to figure out a way to make this fit without being abnormal

PR #10448, opened 2026-07-11
Add filled head of department lockers

BreakpointDebug1234:
## About the PR

Adds filled versions of the CO, CMP, CMO and LO lockers. (only ones present on the Almayer).

## Why / Balance

Reduce map maintaining costs as these lockers are used on all UNMC ship maps.

## Technical details

yaml, mapping on Almayer would be ideal but not critical

## Media

image

## Requirements

**Changelog**

:cl:
- code: Added filled versions of department head lockers for mapping

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Structures/Storage/lockers.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Structures/Storage/lockers.yml
@@ -48,6 +48,22 @@
         - BulletImpassable
 
 - type: entity
+  parent: CMLockerCommander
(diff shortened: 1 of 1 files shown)


PursuitInAshes (commented):
With Department Heads I'd rather try to put things in their vendors if we can if they are missing something, the lockers are there specifically so in round someone can store things away without having to worry about it being taken, so I'm not entirely sure if these are needed

PursuitInAshes:
Thanks for the PR!
Closing per above; you can open a new PR to adjust vendors if anything is missing from them

PR #10472, opened 2026-07-13
[Change] B8 Smart Scope zoom removal.

Thereisabear:
## About the PR
Changes the B8 smart scope itself to parity.
-  **Removed the ability to zoom with the scope.** 
_This always felt clunky due to the DoAfter for Zoom, locking your view in one direction and move penalties whilst zoomed. 
As an attachments aimed towards newer players it's alot to take in and  not abundantly clear what its supposed to do._

When attached to an appropriate gun:
- **20% reduced damage per bullet -> 10% reduced damage per bullet.**
- **+0.2 fall off modifier**  
(_causes more fall off when attached, only really feel this if shooting targets screen edge distance 10+ tiles_)

_TLDR It now just functions like an attachable sight that provides IFF at the cost of 10% reduced damage and more damage fall off outside of optimal range._

### Damage numbers - _Tested on unarmoured shooting target._

**M4SPR with no attachments.**
7,10, 14, 16 tiles(no attachments) = 56 damage

_**Current Smartscope - Has Zoom, No damagefalloff, 20% damage penalty**_
7 tiles = 52 damage
10 tiles = 48 damage
14 tiles (Zoom) = 48 damage
16 tiles (Zoom) = 48 damage

_**Proposed Smartscope - Zoom Removed, +0.2 damage fall off, 10% damage penalty**_ 
7 tiles = 52 damage
10 tiles = 49.4 damage
14 tiles (can't see target) = 46.28 damage
16 tiles (can't see target) = 44.72 damage

## Why / Balance
Parity
Alt IFF is another topic.  Even with our IFF currently, I see little reasons holding us back from changing it to match.

## Technical details
Yaml.

## Media
image

image

## Requirements

**Changelog**

:cl:
- tweak: B8 SmartScopes can no longer zoom; changed modifier values: 20% -> 10% reduced damage and +0.2 range falloff.

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/rail_attachments.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Weapons/Guns/Attachments/rail_attachments.yml
@@ -441,7 +441,7 @@
     actionDesc: This civilian-grade scope is a common sight on hunting rifles due to its cheap price and great optics. Fixed at a modest 2x zoom.
 
 - type: entity
-  parent: RMCAttachmentS42xTelescopicMiniscope
+  parent: RMCRailAttachmentBase
   id: RMCAttachmentB8SmartScope
   name: B8 smart-scope
   description: An experimental B8 Smart-Scope. Based on the technologies used in the Smart Gun by AEGIS, this sight has integrated IFF systems. It can only attach to the M4SPR Battle Rifle and the M44 Combat Revolver.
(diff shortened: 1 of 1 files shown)


BarteG44:
this change, among other things, happened on cm because they completely reworked how the b8 works. There it just prevents you from shooting if a marine is in your line of fire, so labeling this parity is a bit disingenuous. 

Also the B8 is not *only* a noob attachment. It has it's uses in the hands of more experienced players, that it would loose with this PR

PR #10475, opened 2026-07-13
'Crest Lowered' HUD alert for Defenders

SabreML:
## About the PR
Adds a HUD alert for when your crest is lowered as a Defender.
(Better alternative version of #10317)

The alert sprite is, obviously, based on the existing Defender's sprite, but in order to fit it into a 32x32 alert I had to draw it again from scratch using that as a reference, so it might not be quite up to quality standards. (I am very much not an artist)
If not then I'm happy to make some edits or just replace it entirely!

## Why / Balance
It can be difficult to tell what crest state you're currently in without moving around to check your movement speed, especially when facing northwards. There *is* a sprite change and the ability button gets a highlight, but they're both pretty subtle.

## Media
crest 128

https://github.com/user-attachments/assets/c5b779a1-f98a-4764-b353-0b7482b643a4

## Requirements

**Changelog**

:cl:
- add: Added a HUD alert when your crest is lowered as a Defender.


Resources/Textures/_RMC14/Interface/Alerts/xeno_abilities.rsi/meta.json
```diff
@@ -0,0 +1,14 @@
+{
+  "version": 1,
+  "license": "CC-BY-SA-4.0",
+  "copyright": "Sprite by SabreML",
```
PursuitInAshes:
```suggestion
  "copyright": "Sprite by SabreML (Github)",
```
Should also include what its based on

PursuitInAshes (changes requested):
Mild copyright change, but the rest seems good, TY

Resources/Textures/_RMC14/Interface/Alerts/xeno_abilities.rsi/meta.json
SabreML: Oh yeah that would be good to do. 👍

Added the names of the original Defender sprite artists from [defender.rsi](https://github.com/RMC-14/RMC-14/blob/5645d5778b606ac68624b826223e2cbdc3eba5ce/Resources/Textures/_RMC14/Mobs/Xenonids/Defender/defender.rsi/meta.json#L4).

PursuitInAshes (approved):
Thanks for the PR!

PR #10495, opened 2026-07-15
[Fix] Fixes tools placing when trying to deconstruct.

Thereisabear:
## About the PR
Fix strange interaction when trying to deconstruct something that is also a placeable surface e.g Table, Racks.
_Game fights itself trying to decide if you are placing or deconstruct/repairing._

## Why / Balance
Fix.   

## Technical details
adds a guard in the upstream file to check if tool.

## Requirements

**Changelog**

:cl:

- fix: Fixed tools being placed when trying to deconstruct/repair placeable surfaces.


PursuitInAshes (approved):

CatAndHats (changes requested):
While this does fix the bug, it also prevents any tools from being placed on surfaces which is grim, is there any way of making this more specific to only ones needed in the deconstruction steps?

For example, this makes it so you cannot place crowbars/wirecutters/screwdrivers on tables despite those not having any table interactions.

Thereisabear:
> While this does fix the bug, it also prevents any tools from being placed on surfaces which is grim, is there any way of making this more specific to only ones needed in the deconstruction steps?
> 
> For example, this makes it so you cannot place crowbars/wirecutters/screwdrivers on tables despite those not having any table interactions.

This is intentional by me.

CatAndHats (approved):
Discussed in Maint chat. LGTM.

PR #10496, opened 2026-07-15
Collision fixes

Drazamuffin:
## About the PR

Fixed collisions for Machine Frames, Hybrisa Street Lights, Platform corners and Janitor Carts all not being parity

## Why / Balance

Parity.

## Technical details

Hybrisa streetlamps technically do have density in parity however due to the following line of code;
AddComponent(/datum/component/shimmy_around, east_offset = -15, west_offset = -15)
Do not effect mob movement, only object's movement, removed collisions in lue of the actual system as it effects player map traversal too heavily compared to parity in its current state.

## Media

Bounding box tweaks only

## Requirements

**Changelog**

:cl:
- fix: Fixed non-parity collisions for Machine Frames, Hybrisa Street Lights, Platform corners and Janitor Carts

### DIFF
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Misc/floodlights.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Misc/floodlights.yml
@@ -376,7 +376,7 @@
   name: broken colony streetlight
   description: A pasted on note says 'To fix, unscrew the panel, crowbar out the damaged assembly, weld it fixed, add 2 cable, then close the panel.'
   components:
-  - type: Sprit
+  - type: Sprite
     noRot: true
     sprite: _RMC14/Objects/Misc/Lights/streetlight.rsi
     offset: 0.0,0.8
--- a/Resources/Prototypes/_RMC14/Entities/Objects/Misc/floodlights.yml
+++ b/Resources/Prototypes/_RMC14/Entities/Objects/Misc/floodlights.yml
@@ -342,6 +342,9 @@
   id: RMCStreetlight
   name: colony streetlight
   components:
+  - type: Physics
+    bodyType: Static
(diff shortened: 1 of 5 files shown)


MidZik:
I looked into the street lights a bit, and they're annoying to port correctly. Firstly, although you can move through them, they slow you down a bit. Secondly, you cannot use movement abilities through them. Leaping for example gets blocked by street lights. I've been looking into adding mob collision to them to try and simulate the behavior but it's not simple to do.

I removed small platform corner collisions in #10780. It seems you removed the collisions from the big corners as well here?

Drazamuffin:
> I looked into the street lights a bit, and they're annoying to port correctly. Firstly, although you can move through them, they slow you down a bit. Secondly, you cannot use movement abilities through them. Leaping for example gets blocked by street lights. I've been looking into adding mob collision to them to try and simulate the behavior but it's not simple to do.
> 
> I removed small platform corner collisions in #10780. It seems you removed the collisions from the big corners as well here?

I'm aware with street lights as the sidestep mechanic is annoying given engine differences, that portion was more intended as a stop gap for the time being rather then a permanent solution.

As for the corner pieces; CM13 doesn't have diagonal movement, corner pieces of platforms are all flagged as 'platform_decoration' on the mapping end in parity and contain 0 platform flags as by all means they are not platforms. 
All having them block movement does is make it semi-required to destroy an object never mapped with the intent of obstructing player movement.
image

MidZik:
I am referring to these larger corners. You removed the collision from these as well.
image

I agree with the small corner pieces not having collision, although I don't agree with removing their destructibility, even though it might be parity. To me it makes sense to only remove their collisions but still allow them to be destroyed, just to keep them in line with the other platforms while removing their unintentionally large impact on movement.

Drazamuffin:
Ahh that, uh that is actually just unintentional lol, I forget to test everything I change sometimes.

Drazamuffin:
I'm just going to close this for now as i'm busy with other things

PR #10511, opened 2026-07-17
(very) Roughly halved integration test run-time

SabreML:
## About the PR
Reduces how long the 'Build & Test Debug / build (ubuntu-latest) (pull_request)' CI test takes to run by speeding up `Content.IntegrationTests.Tests.EntityTest` by roughly 60%.

I noticed that upstream's CI tests are significantly faster than here so I thought I'd try to mess around investigating a bit. I saw while profiling RMC's `SpawnAndDirtyAllEntities` test that ~1/3rd of its CPU usage was being taken up by `System.GC.Collect()`, and that Wizden's version by comparison doesn't call it at all.
GC Collect 1
GC Collect 2

I'm not going to pretend that I fully understand how the integration test system works, but removing the garbage collection from `SpawnAndDeleteAllEntitiesOnDifferentMaps` & `SpawnAndDirtyAllEntities` does evidently speed things up significantly, still runs the tests as normal, and doesn't seem to cause any increase in memory usage either. At least from my own local testing.

### Before:
**Run 1**
Before 1
**Run 2**
Before 2
**Run 3**
Before 3
**Avg. 14.4 minutes.**
<hr>

### After:
**Run 1**
After 1
**Run 2**
After 2
**Run 3**
After 3
**Avg. 5 minutes.**

## Requirements

## Changelog
N/A

### DIFF
--- a/Content.IntegrationTests/Tests/EntityTest.cs
+++ b/Content.IntegrationTests/Tests/EntityTest.cs
@@ -100,7 +100,6 @@ namespace Content.IntegrationTests.Tests
                     });
                 });
 
-                GC.Collect();
             }
 
             await pair.CleanReturnAsync();
@@ -254,7 +253,6 @@ namespace Content.IntegrationTests.Tests
                     });
                 });
 
-                GC.Collect();
             }
 
             await pair.CleanReturnAsync();


SabreML:
#### Before:
https://github.com/RMC-14/RMC-14/actions/runs/29585702118/job/87902033582
image
https://github.com/RMC-14/RMC-14/actions/runs/29553932258/job/87802105210
image

#### After:
https://github.com/RMC-14/RMC-14/actions/runs/29596463865/job/87937951021
image
holy hell it work

PursuitInAshes (approved):
LGTM
Smug says the entity count in the loop was originally higher, but when she lowered it the collect wasn't removed, so this should be fine I think

CatAndHats (approved):
Looks correct. Smug said it should be good to merge.

PR #10530, opened 2026-07-21
Xeno burrowed behavior adjustments, small surv count tweak

MidZik:
## About the PR

Burrowed larva now fully count for the purposes of determining T2 and T3 slots. (i.e. 30 burrowed larva will now give 6 T3 slots, instead of ~2.2 T3 slots).

Late joining marines before the 15 minute mark now give the same amount of burrowed larva as if they readied up. (From 7.5 late joins per burrowed larva to [3-5.5] late joins per burrowed, based on autobalance value.)

Late joining marines after the 15 minute mark give significantly more burrowed larva. (From 7 late joins per burrowed larva to [2.5-4] late joins per burrowed, based on autobalance value.)

The best-case-scenario autobalance value for xenos has been reduced significantly, from 1/3 of readied up players to 1/4 of readied up players. As an example, instead of 300 readied up players potentially giving 100 xenos, 300 readied up players can only ever produce 75 xenos at most.

Slightly reduced survivor counts at some player counts. Mostly as a consequence of me changing `round` to `floor`. I adjusted the surv ratio to reduce the impact of this change as much as possible.

Added a feature to only adjust the autobalance when there is a streak, but I have it disabled by default in this PR.

## Why / Balance

- **Burrowed larva now fully count for the purposes of determining T2 and T3 slots.**
	This alleviates the issue where the hive is _much_ weaker when there are simply not enough xeno players to take every burrowed larva, which was particularly bad at very high burrowed counts.

- **Late joining marines before the 15 minute mark now give the same amount of burrowed larva as if they readied up.**
	Removes meta-gaming the hive size by purposefully joining late as a marine, which would give significantly less xenos to the hive compared to readying up. Now there is no benefit or downside to late joining as a marine in the first 15 minutes. This is in the same spirit as changing the tank spawning requirements, removing late-joining as a meta-gaming tool.

- **Late joining marines after the 15 minute mark give significantly more burrowed larva.**
	This seems to be the parity value. In CM13, pre-15 minutes every 4 latejoins gives a burrowed larva, while after the 15 minute mark every 2.5 latejoins gives a burrowed larva. Rather than make it a constant 2.5 as in CM13, I made the minimum 2.5. Otherwise it is 1.5 less than the autobalance value. That gives a range of [2.5-4]. Besides "it's parity" as a reason, I do think reducing how much late joining favors marines is good for the game, and dramatically weakens the strategy of having the marines stall in safe, fortified locations picking off xenos while late joins refill their numbers.

- **The best-case-scenario autobalance value for xenos has been reduced significantly, from 1/3 of readied up players to 1/4 of readied up players.**
	The autobalance at its most xeno-favored value was kind of pathological. With 200 readied up players, you could get 67 xenos, 7 survs, and 126 marines. That is 1 xeno for every ~1.9 marines, _including shipside marines_. It's kind of absurd. Now, the best case will be 1 xeno for every ~2.86 marines. This was alleviated by the fact that xenos got way less mid-round larva from late joins, but since I'm changing that in this PR, keeping this case around is probably a bad idea.

- **Slightly reduced survivor counts at some player counts.**
	There was an error in calculating xeno and surv counts. Xenos and survs have a CVar that say how many marines they need per spawn. But the calculation was using `round` instead of `floor`. So even though the config was saying 17 marines per surv, in reality you could get more survs than that because "2.5" survs becomes "3" survs. Now "2.9" survs becomes "2" survs. The xeno calculation was also fixed in this manner but because the xeno count is much higher the effect is much less noticable.

    I have generally heard complaints that there are too many survs for the amount of xenos that there are at low populations, so I don't mind keeping this adjustment around both as a bugfix and a small balance change, but I am also not very interested in balancing surv counts in this PR. I have adjusted the surv ratio to offset the change as much as I could without increasing survivor counts. About 25% of low-population rounds are affected. For example, between 81-84 players, the surv count is one less, while 85-98 players has the same surv count.
	

## Technical details

Refactored and fixed up calculations that determine roundstart xeno counts, surv counts, and marine counts (added +1 in the denominator, changed `round` to `floor`).
Added new CVars and removed obsolete ones.
Added calculation that adds more roundstart burrowed larva when the marine count is higher than expected.
Replaced the static late-join-per-burrowed logic with new logic based on autobalance.
Changed surv ratio from 18 marines per surv to 16 marines per surv to account for the other changes/bugfixes in order to minimize how much surv counts are affected.
Autobalance max value of 6.5 changed to 5.5 to keep the max the same. (5.5 with this PR works the same as 6.5 without this PR.)
I am keeping the default value for CMMarinesPerXeno at 4f. This is, technically, another adjustment to it, because changing it to 3f would keep the behavior the same. But it's autobalanced anyway so it should be fine.
The rest is pretty straightforward.

## Requirements

**Changelog**

:cl:
- tweak: Burrowed larva now fully count for the purposes of determining T2 and T3 slots. For example, 30 burrowed larva now give 6 T3 slots, instead of ~2.2 T3 slots.
- tweak: Late joining marines in the first 15 minutes now give the same amount of burrowed larva as if they had readied up. After the first 15 minutes, they give significantly more burrowed larva, based on the autobalance value.
- tweak: Slightly reduced survivor counts at some player counts. In effect, about 25% of rounds with player counts between 45-120 will have one less survivor.


MidZik:
Also I apologize for putting so many things in one PR. I figured since they all touch on the same area it will be easier to discuss their cumulative effect if they're all in one place.

Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Helpers.cs
```diff
@@ -378,6 +378,41 @@ private Spawners GetSpawners()
...
+    /// Returns the maximum amount of roundstart xenos possible given the number of readied players.
+    /// </summary>
+    /// <param name="readyupCount"></param>
+    /// <returns></returns>
+    /// <remarks>This limit assumes the marine count is at the minimum, which will happen if every xeno slot is taken. If not every xeno slot is taken,
+    /// the xeno + burrowed count can exceed this limit to account for the extra marines.</remarks>
+    private int GetRoundstartXenoLimit(int readyupCount)
+    {
+        // TODO RMC14 xeno limit should be changed to be based on number of marines rather than number of readied players, but this is a decent estimate.
+        return (int)Math.Floor(Math.Max(1, readyupCount / (_marinesPerXeno + 1)));
```
yagwog, line 394: How does this calculation differ from CM currently?

Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Helpers.cs
```diff
@@ -378,6 +378,41 @@ private Spawners GetSpawners()
...
+
+    /// <summary>
+    /// Returns the maximum amount of roundstart survs possible given the number of readied players.
+    /// </summary>
+    /// <param name="readyupCount"></param>
+    /// <returns></returns>
+    private int GetRoundstartSurvLimit(int readyupCount)
+    {
+        // TODO RMC14 surv limit should be changed to be based on number of marines rather than number of readied players. This isn't a good estimate since it includes xenos.
+        return (int)Math.Clamp((int)Math.Floor(readyupCount / (_marinesPerSurvivor + 1)), _minimumSurvivors, _maximumSurvivors);
```
yagwog, line 418: Honestly, this can probably be simplified a bit but will need to ask smug, the limit is practically pinned at max count regardless of pop nowadays.

Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Helpers.cs
MidZik: When the autobalance is at it's most xeno-favored, it's nearly exactly like CM. CM tries to make 1/4 of the lobby xenos. When CMMarinesPerXeno is 3 (the top-end for xenos), that's equivalent to 1/4 of the lobby. Before this PR, it was 1/3 of the lobby at the top-end.

I say _nearly_ exactly like CM because CM doesn't count shipside roles as heavily. We have that logic for mid-round late joins giving burrowed, but not for roundstart.

Actually, I think I can make a sort of workaround for this by calculating burrowed after marines are spawned. This may lead to a "negative" burrowed count at round start if every xeno slot is taken but late joins should rapidly get it to zero.

rafapecogames:
amazing work here, it did really feel sometimes that xenos had too much numbers roundstart and received almost nothing mid round. i think this is great rebalancing

MidZik:
Added much more accurate xeno count correction logic. If there aren't enough xeno players, xenos will get the appropriate amount of burrowed, respecting shipside role weight limits. If there are too many xeno players spawned in, a "larva debt" will be placed on the hive that prevents late joins from giving larva until the debt is paid.

Unfortunately there isn't an easy way to prevent too many xenos from spawning in the first place, the current spawning system would have to be significantly changed to support that. The "larva debt" system is the best I can come up with without having to rip apart the spawning logic.

Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Helpers.cs
MidZik: The TODO is mostly referring to the mismatch between the logic and the cvar name. The logic itself does match how it works in CM13, although our configured numbers are different. It also has the same issue where shipside roles are counted too heavily when determining surv count but that isn't easy to fix.

Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Helpers.cs
MidZik: Added the workaround. This logic is still "flawed" but only serves as an estimate given the number of readied players. Xenos will get the appropriate amount of burrowed (or have a debt applied to them) after all players are spawned in, since at that point we know the exact role weights of every player in the round.

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
```diff
@@ -442,6 +442,26 @@ private void SetHiveBurrowedLarva(Entity<HiveComponent> hive, int larva)
...
     }
 
+    public void ChangeBurrowedLarvaDebt(int amount)
+    {
+        var hives = EntityQueryEnumerator<HiveComponent>();
+        while (hives.MoveNext(out var uid, out var hive))
+        {
+            ChangeBurrowedLarvaDebt((uid, hive), amount);
+        }
+    }
```
Arendian, line 452: Applying a burrowed larva debt will not balance the initial hive size. If I am understanding correctly how it works, you can end up with an oversized initial hive compared to the amount of marines if people ready up, don't get their role, then either quit or late join as a burrowed larva. It doesn't matter that the xenos then don't get any burrowed for the next x latejoins if the debt is not cleared before the first drop.

If it only checks spawned marines you can ignore this, I will check myself if this is the case once I have a bit more time.

Arendian (changes requested):

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
MidZik: Yes, that is correct. Fixing this is hard though, as xenos would have to be selected and spawned after all marines have their role. I'm not sure the best way to do that at this time so this is my "workaround" for the time being. It also is better than the system we currently have, where xenos just keep the extras with no downside.

MidZik: Ah but I should mention, if xenos _do_ have too many roundstart xenos, there will not be any burrowed larva available. So if a marine readies up but doesn't get a role, and there are too many xenos, it is impossible for them to latejoin as xeno. The logic that gives roundstart larva is the same as the logic that gives debt. Xenos will either gain larva if there aren't enough of them, or gain debt if there are too many.

There were two methods I was thinking of doing for trying to account for this:

1) Add slight penalty to the roundstart xeno limit estimate. I.e. if the ratio is 3:1 make it 3.2:1 or some other small adjustment. This would not change how many round start burrowed xenos would get since that calculation would not have the adjustment. This is dead simple to write but still an estimate, and as long as the adjustment is not too large it would not cause issues. If the adjustment is too large you'd end up with xenos not getting a roundstart larva but still being able to late join as xeno, which is weird behavior but not game breaking.

2) Redo the whole spawning logic to work more like it does in CM, or otherwise calculate the role weights as roles are being selected. Lots of possible ways to rewrite the logic to work, but regardless of which one is chosen I think it would be significant work.

MidZik:
I believe I have an idea on how to rework the spawning system in a way that would support getting the roundstart surv and xeno count exactly correct. Might take a while to get it working though and I'll submit it as a separate PR.

MidZik:
I have a rework to how spawning works in #10754. That PR would resolve the issue of xeno counts not being correct for the number of players who have actually spawned in.

Content.Shared/_RMC14/Xenonids/Hive/SharedXenoHiveSystem.cs
MidZik: This is resolved in #10754. With that PR, round start xeno and surv count is based on who actually spawns in the round, and considers their role weight so shipside roles don't contribute as much. I've removed the debt system as it won't be necessary. (I've also removed the granting of burrowed larvas here, but that is being added in the other PR as well.)

PR #10536, opened 2026-07-21
Adds: Button Coordinates Copy Range in Rangefinder UI

wheelwrightt:
## About the PR
Adds a button to the range ui to copy coordinates coming out to: "Longitude (x), Latitude (-y)"

## Why / Balance
parity

## Media
image

## Requirements

**Changelog**
:cl:
- add: Coordinates can now be copied to clipboard when using rangefinder.

### DIFF
--- a/Content.Client/_RMC14/Rangefinder/RangefinderBui.cs
+++ b/Content.Client/_RMC14/Rangefinder/RangefinderBui.cs
@@ -11,13 +11,19 @@ namespace Content.Client._RMC14.Rangefinder;
 [UsedImplicitly]
 public sealed class RangefinderBui : BoundUserInterface
 {
+    [Dependency] private readonly IClipboardManager _clipboard = default!;
+
     private RangefinderWindow? _window;
 
     private readonly AreaSystem _area;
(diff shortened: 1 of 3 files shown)


Content.Client/_RMC14/Rangefinder/RangefinderBui.cs
```diff
@@ -27,6 +33,9 @@ protected override void Open()
         base.Open();
         _window = this.CreateWindow<RangefinderWindow>();
         _window.Header.SetMarkupPermissive(Loc.GetString("rmc-rangefinder-header"));
+
+        _window.CopyButton.OnPressed += _ => _clipboard.SetText($"Longitude ({_lastX}), Latitude ({_lastY})");
```
PursuitInAshes, line 37: It may be easiest for understanding to leave off "Long, Lat" and just do X,Y like most people end up sending. I see a lot of people get tripped up when Long/Lat is attached to coords already

PursuitInAshes (changes requested):
Preference thing, code seems okay but I'm not familiar with UI stuff.
TY for the PR

PR #10540, opened 2026-07-22
Leave context menu open when repeating actions

### LINKED ISSUE #10539: Xeno acid action problem
Aciding a container or chest full of items with the xeno acid seems to be a chore. Need to press acid => right click for verb => the action gone =>  and then acid again and repeat. 

Unless I miss something, this shouldn't be this tedious.

MidZik:
## About the PR

The context menu no longer closes when you use an ability on an entity in it, if the ability is repeatable. For example, such as with aciding items.

Fixes #10539.

## Why / Balance

QOL. We made acid repeatable as a QOL change but it was actually anti-QOL when the context menu was involved, because upon aciding an item in the context menu, the context menu closed while the ability was still selected. Right clicking then cancelled the ability usage without opening the context menu. So you'd acid something, right click to cancel the ability, right click to open the context menu, use your ability again, and then you can acid something again.

Now you can right click to open the context menu, and acid as much as you want without having to double right click and reuse the ability.

This applies to any repeatable ability, so it also includes things like spits.

## Technical details

The context menu keybind handler will not close if all the following are true:

- The use bind is being used on an entity inside the context menu.
- An ability is selected.
- The ability can target things.
- The ability is repeatable.

If any of the above isn't true the context menu will close the same way as before.

## Media

https://github.com/user-attachments/assets/9794c9a1-8bb7-41c3-bda5-b05ab7eaf22a

## Requirements

**Changelog**

:cl:
- add: Using repeatable abilities, such as applying acid, on entities in a context menu will no longer close the context menu.


Content.Client/ContextMenu/UI/EntityMenuUIController.cs
```diff
@@ -155,6 +162,19 @@ public void OnKeyBindDown(ContextMenuElement element, GUIBoundKeyEventArgs args)
                     inputSys.HandleInputCommand(session, func, message);
                 }
 
+                // RMC start
+                // Don't close the menu on use if the user is currently using a repeatable action.
+                if (args.Function == EngineKeyFunctions.Use
+                    && _action.SelectingTargetFor is { } selectedActionId
+                    && _actionsSystem.GetAction(selectedActionId) is { } action
+                    && EntityManager.TryGetComponent<TargetActionComponent>(action, out var target)
+                    && target.Repeat)
```
Thereisabear, line 168:
```suggestion
                if (args.Function == EngineKeyFunctions.Use
                    && _action.SelectingTargetFor is { } selectedActionId
                    && _entityManager.TryGetComponent<TargetActionComponent>(selectedActionId, out var target)
                    && target.Repeat)
```

Content.Client/ContextMenu/UI/EntityMenuUIController.cs
```diff
@@ -1,10 +1,13 @@
 using System.Linq;
 using System.Numerics;
+using Content.Client.Actions;
 using Content.Client.CombatMode;
```
Thereisabear, line 3: remove since we not calling GetActions 
```suggestion
using System.Numerics;
using Content.Client.CombatMode;
```

Content.Client/ContextMenu/UI/EntityMenuUIController.cs
```diff
@@ -52,6 +55,10 @@ public sealed partial class EntityMenuUIController : UIController, IOnStateEnter
         [UISystemDependency] private readonly TransformSystem _xform = default!;
         [UISystemDependency] private readonly CombatModeSystem _combatMode = default!;
 
+        // RMC
+        [Dependency] private readonly ActionUIController _action = default!;
+        [UISystemDependency] private readonly ActionsSystem _actionsSystem = default!;
```
Thereisabear: sames again.
```suggestion
        [Dependency] private readonly ActionUIController _action = default!;
```

Thereisabear (changes requested):
Calling GetAction in the block isnt needed

MidZik:
Thanks for the review.

Thereisabear (approved):
LGTM

PR #10273, opened 2026-06-23
Plasma Trees give plasma to ovi Queen + change doafter visual

Vermidia:
## About the PR

Title.

## Why / Balance

Request by Smug on #9471 and is currently blocking it (The ovi part). The plasma part is just nice and consistent.

## Technical details

Plasma Tree checks for xenoResting or attachedOvipositor. Changed the effect to use the one used for plasma transfer.

## Media

https://github.com/user-attachments/assets/35705cae-640a-42c3-92b2-4af5f86c5374

(effect not pictured)

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- tweak: Plasma Trees now check for if a xeno is resting or ovi for giving plasma.
- tweak: Plasma Tree doafter now uses the transfer plasma action's effects.

PR #10560, opened 2026-07-24
Recovery node refactor

SabreML:
## About the PR
Merged `RecoveryNodeComponent` and `PlasmaTreeComponent` together along with their associated systems, since they were both essentially identical.

The new system works more or less the same as before but now checks the component's `RecoveryType` enum to decide whether it recovers health or plasma.
I also noticed that there were no `Dirty()` calls despite the component's fields being changed pretty often, so that's fixed as well.

I *had* wanted to make it all predicted too while I'm messing with the code because recovery nodes can feel really sluggish sometimes, but I wasn't able to figure out how to do that with the `_random.Pick(possibleTargets)` line.

## Why / Balance
Having two components + two systems which are probably 95% identical to each other is just silly. This also opens up the possibility of adding more structure variants with different effects in the future.
No in-game changes other than maybe some benefits from the added `Dirty()`ing.

## Technical details
The `HasComp<MobStateComponent>()` on L71 was removed since the following `_mob.IsDead()` method checks for that itself.
All of the other `HasComp` checks were changed to use entity queries since they're in a foreach loop over every entity in range.

In the yaml file, `HivePlasmaTreeXeno` and `HivePlasmaTreeXenoConstructionNode` have been changed to be children of the above recovery node prototypes so that most of their component definitions can just be inherited.

## Requirements

**Changelog**

N/A


Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
Vermidia: Can you put this effect into the component? For #10273 I gave plasma trees that plasma recovery effect instead so it'd be good to account for something like that.

Vermidia (changes requested):

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
SabreML: Oh yes that's a good idea, thank you.

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeComponent.cs
```diff
@@ -1,24 +1,37 @@
...
     [DataField, AutoNetworkedField]
-    public TimeSpan NextHealAt;
+    public TimeSpan NextRecoveryAt;
 
     [DataField]
-    public DoAfterId? HealDoAfter;
+    public DoAfterId? DoAfter;
+}
+
+public enum RecoveryType
```
Thereisabear, line 35: needs Serializable, NetSerializable

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/Hive/xeno_hive_recovery.yml
```diff
@@ -68,3 +70,40 @@
...
+  parent: HiveRecoveryNodeXeno
+  id: HivePlasmaTreeXeno
+  name: plasma tree
+  description: A tree-shaped node whose "berries" are filled with glowing blue liquid that drips from them.
+  components:
+  - type: Sprite
+    sprite: _RMC14/Structures/Xenos/hive_structures.rsi
+    state: recovery_plasma
+  - type: RecoveryNode
+    recoveryType: Plasma
```
Thereisabear, line 85: Specify the different recovery value in the yml otherwise youve silently nerfed plasma regen.
```suggestion
  - type: RecoveryNode
    recoveryType: Plasma
    recoveryAmount: 75
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -1,42 +1,46 @@
...
+    [Dependency] private readonly SharedXenoHealSystem _heal = default!;
+    [Dependency] private readonly SharedXenoHiveSystem _hive = default!;
+    [Dependency] private readonly XenoPlasmaSystem _plasma = default!;
+
+    private EntityQuery<DamageableComponent> _damageableQuery;
+    private EntityQuery<XenoPlasmaComponent> _plasmaQuery;
+    private EntityQuery<XenoRestingComponent> _restingQuery;
+    private EntityQuery<XenoComponent> _xenoQuery;
+
     public override void Initialize()
```
Thereisabear, line 39: Since you done a refactor perfect time doing a minor optimisation.
```suggestion
    private EntityQuery<XenoComponent> _xenoQuery;

    private readonly HashSet<EntityUid> _nearbyEntities = new();
    private readonly List<EntityUid> _possibleTargets = new();

    public override void Initialize()
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -49,63 +53,96 @@ public override void Update(float frameTime)
...
     }
 
-    private void TryHealRandomXeno(Entity<RecoveryNodeComponent> recoveryNode)
+    private void TryRecoverRandomXeno(Entity<RecoveryNodeComponent> recoveryNode)
     {
-        var (ent, comp) = recoveryNode;
-        var nearbyEntities = _lookup.GetEntitiesInRange(ent, comp.HealRange);
+        var nearbyEntities = _lookup.GetEntitiesInRange(recoveryNode, recoveryNode.Comp.Range);
         var possibleTargets = new List<EntityUid>();
         foreach (var nearbyEntity in nearbyEntities)
```
Thereisabear: And then change this
```suggestion
        _nearbyEntities.Clear();
        _possibleTargets.Clear();
        _lookup.GetEntitiesInRange(recoveryNode.Owner, recoveryNode.Comp.Range, _nearbyEntities);
        foreach (var nearbyEntity in _nearbyEntities)
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -49,63 +53,96 @@ public override void Update(float frameTime)
...
-            if (!_hive.FromSameHive(ent, nearbyEntity) ||
-                !HasComp<XenoComponent>(nearbyEntity) ||
-                !HasComp<XenoRestingComponent>(nearbyEntity) ||
-                !TryComp<DamageableComponent>(nearbyEntity, out var damageComp) ||
-                damageComp.TotalDamage <= 0 ||
-                !HasComp<MobStateComponent>(nearbyEntity) ||
+            if (!_hive.FromSameHive(recoveryNode.Owner, nearbyEntity) ||
+                !_xenoQuery.HasComp(nearbyEntity) ||
+                !_restingQuery.HasComp(nearbyEntity) ||
                 _mob.IsDead(nearbyEntity))
```
Thereisabear: Then this ( this another minor optimisation but now is good time since refactor)
```suggestion
            if (!_xenoQuery.HasComp(nearbyEntity) ||
                !_restingQuery.HasComp(nearbyEntity) ||
                _mob.IsDead(nearbyEntity) ||
                !_hive.FromSameHive(recoveryNode.Owner, nearbyEntity))
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -49,63 +53,96 @@ public override void Update(float frameTime)
...
+                        continue; // continue the foreach loop
+                    break; // break the switch statement, not the foreach loop
+
+                case RecoveryType.Plasma:
+                    if (!_plasmaQuery.TryComp(nearbyEntity, out var plasmaComp) || plasmaComp.Plasma == plasmaComp.MaxPlasma)
+                        continue;
+                    break;
+            }
+
             possibleTargets.Add(nearbyEntity);
```
Thereisabear:
```suggestion
            _possibleTargets.Add(nearbyEntity);
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -49,63 +53,96 @@ public override void Update(float frameTime)
...
         }
 
-        recoveryNode.Comp.NextHealAt = _time.CurTime + recoveryNode.Comp.HealCooldown;
+        recoveryNode.Comp.NextRecoveryAt = _time.CurTime + recoveryNode.Comp.Cooldown;
 
         if (possibleTargets.Count == 0)
```
Thereisabear:
```suggestion
        if (_possibleTargets.Count == 0)
```

Content.Shared/_RMC14/Xenonids/Construction/RecoveryNode/RecoveryNodeSystem.cs
```diff
@@ -49,63 +53,96 @@ public override void Update(float frameTime)
...
+        {
+            Dirty(recoveryNode);
             return;
+        }
 
         var selectedTarget = _random.Pick(possibleTargets);
```
Thereisabear:
```suggestion
        var selectedTarget = _random.Pick(_possibleTargets);
```

Thereisabear (changes requested):
See changes, just minor optimisation.
Also check verms PR regarding the Ovi check for Queen. 

They made changes onto the plasmatree system but in this refactor that file gone.

Resources/Prototypes/_RMC14/Entities/Structures/Xeno/Hive/xeno_hive_recovery.yml
SabreML: oh heck thank you

SabreML:
> [...] Also check verms PR regarding the Ovi check for Queen.
> 
> They made changes onto the plasmatree system but in this refactor that file gone.

I'll make sure that I remember to add the changes from that PR if it gets merged first, since it might not get flagged as a merge conflict with the file being deleted here. 👍

SabreML:
Tested and the changes from the other PR all still work! (Plasma trees use a different effect and are able to give plasma to the queen when she's ovi'd and not resting.)

Due to `PlasmaTreeSystem` and `RecoveryNodeSystem` being merged in this PR, this *does* now also allow recovery nodes to heal the queen in ovi too, which is a change from the original PR only modifying plasma trees.
It makes sense to me to keep it like this so that both structures work the same, but if that's an undesired change then it would be easy to change it back.

Thereisabear:
> merged in this PR, this _does_ now also allow recovery nodes to heal the queen in ovi too, which is a change from the original PR only modifying plasma trees.
> It makes sense to me to keep it like this so that both structures work the same, but if that's an undesired change then it would be easy to change it back.

This should be fine, its a very edge case where the heal would actually affect a queen in Ovi

Thereisabear (dismissed):
Changes look good.

PR #10568, opened 2026-07-25
Health analyzer update restrictions

Vermidia:
## About the PR

Title. Health analyzers no longer update if the user holding one would require a doafter to use it. They also stop updating if the user gets out of range of the target.

## Why / Balance

Mostly to match parity without killing the QoL.

## Technical details

Changed the brute and burn protos for wound types to be a dict to be easier to modify if needed (and makes it easier to pass to the health scanner states)

health scanner states now take a mobstate, damage spec. wound list, and wound type dict to store said information without having to update due to the target wandering off etc.

Everything but the holocard now updates when looking at the health analyzer (I don't want to deal with storing that too right now).

## Media

https://github.com/user-attachments/assets/557f8e8f-b138-4320-b225-0c294e1a307d

## Requirements

**Changelog**

:cl:
- tweak: Health analyzers no longer update continuously if you are too far from the scanned target, or if the user would require a doafter to use the analyzer.

### DIFF
--- a/Content.Client/_RMC14/Medical/Scanner/HealthScannerUiData.cs
+++ b/Content.Client/_RMC14/Medical/Scanner/HealthScannerUiData.cs
@@ -285,7 +285,7 @@ public sealed class HealthScannerUiData
             return;
 
         // Defibrillation related
-        if (_mob.IsDead(target))
+        if (uiState.State == Shared.Mobs.MobState.Dead)
         {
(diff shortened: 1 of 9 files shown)


PursuitInAshes (approved):
This all seems reasonable but I don't feel comfortable fully reviewing the C#, so just conceptual approval here
TY for the PR

PR #10579, opened 2026-07-27
EORG clarification.xml

Beansies:
## About the PR
Updated EORG for clarity to reflect Admin enforcement 

## Why / Balance
Self damage is currently enforced as EORG and this makes it clear in the rules

## Technical details
Added that friendly fire includes damage done to yourself

## Media
N/A

## Requirements

### DIFF
--- a/Resources/ServerInfo/Guidebook/_RMC14/Rules/RMCZeroTolerance.xml
+++ b/Resources/ServerInfo/Guidebook/_RMC14/Rules/RMCZeroTolerance.xml
@@ -14,7 +14,7 @@
 
   - End of round grief (EORG) will result in an instant unappealable 3 hour ban.
   \n[bold]This includes suicides without a valid roleplay reason.[/bold]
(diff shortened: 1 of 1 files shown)


PursuitInAshes:
Thanks for the PR!
Consensus is this isn't needed, the line above what you have altered covers this with reasonable clarity:
`\n[bold]This includes suicides without a valid roleplay reason.[/bold]`

There is additionally a popup that must be dismissed that covers EORG rules, with a checkbox (Unticked by default) to make it not display again. The text can be found at `Resources/Locale/en-US/_RMC14/roundend/no-eorg-popup.ftl`.

Beansies:
> Thanks for the PR! Consensus is this isn't needed, the line above what you have altered covers this with reasonable clarity: `\n[bold]This includes suicides without a valid roleplay reason.[/bold]`
> 
> There is additionally a popup that must be dismissed that covers EORG rules, with a checkbox (Unticked by default) to make it not display again. The text can be found at `Resources/Locale/en-US/_RMC14/roundend/no-eorg-popup.ftl`.

If talked to many players who didn’t know that self damage that doesn’t result in suicide still constitutes EORG. And this wasn’t clear to me personally either. I feel like adding this would make it clear that any self damage even if it’s to test game mechanics or damage values are still considered EORG.

PR #10078, opened 2026-06-08
Fix: Toxic water damage

IceNoobXD:
## About the PR

Toxic water damage has been reduced 
| Xeno| Larva| Marine|
|--------|--------|--------|
| 88 Burn| 4.91 Burn | 15 Burn|
| 17.12 Heat| 0.98 Heat| 7.5Heat|

cmss13 stats are 34, 2 and 15 heat

> This brings all values to parity no nerfs or buffs.
RMC systems damage ticks every 1second, CMs damage values are calculated on ticks of 2 seconds.
So the values you have are per 2 seconds.

## Why / Balance

Parity

## Technical details

Changed the damage type to heat instead of group: burn which caused to player to take cold damage also, changed the 2 multipliers 

## Requirements

- [x] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- fix: The damage per second of the toxic water has been greatly reduced to only be 7.5 for marines, 17 for xenonids and 1 for larvae!

PR #10581, opened 2026-07-27
Add Trijent Dam toxic water filtration mechanics

GrigoriyGalintano:
## About the PR

Adds a reusable CMSS13-style toxic water filtration system for Trijent Dam.

Includes linked filter controls, purification propagation, toxic water effects, transition visuals, and ready-to-map prototypes. No maps are modified.

## Why / Balance

Ports the Trijent Dam water filtration mechanic from CMSS13.

There is no current round impact because the system is not placed on any map.

## Technical details

- Adds arbitrary-ID filter networks with buttons, initiators, relays, and delayed relays.
- Implements eight-directional purification propagation and `Toxic → Dispersing → Purified` states.
- Toxic water damages and irradiates mobs, releases xeno pulls, damages weak vehicles, and respects catwalk cover.
- Purification removes weeds, resin, and fire, blocks new xeno construction, and can throw loose items.
- Extends RMC power receivers with requested load modes and one-cycle power consumption.
- Extends damage-over-time support with per-source timers, radiation, and popup filtering.
- Adds Filter 1/2 water prototypes and licensed CMSS13 transition sprites.

## Media

https://github.com/user-attachments/assets/b43bc06d-d9c5-46b3-ad55-d5124e217147

## Requirements

**Changelog**

:cl: GrigoriyGalintano
- code: Added reusable toxic water filtration mechanics and mapping prototypes.


IceNoobXD:
Firstly I want to thank you for this but I have noticed a few problems with it that I would like to share.

The button uses the APC as power source and not gens meaning with a charged battery you can activate the filtration process without even touching a gen, correct me if im wrong in cmss13 it gens need to be active for the button to be pressable

The damage seems rather high compared to already existing toxic water prototype
The water damage was already discussed in this PR #10078 
> RMC systems damage ticks every 1second, CMs damage values are calculated on ticks of 2 seconds.
So the values you have are per 2 seconds.

and I am also uncertain about it dealing toxin damage.
(Would also be lovely if you could update the pre-existing toxic water with your new code if you didnt already.)

The animation direction for the edge tiles is a bit funky and it appears all animations play twice.

Shore tiles are not affected by the purification system, this would only be needed for appearence consistency.

https://github.com/user-attachments/assets/b35bab84-8b08-4af5-9345-7f60f1538d98

GrigoriyGalintano:
> Firstly I want to thank you for this but I have noticed a few problems with it that I would like to share.
> 
> The button uses the APC as power source and not gens meaning with a charged battery you can activate the filtration process without even touching a gen, correct me if im wrong in cmss13 it gens need to be active for the button to be pressable
> 
> The damage seems rather high compared to already existing toxic water prototype The water damage was already discussed in this PR #10078
> 
> > RMC systems damage ticks every 1second, CMs damage values are calculated on ticks of 2 seconds.
> > So the values you have are per 2 seconds.
> 
> and I am also uncertain about it dealing toxin damage. (Would also be lovely if you could update the pre-existing toxic water with your new code if you didnt already.)
> 
> The animation direction for the edge tiles is a bit funky and it appears all animations play twice.
> 
> Shore tiles are not affected by the purification system, this would only be needed for appearence consistency.
> 
>  test.mp4

Thanks for pointing this out. I double-checked the CMSS13 implementation: filtration_button/attack_hand() only checks inoperable(), which covers NOPOWER and BROKEN, and then calls use_power(5). It does not directly check whether the generators are running.
For that reason, I kept the powered APC channel as the requirement, including power supplied by the APC battery. The generators may be how the area is normally supplied, but they are not a separate requirement of the button itself. If there is additional map-specific logic I have missed, I would be happy to check it.

PR #10593, opened 2026-07-29
Xeno fishing

Vermidia:
## About the PR

Title.

## Why / Balance

It's fun.

## Technical details

Made a shared system for fishing, and a client system. Moved some stuff over to shared, mainly spearfishing stuff.

Made a general spearfishing function, and a public function for xeno fishing. Tail stab on miss calls the function, and tail seize using a new event, calls it before max range deletion if it didn't hit anything (idk if spent is actually set right but its there in case). Also flurry attempts to catch fish too, why not.

Tail stab makes a sound if a fish was caught. I did check, normal fishing looks fine.

## Media

https://github.com/user-attachments/assets/505b9004-c459-43b7-9fa3-500b9674df95

## Requirements

**Changelog**

:cl:
- add: Xenos can now fish using tailstab on water (this only triggers on a tail stab miss). Additionally tail seize can also be used to fish, if it fails to hit anyone. Flurry can also be used to fish somewhat. Defender tail slam can fish, but is much worse than other xenos' tail stabs.

### DIFF
--- a/Content.Client/_RMC14/Fishing/RMCFishingSystem.cs
+++ b/Content.Client/_RMC14/Fishing/RMCFishingSystem.cs
@@ -1,0 +1,7 @@
+using Content.Shared.Fishing;
+
+namespace Content.Client._RMC14.Fishing;
+
+public sealed partial class RMCFishingSystem : SharedRMCFishingSystem
+{
+}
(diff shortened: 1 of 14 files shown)


GrigoriyGalintano:
LOL

PR #10368, opened 2026-07-04
Prevent syringe use without medical 2

Vermidia:
## About the PR

Title.

## Why / Balance

parity!

## Technical details

yml

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- fix: Fixed being able to use normal syringes without medical skill 2.

PR #10605, opened 2026-07-30
Syringe parity

### LINKED ISSUE #3978: Syringe Improvements
## Description

Syringes require medical skill 2 to use in cm. They are insta draw. You can also instant inject yourself. You can instant inject on harmmode but this breaks the syringe making in unusable. You can also inject blood into the blood stream, which won't treat it as injested as it does currently.

**Reproduction**

**Screenshots**

**Additional context**

Vermidia:
## About the PR

Title.

Added instant self inject, instant drawing, skill based delays. Forced injects in combat mode and breaking of the syringe, and ways that can be cancelled.

Toggling injector state when in a doafter now cancels the doafter. Parity just preserves previous state but I think this is more intuitive and good if someone wants to quickly cancel an inject.

## Why / Balance

Parity! Fixes #3978. (Note the skill lock is done in a diff PR)

## Technical details

New RMCInjector system which handles a new event to change syringe inject delay, for instant self injects and instant draws in most cases. Also skill reducing delay! With some other options. Also implements combat injections, where if the syringe is used as a weapon it can forcefully inject some of the chems, even with no skill!

Some changes were made to the upstream injector system, mostly to prevent chem draw from drawing blood, fixing try inject bloodstream so it works and can directly inject blood, and actually care about the delay.

Might have to rework this if the hypo-injector merge upstream happens but that's fine, it'll prob be awhile...

## Media

https://github.com/user-attachments/assets/32322981-92d9-4c36-b50e-9bb52e82bd21

https://github.com/user-attachments/assets/8ea2dde8-883f-48db-b24c-d7a657309711

## Requirements

**Changelog**

:cl:
- fix: Fixed syringe inject and draw times. They now instant draw and instant self inject. Otherwise their inject time is 2s, and is a bit quicker with very high skill.
- fix: Fixed syringes removing non-blood chems from mobs. They only remove blood now.
- fix: Fixed syringes being unable to direct inject blood properly. It now properly refills blood and no longer causes vomiting.
- fix: Fixed being able to freely switch modes while a syringe doafter was occuring. Switching modes now cancels the doafter.
- add: Syringes can now be forced injected (even without skill) by using them in combat mode. This breaks the syringe, and injects a random amount of chems present. This can be cancelled by the target having high cqc skill and being in combat mode or by armor.


DelftModiste:
`fix: Fixed syringes removing non-blood chems from mobs. They only remove blood now.`
This would effect syringes ability to treat ODs outside of medbay, since syringing blood out and using a blood bag to Iv blood back in allowed excess chems in a system to be brought down faster.

Not arguing that functionality _should_ stay, just wanted to bring up a gameplay side effect of the change.  In worst case scenarios, medevac to medbay may end up being required, but would still be fixable there.

x5tink:
Instant drawing of blood? this seems very open to abuse... 
You could literally drain all the blood in a person by spam alternating Click and C to pour the syringe into a bucket/jug in your other hand and remove every drop from a person in seconds. Or for example, you do this to everyone on the alamo, without warning or them being able to do anything. Is this something we really want?

x5tink:
Also based on the code, a person with zero skill, ie a rifleman can instantly inject themselves with chems. Kind of making auto injectors irrelevant, as they can just take a bottle and a syringe and achieve the same effect.

Vixting:
> Also based on the code, a person with zero skill, ie a rifleman can instantly inject themselves with chems. Kind of making auto injectors irrelevant, as they can just take a bottle and a syringe and achieve the same effect.

There is a sperate pr adding a skill lock
Only way to do this for non skilled would cause the syringe to break

x5tink:
> > Also based on the code, a person with zero skill, ie a rifleman can instantly inject themselves with chems. Kind of making auto injectors irrelevant, as they can just take a bottle and a syringe and achieve the same effect.
> 
> There is a sperate pr adding a skill lock Only way to do this for non skilled would cause the syringe to break

Could you not just apply the MedicallyUnskilledDoAfter component to it? the same way the trauma and burn kits work?

x5tink:
> Instant drawing of blood? this seems very open to abuse... You could literally drain all the blood in a person by spam alternating Click and C to pour the syringe into a bucket/jug in your other hand and remove every drop from a person in seconds. Or for example, you do this to everyone on the alamo, without warning or them being able to do anything. Is this something we really want?

https://github.com/user-attachments/assets/810cd4ce-4a95-4c25-b926-348d671c9ded

This is on this branch, just to show that a rifleman can become a vampire and steal everyones blood in moments....

Thereisabear:
> Instant drawing of blood? this seems very open to abuse... 
> 
> You could literally drain all the blood in a person by spam alternating Click and C to pour the syringe into a bucket/jug in your other hand and remove every drop from a person in seconds. Or for example, you do this to everyone on the alamo, without warning or them being able to do anything. Is this something we really want?

This raises a good point about the stealing of everyone's blood in seconds .. oh Jesus, just saw the video.

Resources/Prototypes/_RMC14/Entities/Objects/Medical/syringes.yml
```diff
@@ -72,3 +98,28 @@
         reagents:
         - ReagentId: RMCChloralhydrate
           Quantity: 50
+
+- type: entity
+  parent: BaseItem
+  id: RMCSyringeBroken
+  name: syringe
+  description: A syringe. It's broken.
+  suffix: RMC14
```
x5tink: Just a suggestion, add an additional suffix of 'Broken', so its easier to tell the difference in the entity spawner menu. Or change its name.

AverageNotDoingAnythingEnjoyer:
> This is on this branch, just to show that a rifleman can become a vampire and steal everyones blood in moments....

I think there are better ways to raid if that's the main concern

Vermidia:
Yeah I'm aware this makes ODs worse, but that's parity (if there's no field triage with a portal dialysis machine or etc).

See #10368 for the skill lock, and also because I couldn't get the no draw on hostiles to be ok with factions (it was preventing drawing from monkeys) I changed it to check combat mode instead so prob pretty safe?

Vixting (approved):

PR #10608, opened 2026-07-30
Add blitz lag comp

Vermidia:
## About the PR

Title. Inspired by/taken from #10607.

## Why / Balance

Same reason as why tail sweep benefits from it. Blitz is a very close range move that often fails to hit due to no lag comp on the vanguard's side. It should make it more reliable.

## Technical details

Basically the same thing as the above PR. It's good.

## Media

https://github.com/user-attachments/assets/cebcee70-8553-47d6-b149-f19217252eee

Didn't do too much testing but it should work the same as the other PR probably??

## Requirements

**Changelog**

:cl:
- fix: Vanguard blitz is now lag compensated.

### DIFF
--- a/Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
@@ -105,7 +105,7 @@ public sealed class XenoBlitzSystem : EntitySystem
 
         var session = CompOrNull<ActorComponent>(xeno)?.PlayerSession;
 
-        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range + xeno.Comp.LagCompensationLookupMargin, LookupFlags.Dynamic | LookupFlags.Approximate))
+        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range + xeno.Comp.LagCompensationLookupMargin, LookupFlags.Dynamic | LookupFlags.Static | LookupFlags.Approximate))
         {
             if (!_xeno.CanAbilityAttackTarget(xeno, hit))
                 continue;
--- a/Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
+++ b/Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
@@ -105,7 +105,7 @@ public sealed class XenoBlitzSystem : EntitySystem
 
         var session = CompOrNull<ActorComponent>(xeno)?.PlayerSession;
 
-        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range + xeno.Comp.LagCompensationLookupMargin))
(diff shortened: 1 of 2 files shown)


Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
```diff
@@ -97,16 +99,21 @@ private void OnAttackBlitz(Entity<XenoBlitzComponent> xeno, ref XenoBlitzEvent a
...
 
-        //Note doesn't seem to work here
         EnsureComp<XenoSweepingComponent>(xeno);
 
         var hits = 0;
 
-        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range))
+        var session = CompOrNull<ActorComponent>(xeno)?.PlayerSession;
+
+        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range + xeno.Comp.LagCompensationLookupMargin))
```
MidZik: As in #10607, dramatically reduce how much time is spent here in the worst cases.
```suggestion
        foreach (var hit in _lookup.GetEntitiesInRange<MobStateComponent>(_transform.GetMapCoordinates(xeno), xeno.Comp.Range + xeno.Comp.LagCompensationLookupMargin, LookupFlags.Dynamic | LookupFlags.Approximate))
```

Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
```diff
@@ -97,16 +99,21 @@ private void OnAttackBlitz(Entity<XenoBlitzComponent> xeno, ref XenoBlitzEvent a
...
         {
             if (!_xeno.CanAbilityAttackTarget(xeno, hit))
                 continue;
 
+            // Range check against the target's lag-compensated position
+            if (!_rmcLagCompensation.IsWithinMargin(xeno.Owner, hit.Owner, session, xeno.Comp.Range))
+                continue;
```
MidZik: As in #10607, since we don't rely on client input we can probably remove the margin.
```suggestion
            // Range check against the target's lag-compensated position, without margin
            var hitCoords = _rmcLagCompensation.GetCoordinates(hit, session);
            if (!_transform.InRange(xeno.Owner.ToCoordinates(), hitCoords, xeno.Comp.Range))
                continue;
```

MidZik (changes requested):
Same changes as requested in the defender lag comp PR.

Content.Shared/_RMC14/Xenonids/Blitz/XenoBlitzSystem.cs
Vermidia: Unlike tail sweep, blitz actually does run fully on server/client, does that matter in this case?

MidZik: I believe the margin exists for when the client sends hits to the server. That lets the server give the client some leeway when the client has small mispredicts and the hits don't exactly align. Since the client isn't sending any hits to the server, we don't need the margin.

That does mean there will definitely be mispredicts though. The client will mispredict blitz hits due to the action running out of simulation. I believe that would be mostly if not entirely fixed by #10103. Changing if or how the margin is used would not help with this.

We could solve the mispredicts without #10103 by having the client network the blitz hits and then using the margin only on the server, but I think this lag comp method is preferable.

MidZik:
Tested and seems to work as expected, with the same caveats as in #10607.

I'll note here that there was a bug where the lag comp margin was being applied against moving targets incorrectly. That was giving blitz (and defender's tail sweep) an additional 0.25 range against moving targets. That is fixed in these PRs, so their radius is now smaller against moving targets.

MidZik (approved):

PR #10614, opened 2026-07-31
no more heroes: SGO

Ian321:
## About the PR

I heard that we (eventually) want to move away from specs and have already thought of some ways. This PR adds a "smartgun conversion kit" for the M54CE2 that aims to replace SGOs. It is currently only available via the spawn menu and adds a case that contains the attachment for the M54CE2 and a harness that has to be worn in the backpack slot in order to use it.

My vision would be to add this to req with a lower scale than the gun itself and to remove the SGO role. Same goes for the WS role; to create some "nerfed" version of their guns/etc as restricted weapons and to remove the role.

## Why / Balance

- The M54CE2 was already quite similar to the ML66A.
- Add a alternative play style to it compared to "bipod deployed!"
- Harnes in the back-slot, as to mirror the ML66A blocking it.

## Technical details

- New component for attachments to add/remove a component to the gun on attach/detach.
- New attachment for the M54CE2.
  - Grants IFF
  - -10 damage, -300° scatter and -2 recoil to bring it on par with the ML66A
  - Adds requirement for the harness.
- New harness "backpack".

## Media

[Screencast From 2026-07-31 09-26-15.webm](https://github.com/user-attachments/assets/2ee78fa5-ba9d-4585-a02a-1ed166944e53)

## Requirements

**Changelog**

Only admeme for now.


BarteG44:
really don't see the point of this or moving away from specs being their own roles. This being added achieves the same exact effect as SGOs but without the possibility of hourlocking the role and by replacing the purely chance based roll at roundstart with a lag-dependant rush to req when you wake up. At best it means that the gun isn't locked to being used by just one person, meaning that their death isn't a huge loss, but that also makes it 100x more likely that people will just steal these guns

Gallagin:
adding more stuff to the RFNs all rushing to get kits from req isnt a good call imo

iirc the whole removing specs talk was more about potentially making WS more about multiple people doing a crew served weapon sort of thing

I'm more of the opinion that post parity SGO should get more weapon options like TGMC has and then potentially increasing it to maybe 2 per squad to make it feel less exclusive and then balancing the weapons to not be so oppressive when multiple of them are together shooting the same target. Then adding more backpack slot options to buy for points such as a no inventory space backpack phone since they get assigned as aSL so commonly

PR #10625, opened 2026-08-01
Clarify/Adjust the Synthetic MLaw Page for Survivor Synths

PursuitInAshes:
## About the PR
Added a line to hopefully prevent RP of CLs trying to sell off recovered synths, or the like, not desirable RP.
"Survivor Synthetics" has been replaced with "Third-Party Synthetics" and adjusted to match. This removes the line where these Synthetics are granted "Similar rights as UNMC Synthetics" to hopefully make it easier for MPs and Survs alike when these are present.

## Why / Balance
Discussion with Synths has determined that the above clause was too vague, the options were to remove it or expand/clarify it, the concern then was by expanding it would cause complexity issues and most likely cause some gaps that would invite strange issues. So instead, it was removed and clarified that you can just treat recovered synthetics like any other Survivor and that you may still "contract" them.

## Technical details
XAML

## Media
image

--------------

image

## Requirements

**Changelog**
:cl: PursuitinAshes
- tweak: (MLaw) Clarified how Third-Party Synthetics should be handled/processed.
- tweak: (MLaw) The Chief MP or Commander may now declare Synthetics rogue if they have been consistently detrimental to the operation, this should be used as a last resort.


CatAndHats (approved):
Requires Whisper approval, all LGTM though

PR #10633, opened 2026-08-02
Fix xenos despawned by hijack being offered to players in the larva queue

### LINKED ISSUE #10617: Post hijack queue attempts to put player in dead body
## Description
After Queen hijacks and the xenos that remain on the surface are forced to join the queue and come back as a larva, however during round 13530 the larva queue showed up for a fraction of a second before being replaced by the alert that the body I was attempting to join was dead. This put me back in the queue in position 1, but I accidentally left the queue before I was put in a body and rejoined at the end of the queue instead. 

**Reproduction**
I do not know the exact steps to reproduce, but I assume this is related to "Changing castes/strain no longer gives a fake larva queue offer #10150" in that a body becomes available to join while it should not. 

**Screenshots**

Image

Image

**Additional context**

Image

Arendian:
## About the PR

Bodies of planet side xenonids that are despawned because of hijack no longer get offered to players in the larva queue.

## Why / Balance

Resolves #10617

## Requirements

**Changelog**

:cl: Dygon
- fix: Xenonids despawned because of hijack no longer get offered to players in the larva queue.

### DIFF
--- a/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Hijack.cs
+++ b/Content.Server/_RMC14/Rules/DistressSignal/CMDistressSignalRuleSystem.Hijack.cs
@@ -81,6 +81,8 @@ public sealed partial class CMDistressSignalRuleSystem
                 if (comp.CountedInSlots)
                     larva++;
 
+                RemComp<CanBeLarvaQueuedComponent>(xeno);
+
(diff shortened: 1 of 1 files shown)


PursuitInAshes (approved):
Should be okay, Thanks

CatAndHats (approved):
LGTM

PR #10654, opened 2026-08-04
Blood Queen

Vermidia:
## About the PR

Admin spawn thingy!

## Why / Balance

It was a fun concept to mess around with while vaguely trying to balance it.

Changes:
Loses screech, spit abilities, ovi, building abilities, planting weeds. Acid downgraded to weak, pheros downgraded to strong (x3). Also they can't hold eggs, I guess.

Gains 500 health, 10 armor, king-level speed, and 25% more damage to slashes + tail stab.

Blood queen gets king's doom (no changes) and crusher's defensive shield (renamed brace, lasts 2 less seconds, no explosion resist, shield visuals and only reduces damage by 5, not 10).

They also get a modified crusher brutalize (no cooldowns reduced, does 80% of base damage).

Instead of gut it gets massacre, which gibs in an area. It has the same cooldown as normal gut, and the bodies must be dead. For every gib they get a burrowed larva.

They also get blood drain, which on a pulled down target removes some blood, does some damage, and gives the hive's xenos some evolution points per drain.

The basic idea is she can't actually do any support and has to rely on deaths or the occasional host sacrifice to keep the hive alive.

## Technical details

YML for most changes.

New systems/comps for the new gib ability and the blood drain.
Slight change to crusher shield to not show explosion resist stuff if it isn't present + shield visual effects.

~~Theres a doom fix in here that's been split off - don't want to bother to remove it in this PR should be fineeee.~~

## Media

[It's in discord I made it too long](https://discord.com/channels/1168210010233376858/1386929244168392744/1534267532855349248)

## Requirements

**Changelog**


ChaMaganera:
Actual fire concept to shake up the game from time to time 🔥 🔥 🔥 🔥  (or when admins feels like doing something different)

PursuitInAshes (dismissed):
Thanks for the PR!
Could you atomize the doom fix from blood queen just so they can be considered separately and not potentially held up?

PR #10667, opened 2026-08-06
Cleanup bone chips/shard resistance

Vermidia:
## About the PR

Title.

## Why / Balance

Unintuitive that it resists damage without actually marking itself as ignoring resistances.

## Technical details

YML to give them RMCBullet, which is used to mark stuff as "bullets", making it properly hit bullet armor, if a proto down the line needs to parent it and consider AP. And properly sets the ignore resistances field.

Also removed an unused projectile.

## Requirements

**Changelog**

no cl just code fix


Thereisabear:
image

Our bone chips are currently parity, they ignore armor which is why they have no comp.

Yes this means hedgehog shards are intended to ignore armor too.

If they are somehow hitting bio armor then that's the bug that would need to be fixed.

Vermidia:
Ok so ours ignored armor simply by the chance that Bio specifically only considers the burn types of damage. Added it to the yml so it's actually clear that's what it's supposed to do.

PR #10680, opened 2026-08-09
Allow exiting a destroyed vehicle even when the normal exit is blocked.

### LINKED ISSUE #10594: Xenos/Marines entering a tank against a wall can never get out again (without a VC)
## Description

Title. See round #13484.

CM doesn't solve this issue! It even makes it worse because marines can't enter the tank at all since the entrance is at the  very back and they can't reach it. Idk if the tank can't move maybe there could be at least one long doafter to leave it anyway into a free tile at least, or something.

**Reproduction**

**Screenshots**

**Additional context**

Arendian:
## About the PR

The game  will now try to place an entity on any vehicle adjacent tile if the normal exit is blocked. If even that fails the entity will be placed on top of the vehicle if the vehicle supports it.

## Why / Balance

Prevent being locked in a vehicle if the exit is blocked.

## Media

https://github.com/user-attachments/assets/ef2578b9-9132-48d3-a214-351dfee3adce

## Requirements

**Changelog**

:cl: Dygon
- fix: You can now exit a destroyed vehicle even if it's exit is blocked.

### DIFF
--- a/Content.Shared/_RMC14/Vehicle/Core/VehicleSystem.cs
+++ b/Content.Shared/_RMC14/Vehicle/Core/VehicleSystem.cs
@@ -1099,7 +1099,7 @@ public sealed class VehicleSystem : EntitySystem
         if (!TryComp(vehicle, out HardpointIntegrityComponent? frameIntegrity))
             return false;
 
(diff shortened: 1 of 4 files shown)


Vermidia:
Fixes #10594

Vixting (approved):

PR #10696, opened 2026-08-12
Allow the queen to remotely remove structures and structure ghosts

Arendian:
## About the PR

- The queen can now remotely remove strucures by clicking their icon on the map, built structures take 5 minutes to be removed and unbuilt ones are removed instantly. Built structure removal can be cancelled by clicking the map icon again before the timer is finished.
- Xeno structure ghosts now have a transparent icon on the map.

## Why / Balance

Allows the queen to remove unreachable structures so they don't take up a build slot.

## Media

image

## Requirements

**Changelog**

:cl: Dygon
- tweak: The Queen can now remotely remove structures by clicking their icon on the map. Built structures will be removed after 5 minutes, unbuilt structures will be removed instantly.
- tweak: Unbuilt xenonid structures now have a transparent icon on the map.


QuietlyWhisper:
This won't work for structures on shipside, if for example somehow a drone sets up a small hive on the ship, is killed, but the hive for whatever reason isn't found

Or just on hijack generally speaking, which is a smaller concern

If you could have for example, another map tab on the ui with the ship map, that would solve this issue, but is out of scope for this PR in particular and isn't required

PR #10592, opened 2026-07-29
Extends harm mode to match pulling target range

ChaMaganera:
## About the PR
Extends harm mode to match pulling target range

## Why / Balance
QOL change specially for xenos, there is nothing more frustating than dragging but going "Oops i was too far when i started dragging now i cant tackle the tall, guess i lose the cap :("

## Technical details
Checks if the target ID from RMCGetUserDisarmRange and RMCMeleeUserGetRangeEvent is the same as the Pulling ID, if yes extends the action of the attack/disarm to 4 tiles.

## Media
When pulling something, your harm range to it gets extended to only the thing your pulling

https://github.com/user-attachments/assets/05f93603-54d7-4262-a82c-e7986c10d586

Marines are also benefited from this

https://github.com/user-attachments/assets/bda265d4-ab59-402c-ad5a-72236f538fcd

Ps: Shoutout to sister EF-EE for helping me out in my first C# project
## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl:
- add: Makes the pulling target always be able to be targeted in harm mode.

PR #10722, opened 2026-08-16
Fix for xenos dragging marines at max range being untacklable

Ripofin:
## About the PR
There's some innate range 'slack' built into the SS14 dragging system (joint.Length + **0.15f**) that can allow for a dragged marine to become out of range of a xeno's tackle while still being dragged. This PR allows xenos to be able to tackle marines that they're dragging regardless of range. Only applies to xeno's tackling and only the marine they're dragging. Still respects LoS, just removes the range component.

## Why / Balance
This is simply a quality of life fix, but can be argued to be a xeno buff. Seemed like an oversight that someone you are dragging can become out of range. An alternative fix would be to remove the 'slack' for xenos when dragging, or reduce the range that they can initiate a grab, but that could have some larger physics repercussions than simply removing the range component.

## Technical details
Added a new event, RMCMeleeUserGetDisarmRangeEvent, which can be used to adjust the range of the disarm event. Ensured that the base SS14 code that calculates disarm range uses the new GetUserDisarmRange() function which leverages this new event, but will use the pre-existing meleeComponent.Range value for non XenoComponents. 

## Media

Tackle

## Requirements

**Changelog**

:cl:
- fix: Fixed xenos being unable to tackle marines grabbed and dragged when at max range.

### DIFF
--- a/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
+++ b/Content.Client/Weapons/Melee/MeleeWeaponSystem.cs
@@ -219,7 +219,9 @@ public sealed partial class MeleeWeaponSystem : SharedMeleeWeaponSystem
             target = screen.GetClickedEntity(mousePos);
 
         var attackerPos = TransformSystem.GetMapCoordinates(attacker);
-        if (mousePos.MapId != attackerPos.MapId || (attackerPos.Position - mousePos.Position).Length() > meleeComponent.Range)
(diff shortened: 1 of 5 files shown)


SabreML:
Same idea as #10592

Ripofin:
> Same idea as #10592

Ahh I searched and couldn't find anything similar. Looks like I can close this.

PR #10726, opened 2026-08-16
[Fix] Open doors no longer block some mechanics

carccborg:
## About the PR

Minor C# change, added an if statement. Linesystem counted two opening doors as blockers on the diagonal.
## Why / Balance

Bugfix.
## Technical details

Check to see if a potential blocker is a door, and if said door is not closed, denying, or welded, skip further blocker checks on that door.
## Requirements

**Changelog**

:cl:
- fix: Open doors no longer block some mechanics, including Oppressor's Abduct and the M34 incinerator


carccborg:
This is visible if you try to use Oppressor's Abduct through one door, then try to do the same for two doors on the same axis such that the raycast would pass through both doors as they're opening. It's not visible unless you do it quickly while both doors are still opening

CatAndHats (approved):

PR #10728, opened 2026-08-17
Motion Detectors now detect Hugged/Infected Marines

EndPage:
## About the PR

Motion Detectors now detect hugged/infected marines if they are moving. This also causes you to detect yourself if you are infected.

## Why / Balance

Since there is a Xeno inside you, the Motion Detector should be able to detect it. This should also make first contact and roleplay more interesting. The Discord feature request thread is here:
https://discord.com/channels/1168210010233376858/1535768081291542578

## Technical details

- Marines don't normally have `MotionDetectorTrackedComponent`, so I add it when they become infected and remove it when the infection (`VictimInfectedComponent`) is removed.
- Some entities already have `MotionDetectorTrackedComponent`, so entities that did not have the component before infection are tracked in a `HashSet`. Upon removal of `VictimInfectedComponent`, a function in `SharedXenoParasiteSystem.cs` is called to remove `MotionDetectorTrackedComponent` from the ones I saved.
- I determined infection by either an entity having a `VictimInfectedComponent `or an `InfectableComponent` with `BeingInfected = true.` 

## Media

Motion Detector detects infected marine
https://github.com/user-attachments/assets/9a1b68aa-6bae-423f-ab4b-cfce00b0bd56

Motion detector detects the user after infection
https://github.com/user-attachments/assets/3d356b7b-e3ca-4c48-83d0-ed40dabe8253

Motion Detector still doesn't detect uninfected marines
https://github.com/user-attachments/assets/42cbeb62-48a3-43d6-9ad0-5d1a6476139c

## Requirements

**Changelog**

:cl:
- tweak: Motion Detectors now detect infected/hugged marines if they are moving, including the user if they are infected.


Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
```diff
@@ -76,6 +77,27 @@ public override void Initialize()
...
+            return;
+
+        var trackedEnt = new Entity<MotionDetectorTrackedComponent>
+        {
+            Owner = ev.Target,
+            Comp = AddComp<MotionDetectorTrackedComponent>(ev.Target)
+        };
+
+        _addedTracked.Add(ev.Target);
+    }
```
SabreML: don't need the `trackedEnt` `Entity<>` var here since `_addedTracked` is a `HashSet<EntityUid>`, so you can just call `AddComp()` on its own:

```suggestion
        if (!HasComp<MotionDetectorTrackedComponent>(ev.Target))
        {
            AddComp<MotionDetectorTrackedComponent>(ev.Target);
            _addedTracked.Add(ev.Target);
        }
    }
```

(As a side note, `_toUpdate` and `_tracked` both storing `Entity<MotionDetectorTrackedComponent>` *does* make it seem like this one should do the same instead of `EntityUid`s, but it's actually pretty bad practice to store full `Entity<T>`s since there's no built-in checks to make sure that the component `T` is still valid on the entity.)
(so essentially `HashSet<EntityUid> _addedTracked` is fine as-is)

Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
```diff
@@ -76,6 +77,27 @@ public override void Initialize()
...
+
+    public void RemoveAddedTrackedComponent(EntityUid ent)
+    {
+        // Remove the MotionDetectorTrackedComponent from the entity upon removal of VictimInfectedComponent if it didn't have it before addition.
+        if (!_addedTracked.Remove(ent))
```
SabreML: Might be worth changing the name from 'added tracked' to something like 'temp tracked'? Just because the method name `RemoveAddedTrackedComponent` was a bit confusing initially lol.

Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
```diff
@@ -379,14 +401,23 @@ public override void Update(float frameTime)
...
 
-                if (hasFaction && _userFactions.Any(f => _gunIFF.IsInFaction(tracked.Owner, f)))
-                    continue;
+                // Motion detectors always detect infected marines, including if the user is infected regardless of faction. 
+                var infected =
+                    HasComp<VictimInfectedComponent>(tracked.Owner) ||
+                    (TryComp<InfectableComponent>(tracked.Owner, out var infectable) &&
+                        infectable.BeingInfected);
+
+                if (!infected)
```
SabreML, line 412: Since this is running inside `Update()` and it's per motion detector + per MD-detected entity, it'd be good to have these as cached `EntityQuery`s to let it skip a couple of steps fetching them each time:

```suggestion
                var infected = _infectedQuery.HasComp(tracked) ||
                    (_infectableQuery.TryComp(tracked, out var infectable) && infectable.BeingInfected);

                if (!infected)
```

Then above, to get the queries in the first place:
```diff
    private EntityQuery<StorageComponent> _storageQuery = default!;
+    private EntityQuery<VictimInfectedComponent> _infectedQuery;
+    private EntityQuery<InfectableComponent> _infectableQuery;

    private readonly HashSet<Entity<MotionDetectorTrackedComponent>> _toUpdate = new();
    private readonly HashSet<Entity<MotionDetectorTrackedComponent>> _tracked = new();
    private readonly HashSet<EntityUid> _addedTracked = new();
    private readonly HashSet<EntProtoId<IFFFactionComponent>> _userFactions = new();

    public override void Initialize()
    {
        _detectorQuery = GetEntityQuery<MotionDetectorComponent>();
        _storageQuery = GetEntityQuery<StorageComponent>();
+        _infectedQuery = GetEntityQuery<VictimInfectedComponent>();
+        _infectableQuery = GetEntityQuery<InfectableComponent>();
```

SabreML (commented):
This is a really fun idea.
All seems to work fine on my end, just a couple of minor things.

Wuattt:
What is the reason for this change and how it's explained in lore? Detector ignores marines if they have dogtags. Does it detect alien through ultrasound or x-rays? If to add checking for parasites, better off adding this functionality to health analyser, since it already scans the whole body.

Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
EndPage: Thank you!

Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
EndPage: Yeah I agree, I'll change that.

Content.Shared/_RMC14/MotionDetector/MotionDetectorSystem.cs
EndPage: Thank you! I didn't know that this existed, good to know for the future.

EndPage:
> What is the reason for this change and how it's explained in lore? Detector ignores marines if they have dogtags. Does it detect alien through ultrasound or x-rays? If to add checking for parasites, better off adding this functionality to health analyser, since it already scans the whole body.

The motion detector can already detect xenos and survivors through any walls and objects, so I figured the marine's body wouldn't necessarily prevent it from picking up the parasite's movement either. The dogtags would still cause the marine themselves to be ignored, but the movement of the xeno inside them is causing the signal.

In lore this could be explained by the dogtags having some sort of biometric tag with the marine they are assigned to, and the motion detector detects something (using some sort of physics mechanism) inside the marine that doesn't have that biometric tag or something. 
 
I agree that the health analyzer detecting infections would also make sense, though I think that could be handled separately in another PR.

In terms of actual gameplay, I don't really expect this having much of an impact. I see it mostly as a scenario that creates some interesting roleplay opportunities.

ThanosDeGraf:
> The motion detector can already detect xenos and survivors through any walls and objects

Only when they move, it is a MOTION detector after all.

> In terms of actual gameplay, I don't really expect this having much of an impact. I see it mostly as a scenario that creates some interesting roleplay opportunities.

There would be cases where xeno manage to infect a marine and drag them back away from the frontline. While a xeno is standing still trying to dig away, marines would be able to pinpoint exactly where they are through the infected marine in its stomach.

It's a far fetched scenario, but that uncertainty whether a infected marine is still in reach of being saved or not is an important part of the marine v xeno mental warfare.

SabreML:
> While a xeno is standing still trying to dig away, marines would be able to pinpoint exactly where they are through the infected marine in its stomach.

That scenario at least shouldn't be a problem, since again it would only be when the infected marine is moving. Plus I just double checked on the branch and if a xeno has a marine swallowed it only gets picked up as a single contact.

Wuattt:
>The motion detector can already detect xenos and survivors through any walls and objects, so I figured the marine's body wouldn't necessarily prevent it from picking up the parasite's movement

How it differentiates a xeno from bowel movements?

> would only be when the infected marine is moving. 

So it detects movement or some bluespace radiation?

>That scenario

Another scenario would be infected SSD marine hidden behind a wall, but pinging on motion detector

EndPage:
> Only when they move, it is a MOTION detector after all.
>

> Another scenario would be infected SSD marine hidden behind a wall, but pinging on motion detector
>
I think a big miscommunication here is that the Motion Detector detects infected marines even if they are not moving. My changes don't cause the MD to always ping infected marines. They will only be pinged if they are moving. Basically if marine is infected AND moving the MD will ping. So these scenarios provided will not happen.

I edited my PR to reflect these changes.

Wuattt:
> In lore this could be explained by the dogtags having some sort of biometric tag with the marine they are assigned to, and the motion detector detects something (using some sort of physics mechanism) inside the marine that doesn't have that biometric tag or something.

Please explain on this more. If motion detector doesn't detect eating a watermelon (it's not a part of biometric tag data), how it detects a parasite (also not a part of biometric tag data)? From movies we know that parasites don't really move inside until they bust chest.

EndPage:
> > In lore this could be explained by the dogtags having some sort of biometric tag with the marine they are assigned to, and the motion detector detects something (using some sort of physics mechanism) inside the marine that doesn't have that biometric tag or something.
> 
> Please explain on this more. If motion detector doesn't detect eating a watermelon (it's not a part of biometric tag data), how it detects a parasite (also not a part of biometric tag data)? From movies we know that parasites don't really move inside until they bust chest.

Frankly my explanation above was just me writing my thoughts down - not an official explanation. I don't know enough about the lore to give a satisfactory answer here, but perhaps another PR could add lore explaining exactly how the MD works. Right now I'm more focused on the gameplay elements. I mainly added this change because of a Discord feature request thread that I thought could be fun.

Ozehk:
I wouldn't overthink it, the marine just has an immature larva (by extension a xenonid) inside of them.

carccborg:
> I wouldn't overthink it, the marine just has an immature larva (by extension a xenonid) inside of them.

We are indeed bikeshedding

PursuitInAshes:
@Wuattt @EndPage 
As a general reminder; PR comments are primarily for reviewing code and not the specifics of a feature or if it should be implemented; maintainers can permit this as required but the former is the standard and expectation. You should use discord for discussions like this as it has dedicated places for such.
Thanks for the PR EndPage, and discussion though

PR #3767, opened 2024-08-20
Keep Morgue Quiet

blueDev2:
## About the PR

Fixes #3748

## Media

- [X] I have added screenshots/videos to this PR showcasing its changes ingame, **or** this PR does not require an ingame showcase

**Changelog**

:cl: blueDev2
- tweak: Morgue slabs no longer beep when there is a soul in them.

PR #10738, opened 2026-08-19
Fix Morgue Beeping

### LINKED ISSUE #3748: Morgues will sound the alarm for bodies if a ghost is inside, regardless of if the body is unrevivable or not

MrFish-100:
## About the PR
Added a check for revivability on the morgue to stop if from beeping when a ghost is in a body.

## Why / Balance
This is an unneeded feature from upstream.

Closes #3748 
## Requirements

**Changelog**
:cl:
- fix: Fixed morgue's unintended beeping

### DIFF
--- a/Content.Server/Morgue/MorgueSystem.cs
+++ b/Content.Server/Morgue/MorgueSystem.cs
@@ -3,7 +3,7 @@ using Content.Shared.Body.Components;
 using Content.Shared.Examine;
 using Content.Shared.Morgue;
 using Content.Shared.Morgue.Components;
-using Content.Shared.Traits.Assorted;
+using Content.Shared.Traits.Assorted; //RMC14
 using Robust.Shared.Audio.Systems;
 using Robust.Shared.Player;
 
--- a/Content.Server/Morgue/MorgueSystem.cs
+++ b/Content.Server/Morgue/MorgueSystem.cs
@@ -3,6 +3,7 @@ using Content.Shared.Body.Components;
 using Content.Shared.Examine;
 using Content.Shared.Morgue;
 using Content.Shared.Morgue.Components;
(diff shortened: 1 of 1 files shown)


CatAndHats (dismissed):
Ty for the PR. For any non-RMC files you'll need to comment that it is an RMC change, as per coding standards here https://github.com/RMC-14/RMC-14/blob/master/coding-standards.md#upstream-code-changes.

PursuitInAshes (changes requested):
Thanks for the PR!

Instead of any C# changes it would be best to instead edit the `CMMorgue` entity to either not have the `Morgue` component or to set the `DoSoulBeep` field to false if just outright removing it causes functionality issues.

MrFish-100:
Setting `DoSoulBeep` to false was already tried in #3767, from where @DrSmugleaf suggested an unrevivable check.

A bonus to this approach is that the beep retains its functionality, now beeping when a non-permadead entity is locked inside.

PursuitInAshes (dismissed):

Content.Server/Morgue/MorgueSystem.cs
```diff
@@ -3,6 +3,7 @@
 using Content.Shared.Examine;
 using Content.Shared.Morgue;
 using Content.Shared.Morgue.Components;
+using Content.Shared.Traits.Assorted;
```
CatAndHats:
```suggestion
using Content.Shared.Traits.Assorted; //RMC14
```

CatAndHats (changes requested):
Can approve after this. Ty

CatAndHats (approved):

PursuitInAshes (approved):
Thanks for the PR!
Triggering tests but it seems fine

PR #9459, opened 2026-03-07
multi character selection

Vixting:
## About the PR

ports https://github.com/space-wizards/space-station-14/pull/36493 with fixes made to work in RMC & the addition of map preferences

### From the PR, made by quantum-cross:
This is my as of now successful attempt at implementing multi-character selection. I wanted to get a draft on the books so I can start making incremental commits/changes to fix individual issues.

It allows each character slot to be enabled or disabled, as well as having binary job selections similar to how antags are selected.

Job priorities (never -> high) are selected globally on your player. This is represented by a new button above your characters that says "Edit job priorities".

All of these new configuration elements are saved on the server database.

When the round begins, the server will first create a set of enabled jobs on all your characters, and find the intersection of that with your player job priorities > "never". It will then use that resultant list of jobs to consider for your assignment.

After assignments are complete and the players are to be spawned, the server will look at all of your enabled characters who can take that job and pick one randomly for you to play.

## Technical details

Changes made from original pr:

- Surv preferences for map variants, show  variant % chance of rolling. Show requirements for map variants.
- Removed static variables
- Fix some issues with rebuilding tree
- Drag ghost instead of reparenting icon. Drag now uses popup ghost while the real icon stays in layout. Removes a lot of tree mutation issues that could occur
- Job order indexes cached instead of repeated lookups
- One source for the prioritizes in panel move ordering, remove some issues when something else edits them

## Media

[chars.webm](https://github.com/user-attachments/assets/802fd66d-f6ef-426b-993d-ad5bb5cce949)

image

image

## Requirements

- [X] I have read and am following the [Pull Request and Changelog Guidelines](https://docs.spacestation14.com/en/general-development/codebase-info/pull-request-guidelines.html).
- [X] I have added media to this PR or it does not require an ingame showcase.

- [X] By submitting this code and/or assets, I confirm that I either own them or have provided the correct necessary licenses to use and distribute them. I agree to be fully responsible for any legal claims or issues arising from the use of these materials.

**Changelog**

:cl: Quantum-Cross, Vixting
- add: You can now disable rolling inserts / base survs
- add: You can now choose specific squads & choose to stay within squad if they aren't selected
- add: Job priorities are now bound to your player
- add: You can now set multiple squad preferences & choose to remain in the lobby if your selected squad(s) are not available. 
- add: You can now set preferences for individual surv jobs & see % chance of a job occuring
- tweak: Job preferences on characters are "yes" or "no" like antag selection
- add: Characters can be individually enabled or disabled
[...]

PR #10754, opened 2026-08-21
Job preferences now work like actual preferences

MidZik:
## About the PR

Roundstart role selection is changed to behave more intuitively. You can now select multiple high-demand jobs with varying preferences, and the game will choose a job for you following those preferences when possible. Most notably, your job preferences don't compete with other player's preference. Players that have a job with HIGH preference no longer have priority for the job over those who have it set lower.

This works by assigning players a random priority. Those with high priority get assigned jobs first. Those with lower priority only get assigned to jobs that haven't been taken by those with higher priority.

Xenos and survivors no longer have an innate priority to be chosen. WS high, survivor medium, xeno low is now a perfectly acceptable set of preferences that will give you WS if possible, survivor only if you failed to roll WS, and xeno only if you failed to roll both WS and survivor.

Certain command jobs will still have priority, so it is possible that you lose a higher preference job in order to fill in for one of these more important jobs. However, these are chosen in _reverse_ priority order. The player with the lowest priority that has such a job marked higher than "never" will be selected for these jobs in the case that no one was selected for them using their normal priority preferences. This is the only time you might receive a lower priority job than you otherwise would have gotten. Note that this means that QUEEN: LOW players MAY receive queen before QUEEN: MEDIUM players, if the LOW player was lower in priority order and the QUEEN: MEDIUM player was assigned a more preferable role.

Additionally, the number of roundstart xenos and survivors will no longer be based on the number of readied players, but instead on the actual roles that have spawned into the round. Namely, shipside roles no longer count as a full player for the purposes of determining roundstart xeno and survivor counts. This will reduce the amount of larvas at the start of most rounds, and will probably reduce survivor counts at lower populations. See #10530 for changes that compensate for this.

## Why / Balance

Turns your job preferences into actual preferences. Players no longer have to decide which high-demand role they want a chance at getting. They can select all the ones they want to try rolling for. Players with a job selected as "High" no longer have an advantage to take the slot over players who have it selected lower.

This will also allow a greater variety of players to be seen in a variety of higher demand roles, because "WS High" no longer requires abandoning rolling for other roles.

The change in how xeno and survivor counts are calculated is parity. Both of their numbers should be based on the roles that will actually spawn in, so that shipside roles don't contribute as much to how many xenos or survs appear. This will usually reduce the number of xenos that appear at round start, and will reduce survivor counts at low populations. See #10530 for changes that compensate for this. 

## Technical details

The player job selection algorithm has been completely changed. The new algorithm assigns players a random priority by shuffling all ready players and assigning them jobs in order.

In order to make sure that the amount of each role available is correct, the amount of slots available is re-calculated after every player is assigned. Afterwards, all players are again re-assigned roles given these new role counts. This loop continues until every ready player has been processed.

## Requirements

**Changelog**

:cl:
- tweak: Round start role selection has been reworked. Each player is now assigned a random priority. Players with higher priority are assigned roles first. Each player receive their most preferred role that remains after all players with higher priority are assigned roles. For example, WS high, Survivor medium, Xeno low will have the game try to assign you WS if possible first, Survivor only if WS isn't possible, then Xeno only if WS and Survivor isn't possible. Players with a role selected with High preference no longer have an advantage over players who have the role as Medium preference.
- fix: Fixed xenos sometimes not getting enough burrowed larva when the amount of round start xeno players was low.
- fix: Fixed xenos getting too many larva when the amount of round start xeno players was high.
- fix: Fixed too many survivors sometimes spawning at lower populations.


Thereisabear:
Might be duplicate or very similar in scope? you'd have to check in with vixting  PR #9459

MidZik:
I don't believe there is functionality overlap. There will be a lot of merge conflicts though.

Gallagin:
If this gets added im looking forward to WS rolling going from ~9% it is currently on average to like a ~1% chance 

Always really wanted to roll for WS without sacrificing my SGO rolling and I feel like this is just going to decrease the amount of me getting either compared to when I flip back and forth which one im rolling every other round.

Curious how it'll fully play out in game as this is probably just going to make all highly demanded jobs even harder to get since now everyone can always roll for them without having to pick and choose

MidZik:
> Always really wanted to roll for WS without sacrificing my SGO rolling and I feel like this is just going to decrease the amount of me getting either compared to when I flip back and forth which one im rolling every other round.

Since the number of slots are the same, any decrease in odds you experience is due to other players getting the role more. Those players would have never rolled SGO in the first place. A consequence of this is that "mains" of certain roles will probably get the role less often, but the upside as I said is that more people can finally try a role without sacrificing their attempt to get their favorite role.

MidZik:
I believe this should work as expected now. It is, however, impossible for me to test higher player counts. I might refactor it so I can write some integration/unit tests.

MidZik:
I forgot that surv counts will drop as well, since their calculation is based on total role weights that spawn in just like xenos are. Surv ratio should probably be changed to compensate for this. Additionally, parity surv count depends on how many xenos there are as well, so ideally that would be changed too so only marine count determines surv count, to prevent metagaming.

PR #10762, opened 2026-08-22
larvae burst victim priority fixes

Vixting:
## About the PR

getting victim priority is now a setting
no longer have to remain in body ( can ghost)
only applies if your body is nested
reduced victim time to 15s to accept / decline
## Why / Balance

parity
prevents extending your burst timer
## Requirements

**Changelog**

:cl:
- tweak: Getting your larvae from yourself bursting is now a setting. You no longer have to remain in your body (can ghost). Only applies if your body is nested. 
- tweak: Reduced larvae offer time to 15s

### DIFF
--- a/Content.Client/Options/UI/Tabs/MiscTab.xaml
+++ b/Content.Client/Options/UI/Tabs/MiscTab.xaml
@@ -13,6 +13,7 @@
                 <CheckBox Name="RMCAutoEjectMagazines" Text="{Loc 'rmc-ui-auto-eject-magazines'}" Margin="0 0 0 5" />
                 <CheckBox Name="RMCDamageYourself" Text="{Loc 'rmc-ui-damage-yourself'}" Margin="0 0 0 5" />
                 <CheckBox Name="RMCShowNewPlayerIcons" Text="{Loc 'rmc-ui-show-new-player-icons'}" Margin="0 0 0 5" />
+                <CheckBox Name="RMCLarvaQueueVictimPriority" Text="{Loc 'rmc-ui-larva-queue-victim-priority'}" Margin="0 0 0 5" />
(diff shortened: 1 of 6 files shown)


Beansies:
the burst victim priority did kinda add another 20 seconds to your burst timer, making shipside bursts a lot more low stakes. 
So before the larva queue you had at least (usually, almost exactly) 10 seconds after the larva was fully formed before a player could take it (ghost role)
after the queue you had over 30 seconds guaranteed with victim priority,
with the change you have 0 seconds guaranteed.

well, I think it will be fine regardless. I believe player attentiveness has lapsed with the queue and think in practice there will be anywhere from a 5-30+ second delay dependent entirely on how attentive the pos 1 of the larva queue is. I would consider reducing the accept time to 15 seconds for all bursts to prevent metagaming players or AFK players letting the timer run the full 30 seconds but that might be splitting hairs. 

regardless I like the PR and it fixes an unintentional buff to hugged marines.

Vixting:
> the burst victim priority did kinda add another 20 seconds to your burst timer, making shipside bursts a lot more low stakes. So before the larva queue you had at least (usually, almost exactly) 10 seconds after the larva was fully formed before a player could take it (ghost role) after the queue you had over 30 seconds guaranteed with victim priority, with the change you have 0 seconds guaranteed.
> 
> well, I think it will be fine regardless. I believe player attentiveness has lapsed with the queue and think in practice there will be anywhere from a 5-30+ second delay dependent entirely on how attentive the pos 1 of the larva queue is. I would consider reducing the accept time to 15 seconds for all bursts to prevent metagaming players or AFK players letting the timer run the full 30 seconds but that might be splitting hairs.
> 
> regardless I like the PR and it fixes an unintentional buff to hugged marines.

15s globally, cm its instant but I was concerfned about afk people but hopefully this should be fine

PR #10782, opened 2026-08-26
fix: flamer nozzle fixes

Ian321:
## About the PR

Fix my flamer nozzle PR.

## Why / Balance

- Someone mapped the old proto and I forgot to add it to migrations.yml.
- While testing I also noticed that the surv flamer is also supposed to have it. See: https://github.com/cmss13-devs/cmss13/blob/master/code/modules/projectiles/guns/flamer/flamer.dm#L996

## Technical details

Just some yaml.

## Requirements

**Changelog**

No

### DIFF
--- a/Resources/Maps/_RMC14/hybrisa.yml
+++ b/Resources/Maps/_RMC14/hybrisa.yml
@@ -82831,7 +82831,7 @@ entities:
     - ContainerContainer
     - Speech
     - RequisitionsComputer
-- proto: CMAttachmentXMVESG1FlamerNozzle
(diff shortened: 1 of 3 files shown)


CatAndHats (changes requested):
this PR is fine however you seem to have also removed the SLs flamethrower in their case in the original PR?? and not actually added the nozzle which may need to be there from what the TODO says

Image

CatAndHats (approved):
Nevermind I cannot read. LGTM

kleinerstation13 (approved):
sure

PR #10819, opened 2026-09-04
Multi-directional ladders

### LINKED ISSUE #8776: The ladder ID system is broken
## Description

The ladder ID system is probably not working as intended currently. You can assign as many ladders as you want to the same ID if you go about it a certain way.

**Reproduction**

Create two ladders and then link them with an ID of X.
Create another two ladders and link them with the same ID of X.
The ladders will now all share an ID of X.

**Video**

https://github.com/user-attachments/assets/c9592749-2f80-4484-ba93-8ea466c3fce0

**Additional context**

The reason for this behaviour is that ladder IDs that are checked are only the most recently created ladders. Everything that was created before the last spawned ladder will not be included in the ID checking.

SabreML:
## About the PR
Adds the ability for ladders to be connected to multiple other ladders at the same time, allowing you to climb up *and* down! (wow)
Interacting with a ladder that has multiple others connected (up and down) will open a radial menu, allowing you to pick which way you want to climb (or look).
In order to actually implement this, the PR comes with a breaking change of a new `Level` datafield on `LadderComponent` to indicate the "floor level" that the ladder is on. This is required so that the ladder can tell what should be connected to what.

I'm not a mapper, but I tried to keep the process of creating+linking new ladders simple:
1. Place the ladders on the map wherever you want them as usual.
2. For each group of ladders you want to connect together, choose their intended "floor level". The default value is 0, so you could have one "above" that at 1, another above *that* at 2, etc. (negative numbers should also work)
3. Run the command `ladder:set_level {uid of ladder} {floor level num}` (Or VV their ladder component) to set the "floor level" of each ladder in the group.
Every ladder in a group *must* have a unique `Level` value, otherwise the system won't know what's up and what's down. (this shouldn't be a problem since, you know, that's how ladders work)
5. Run the command `ladder:group_add {uid of ladder} "{group name}"` with each ladder's UID to add them all to the same group.
6. Done! On mapload, `LadderSystem` will go through each ladder group and link them all together.

Resolves #8776.

## Why / Balance
Cool feature requested by map people :)

## Technical details
I ended up refactoring most of the ladder System so there's a few things to mention here.

### LadderBoundUserInterface.cs:
- There's a weird hacky workaround in here at the moment to bypass `SimpleRadialMenu` not supporting buttons with both an icon and a style class. (In this case `"RadialMenuButton"` for the square grey background)
https://github.com/space-wizards/space-station-14/pull/39223 fixes the problem, and I got halfway through cherry-picking it but eventually decided that maybe reworking every radial menu in the game was a bit out of scope of this PR lol. I'll do that in a separate PR if I remember.
- Whether the radial menu selects a ladder to climb or to view is dependent on the `LadderRadialAction` used to open it, which acts as its UI key. I wasn't able to find any other existing interfaces that had multiple key variants pointing to the same base BUI, but there's no reason it *shouldn't* work really. (Seems to go fine here)
<hr>

### LadderCommand.cs:
The `linkladders` command has been replaced with three subcommands, `ladder:group_add`, `ladder:group_rem`, and `ladder:set_level`.
The main thing of note there is the copy-pasted checks and error messages from the ladder system methods.
I wanted any errors to be visible in the console so that the user actually knows what happened, but since it's all in the server-side `LadderSystem` they only come up in the server console.
The current solution definitely isn't ideal, but it's probably fine since it only happens, assuming there isn't a "correct" way to pass error messages through.

I *had* wanted to add an optional `int? level` parameter to the `ladder:group_add` command so that you could skip needing to use `ladder:set_level` and do it all in one go, but I wasn't able to figure out how optional params work for `ToolshedCommand`s.
(all existing commands with optional params seem to use a different older system, not `ToolshedCommand`/`CommandImplementation`)
<hr>

### LadderSystem.cs:
Mostly self-explanatory, I think. I moved the `Update()` loop and related methods to server-side since it wasn't predicted in the first place.
<hr>

### LadderComponent.cs:
- Renamed `Id` to `GroupId` since it that fits a bit better with the new system.
- Changed `Other` to `Above` and `Below`, which are automatically set based on the ladder's `Level` position in its group.
Ideally these would be completely skipped over when saving a map if they were accidentally set, but I'm not sure if that's possible.
- Simplified the `DoAfter` fields.
<hr>

### SharedLadderSystem.cs
Refactored pretty much everything other than the 'Watch' stuff. The github diff is all jumbled up and very unhelpful so I suggest reviewing this as just the full files before/after, since it isn't too bad then.
There's really too much changed for me to go into any detail here, but one semi-unrelated thing I added is directional text for the popup messages, so it says "climbing up" or "climbing down" depending on the relative `Layer`s instead of always being "climbing up". (Inspired by #9136)
<hr>

### Various map files
Edited every existing ladder to work with the new system.

**IMPORTANT NOTE:**
This *doesn't* include ladders on the private maps like the Almayer, so those will be broken unless someone with access can get to them.
My process for fixing the existing maps was to temporarily re-add the `Id` datafield and remove the `[ViewVariables(VVAccess.ReadOnly)]` from `GroupId` in `LadderComponent`, then open the map with `mapping`, run `entities with Ladder visualize`, then manually VV and set the `Level` of the "upstairs" ones to 1, and cut+paste their old `Id` value into their `GroupId`.
Then `savemap`, and that should in theory result in a map diff similar to the ones in this PR.

Alternatively you could just directly edit the fields in the map's yml file, swapping out `id` for `groupID` and adding a `level`, but that does make it more difficult to to tell which of the ladders should be "above" and which should be "below".
<hr>

## Media

https://github.com/user-attachments/assets/65f86b24-52c1-4e99-ac7c-5fbfaf48fd79

(The slight delay when opening the UI has been fixed now)

## Requirements

**Changelog**

:cl:
- add: Added support for ladders to have multiple connections.
- tweak: Added directional text ("up" or "down") for ladder climbing.
- code: Connected ladders now each require a unique 'Level' value. (See PR for mapping details)


Content.Client/_RMC14/Ladder/LadderBoundUserInterface.cs
```diff
@@ -0,0 +1,139 @@
...
+                return;
+
+            // In order of the buttons' creation above.
+            AddButtonIcon(container.Children.ElementAt(0), UpIcon);
+            AddButtonIcon(container.Children.ElementAt(1), DownIcon);
+            break;
+        }
+    }
+
+    // Workaround until `RadialMenuIconSpecifier` is ported from upstream.
```
Vixting, line 97: add a TODO RMC comment here

Content.Client/_RMC14/Ladder/LadderBoundUserInterface.cs
Vixting, line 1: some of this could be moved to the component buts its probably fine for now

Content.Server/_RMC14/Ladder/LadderSystem.cs
```diff
@@ -9,34 +12,154 @@ public sealed class LadderSystem : SharedLadderSystem
...
+            return false;
+
+        var group = GetLadderGroup(newGroupId);
+        if (group.TryFirstOrNull(l => l.Comp.Level == ladder.Comp.Level, out var sameLevelLadder))
+        {
+            Log.Error($"Failed to add {ToPrettyString(ladder)} to group '{newGroupId}'. {ToPrettyString(sameLevelLadder)} has a duplicate Level value of {ladder.Comp.Level}!");
+            return false;
+        }
+
+        ladder.Comp.GroupId = newGroupId;
```
Vixting:
```suggestion
        ladder.Comp.GroupId = newGroupId;
        Dirty(ladder);
```

Vixting (changes requested):
minor issue

Content.Client/_RMC14/Ladder/LadderBoundUserInterface.cs
SabreML: Oh, yeah looking again at the coding standards page it does directly say to avoid using BUI states. I will try moving some over if only for the sake of doing it "properly".

Vixting (approved):

CatAndHats:
Do Not Merge (yet) label because Smug would need to fix the Almayer ladders if this gets merged.

PR #10830, opened 2026-09-06
Ravagers now have a chance to hit the target instead of tackling

BreakpointDebug1234:
## About the PR

Ravagers now have an 25% chance to slash their target instead of tackling them.

## Why / Balance

Parity change see PR: https://github.com/cmss13-devs/cmss13/pull/12742

Nerfs the ability of ravagers to capture while still allowing them to try to disarm hosts, this will force hives to diversify their castes instead of using ravagers to both kill and capture.
## Technical details

Adds an slash chance to the component which will trigger an attack instead of an tackle based on an 25% chance, similar to how warrior behaves on lunged targets. 

Modifies TackleSystem to accommodate for this by interacting with the MeleeAttackAttemptEvent and changing the attack to an LightAttackEvent if the players fails the roll.

Adds necessary admin logs.
## Media

image

https://github.com/user-attachments/assets/525d2ad5-696c-41e9-8a1c-bb7ae4593d3e

## Requirements

**Changelog**

:cl:
- tweak: Ravagers now have an 25% chance to hit their target on tackle.

### DIFF
--- a/Content.Shared/_RMC14/Tackle/TackleComponent.cs
+++ b/Content.Shared/_RMC14/Tackle/TackleComponent.cs
@@ -20,4 +20,7 @@ public sealed partial class TackleComponent : Component
 
     [DataField, AutoNetworkedField]
     public TimeSpan StunMax = TimeSpan.FromSeconds(3);
+
+    [DataField, AutoNetworkedField]
+    public float SlashChance;
 }
--- a/Content.Shared/_RMC14/Tackle/TackleSystem.cs
+++ b/Content.Shared/_RMC14/Tackle/TackleSystem.cs
@@ -315,6 +315,11 @@ public sealed class TackleSystem : EntitySystem
         if (!_random.Prob(ent.Comp.SlashChance))
             return;
 
+        var user = ent.Owner;
+        var target = GetEntity(args.Target);
+
+        _adminLog.Add(LogType.RMCTackle, $"{ToPrettyString(user)} tried to disarm {ToPrettyString(target)} but failed, slashing them instead.");
+
         args.Attack = new LightAttackEvent(disarm.Target, args.Weapon, disarm.Coordinates);
         _pendingTackleSlash[ent.Owner] = (GetEntity(args.Target), _timing.CurTime);
     }
--- a/Content.Shared/_RMC14/Tackle/TackleSystem.cs
+++ b/Content.Shared/_RMC14/Tackle/TackleSystem.cs
@@ -2,6 +2,7 @@ using Content.Shared._RMC14.Hands;
 using Content.Shared._RMC14.Marines.Skills;
 using Content.Shared._RMC14.Pulling;
 using Content.Shared._RMC14.Xenonids.Parasite;
+using Content.Shared._RMC14.Weapons.Melee;
 using Content.Shared.Administration.Logs;
 using Content.Shared.Buckle.Components;
 using Content.Shared.Damage.Systems;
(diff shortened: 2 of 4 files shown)


Randomplayer2or3:
This is a bad change, as ravs killing while tackling is a part of modern CM balancing (on CM bursts are 1.5 while you can't tunnel caps anymore).

BreakpointDebug1234:
This should impact captures that much, it just means the xenos will have to use other castes as they are already (defenders, warrior, etc.)

MidZik:
> This should impact captures that much, it just means the xenos will have to use other castes as they are already (defenders, warrior, etc.)

Ravagers are generally considered the backbone of a hive here, and a large part of that is because of their ability to capture, especially capturing marines from queen screeches against the marine ball. Saying it won't impact captures that much is disingenuous.

Kvantos0:
This absolutely will impact captures a lot, you almost always do at least 5 tackles while dragging off a scream effectively guaranteeing that you'll slash at least once, vest users especially will often just get denied arbitrarily while being tackled, this won't diversify hive composition, rather it will confuse new ravagers and encourage defender and dancer spam for dragging instead of ravagers - making explosions and CAS even stronger than it already is. You're already giving up wiping power as a hedgehog to specialize into capturing, berserker is not nearly as good at capturing. I don't see how a recent balance change on CM13 makes sense for the current version of RMC14.

Randomposter12:
> This is a bad change, as ravs killing while tackling is a part of modern CM balancing (on CM bursts are 1.5 while you can't tunnel caps anymore).

So i just want to point out something. Is that the CM13 pr itself mentions it that the change was done because Ravs by themselves have a lot on their toolkit and can do a lot, it was not inherently made because of the modern larva burst and haul changes in CM13, trying to make it the focus is a lie of omission.

And to talk about modern parity changes. While it is a modern change, it is up to the maints and hosts to decide if it benefits RMC. It is not uncommon for modern changes to be merged if it befenits our community or the vision of the hosts.

My only problem would be if this were merged without proper newmed to actually reinforce ravs power as a killing xeno such as they are based off the lurker style of massive body harm. But i trust the hosts/maints know about that and can hold off on such a change.

rafapecogames:
although of course all initiative and prs are welcome, i dont know why we should nerf ravager. ravager, boiler and valk are the only complete and balance kits imo. all others are too weak (all xeno castes) or too strong (queen). making ravager capture harder will just make it way more harder to make captures for xenos. it will impact A LOT on capturing. t1s and t2s mostly fear dragging because too few health (understandable) and praes basically have this problem aswell, just vanguard and dancer that can drag but even then it isnt that reliable. ravagers are the only ones that can drag consistently, albeit slower than praes.

this is not making things more diverse. its just forcing it. instead we should focus on buffing praes to get on the level of a ravager, i feel.

 besides, there is something somebody already commented that is modern balance changes arent exactly considered parity. i already think listening to cm13 is a bad decision, they have a lot of questionable choices, but there is also the question that cm13 is already far more advanced in development than RMC is.
 
 this will also just make xenos have more fights with themselves. it will not stop ravagers from dragging, because it will still be the best caste for it, but whenever they slash xenos will just have problems between themselves which could just cause more conflicts. the previous comments already show their disproval on it but i feel i should comment all of these aswell

CatAndHats:
We will wait for Smug's opinion before making a final decision, but the Maintainers seem pretty agreed that this would not be a good change for various reasons. More recent CM13 changes are under higher scrutiny, which is why this seems likely to not make the cut.

QuietlyWhisper:
Please reserve github for development comments and reviews, opinions and discussions can be directed to a feature post

PR #10843, opened 2026-09-11
Fix CAS missiles offsetting up to 6 tiles away

butchcats:
## About the PR
Fixes an issue with PR #8860 that allowed 'double-offsetting', resulting in missiles being offset up to 6 tiles away from the original CAS mark.

## Why / Balance
Bugfix. Missile offset now behaves as intended.

## Technical details
If the initial tile selected by the missile offset is an indestructible wall, tiles within 3 tiles of *that* tile become valid search targets for FindAlternateLandingTile.  
- Added a range limit to FindAlternateLandingTile to only select tiles within both 3 tiles of the initial tile *and* the offset tile.  
- If no valid tiles exist within this bound (i.e. the initial offset is sent three tiles into cave wall), searches all tiles around the original flare until a valid tile is found.
- If no valid tiles are found (physically impossible outside of a dev environment, as cas marks can't be placed deep into indestructible walls), missile fires on the original offset.

## Media
Not sure how to display this in a way that reliably demonstrates that missiles *can't* offset anymore, but I have thoroughly tested it.

## Requirements

**Changelog**
:cl:
- fix: Fixes an issue with PR #8860 that allowed 'double-offsetting', resulting in missiles being offset up to 6 tiles away from the original CAS mark.

### DIFF
--- a/Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
+++ b/Content.Shared/_RMC14/Dropship/Weapon/SharedDropshipWeaponSystem.cs
@@ -744,12 +744,13 @@ public abstract class SharedDropshipWeaponSystem : EntitySystem
         Dirty(weapon.Value, weaponComp);
 
         var spread = ev.Spread;
+        var originalCoords = coordinates;
         var targetCoords = coordinates;
         if (spread != 0)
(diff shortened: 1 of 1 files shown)


CatAndHats:
TY for the PR. Please use the changelog format that is given, to prevent any issues with the automatic changelog system, i've fixed it for this PR

PR #10847, opened 2026-09-14
Change Pressurized Container Pouch's Flush Container to Normal Verb

ShepardToTheStars:
## About the PR

Made Flush Container a regular verb for the pouch.
 
## Why / Balance

There's no reason to have this the first interaction when alt-clicking a pouch. Flushing a pouch should be more intentional rather than it being a 'gotcha' because SS14 controls.

This should only affect these pouches.

## Technical details

none really. Just changed AlternativeVerb to Verb.

## Media

image

## Requirements

**Changelog**

:cl:
- tweak: The pressurized reagent container pouch will no longer flush when alt-clicked. It can still be flushed from the interactions menu.


Ozehk:
I find it much more intuitive and practical for the alternative verb to be ejecting the canister. Do you think you might want to include that here?

PR #10853, opened 2026-09-15
M20 colony rifle

kleinerstation13:
## About the PR
Adds the M20 colony rifle from CM, currently not mapped or in loadouts

Basically the same stats as the M54C except for worse mag capacity, worse damage and no access to AP

## Why / Balance
Parity

CM wants to move away from the M16 and have a UN weapon that isn't the M54C to be mapped for survivors, we'll probably be killing it for similar reasons

Not mapping it anywhere in this PR but it's supposed to be on Solaris, and on NV in the LACN rework

## Technical details
Sprites and yaml for the mag, gun, stock and magbox

## Media
https://github.com/user-attachments/assets/6a43ae3c-eae1-4caa-8719-26d359dd6852

obraz

The name is a SOMA reference, there's no other reason why I changed it

## Requirements

**Changelog**
:cl:
- add: Added the M20 Assault Rifle, currently admin spawn only


kleinerstation13:
Would also like a lore check on this because I changed the description

kleinerstation13:
Three heisentests in a row...

PR #10858, opened 2026-09-17
[Fix] Sentry Laptop Leak & Errors

Thereisabear:
## About the PR
Should fix sentry laptops being evil. 
The Sentry camera view wasn't closing properly if closed via the X, moving out of range or picking up the laptop.
SentryLaptopWatcherComponent was not being removed properly.

## Why / Balance
Fixes and it was causing server to its knees at times.

## Technical details
SentryLaptopWatcherComponent was not being removed properly.
Adds a OnBoundUIClosed that handles that properly which ClearsWatcher.

Target Name console error/Local Crash 
Sentry was still holding stale ref causing throws log spam. (Added to the server crawling.)
Def guard in GetSentryTarget, GetSentryTargetName. If truly not existing, it'll just show Unknown instead of throwing.

Stale SoundOnDeath ref, pointed it back to its originating so when one goes the ref clears. 

## Requirements

**Changelog**
:cl:
- fix: Fixed sentry laptops causing server issues. No in-game behavioural change otherwise.


PursuitInAshes (approved):
Seems okay

PR #10860, opened 2026-09-17
minor simple mob fixes

Drazamuffin:
## About the PR

Removes defib comp from most simplemobs
Gives simple mobs that marines typically have easy access to slowdown modifiers when worn to prevent marines wearing them in combat.

## Why / Balance

Simple mobs being non-reviveable is parity
Doc has been omitted because that stupid mouse dies a lot due to drinking coffee

Easy to access pets now slowdown marines. Colony pets have been omitted from this for the most part as survs have better mechanical reasons for needing to place a cat on their head

## Requirements

**Changelog**

:cl:
- tweak: Easy to access simple mobs for marines now have a slowdown modifier when worn.
- tweak: Simple mobs are no longer reviveable... except Doc.


Vermidia:
Small thing for Doc - it might be better to instead of make them revivable, replace their animal stomach with a normal human one (so it can eat whatever marines eat just fine) or just remove their hunger or the AI to eat things (I think in CM they only eat cheese?).

Drazamuffin:
> Small thing for Doc - it might be better to instead of make them revivable, replace their animal stomach with a normal human one (so it can eat whatever marines eat just fine) or just remove their hunger or the AI to eat things (I think in CM they only eat cheese?).

Looking over it, you're right going to draft this as because i totally didnt succumb to scope creep

PR #10864, opened 2026-09-18
[QoL Tweak] Meltable Hydroponics Trays

ffxivsidehoes:
## About the PR

Hydroponics trays are now meltable.

## Why / Balance

Xeno QoL. Could potentially affect balance insofar as xenos having to spend less time manually destroying trays and having more time to build, but I think this would be an incredibly minor effect.

## Technical details

Added - type: Corrodible to CMHydroponicsTray

## Media

https://github.com/user-attachments/assets/8abbfd18-69eb-44a5-ae18-5700726d24c3

## Requirements

**Changelog**

:cl:
- tweak: Hydroponics trays are now meltable.


CatAndHats (changes requested):
When looking at changing an object for QoL please first check Parity to see if how it works in RMC is correct, for example in CM13 Hydroponics trays are actually destroyed by xenos in one hit. This would be preferred to your current implementation, as them being un-acidable is correct, but making them be able to be destroyed easily serves the same purpose.

CatAndHats:
To elaborate, you'd want to do something like what https://github.com/RMC-14/RMC-14/pull/9118 did.
image
(You'd want to check if Hydroponics trays should drop metal, I don't think they do)

CatAndHats (approved):
After some investigation it turns out that they ARE actually meant to be meltable in CM13 - however some maps make them not meltable manually... I hate this. So I'm approving this now, we can decide if we want to make the specific map ones indestructible later.

